PR #2, opened 2016-02-18
Fix compile error on Linux

gordonchaffee:
error: format not a string literal and no format arguments
   [-Werror=format-security]


sterlinghughes:
Hi Gordon, Thanks for the pull request.  apache:master is locked for release, we've switched to apache:develop as the main branch for development.  That gets regularly merged into master.

PR #7, opened 2016-02-25
Mynewt 9

paulfdietrich:
This is a basic BSP for arduino zero. It includes the atmel samd drivers integrated into the mynewt framework with a BSP for the arduino zero board.


mkiiskila:
There's some changes to stm32f4xx UART driver here which look accidental, get rid of those.

Another thing is changes to libs/cmsis-core/src/cmsis_nvic.c break the build for NRF51, which is another Cortex-M0. Maybe change the #ifdef for __VTOR_PRESENT check for the presence of the define instead of the value of the define?
I.e. #ifdef __VTOR_PRESENT as opposed to #if (__VTOR_PRESENT == 0)

Also, could you pull in the latest changes from develop branch and merge with this stuff? YAML file stuff has changed there.

mkiiskila:
Yet another thing that does not build anymore is slinky when building for native BSP (on MacOS).
Try building this target:
[marko@IsMyLaptop:~/src/larva]$ newt target show slinky_sim 
slinky_sim
	arch=sim
	bsp=hw/bsp/native
	compiler=sim
	compiler_def=debug
	name=imgmgr_sim
	project=slinky
Passing uint32_t to %lx in printf is not accepted by native compiler, I would typecast the value to unsigned long.

paulfdietrich:
Marko,

I made the changes to the things you recommended. I think this is ready.  NOTE: there is a small patch 4fc6b7a for a uninitialized data in a fix Chris added which I picked up when I pulled from develop. I reviewed with Chris offline and he though the patch was fine until he gets in to take a look

PR #9, opened 2016-03-04
Mynewt 230

paulfdietrich:
The arduino zero build was complaining about the implicit cast from char to int.  This adds the explicit cast.


paulfdietrich:
i screwed up and did this with respect to master instead of develop.  I'll resubmit.

PR #10, opened 2016-03-04
MYNEWT-230

paulfdietrich:
explictely cast to an integer since some compilers don't like to implicitelty
cast a char to an int

### DIFF
--- a/libs/util/src/datetime.c
+++ b/libs/util/src/datetime.c
@@ -211,7 +211,7 @@ parse_number(const char *str, int digits, int *val)
     cp = str;
     end = str + digits;
     while (cp < end) {
-        if (!isdigit(*cp)) {
+        if (!isdigit((int) *cp)) {
             return (NULL);
         }
         *val *= 10;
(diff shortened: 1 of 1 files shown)


neelnatu:
The change looks good but the warning surprised me so I googled around.

I found the following email thread that explains it very well:
https://sourceware.org/ml/newlib/2010/msg00245.html

> warning: array subscript has type 'char'
> 
> Most, if not all, of these appear to be from
> using ctype.h methods which is caused by
> a missing (int) case in line 57 in ctype.h:
> 
> #define __ctype_lookup(__c) ((__ctype_ptr__+sizeof(""[(__c]))[(int)(__c)])
> 
> Changing this to the following resolved the
> warnings.
> 
> #define __ctype_lookup(__c)
> ((__ctype_ptr__+sizeof(""[(int)__c]))[(int)(__c)])
> 
> This is with newlib 1.18.0 with the current RTEMS
> patch.
> 
> Does this cast need to be added?

Absolutely not.  The warning is _intentional_.  You have a bug in your
code - you are calling the ctype methods with a char argument, but char
is signed, and an 8-bit char is out of range and causes unspecified
results according to the standards.  In particular, that warning was
added because we had a demonstration of where isspace(255u) and
isspace(EOF) intentionally give different results in some locales, but
where the calling app had been doing isspace((char)255) and going into
an infinite loop because of their bug.  You should fix your code to use
unsigned char or int, as appropriate, in the first place.

PR #11, opened 2016-03-05
Support building 'sim' with optimization levels higher than -O0

neelnatu:
0796ef213444be48b59d31e378adca2efc61a3dc:
Support building the 'sim' with optimization levels higher than -O0. The
higher optimization levels emit useful warnings that are masked at -O0.

The basic problem with the implementation of 'os_arch_task_stack_init()'
is that we are switching stacks from underneath the compiler without any
ability to influence how the compiler uses the stack. Thus when setjmp()
returns the compiler doesn't know that the stack frame it allocated for
local variables doesn't exist anymore on the new stack.

This happens to work with the -O0 optimization level because the compiler
addresses local variables using the frame pointer (%ebp). Higher optimization
levels use offsets from the stack pointer and immediately expose the issue
mentioned above.

Fix this by switching stacks in 'os_arch_frame_init()' which is implemented
in assembly so we can control the use of stack after setjmp() returns.

fb3520e2ca4aadc0d47928a0a54ebcd54a433f6e:
Don't drop the critical section between selecting the task to run and
actually switching to it. This creates a window where an interrupt might
make a higher priority task runnable but won't switch to it immediately.

### DIFF
--- a/compiler/sim/pkg.yml
+++ b/compiler/sim/pkg.yml
@@ -29,15 +29,16 @@ pkg.keywords:
     - gcc
 
 compiler.path.cc: "/usr/local/bin/gcc-5" 
+compiler.path.as: "/usr/local/bin/gcc-5 -x assembler-with-cpp"
 compiler.path.archive: "ar"
 compiler.path.objdump: "gobjdump"
 compiler.path.objsize: "objsize"
 compiler.path.objcopy: "gobjcopy"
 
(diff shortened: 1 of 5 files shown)


wes3:
Merged the pull request. Code seemed fine but I admit I did not go through all the assembly code entirely to make sure it works.

neelnatu:
Will merged the pull request.

PR #15, opened 2016-03-08
a library to control the HC-SR04 ultrasonic distance sensor

paulfdietrich:
a small library to control an ultrasonic range finder.

### DIFF
--- a/libs/HCSR04/include/hcsr04.h
+++ b/libs/HCSR04/include/hcsr04.h
@@ -1,0 +1,50 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


mkiiskila:
Hi Paul, do you feel that the ultrasonic driver should be part of the core OS?
I wonder if the Arduino repository would be a better place to keep this.

Also, setting identities in pkg.yml for this library does not seem right. What's the goal with that setting?

ccollins476ad:
I think this library does need to stay in core because we have a tutorial in the mynewt docs which relies on it (http://mynewt.apache.org/os/tutorials/air_quality_sensor/)

I do agree that the PERIPHERAL feature in libs/HCSR04/pkg.yml seems to serve no purpose, and it should probably be removed.

paulfdietrich:
I will close this as I will put it into a remote repository. I think the tutorial is on an air sensor and not a distance sensor.

sterlinghughes:
Cool, that's great.  Please add a pointer to it when you do so!

PR #16, opened 2016-03-11
Fix building and running 'sim' on Linux

neelnatu:


mkiiskila:
Looks good. Only thing I'm ambivalent about is the task stack increases for task1 and task2. I would tone those down to maybe 256 dwords each. What I'm worried about is the RAM use when running in low memory platforms.

ccollins476ad:
Can we just define the stack sizes conditionally on the architecture?  E.g.,

```
#ifdef ARCH_sim
#define TASK1_STACK_SIZE    OS_STACK_ALIGN(1024)
#else
#define TASK1_STACK_SIZE    OS_STACK_ALIGN(128)
#endif
```

By the way, this merge works with the latest newt and core/develop.  I just had to make a few minor changes to the testutil library to ensure the sigset gets properly initialized in case the OS never gets started (some unit tests don't actually use the OS, but they use some libs/os functions, e.g., os_memblock_get()).

neelnatu:
How about we go a step further and do it in the sim-specific
OS_STACK_ALIGN() by:

- adding a fixed amount to all stacks (say 4096 bytes) or

- allocating a multiple of the requested size (say __nmemb * 8)

Thoughts?

best
Neel

neelnatu:
I reverted the stack size changes and instead scaled them up in
arch-specific code.

If this looks ok then I can update the pull request.

best
Neel

diff --git a/libs/os/include/os/arch/sim/os/os_arch.h
b/libs/os/include/os/arch/sim/os/os_arch.h

index cfe737f..59096f8 100644

--- a/libs/os/include/os/arch/sim/os/os_arch.h

+++ b/libs/os/include/os/arch/sim/os/os_arch.h

@@ -40,8 +40,20 @@ typedef unsigned int os_stack_t;

 #define OS_SANITY_STACK_SIZE (1024)

 #define OS_IDLE_STACK_SIZE (1024)


+/*

+ * The 'sim' architecture-specific code does not have as much control on

+ * stack usage as the real embedded architectures.

+ *

+ * For e.g. the space occupied by the signal handler frame on the task

+ * stack is entirely dependent on the host OS.

+ *

+ * Deal with this by scaling the stack size by a factor of 16. The scaling

+ * factor can be arbitrarily large because the stacks are allocated from

+ * BSS and thus don't add to either the executable size or resident

+ * memory.

+ */

 #define OS_STACK_ALIGN(__nmemb) \

-    (OS_ALIGN((__nmemb), OS_STACK_ALIGNMENT))

+    (OS_ALIGN(((__nmemb) * 16), OS_STACK_ALIGNMENT))


 /* Enter a critical section, save processor state, and block interrupts */

 #define OS_ENTER_CRITICAL(__os_sr) (__os_sr = os_arch_save_sr())

diff --git a/project/slinky/src/main.c b/project/slinky/src/main.c

index 4534097..edaad80 100755

--- a/project/slinky/src/main.c

+++ b/project/slinky/src/main.c

@@ -45,24 +45,24 @@ int init_tasks(void);


 /* Task 1 */

 #define TASK1_PRIO (1)

-#define TASK1_STACK_SIZE    OS_STACK_ALIGN(1024)

+#define TASK1_STACK_SIZE    OS_STACK_ALIGN(128)

 struct os_task task1;

 os_stack_t stack1[TASK1_STACK_SIZE];

 static volatile int g_task1_loops;


 /* Task 2 */

 #define TASK2_PRIO (2)

-#define TASK2_STACK_SIZE    OS_STACK_ALIGN(1024)

+#define TASK2_STACK_SIZE    OS_STACK_ALIGN(128)

 struct os_task task2;

 os_stack_t stack2[TASK2_STACK_SIZE];


 #define SHELL_TASK_PRIO (3)

 #define SHELL_MAX_INPUT_LEN     (256)

-#define SHELL_TASK_STACK_SIZE (OS_STACK_ALIGN(1024))

+#define SHELL_TASK_STACK_SIZE (OS_STACK_ALIGN(384))

 os_stack_t shell_stack[SHELL_TASK_STACK_SIZE];


 #define NEWTMGR_TASK_PRIO (4)

-#define NEWTMGR_TASK_STACK_SIZE (OS_STACK_ALIGN(1024))

+#define NEWTMGR_TASK_STACK_SIZE (OS_STACK_ALIGN(512))

 os_stack_t newtmgr_stack[NEWTMGR_TASK_STACK_SIZE];


 struct log_handler log_console_handler;

neelnatu:
I have updated the pull request to revert the stack size changes.

https://github.com/apache/incubator-mynewt-core/pull/16

best
Neel

PR #22, opened 2016-03-22
Hal pwm

paulfdietrich:


paulfdietrich:
I just updated this to include the same model as the hal_adc

paulfdietrich:
I submitted an alternate with a better (IMO) API

PR #25, opened 2016-03-25
After reviews of the three hal ADC prototypes, this is what I think i…

paulfdietrich:
Thanks for all the help everyone.  I do like this implementation the best.

Some special things to note about the HAL:

1) For ADC, I made the hal_device_s not constant.  This allows drivers to wrap this with state.  My implementations take up 1 or 2 words per device depending on whether the have state.
2) There is still a system ID and mcu ID.  I made special include files for them.  We can't do the simple thnk that MBED does since they use PINs and we wanted to support external devices in our HAL
3) Unlike the other drivers though, all the glue for the device factory lives in one function in the BSP.
4) Although the systemID are not necessarily iterable, in an application you can place their pointers in an array and iterate them nicely.
5) The MCU Device IDs can be specified differently for each MCU and need not be simple numbers (could be pins for the MBED hal for example)


hw/bsp/native/include/bsp/bsp_sysid.h
```diff
@@ -0,0 +1,44 @@
...
+
+/* This file defines the system device descriptors for this BSP.
+ * System device descriptors are any HAL devies */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* this is a native build and these are just simulated */
+enum SystemDeviceDescriptor {
```
sterlinghughes: This should be enum system_device_descriptor re coding conventions. (they are unfortunately not documented yet.)
enum system_dev_descriptor might even be better :-)

hw/bsp/native/src/hal_bsp.c
```diff
@@ -35,3 +39,26 @@ bsp_flash_dev(uint8_t id)
     }
     return &native_flash_dev;
 }
+
+
+/* This is the factory that creates adc devices for the application.
+ * Based on the system ID defined, this maps to a creation function for
+ * the specific underlying driver */
+extern struct hal_adc_device_s *
+bsp_get_hal_adc_device(enum SystemDeviceDescriptor sysid) {
```
sterlinghughes: { goes on the line after the function declaration.

hw/hal/include/hal/hal_adc.h
```diff
@@ -0,0 +1,60 @@
...
+#define HAL_ADC_H
+
+/* for the pin descriptor enum */
+#include <bsp/bsp_sysid.h>    
+
+/* This is the device for an ADC. The application using the ADC device
+  *does not need to know the definition of this device and can operate 
+ * with a pointer to this device.  you can get/build device pointers in the 
+ * BSP */
+struct hal_adc_device_s;
```
sterlinghughes: struct names shouldn't be postfixed with the type (i.e. _s).

hw/mcu/native/include/mcu/mcu_devid.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef MCU_DEVID_H
+#define MCU_DEVID_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* this is a native build and these pins are not real.  They are used only
+ * for simulated HAL devices */
+enum McuDeviceDescriptor {
```
sterlinghughes: mcu_device_descriptor

hw/mcu/native/src/hal_adc.c
```diff
@@ -0,0 +1,315 @@
...
+};
+
+struct hal_adc_device_s*
+native_adc_file_create(enum McuDeviceDescriptor devid, char *fname) 
+{
+    struct file_adc_device_s *padc = NULL;
+    if(devid == MCU_ADC_CHANNEL_5) {
+        padc = malloc(sizeof(struct file_adc_device_s));
+        
+        if(padc) {
```
sterlinghughes: if ( not if(

hw/mcu/native/src/hal_adc.c
```diff
@@ -0,0 +1,315 @@
...
+            fclose(pfileadc->native_fs);
+        }      
+    }
+    
+    return -1;    
+}
+
+
+
+#if 0
```
sterlinghughes: obviously, cull out the #if 0 code prior to commit.

hw/mcu/native/src/hal_adc.c
```diff
@@ -0,0 +1,315 @@
...
+struct mmm_adc_device_s {
+    struct hal_adc_device_s parent;
+    enum McuDeviceDescriptor devid;   
+};
+
+struct hal_adc_device_s*
+native_adc_mmm_create(enum McuDeviceDescriptor devid) 
+{
+    struct mmm_adc_device_s *padc = NULL;
+    if(        (devid == MCU_ADC_CHANNEL_2)
```
sterlinghughes: standard is if ((devid == MCU_ADC_CHANNEL_2) ||
                      (devid == MCU_ADC_CHANNEL_3) ||
matching the inner parentheses with the start of the line, but including the || statement at the end of the line.

mkiiskila:
No architecture besides native has bsp_sysid.h present in this pull request, so nothing except native targets would build after merging this in. Could you add stubs for all BSPs so that those platforms would still be buildable?

PR #30, opened 2016-04-01
correction of the the failure of the lib/json test.

NgesBrian:
This is to correct the the error encountered when testing  the libs/json.
this lib/json is use for data transfer.


mkiiskila:
Good catch. However, this was fixed in develop branch with commit id e8a365bd766ac2963ca9c757175889caf5f12461.

The way we work in this code base is by having a develop branch, where most of the activity takes place. And this branch is then merged in to 'master' just before release. This allows us to have a relatively stable master branch, while still working on new features/bug fixes together on develop.

PR #34, opened 2016-04-07
Implement tickless mode during idle for nRF51 and nRF52 microcontrollers

neelnatu:
Implement tickless mode during idle for nRF51 and nRF52 microcontrollers.

When the idle task executes it gets from the OS the number of ticks until
there is useful work to be done (e.g. a callout handler needs to execute).
It then calls 'os_ticks_idle(idleticks)' which in turn idles the CPU for a
duration that does not exceed 'idleticks'.

'os_ticks_idle()' disables the periodic timer interrupt to enter the
tickless mode during idle. It sets up a one-shot timer to wake up the
CPU after a duration equivalent to 'idleticks' has elapsed. Finally,
it enters the low power idle state using a CPU-specific mechanism.
For e.g. ARM processors use the 'WFI' instruction for this purpose.

Add HAL APIs for the periodic OS timer:
- os_tick_init() is used to initialize the periodic timer.
- os_tick_idle() is used to enter tickless mode and halt the CPU when idle.

A related change is to define 'OS_TICKS_PER_SEC' in MCU-specific header file.
This is needed because the capabilities of the underlying periodic timer
dictate the appropriate timer frequency.

Increase the idle task's stack size to account for the increase in function
call depth due to os_tick_idle().

Add APIs 'os_callout_wakeup_ticks()' and 'os_sched_wakeup_ticks()' that
return the number of ticks until which the callout and sched subsystems
do not have useful work. In other words these APIs indicate for how long
the OS can safely idle.

Implement context switching in 'sim' in a similar manner to how it is done
on ARM Cortex processors. More information in commit http://bit.ly/1MWxuuG


libs/os/src/arch/sim/os_arch_sim.c
```diff
@@ -179,18 +196,53 @@ os_arch_in_critical(void)
...
+
+    OS_ASSERT_CRITICAL();
+
+    if (ticks > 0) {
+        /*
+         * Enter tickless regime and set the timer to fire after 'ticks'
+         * worth of time has elapsed.
+         */
+        it.it_value.tv_sec = ticks / OS_TICKS_PER_SEC;
+        it.it_value.tv_usec = (ticks % OS_TICKS_PER_SEC) * OS_USEC_PER_TICK;
```
mkiiskila: I would do this (ticks * OS_USEC_PER_TICK) % OS_TICKS_PER_SEC. But this change is not a must.

mkiiskila: Sorry, misread that code. I was reading % as /. Ignore my previous comment.

PR #40, opened 2016-04-13
Log changes

vrahane:
- Adding/Changing fields(msg, ts, entries) to log entries
- Fixing json encode for 64 bit data types
- Adding os_time function to get uptime in microsenconds
- Changing json responses to contain multiple log entries

### DIFF
--- a/libs/baselibc/src/tinyprintf.c
+++ b/libs/baselibc/src/tinyprintf.c
@@ -103,15 +103,15 @@ static void li2a(long num, struct param *p)
 }
 #endif
 
-static void ui2a(unsigned int num, struct param *p)
+static void ui2a(unsigned long long int num, struct param *p)
 {
     int n = 0;
-    unsigned int d = 1;
+    unsigned long long int d = 1;
     char *bf = p->bf;
     while (num / d >= p->base)
         d *= p->base;
     while (d != 0) {
-        int dgt = num / d;
+        unsigned long long  dgt = num / d;
         num %= d;
         d /= p->base;
         if (n || dgt > 0 || d == 0) {
@@ -122,7 +122,7 @@ static void ui2a(unsigned int num, struct param *p)
     *bf = 0;
 }
 
-static void i2a(int num, struct param *p)
+static void i2a(long long int num, struct param *p)
 {
     if (num < 0) {
         num = -num;
@@ -282,7 +282,7 @@ size_t tfp_format(FILE *putp, const char *fmt, va_list va)
                     uli2a(va_arg(va, unsigned long int), &p);
                 else
 #endif
-                    ui2a(va_arg(va, unsigned int), &p);
+                    ui2a(va_arg(va, unsigned long long int), &p);
                 written += putchw(putp, &p);
                 break;
             case 'd':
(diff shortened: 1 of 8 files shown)


libs/os/src/os_time.c
```diff
@@ -141,3 +141,19 @@ os_gettimeofday(struct os_timeval *tv, struct os_timezone *tz)
 
     return (0);
 }
+
+int64_t
+os_get_uptime_usec(void) {
```
sterlinghughes: the { should be on the next line.

libs/os/src/os_time.c
```diff
@@ -141,3 +141,19 @@ os_gettimeofday(struct os_timeval *tv, struct os_timezone *tz)
...
+  struct os_timeval tv;
+  os_time_t delta;
+  os_sr_t sr;
+
+  OS_ENTER_CRITICAL(sr);
+
+  delta = os_time_get() - basetod.ostime;
+  os_deltatime(delta, &basetod.uptime, &tv);
```
sterlinghughes: This is a long period to disable a critical section.  can you just grab basetod.os_time, then do the calculation?

sys/log/include/log/log.h
```diff
@@ -51,8 +51,9 @@ struct log_handler {
 
 struct log_entry_hdr {
     int64_t ue_ts;
-    uint16_t ue_level;
     uint16_t ue_module;
+    uint16_t ue_index;
+    uint8_t ue_level;
```
sterlinghughes: This is unaligning the structure, please align the structure.  you can add a ue_arg0, or unused field if you want to, but the behavior should be defined here.

libs/baselibc/src/tinyprintf.c
```diff
@@ -103,15 +103,15 @@ static void li2a(long num, struct param *p)
 }
```
sterlinghughes: In general, please remove whitespace changes from patch submissions.  Once you get the commit bit, you can make those changes, but its hard to read through the patches and distinguish what is WS and what isn't -- and those WS changes are probably not that critical.

sys/log/src/log_nmgr.c
```diff
@@ -76,17 +76,30 @@ log_nmgr_add_entry(struct log *log, void *arg, void *dptr, uint16_t len)
...
 
+    JSON_VALUE_UINT(&jv, ueh.ue_level);
+    rc = json_encode_object_entry(encoder, "level", &jv);
+    if (rc != 0) {
+        goto err;
+    }
+
+    /* Dummy index value for now */
+    JSON_VALUE_UINT(&jv, ueh.ue_index);
+    rc = json_encode_object_entry(encoder, "index", &jv);
```
sterlinghughes: Why is the index in the log, if its only a dummy index value?

libs/os/src/os_time.c
vrahane: Will fix this.

libs/os/src/os_time.c
vrahane: Will fix this. I can move os_deltatime() outside the critical section.

sys/log/include/log/log.h
vrahane: Fix is to have a log_level and index be a byte each. The problem is also that the alignment is going to be 8 bytes and so the sizeof the structure is going to be 16 bytes. I am going to pack it sop that it uses less space as this is going to go on the flash.

libs/baselibc/src/tinyprintf.c
vrahane: Point taken.

sys/log/src/log_nmgr.c
vrahane: It is for the ordering for events that happen in a millisecond. I am changing it to be a byte as per the discussion and suggestion.

vrahane:
Changing as per review comments.

PR #41, opened 2016-04-14
Log changes

vrahane:
- Adding/Changing fields(msg, ts, entries) to log entries
- Fixing json encode for 64 bit data types
- Adding os_time function to get uptime in microsenconds
- Changing json responses to contain multiple log entries
(Also addressing comments on github)

### DIFF
--- a/sys/log/src/log_nmgr.c
+++ b/sys/log/src/log_nmgr.c
@@ -76,13 +76,26 @@ log_nmgr_add_entry(struct log *log, void *arg, void *dptr, uint16_t len)
     json_encode_object_start(encoder);
 
     JSON_VALUE_STRINGN(&jv, data, rc);
-    rc = json_encode_object_entry(encoder, "log", &jv);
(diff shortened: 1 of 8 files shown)


sys/log/src/log_nmgr.c
```diff
@@ -76,17 +76,30 @@ log_nmgr_add_entry(struct log *log, void *arg, void *dptr, uint16_t len)
...
         goto err;
     }
 
+    JSON_VALUE_UINT(&jv, ueh.ue_level);
+    rc = json_encode_object_entry(encoder, "level", &jv);
+    if (rc != 0) {
+        goto err;
+    }
+
+    /* Dummy index value for now */
```
sterlinghughes: remove this comment.  we can still commit this PR, but please correct in next PR.

vrahane: Yes. I have to change something else. So, I can remove it.

PR #44, opened 2016-04-19
I2C interface supporting blocking read and write

paulfdietrich:
An interface to support blocking I2c read and write.  This allows for read transactions, write transactions, and combined read/write (Like random access eeprom)


paulfdietrich:
I pushed this change to develop. If there are issues, let me know.

PR #47, opened 2016-05-06
this is the task given in the mail thread

NgesBrian:
I doubt know whether I have done it correctly but I will be very happy to receive corrections .
thanks

### DIFF
--- a/net/nimble/host/include/host/ble_gap.h
+++ b/net/nimble/host/include/host/ble_gap.h
@@ -198,5 +198,6 @@ int ble_gap_wl_set(struct ble_gap_white_entry *white_list,
                    uint8_t white_list_count, ble_gap_wl_fn *cb, void *cb_arg);
 int ble_gap_update_params(uint16_t conn_handle,
                           struct ble_gap_upd_params *params);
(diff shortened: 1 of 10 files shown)


libs/json/src/test/test_json_simple.c
```diff
@@ -205,4 +209,32 @@ TEST_CASE(test_json_simple_decode){
 
     rc = memcmp(string2, "foobarlongstring", 10);
     TEST_ASSERT(rc==0);
-}
+        
+        //second stuffs to attr
```
sterlinghughes: Hi nges, according to Coding style, C++ style comments (//) are not permitted.  Please use C style comments /* */

libs/json/src/test/test_json_simple.c
```diff
@@ -205,4 +209,32 @@ TEST_CASE(test_json_simple_decode){
...
+            .type = t_uinteger,
+            .addr.uinteger = &uint_val2,
+            .nodefault = true
+        },
+        
+         
+      };
+     
+     
+     //writing and testing
```
sterlinghughes: See above.

PR #48, opened 2016-05-10
Patches for exposing HCI over UART

jhedberg:
These patches add an Arduino 101 BSP as well as a simple app to expose HCI over UART. With the app one can use the native Bluetooth LE host stack in Zephyr on the Arduino 101, with MyNewt running on the nRF51 and Zephyr on the Quark SE.

### DIFF
--- a/apps/blehci/pkg.yml
+++ b/apps/blehci/pkg.yml
@@ -1,0 +1,29 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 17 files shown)


wes3:
A comment regarding the nrf51822 on the 101. I was trying to find the part number of the chip and was not having any luck. I presume that this is the 256K flash/16KB RAM variant? If this is the 128KB flash variant the linker scripts will need to be changed as they are assuming a 256KB flash.

One other issue is that the nrf51dk-16kbram_no_boot linker script has an incorrect RAM size allocated (it is allocating 32KB as opposed to 16KB). This was a mynewt error for existing the existing nrf51 bootloader. I will fix it there.

I am going to merge the pull request and we can address these issues in later commits if needed.

PR #55, opened 2016-05-17
Mynewt add

vrahane:


sys/log/include/log/log.h
```diff
@@ -24,6 +24,12 @@
 
 #include <os/queue.h>
 
+/* Global log info */
+struct log_info {
+    int64_t li_timestamp;
+    uint8_t li_index;
+}g_log_info;
```
sterlinghughes: space here.

sys/log/include/log/log.h
```diff
@@ -72,11 +86,22 @@ struct log_entry_hdr {
...
+#define LOG_MODULE_MAX              (255)
+
+#define LOG_MODULE_STR(module) \
+    (LOG_MODULE_DEFAULT     == module ? "DEFAULT"     :\
+    (LOG_MODULE_OS          == module ? "OS"          :\
+    (LOG_MODULE_NEWTMGR     == module ? "NEWTMGR"     :\
+    (LOG_MODULE_NIMBLE_CTLR == module ? "NIMBLE_CTLR" :\
+    (LOG_MODULE_NIMBLE_HOST == module ? "NIMBLE_HOST" :\
+    (LOG_MODULE_NFFS        == module ? "NFFS"        :\
+     "UNKNOWN"))))))
```
sterlinghughes: Why don't you just use a macro to string expansion on "module", which will give you the name of the define.  Also, if the condition checking remains, this should be a function, not a macro.

sys/log/src/log_nmgr.c
```diff
@@ -71,96 +101,317 @@ log_nmgr_add_entry(struct log *log, void *arg, void *dptr, uint16_t len)
...
 }
 
+/**
+ * Log encode entries
+ * @param log structure, the encoder, timestamp, index
+ * @return 0 on success; non-zero on failure
+ */
+static int
+log_encode_entries (struct log *log, struct json_encoder *encoder,
+                    int64_t ts, uint32_t index)
```
sterlinghughes: No space before the (

sys/log/src/log_nmgr.c
```diff
@@ -180,7 +431,7 @@ log_nmgr_clear(struct nmgr_jbuf *njb)
         }
 
         rc = log_flush(log);
-        if (rc != 0) {
+        if (rc) {
```
sterlinghughes: Why the change here -- same use: the majority of the code is written the way it was previously checked.

sys/log/src/log_nmgr.c
```diff
@@ -208,7 +460,7 @@ log_nmgr_register_group(void)
     log_nmgr_group.ng_group_id = NMGR_GROUP_ID_LOGS;
 
     rc = nmgr_group_register(&log_nmgr_group);
-    if (rc != 0) {
+    if (rc) {
```
sterlinghughes: ibid

sys/log/src/log.c
```diff
@@ -103,20 +103,27 @@ log_append(struct log *log, uint16_t module, uint16_t level, void *data,
...
-
     ue = (struct log_entry_hdr *) data;
 
+    /* Resetting index every millisecond */
+    if (g_log_info.li_timestamp > ue->ue_ts + 1000) {
+        g_log_info.li_index = 0;
+    }
+
+    g_log_info.li_index++;
+
```
sterlinghughes: Why is this not stored on the log itself?  Presumably this would be per-log.

sys/log/include/log/log.h
vrahane: - Macro to string expansion(Stringification) converts the define(constant) into string literal. It does not really give the name of the define. More here: https://gcc.gnu.org/onlinedocs/cpp/Stringification.html

The MACRO is close to the actual definitions of the constants which makes it easier to edit in the future if new modules or levels are added. This way someone doesn't have to go searching for the function to add one more string. C Source file remains intact, only the header file changes.

sys/log/src/log_nmgr.c
vrahane: Yes, I agree but I prefer keeping it clean. Doesn't hurt and I am not violating the coding standards :-)

sys/log/src/log.c
vrahane: The index is now now treated as a secondary key instead of a primary key for the logs. Each log doesn't have it's own index anymore. It is a global index. Spoke to Will, Chris and Paul. The reason for making it a secondary key is that the index would roll over only after 1 millisecond elapses since the last timestamp. If you see multiple logs from different modules, you can actually relate the events looking at the index for the same timestamp. Also, the belief is we would not ever see more than 255 events being logged in a millisecond. Hope this makes sense.

PR #64, opened 2016-06-20
Cache lastblock ID for each file in on-flash data structure

peterfs:
Change on-flash data structures including the addition of ndi_lastblock_id
in the nffs_disk_inode struct.  This field will store the last block of each
file for fast restores. That change is not enabled in this set of changes,
it will come in a subsequent commit.
Other changes include:
* Deleting the per inode/block magic number (ndi_magic, ndb_magic)
* Modifying the code to read in disk objects during restore and in ffs2native.
* Incrementing the filesystem version number stored in struct nffs_disk-area
* Add "-s" flag to ffs2native to parse NFFS flash area images (V1 format for now)

### DIFF
--- a/apps/ffs2native/src/main.c
+++ b/apps/ffs2native/src/main.c
@@ -40,8 +40,18 @@
 #include <mcu/mcu_sim.h>
 #endif
 
+#include <stdlib.h>
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+
+static struct log_handler nffs_log_console_handler;
+struct log nffs_log;
 static const char *copy_in_dir;
 static const char *progname;
+
+char *file_flash_area;
+int file_scratch_idx;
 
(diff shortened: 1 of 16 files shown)


ccollins476ad:
It looks great. Thanks for fixing the broken CRC offset!  I just have an annoying comment to make.  Some of the changes don't conform to the CODING_STANDARDS.md document.  In particular, there are some tabs instead of spaces, and a few single-line if statements that lack curly braces.  If merge the request if you can take care of those issues.

peterfs:
I've changed the tabs to spaces and fixed the single-line if statements.

I've added two more commits to this pull request - the first implements the lastblock support, removing the "find_file_ends" way to restore NFFS flash areas. The second commit adds support for V0 NFFS layouts to the ffs2native utility.

peterfs:
Added fix for ffs2native for reading in V0 objects.

PR #70, opened 2016-06-30
adding more test to os_eventq_poll

NgesBrian:


mkiiskila:
It would be better to derive time quantities from OS_TICKS_PER_SEC when calling os_time_delay(), or other os functions which take the number of ticks as a parameter. At the moment the ticks per second is 1000 for simulator, but there are reasons why we might want to adjust this.
If it gets adjusted, all the time delays in the code will change. I.e. wait of half a second might turn into a wait of 5 seconds. And that kind of a wait will not be acceptable within unit tests.

That's something to keep in mind for future.

PR #71, opened 2016-06-30
Shell/Console prompt

davidgs:
Prints ‘> ‘ as a prompt for the shell/console


libs/shell/src/shell_priv.h
```diff
@@ -23,5 +23,5 @@
 int shell_os_tasks_display_cmd(int argc, char **argv);
 int shell_os_mpool_display_cmd(int argc, char **argv);
 int shell_os_date_cmd(int argc, char **argv);
-
+void console_print_prompt();
```
mkiiskila: This declaration should not be needed, as function prototype is in console.h.

libs/console/full/include/console/console.h
```diff
@@ -29,7 +29,7 @@ void console_write(const char *str, int cnt);
 int console_read(char *str, int cnt);
 void console_blocking_mode(void);
 void console_echo(int on);
-
+void console_print_prompt(void);
```
mkiiskila: We should keep the API in the other console package exactly the same as here. I.e. add a matching stub to console/stub/include/console.h.

libs/console/full/src/cons_tty.c
```diff
@@ -179,6 +179,15 @@ console_file_write(void *arg, const char *str, size_t cnt)
 }
 
 void
+console_print_prompt()
```
mkiiskila: My coding style comment. I would add 'void' inside function declaration when there are no arguments.

mkiiskila:
Good, I want this functionality.

Here's what it looks like in use:
----8<-----cut-start------
> ?                                                                             
552872:   config       log        ls        rm     mkdir        mv              
552878:      cat      echo         ?     tasks  mempools      date              
552884:     stat     flash                                                      
>                                                                               
                                                                                
?                                                                               
554707:   config       log        ls        rm     mkdir        mv              
554712:      cat      echo         ?     tasks  mempools      date              
554719:     stat     flash                                                      --
---8<-----cut-end----

Can we have prompt appear even when there is no incoming command?
Also, I'd prefer if the prompt had a timestamp in front (like the output lines).

davidgs:
I have been trying to figure out why the prompt doesn't appear when no character is entered. Something gobbles the \r somewhere. 

I'd actually like to see the prompt a bit more configurable so I may look at enable/disable time stamp and changing the prompt character. 

Best regards,
dg
--
I'll take credit For the funny typos, the rest I blame on the iPhone spell wrecker.

PR #75, opened 2016-07-17
att: use offset when preparing Read Blob Response

mikeryan:
This remediates a bug wherein the offset of a Read Blob Request (from
the central) is ignored and the response always includes first MTU-1
bytes of the attribute value regardless of offset requested. As a
result, most hosts sit in an infinite loop incrementing the offset but
continually receiving the first MTU-1 bytes.

Note that this code has not been carefully tested for boundary
conditions.

### DIFF
--- a/net/nimble/host/src/ble_att_svr.c
+++ b/net/nimble/host/src/ble_att_svr.c
@@ -1589,7 +1589,8 @@ ble_att_svr_rx_read_blob(uint16_t conn_handle, struct os_mbuf **rxom)
         goto done;
     }
 
-    rc = ble_att_svr_build_read_blob_rsp(ctxt.attr_data, ctxt.data_len, mtu,
+    rc = ble_att_svr_build_read_blob_rsp(ctxt.attr_data + ctxt_offset,
+                                         ctxt.data_len - ctxt_offset, mtu,
                                          &txom, &att_err);
     if (rc != 0) {
         err_handle = req.babq_handle;


mikeryan:
This affects the devel branch as well.

ccollins476ad:
Hi Mike,

Thanks for the pull request!  I think this is a case of ambiguity caused by a lack of documentation.  The expectation is that the characteristic access callback looks at the offset and takes it into account.  This is how the ble_att_svr_test_read_blob unit test behaves (net/nimble/host/src/test/ble_att_svr_test.c).

I think there is a good practical reason for delegating the offset processing to the application callback.  The stack provides a fixed length buffer that the callback can used to dynamically generate the attribute value if it needs it.  I don't think the application should be required to generate the entire attribute if the stack will only be reading a portion of it.  Moreover, the buffer size is based on the maximum ATT MTU that the stack was built with, so it may be impossible to fit the entire attribute in the buffer.

mikeryan:
I can provide Bluetooth logs from a known good certified Bluetooth stack (BlueZ 5 on Linux) having issues reading a long attribute using a Read Blob Request if that helps you understand the specific problem I'm solving here. I suspect that particular test case is inadequate to tickle the bug.

ccollins476ad:
I think I would need to see the GATT characteristic access callback.  In particular, does it take the read offset into account when generating the response?  If it assumes an offset of 0, I can see how it would result in the described behavior.

mikeryan:
The final access callback does not take into account the offset, and indeed does not receive the offset.

The read blob request originally comes in here:

https://github.com/apache/incubator-mynewt-core/blob/master/net/nimble/host/src/ble_att_svr.c#L1541

The offset gets shoved into the ```ble_att_svr_access_ctxt```, and the request is routed through ```ble_att_svr_read```:

https://github.com/apache/incubator-mynewt-core/blob/master/net/nimble/host/src/ble_att_svr.c#L284

The request hits an internal callback, in this case ```ble_gatts_chr_val_access``` since we're accessing a characteristic value:

https://github.com/apache/incubator-mynewt-core/blob/28e2a8b4efcda9d241720f25749137de37ad71e8/net/nimble/host/src/ble_gatts.c#L295

Note that up to and including this point the offset hasn't been touched. In this handler the ```ble_gatt_access_ctxt``` that is passed to the user-defined callback is created, and it does not include the offset. It doesn't even have a field for it:

https://github.com/apache/incubator-mynewt-core/blob/28e2a8b4efcda9d241720f25749137de37ad71e8/net/nimble/host/include/host/ble_gatt.h#L195

The data returned from the user-defined callback, which is the entire data of the characteristic, is set in the ```ble_att_svr_access_ctxt``` passed in from the stack-internal parent function (not the ble_gatt_access_ctxt sent to the user callback).

As you can see, there is no opportunity for the user-defined callback to do anything with this, and based on the internal design of the stack it appears that it's expected that the stack will internally deal with offsets, MTUs, and so forth.

The code I used to trigger this is based on bleprph. You can reproduce it by changing one of the characteristics to have a very long value and then attempt to read blob with a non zero offset. For example, change ```gatt_svr_nimble_test_read_val``` to be an array of 64 bytes and use an ordinary stack to read the value. The reading stack will likely get stuck in an infinite loop.

I can produce code that will trigger this issue and include BlueZ packet logs in a few hours if the above doesn't convince you.

ccollins476ad:
OK, I see what you mean.  Thanks for being patient with me.  It appears this was indeed an issue in earlier versions of the code, but has since been fixed in the develop branch.  The latest version of ble_gatt_access_ctxt contains an instance of the ble_att_svr_access_ctxt, as shown here: https://github.com/apache/incubator-mynewt-core/blob/develop/net/nimble/host/include/host/ble_gatt.h

I have been immersed in the develop branch, so I didn't apply the appropriate context to your comments.  I should have looked at the current release before replying!

I do believe the implementation in develop is the correct solution for the reason I mentioned in my first comment.  Would you agree that making the callback deal with the offset is a suitable solution?

mikeryan:
No worries re: patience. It's a good instinct to assume user error rather than bugs in the stack :)

It's unclear to me that the user-defined access handler should have to or want to care about this sort of thing. One benefit of the implementation in this PR is that the user's GATT code can be extremely stupid, letting the stack do the heavy lifting. It doesn't have to have any concept of MTU, bounds checking, and other brittle areas where bugs are likely to appear.

The only scenario in which this approach seems non-beneficial is if the user's GATT code performs an expensive calculation on each access to the characteristic.

ccollins476ad:
I've been giving it some thought, and I am pretty sure I agree with you - the memory savings and gains in efficiency are not worth the added cost of a more complex interface.

I'll try to come up with a fair pro / con comparison and send a note to the dev list.

Thanks for the heads up.  If the urge strikes you to find any more bugs, please knock yourself out!

PR #79, opened 2016-07-25
ATT: send a Read By Group Type Response in the case of BLE_HS_EMSGSIZE

mikeryan:
Read By Group Type Response messages can raise BLE_HS_EMSGSIZE when the
response is large and the MTU is small. In this case the message built
by ble_att_svr_build_read_group_type_rsp() is valid and complete
(despite the error-looking return value), and should be sent to the
remote party.

### DIFF
--- a/net/nimble/host/src/ble_att_svr.c
+++ b/net/nimble/host/src/ble_att_svr.c
@@ -2071,7 +2071,7 @@ ble_att_svr_rx_read_group_type(uint16_t conn_handle, struct os_mbuf **rxom)
 
     rc = ble_att_svr_build_read_group_type_rsp(conn_handle, &req, uuid128,
                                                &txom, &att_err, &err_handle);
-    if (rc != 0) {
+    if (rc != 0 && rc != BLE_HS_EMSGSIZE) {
         goto done;
     }
 


mikeryan:
This bug can be easily triggered by creating four primary services with 16-bit UUIDs. Connect to the BLE peripheral using a tool with an MTU of 23, 24, or 25, and issue a Primary services request (by issuing a Read By Group Type Request for UUID 2800).

Linux's gatttool in its default configuration will trigger this bug when issuing the "primary" command. The symptoms are that the request times out, and when inspecting Bluetooth packet logs the Read By Group Type Request is sent but no Read By Group Type Response is returned.

This bug appears to affect devel as well.

mikeryan:
The code should read rc != BLE_HS_EMSGSIZE. I will update the pull request in a few hours.

ccollins476ad:
Hi Mike,
In case you didn't see it, I responded to your email on the dev list: https://lists.apache.org/thread.html/0d940f8ca9a32d7fc3e3da7b224b3244349d5638af4bce28f8c79725@%3Cdev.mynewt.apache.org%3E

Please let me know if you think this is still an issue.  Also, if you agree that the bug is fixed, could you please close this pull request?

Thanks,
Chris

PR #80, opened 2016-07-25
Os callout1

NgesBrian:
This is a test to the os_callout

### DIFF
--- a/libs/os/src/test/callout_test.c
+++ b/libs/os/src/test/callout_test.c
@@ -29,40 +29,61 @@
 struct os_task callout_task_struct_send;
 os_stack_t callout_task_stack_send[CALLOUT_STACK_SIZE];
 
-/* Initializing the callout 
-struct os_event callout_event;
-struct os_eventq callout_eventq;
-struct os_eventq callout_eventq_delivered;
-struct os_callout struct_callout;
-*/
+#define RECEIVE_CALLOUT_TASK_PRIO        (2)
+struct os_task callout_task_struct_receive;
+os_stack_t callout_task_stack_receive[CALLOUT_STACK_SIZE];
 
 /* Delearing variables for callout_func */
-struct os_callout_func callout_func;
-struct os_eventq callout_eventq;
+struct os_callout_func callout_func_test;
+struct os_eventq *callout_eventq;
 void native_cputimer_cb(void *arg);
+
+/* The event to be sent*/
+struct os_eventq callout_evq;
+struct os_event callout_ev;
 
 /* This is a callout task to send data */
 void
 callout_task_send()
 {
-    int i;
-    /*
-    struct_callout.c_ev = callout_event;
-    struct_callout.c_evq = &callout_eventq;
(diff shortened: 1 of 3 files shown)


libs/os/src/test/callout_test.c
```diff
@@ -0,0 +1,123 @@
...
+    /* should say whether os is armed or not */
+    i = os_callout_queued(&callout_func_test.cf_c);
+    TEST_ASSERT(i == 1);
+
+    /* Send the callout */ 
+    os_time_delay(OS_TICKS_PER_SEC );
+    TEST_ASSERT(i == 0);
+
+     /* should say whether os is armed or not */
+    i = os_callout_queued(&callout_func_test.cf_c);
```
mkiiskila: This code will never be executed, because the callout_task_receive() terminates the test before callout_task_send() gets here.

Move this test to line 87. This is after callout has fired, and has been delivered to receiving task.

NgesBrian:
Please I dont understand what the part of the send task you are talking about

libs/os/src/test/callout_test.c
```diff
@@ -0,0 +1,123 @@
...
+    TEST_ASSERT(i==4);
+}
+
+/* This is a callout task to send data */
+void
+callout_task_send()
+{
+    int i; 
+    /* should say whether os is armed or not */
+    i = os_callout_queued(&callout_func_test.cf_c);
```
mkiiskila: This tests that os_callout_queued() will return 0 after calling os_callout_func_init(), before arming the callout. Given that it is more associated with the os_callout_func_init(), you should do the check right after that call.

libs/os/src/test/callout_test.c
```diff
@@ -0,0 +1,123 @@
...
+    TEST_ASSERT(i == 0);
+
+}
+
+void
+callout_task_receive(void *arg)
+{
+    struct os_event *event;
+    os_time_delay(OS_TICKS_PER_SEC / 20);
```
mkiiskila: This os_time_delay() is not necessary. os_eventq_poll() would make the receiving thread sleep until the event is queued (aka callout fires).

libs/os/src/test/callout_test.c
mkiiskila: I'm talking about the code after os_time_delay(OS_TICKS_PER_SEC). os_test_restart() will get called by the other task before this call to os_time_delay() returns.

NgesBrian:
ok I understand.
Will it be better for us to set a delay before the os_restart or for us to remove it to line 87 ??

NgesBrian:
because removing the code to line 87 means we will set i a global variable. so that its value should be maintained in the receive task if not the TEST_ASSERT will be checking for a different i.

so which is preferable??

NgesBrian:
os_callout_init_func returns void.

and that the os_callout_queued is the function that follows it immediately .

NgesBrian:
Thanks just did the correction.

PR #77, opened 2016-07-22
Added the profiles dir. Added Link Lost Service

bgiori:
Added the "net/nimble/profiles" directory to hold adopted profiles and services which do not belong under host. 

Additionally, I added the Link Lost Service (LLS). The LLS specification can be found online [here](https://developer.bluetooth.org/gatt/services/Pages/ServiceViewer.aspx?u=org.bluetooth.service.link_loss.xml). The full specification can be downloaded [here](https://www.bluetooth.org/DocMan/handlers/DownloadDoc.ashx?doc_id=303198).

The specification is pretty vague so my implementation may be slightly wrong.

PR #83, opened 2016-07-27
Added profiles dir under host. LLS implemented.

bgiori:
This is a new pull request after I screwed up #77 and didn't want to spend the time trying to fix it.

I've changed the location of profiles to be under the host directory and updated it such that each profile/service has its own subdirectory and pkg.yml file.

I've tested it and this works with the new host API.

### DIFF
--- a/net/nimble/host/profiles/lls/include/profiles/lls/ble_svc_lls.h
+++ b/net/nimble/host/profiles/lls/include/profiles/lls/ble_svc_lls.h
@@ -33,10 +33,9 @@ struct ble_hs_cfg;
 typedef int ble_svc_lls_event_fn(uint8_t alert_level); 
 
 uint8_t ble_svc_lls_alert_level_get(void);
-uint8_t ble_svc_lls_alert_level_set(uint8_t alert_level);
+int ble_svc_lls_alert_level_set(uint8_t alert_level);
 void ble_svc_lls_on_gap_event(struct ble_gap_event *event, void *arg); 
 
-int ble_svc_lls_register(void);
 int ble_svc_lls_init(struct ble_hs_cfg *cfg, 
                      uint8_t initial_alert_level,
                      ble_svc_lls_event_fn *cb);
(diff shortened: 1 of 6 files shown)


ccollins476ad:
I have a few comments, but overall I think it looks great.  I think this is a good representation of how future profiles should be implemented... these comments are more nit picks than anything else.

**1.**
```
/* Callback function */
ble_svc_lls_event_fn *cb_fn;
/* Alert level */
uint8_t ble_svc_lls_alert_level;
```

The above variables should probably be declared static since they are not referenced by any other compilation units.

**2.**
```
void
ble_svc_lls_on_gap_event(struct ble_gap_event *event, void *arg) {
    if(event->disconnect.reason ==
                BLE_HS_HCI_ERR(BLE_ERR_CONN_SPVN_TMO)) {
            cb_fn(ble_svc_lls_alert_level);
    }
}
```

I see two different options for this interface:
1. The application determines that the reported GAP event represents a disconnect by checking the GAP event type against BLE_GAP_EVENT_DISCONNECT.  Then, the application extracts the disconnect reason from the event and passes it to this function.

2. The application doesn't determine anything about the GAP event; it just blindly passes the event to this function, and the function guarantees that it will do the right thing (i.e., ignore the event if it does not represent a disconnect).

The current interface is somewhere between these two.  The application needs to check that the event is a disconnect first, but then it passes the entire GAP event to this function.  The application needs to check the event type, because this function assumes the event is a disconnect when it reads `event->disconnect.reason`.  I think this is slightly dangerous, because the function is receiving more information than it needs, and the information that it doesn't need is invalid (because it is overlaid in the union).

I personally prefer (1) over (2), but I think they are both reasonable.

**3.** Finally, I'm just being annoying here, but there are a few coding standard "noncompliances" (see CODING_STANDARDS.md):

> * Put space after keywords (for, if, return, switch, while).

> * After a function declaration, the braces should be on a newline, i.e. do:
>
> ```
>     static void *
>     function(int var1, int var2)
>     {
> ```
> 
> not:
> 
> ```
>     static void *
>     function(int var1, int var2) {
> ```

ccollins476ad:
Thanks, Brian.  It looks good.

PR #89, opened 2016-08-06
Os callout1

NgesBrian:


ccollins476ad:
Hi Nges,

It looks good.  I just have a few comments:

(1)
In callout_task_stop_send():

    /* Assigning data to send */
    for(k = 0; k<MULTI_SIZE; k++){
        callout_func_stop_test[k].cf_c.c_ev.ev_type = 10+ k;
        callout_func_stop_test[k].cf_c.c_ev.ev_arg = NULL;

         os_eventq_put(callout_func_stop_test[k].cf_c.c_evq,
             &callout_func_stop_test[k].cf_c.c_ev);
    }

This code directly enqueues each callout's event onto an event queue.  This is
not the typical callout usage.  The code circumvents the callout API and
directly uses the implementation details of the callout struct.  While it is
good to understand how something is implemented, it is best if test code uses
the public interface, just like a user would.  It is good to use the public
interface because then the test code will continue to work if the callout
implementation changes.

So, rather than directly enqueue the event, it is better to call
os_callout_reset().

(2)
In callout_task_stop_speak():

    /* should say whether callout is armed or not */
    i = os_callout_queued(&callout_func_speak.cf_c);
    TEST_ASSERT(i == 1);

    os_callout_stop(&callout_func_speak.cf_c);
    TEST_ASSERT(i == 1);

The second TEST_ASSERT is extraneous.  Maybe this is just a typo?

(3)
In my_callout_func() (also my_callout_stop_func() and my_callout_speak_func()):

    void
    my_callout_func(void *arg)
    {
        int p;
        p = 4;
        TEST_ASSERT(p == 4);
    }

This TEST_ASSERT() does not verify any Mynewt functionality.  Rather, it
just ensures that the compiler is working properly.  It would be better
to assign a value to a global variable (or a variable pointed to by
"arg").  Then, the test case can ensure that the callback was executed
by comparing the variable with the expected value.  The comparison must
be done outside of the callout function, after the callout has executed.

(4)
I noticed there are a few global variables that don't need to be global:

    int i;
    int j;

You should prefer local variables when possible.  These two variables don't
need to maintain their values across multiple functions, so they can be made
local.

(5)
in callout_task_receive():

    struct os_event *event;
    os_time_t now;
    os_time_t tm;
    os_sr_t sr;
    /* Recieve using the os_eventq_poll */
    event = os_eventq_poll(&callout_func_test.cf_c.c_evq, 1, OS_WAIT_FOREVER);
    TEST_ASSERT(event->ev_type ==  OS_EVENT_T_TIMER);
    TEST_ASSERT(event->ev_arg == NULL);

    TEST_ASSERT(i == 1);

Careful, i has not been assigned yet.  The last assert in this excerpt only
succeeds by chance.  This is a good example of why global variables can cause
difficulty!

NgesBrian:
Thanks very much Sir. I wrote a mail in the morning with all the above changes.
It seems as if the function is not actually called by the initialization of the callout.

Please look at the code and I am waiting to here from you
Thanks

PR #91, opened 2016-08-10
Boot serial

NgesBrian:


mkiiskila:
This pull request had a number of other unrelated commits in it. So I'm not going to bring it in. However, as the fix is a one-liner, I picked it separately and committed it. Thanks, for reporting this. You can close this pull request.

NgesBrian:
OK  great .

Thanks very much.

PR #96, opened 2016-08-19
Fixed bug where ID was always wrong in response

bgiori:
Fixes The newtmgr response notification's ID field always being 0. The ID field is only 1 byte so there is no reason to reverse the endianness.


ccollins476ad:
Looks good.  Thanks, Brian.

PR #100, opened 2016-08-30
Add shell/console history

utzig:
This adds initial history support the shell.

### DIFF
--- a/libs/console/full/src/cons_tty.c
+++ b/libs/console/full/src/cons_tty.c
@@ -68,6 +68,7 @@ struct console_tty {
 struct console_hist {
     uint8_t ch_head;
     uint8_t ch_tail;
+    uint8_t ch_size;
     uint8_t ch_curr;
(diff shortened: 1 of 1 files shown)


libs/console/full/src/cons_tty.c
```diff
@@ -28,19 +28,21 @@ int g_console_is_init;
 /** Indicates whether the previous line of output was completed. */
```
sterlinghughes: Hi Fabio!

Thanks for the PR, generally looks good.

If you can please add this as conditionally compiled, it would be greatly appreciated.

Specifically, look to use a feature in newt, which if enabled, will allow console history to be present -- since this feature takes memory that we probably don't want to spend on lower end platforms like the NRF51.  

A good example of using a feature to make this conditionally compiled exists in the statistics package (sys/stats) -- grep for STATS_NAME_ENABLE.

Thank you for the PR!

PR #106, opened 2016-09-20
Add #ifdef __cplusplus   extern "C" {   guards to all headers.

Timmmm:
This commit adds them to every header, except one:

    libs\baselibc\src\baselibc_test\unittests.h

Because the hack I used looks for normal #include guards for its place to put the C++ ones, and that file doesn't have any.

Also some files don't really have any functions or variables defined in them, so the guards are not strictly necessary, but they
don't hurt either and adding them everywhere makes it easier to see where they are missing.

### DIFF
--- a/apps/blecent/src/blecent.h
+++ b/apps/blecent/src/blecent.h
@@ -22,6 +22,10 @@
 
 #include "os/queue.h"
 #include "log/log.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 struct ble_hs_adv_fields;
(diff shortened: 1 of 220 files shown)


mkiiskila:
This looks good, except that the pull request is made against master, not against develop branch.
Given that the set of include files is different over there, I cannot bring this work easily over to that branch.

Could you use develop branch to do this work, and then do a pull request against that?

Thanks!

Timmmm:
Ok I will do that tomorrow.

Timmmm:
See #110

PR #107, opened 2016-09-22
Stm adc

vrahane:


drivers/adc/adc_stm32f4/src/adc_stm32f4.c
```diff
@@ -0,0 +1,712 @@
...
+
+#define STM32F4_IS_DMA_ADC_CHANNEL(CHANNEL) ((CHANNEL) <= DMA_CHANNEL_2)
+
+static DMA_HandleTypeDef *dma_handle[5];
+static struct adc_dev *adc_dma[5];
+
+struct stm32f4_adc_stats {
+    uint16_t adc_events;
+    uint16_t adc_error;
+    uint16_t adc_DMA_xfer_failed;
```
sterlinghughes: per-coding standards, DMA should be lower cased.

drivers/adc/adc_stm32f4/src/adc_stm32f4.c
```diff
@@ -0,0 +1,712 @@
...
+static GPIO_InitTypeDef
+stm32f4_resolve_adc_gpio(ADC_HandleTypeDef *adc, uint8_t cnum)
+{
+    uintptr_t adc_addr = (uintptr_t)adc->Instance;
+    uint32_t pin;
+    GPIO_InitTypeDef gpio_td;
+
+    switch (adc_addr) {
+        case (uintptr_t)ADC1:
+        case (uintptr_t)ADC2:
```
sterlinghughes: these switches fallthrough, which they shouldn't.  I understand that goto done will exit control here, but a break should be present anyway.

drivers/adc/adc_stm32f4/src/adc_stm32f4.c
```diff
@@ -0,0 +1,712 @@
...
+                    goto done;
+                case ADC_CHANNEL_15:
+                    pin = ADC3_CH15_PIN;
+                    goto done;
+            }
+        default:
+            assert(0);
+    }
+done:
+    gpio_td = (GPIO_InitTypeDef){
```
sterlinghughes: I don't believe this is valid C99.  Please do individual structure element assignment to keep this portable.

drivers/adc/adc_stm32f4/src/adc_stm32f4.c
```diff
@@ -0,0 +1,712 @@
...
+    struct stm32f4_adc_dev_cfg *cfg;
+    GPIO_InitTypeDef gpio_td;
+    struct adc_chan_config *chan_cfg;
+
+    assert(dev != NULL && IS_ADC_CHANNEL(cnum));
+    cfg  = dev->adc_dev_cfg;
+    hadc = cfg->sac_adc_handle;
+    chan_cfg = cfg->sac_chans;
+
+    // Enable DMA, ADC and related GPIO ports clock
```
sterlinghughes: No C++ comments please

drivers/adc/adc_stm32f4/src/adc_stm32f4.c
```diff
@@ -0,0 +1,712 @@
...
+    gpio_td = stm32f4_resolve_adc_gpio(hadc, cnum);
+    hal_gpio_init_stm(gpio_td.Pin, &gpio_td);
+
+
+    cfgdata = (ADC_ChannelConfTypeDef *)cfgdata;
+
+    if((rc = HAL_ADC_ConfigChannel(hadc, cfgdata)) != HAL_OK) {
```
sterlinghughes: space between if and open ()

drivers/adc/include/adc/adc.h
```diff
@@ -146,13 +146,15 @@ struct adc_chan_config {
...
     uint8_t c_configured;
+    uint8_t c_cnum;
 };
 
 struct adc_dev {
     struct os_dev ad_dev;
     struct os_mutex ad_lock;
     struct adc_driver_funcs ad_funcs;
     struct adc_chan_config *ad_chans;
+    void *adc_dev_cfg;
```
sterlinghughes: please keep structure prefixes consistent.  every other element is prefixed with "ad_"

apps/sblinky/pkg.yml
```diff
@@ -24,12 +24,8 @@ pkg.author: "Apache Mynewt <dev@mynewt.incubator.apache.org>"
 pkg.homepage: "http://mynewt.apache.org/"
```
sterlinghughes: Generally looks great, thanks for the submission!  Please change the few comments below, and resubmit, and I'll merge it!

drivers/adc/adc_stm32f4/src/adc_stm32f4.c
vrahane: ADC_1 and ADC_2 share the pins and hence the switches fall through. Also, similar is the case for ADC_1, ADC_2 and ADC_3.

sterlinghughes: not there, but the inner switch statement has the goto done.  I'm talking about adding a break after the inner switch()

drivers/adc/adc_stm32f4/src/adc_stm32f4.c
vrahane: It is valid C99, we use it at multiple places in apache_mynewt_core.

sterlinghughes: If you want to keep that, you can- you're right, it was introduced with C99.  I don't believe we use it elsewhere, I think we only do assignments when declaring variables.

drivers/adc/adc_stm32f4/src/adc_stm32f4.c
vrahane: We could do that for ADC_3 but for ADC1 and ADC_2 it is not possible because ADC_1, ADC_2 and ADC_3 share pins.

PR #110, opened 2016-10-01
Add __cplusplus guards to headers

Timmmm:
This commit adds them to almost every header. There are a few that don't have any normal header include guards because they don't define anything.


Timmmm:
Hmm not exactly encouraging contributions...

sterlinghughes:
Tim, I believe this PR was merged.  We just didn't properly attribute it as closed in the commit message.  If that's not true, please correct, and we'll reopen.

sterlinghughes:
And I should clarify: the reason merges don't happen through GH interface, is that we can't do it, because ASF GIT is mirrored to GH.  So we have to merge the PR using git command line tool, and then in the commit message reference "This closes #110" (for example.)  If we forget to do that, the PR remains open, even if the PR itself is merged.  The attempt here was to clean that up, for precisely the reason here: it's confusing as to which PRs we merged, and which ones we didn't - if we're not on top of this.  Again, I do think we actually merged your PR - but if not, please let us know, and we'll correct.

Timmmm:
Ah right you are - apologies for the attitude! :-)

ccollins476ad:
Sorry Tim, this was my mistake.  I really should have indicated the reason for closing each ticket.  I wasn't actually very confident that the commit would even successfully close the pull requests, so I got lazy.  And as far as attitude goes, I'm not too impressed :).  I would have been a lot worse!

PR #119, opened 2016-11-08
bletiny - Add help command and subcommands

michal-narajowski:
This patch introduces help commands to the bletiny app.
Right now interaction with the app is hard, because
it returns almost no information after executing commands.
To find out how to interact with the app one has to look
directly into the code.

With this patch user can invoke
help commands to list all possible parameters or commands
to pass to the app.

E.g.
b help - lists all toplevel commands like adv, conn, set, etc.
b adv help - lists all adv subcommands and parameters.

### DIFF
--- a/apps/bletiny/src/cmd.c
+++ b/apps/bletiny/src/cmd.c
@@ -256,7 +256,160 @@ static struct kv_pair cmd_adv_filt_types[] = {
     { "scan", BLE_HCI_ADV_FILT_SCAN },
     { "conn", BLE_HCI_ADV_FILT_CONN },
     { "both", BLE_HCI_ADV_FILT_BOTH },
+    { NULL }
 };
+
+static void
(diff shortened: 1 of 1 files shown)


ccollins476ad:
This looks great.  Thanks, Michal!

I just have one comment.  The help functions which take no arguments are defined with empty parameter lists, e.g.,

`
static void
bletiny_tx_help()
`

In C++, would be fine.  In C, an empty parameter list has quirky semantics.  It means the function takes an "unspecified" number of arguments, which is almost never the desired behavior (I would say never).  Instead, these functions should probably have a "void" parameter list, e.g.,

`
static void
bletiny_tx_help(void)
`

Aside from that, I think everything looks great.

Thanks,
Chris

michal-narajowski:
Thanks for the comments, Christopher.
Updated as requested and rebased onto develop.

PR #120, opened 2016-11-10
Added ticks on/off in shell/console

davidgs:
Cleaned up a few things in the prompt stuff and added ticks command to shell. You can now turn on/off printing OS Ticks on the console. New Config CONSOLE_TICKS added to sys/console


ccollins476ad:
Hi David,

First and most important - I like the feature and it seems to work perfectly.  Thanks for implementing this!

That said, there are a few nagging comments that I would like to make:

1. I noticed the default value of CONSOLE_PROMPT got changed from 0 to 1.  I do think most people prefer having the prompt enabled, but I would be leery of changing the default.   My main concern is that we have had a few issues with the prompt interfering with newtmgr commands.  The problem here isn't the prompt; it is that we are overloading the console as a newtmgr transport, which causes no end of weird issues.  As long as we continue to use the console as a newtmgr transport though, I think it is probably best to leave the prompt disabled by default.

2. There are a few coding standard violations, namely:

{noformat}
* Put space after keywords (for, if, return, switch, while).
* After a function declaration, the braces should be on a newline.
{noformat}

It is really easy to accidentally violate a coding standard or two.  I wonder if we should look into using some sort of automated check upon commit.

Do you think you could take a quick look at these issues?  Please feel free to voice any disagreements or questions.

Thanks,
Chris

davidgs:
Hi Chris,

Sorry about the change of the default value. That actually wasn't supposed to be in the commit. 

I've fixed the coding standards violations. Old habits die hard sometimes. I noticed that I do it the old way about 50% of the time, and the new way the other 50%. Getting there.

ccollins476ad:
Hi David,

I saw your other pull request which addresses these minor issues (https://github.com/apache/incubator-mynewt-core/pull/122).  I didn't merge that one because it appears to target the 1.0.0-b1 branch rather than develop.  It might be easiest to just commit the fixes to your develop branch.  That way, they will be part of the same pull request.

Thanks,
Chris

ccollins476ad:
Hi David,

Looks good.  Just don't forget to set the default CONSOLE_PROMPT value back to 0 :).

Chris

ccollins476ad:
It appears there are some unrelated changes in your development branch, which makes it difficult to merge this pull request.  In particular, there are some changes to bleprph.  Github expects there to be a dedicated branch per pull request, so coordination is difficult when the source branch is something like 'develop'.  The source branch needs to be frozen, apart from changes that directly pertain to the new feature or bug fix.

I hate to make more work for you, but do you think you could create a separate branch that contains only the changes pertinent to this pull request?  Then you would close this PR and open a new one against the new branch.

Please let me know if I have misinterpreted something, or if something I've said doesn't seem right.

Thanks,
Chris

PR #125, opened 2016-11-28
Use correct pin for LED

utzig:


mkiiskila:
Hi,
looks like PORTD 12-15 are all LEDs. Should we change from the current one?

utzig:
Forget it. My fail, there was another issue before causing the LED to not work. It's working on PORTD 12 too!

PR #126, opened 2016-11-28
Fix setting value of GPIO output only after init

utzig:
Fix ordering of operations because GPIO value can only be written after GPIO is already configured as output.


mkiiskila:
Ordering is intentional; otherwise initial pin state is incorrect when value to drive it to is 1.
This was done in d3ed09cd9b1836c67465a0ffa4e6bc63b7cc25ac.

utzig:
With the current code the PIN is never set to 1 for me. Have to call hal_gpio_write manually then.

mkiiskila:
I think I know what the issue is; order of events should be: 1) turn on RCC, 2) set output, and 3) enable gpio as output with rest of the settings.
Currently the order is 2-1-3, but 1-3-2 is not right either :)

Good catch though. We should fix this.

utzig:
Ok, done in the suggested order and it also works!

PR #127, opened 2016-11-28
Added console Ticks on/off

davidgs:
Added the ability to turn console ticks on/off. 
Also cleaned up console prompt adding straight on/off command and code cleanup

### DIFF
--- a/repository.yml
+++ b/repository.yml
@@ -20,11 +20,15 @@
 repo.name: apache-mynewt-core
 repo.versions:
     "0.0.0": "develop"
+    "0.0.1": "master"
     "0.7.9": "mynewt_0_8_0_b2_tag"
     "0.8.0": "mynewt_0_8_0_tag"
     "0.9.0": "mynewt_0_9_0_tag"
-    "0.9.1": "master"
-    "0-latest": "0.9.0"
-    "0-dev": "0.9.1"
+    "0.9.9": "mynewt_1_0_0_b1_tag"
+
+    "0-latest": "0.9.9"     # 1.0.0-b1
+    "0-dev": "0.0.0"        # develop
+
     "0.8-latest": "0.8.0"
     "0.9-latest": "0.9.0"
+    "1.0-latest": "0.9.9"   # 1.0.0-b1
--- a/sys/console/full/include/console/prompt.h
+++ b/sys/console/full/include/console/prompt.h
@@ -34,6 +34,8 @@ void console_set_prompt(char);
 
 void console_no_prompt(void);
 
+void console_yes_prompt(void);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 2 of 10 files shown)


ccollins476ad:
Hi David,

It looks good.  The only issue is that the pull request contains two of my commits.  A pull request should only contain your commits, and only commits which implement the feature that is the subject of the pull request.

This issue is easy to fix, though.  You just need to "pull with rebase" from the destination branch (develop).  To do this, you would do something like this from your working directory:

`git pull --rebase origin develop`

(assuming "origin" points to the apache remote).

Then, push again to your davidgs:ticks branch.

Thanks,
Chris

ccollins476ad:
Hmm, it looks like those two extra commits are actually from the master branch, rather than from develop.  I think you may have accidentally pulled the apache master branch into your develop branch at some point, because it appears these problematic commits are in your develop branch.  The other thing I noticed is that some of your commits are duplicated in your 'ticks' branch.  For example:

```
commit 27b5699db2d1ed4a30c67683168c4049a24f328a
Author: David G. Simmons <santafen@mac.com>
Date:   Mon Nov 28 13:45:19 2016 -0500

    missing function declaration

commit 0019763885461261c92ea169c9f5ac55c7da9ecb
Author: David G. Simmons <santafen@mac.com>
Date:   Mon Nov 28 13:45:19 2016 -0500

    missing function declaration
```

Both of these commits consist of the same set of diffs.

I'm afraid this is way over my head :).  Git is way too complicated for me.  I'm sure there is a super-efficient and smart way to fix this, but the only solution I know of is:

1. Checkout the apache develop branch (not your fork).  How to do this depends on how your remotes are named, but it is probably something like `git checkout apache/develop` (I am assuming origin points to your fork, not to the apache server).

2. Create a new branch

`git checkout -b ticks2`

This new branch is split off of the apache develop branch, so it will contain the correct set of commits.

3. Cherry-pick the commits that implement the console ticks feature

`git cherry-pick 0019763885461261c92ea169c9f5ac55c7da9ecb 899d37c74d78d02230f244b9c77403e9c1777213`

4. Push the changes to your fork

`git push <remote> ticks2`

5. Close this pull request and open another one against the ticks2 branch.

Sorry for all the hassle.  Like I said, there is probably a better way to solve this, but I am no git expert.

The other thing you'll probably want to do is fix your develop branch so that it doesn't contain extraneous commits.  If you have changes in that branch that you need to keep, then this might be a bit tricky.  If you don't need anything in that branch, I would replace it with the contents of the apache develop branch.  You can do that as follows (I am assuming 'origin' points to your fork and 'apache' points the apache server):

```
git checkout develop
git reset --hard apache/develop
git push -f origin develop
```

The last command is a "force push", which overwrites the git history on the remote server.  So, be sure you don't need anything in your develop branch before doing this.  To be safe, you can duplicate your remote branch with "checkout -b <branch-name>" first to create a backup.

Chris

PR #129, opened 2016-11-29
L2CAP fixes for PTS testing

rymanluk:
Here is a set of fixes/minor cleaning and minor refactor which came up after testing against PTS.


ccollins476ad:
It looks good.  Thanks, Łukasz!

PR #130, opened 2016-11-30
Various nimble/sm fixes affecting qualification with PTS

andrzej-kaczmarek:


ccollins476ad:
Hi Andrzej,

This is great!  Thanks for fixing all these issues.

There is one problem, though: with your changes, some of the unit tests fail.  You can run the unit tests with the "newt test @apache-mynewt-core/net/nimble/host" command.

The failing tests are:
* ble_sm_test_case_peer_fail_inval
* ble_sm_test_case_us_fail_inval

Both of these tests are implemented in: net/nimble/host/test/src/ble_sm_test.c

The tests fail when they attempt to verify that the host rejects pairing requests and responses that indicate invalid values for any of the following fields:
* OOB flag
* AuthReq
* Minimum key size
* Maximum key size
* Initiator key distribution
* Responder key distribution

Could you point me to the part of the spec which indicates that reserved bits should be ignored?  I wasn't able to find this language anywhere in the spec.  I don't doubt that you are correct, I just want to know where it is written so I can refer to it in the future.

Thanks,
Chris

ccollins476ad:
Actually, I do see where the spec indicates that reserved bits in AuthReq should be ignored (Vol. 3, Part H, 3.5.1, 3.5.2, 3.6.7):
> The reserved 3-bit field shall be set to zero and ignored upon reception.

Also, just to clarify - I agree that nimble should ignore reserved bits if that is what other stacks do.  I was just asking about the spec for my own edification.  The only real issue is the failing unit tests.

Thanks,
Chris

andrzej-kaczmarek:
Hi Chris,

I noticed the failing unit tests after I pushed this pull request and of course I will fix them - they need a bit more work since expected behavior of stack changed.

As for specification references, I also put them in commit messages where applicable and the same applies to PTS test cases which given commit fixes (if any) for future reference.

I'll push updates to unit tests as a separate commit soon (hopefully).

BR,
Andrzej

andrzej-kaczmarek:
Hi Chris,

All unit tests pass now with my fixes. As you can see, I actually removed some of them since they expect pairing to fail in cases which are correct and should succeed. For these scenarios I will prepare new unit tests in proper test group and push in separate pull request.

BR,
Andrzej

ccollins476ad:
Looks good.  Thanks, Andrzej.

PR #132, opened 2016-12-05
gattc: Fix unnecessary status overwrite in read by type err

michal-narajowski:
This status change is not needed. With this patch it is possible to pass TC_GAR_CL_BI_07_C.


ccollins476ad:
Good catch.  Thanks, Michał.

PR #133, opened 2016-12-05
bletiny: Fix parse uuid cmd argument

michal-narajowski:
parse_arg_uint16 extracted UUID argument and if it failed to parse it
parse_arg_byte_stream_exact_length was called, but coudldn't find uuid
argument. Additionally the part that parsed UUID in
"e7add801-b042-4876-aae1112855353cc1" format was ineffective, because
parse_arg_byte_stream_exact_length expects arg name as a first
parameter. Finally parsed stream needed to be reversed.

With this patch it is possible to pass Read Using Characteristic UUID
tests (TC_GAR_CL_BV_03_C, etc.)

### DIFF
--- a/apps/bletiny/src/parse.c
+++ b/apps/bletiny/src/parse.c
@@ -87,6 +87,20 @@ parse_arg_find_idx(const char *key)
 }
 
 char *
+parse_arg_peek(const char *key)
+{
+    int i;
+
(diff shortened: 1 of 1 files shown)


ccollins476ad:
Good catch re: reversed UUID!

PR #134, opened 2016-12-06
Added NRF52_adc Tutorial

davidgs:
Complete Tutorial for adding an eTape Liquid sensor to the NRF52dk board. Docs also added to incubator-mynewt-site.

Required some additions to the bsp.

### DIFF
--- a/apps/nrf52_adc/pkg.yml
+++ b/apps/nrf52_adc/pkg.yml
@@ -1,0 +1,41 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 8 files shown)


aditihilbert:
David, Let me first say that this tutorial covers a lot of ground and is going to be very helpful. There are a couple of issues we will need to address, however. 

Firstly, the Nordic ADC drivers were debundled from @apache-mynewt-core and hosted on runtimeinc/mynewt_nordic because of licensing terms (which we are working on getting aligned with Apache's). Until then we cannot pull in these drivers into the core repo. Since both your app code and bsp code has dependency on these libraries, I suggest this PR be done against the mynewt_nordic repo. 

You should be able to use the same bsp filename in the mynewt_nordic repo without any conflicts because the path is different with the repo reference in the path. However, we haven't tested this yet (there's always a first time for everything :), so I should warn you that you may have to create a new bsp file with a different name.

Also, I'd recommend following the app<-driver<-bsp code flow where there the app calls a simple driver API and the driver code calls the nrf52_adc_dev_init. Then this tutorial would also be a perfect example to link to from Mynewt's driver and hal usage overview in the OS guide. 

thanks!

davidgs:
Yes, it covers a lot of ground. I was made aware of the license issues by Will this morning so I'll be looking into addressing theses issues right away. 

Top of my list.

PR #137, opened 2016-12-09
Initial MMC/SD support

utzig:
I removed the test app because it was basically a copy/paste of the FAT test app with added MMC initialization and was still dependent on modifications to FAT to read/write to MMC instead of hal_flash.

Some issues that remain to be fixed/implemented:

1) After initialization the SPI clock can be switched to 25MHz (from 100KHz). I was not able to make it work probably because of the wire-ups between my boards.
2) There are 3 main types of MMC/SD cards. Two of them use very similar initialization and another uses a different path (older/smaller cards). I only tested the code with newer SDHC/SDXC path.
3) FAT already breaks big reads/writes into sector reads/writes (512 byte) so multi-block read/write commands were only tested with sector sized reads/writes (basic sanity test).
4) All SPI transfers are currently done using polled byte transfers instead of block transfers. Would also be better to do interrupt driven block reads/writes.


ccollins476ad:
Looks great!

PR #138, opened 2016-12-09
Fix FAT FILE access flags usage

utzig:


ccollins476ad:
Thanks, Fabio.  It looks good.  There is an assumption that `unsigned int` is equivalent to `uint32_t`, but I'm not sure if this is a concern or not.  There is probably no chance that we'll ever build on a platform where this assumption doesn't hold.  If the fatfs file system is meant to only work with 32-bit ints, then perhaps the UINT typedef should be redefined to uint32_t (rather than unsigned int).

To be clear, the code I am referring to is the cast from `uint32_t *` to `UINT *` here:

```
static int
fatfs_read(struct fs_file *fs_file, uint32_t len, void *out_data,
           uint32_t *out_len)
{
    FRESULT res;
    FIL *file = (FIL *)fs_file;

    res = f_read(file, out_data, len, (UINT *)out_len);
    return fatfs_to_vfs_error(res);
}
```

Anyway, I really am not sure if this is worth worrying about, but I felt that I should at least mention it.  I see a few options:

1. Just assume we'll always build on platforms where `unsigned int` == `uint32_t`
2. Redefine the UINT typedef to uint32_t.
3. Define a UINT on the stack of this function and pass its address to f_read().  Then assign the contents of the UINT to the contents of `*out_len`.

What do you think?

Thanks,
Chris

utzig:
Regarding the size assumption, I think both MSP430 and AVR have 2-byte sized `unsigned int`. Those architectures could eventually get a port, although the memory on those devices might not allow using FAT at all. The issue I was having was that `out_len` was returning zeroed and I did what you mentioned on `3` and then ended up removing the local variable... leave this here for some more days I will test be testing again and get back with one of the suggested fixes.

utzig:
This fix is already included in #158. Closing it.

PR #143, opened 2016-12-20
nimble: Fix ATT Find By Type Value Request

andrzej-kaczmarek:
This fixes Find By Type Value Request handling which is required to allow discover service by UUID from peer. With current implementation, which is invalid, we can expect a lot of interoperability issues when Nimble is used as GATT Server.


ccollins476ad:
Thanks for fixing the unit tests.  I know the host test code is not easy to work with.

PR #144, opened 2016-12-20
Cbor

davidgs:
Not all of the cbor includes had the tinycbor/ in them, causing errors. Had to back out the ads tutorial to get this to be a clean pr.


utzig:
+1

I had to apply those same changes to build slinky.

Btw @davidgs, it would be look much better if you rebased your branches instead of merging because it would show only your commits instead of all commits applied to develop.

mkiiskila:
Applied the patch. Fabio has a good point regarding history, though.

'git rebase -i HEAD~2'; for example. This would allow you to e.g. combine 2 previous commits you've made. I use that occasionally when I forget to include a file a commit.

PR #147, opened 2016-12-22
First pass at LSM303DLHC sensor driver

microbuilder:
This adds a first pass at an [LSM303DLHC](https://www.adafruit.com/product/1120) ([datasheet](http://www.st.com/content/ccc/resource/technical/document/datasheet/56/ec/ac/de/28/21/4d/48/DM00027543.pdf/files/DM00027543.pdf/jcr:content/translations/en.DM00027543.pdf)) driver using the new sensor API.

The current driver only includes support for reading from the accelerometer. The magnetometer and temperature sensor will be added in the future once the sensor API has stabilized a bit more.

## Initializing the Driver

You can initialize the driver with the following code in main (or elsewhere):

```
os_dev_create((struct os_dev *) &lsm303dlhc_sensor, "lsm303dlhc",
        OS_DEV_INIT_KERNEL, OS_DEV_INIT_PRIMARY, lsm303dlhc_drvr_init, NULL);
```
...
```
static int
lsm303dlhc_drvr_init(struct os_dev *dev, void *arg)
{
    struct lsm303dlhc_cfg cfg;
    int rc;

    rc = lsm303dlhc_init(dev, arg);
    if (rc != 0) {
        goto err;
    }

    cfg.accel_range = LSM303DLHC_ACCEL_RANGE_2;
    cfg.accel_rate = LSM303DLHC_ACCEL_RATE_400;
    cfg.sample_itvl = OS_TICKS_PER_SEC / 200;

    rc = lsm303dlhc_config((struct lsm303dlhc *) dev, &cfg);
    if (rc != 0) {
        goto err;
    }

    return (0);
err:
    return (rc);

}
```

## Displaying Sensor Data in Shell

To display data in the shell enter the following commands:

```
sensor read lsm303dlhc 1 10
```

**Note**: This PR doesn't address the lack of %f/float support in tinylibc in the mynewt repo, so the current code in sensors_shell.c won't display the accelerometer data as-is. You can modify `sensor_shell.c` with the following hacky code to display a float value until a more permanent solution is found:

```
static int
sensor_shell_float_fraction(float f, int precision)
{
    if (f < 0.0F) {
        f *= -1.0F;
    };
    return (int)(f * precision) % precision;
}

static int
sensor_shell_read_listener(struct sensor *sensor, void *arg, void *data)
{
    struct sensor_shell_read_ctx *ctx;
    struct sensor_accel_data *sad;
    float *temp;

    ctx = (struct sensor_shell_read_ctx *) arg;

    ++ctx->num_entries;

    if (ctx->type == SENSOR_TYPE_ACCELEROMETER) {
        sad = (struct sensor_accel_data *) data;
        if (sad->sad_x != SENSOR_ACCEL_DATA_UNUSED) {
            console_printf("x = %i.%05i, ", (int)sad->sad_x,
                sensor_shell_float_fraction(sad->sad_x, 100000));
        }
        if (sad->sad_y != SENSOR_ACCEL_DATA_UNUSED) {
            console_printf("y = %i.%05i, ", (int)sad->sad_y,
                sensor_shell_float_fraction(sad->sad_y, 100000));
        }
        if (sad->sad_z != SENSOR_ACCEL_DATA_UNUSED) {
            console_printf("z = %i.%05i", (int)sad->sad_z,
                sensor_shell_float_fraction(sad->sad_z, 100000));
        }
        console_printf("\n");
    }

    return (0);
}
```

### DIFF
--- a/hw/drivers/sensors/lsm303dlhc/include/lsm303dlhc/lsm303dlhc.h
+++ b/hw/drivers/sensors/lsm303dlhc/include/lsm303dlhc/lsm303dlhc.h
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef __SENSOR_LSM303DLHC_H__
+#define __SENSOR_LSM303DLHC_H__
+
+#include "os/os.h"
+#include "os/os_dev.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum lsm303dlhc_accel_range {
+    LSM303DLHC_ACCEL_RANGE_2            = 0x00 << 4, /* +/- 2g  */
+    LSM303DLHC_ACCEL_RANGE_4            = 0x01 << 4, /* +/- 4g  */
+    LSM303DLHC_ACCEL_RANGE_8            = 0x02 << 4, /* +/- 8g  */
+    LSM303DLHC_ACCEL_RANGE_16           = 0x03 << 4, /* +/- 16g */
+};
+
+enum lsm303dlhc_accel_rate {
+    LSM303DLHC_ACCEL_RATE_POWER_DOWN    = 0x00 << 4,
+    LSM303DLHC_ACCEL_RATE_1             = 0x01 << 4,
+    LSM303DLHC_ACCEL_RATE_10            = 0x02 << 4,
+    LSM303DLHC_ACCEL_RATE_25            = 0x03 << 4,
+    LSM303DLHC_ACCEL_RATE_50            = 0x04 << 4,
+    LSM303DLHC_ACCEL_RATE_100           = 0x05 << 4,
+    LSM303DLHC_ACCEL_RATE_200           = 0x06 << 4,
+    LSM303DLHC_ACCEL_RATE_400           = 0x07 << 4,
+    LSM303DLHC_ACCEL_RATE_1620          = 0x08 << 4
+};
+
+enum lsm303dlhc_mag_gain {
+    LSM303DLHC_MAG_GAIN_1_3             = 0x20 << 5, /* +/- 1.3 gauss */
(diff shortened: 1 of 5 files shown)


hw/drivers/sensors/lsm303dlhc/src/lsm303dlhc.c
```diff
@@ -0,0 +1,398 @@
...
+    /* If the read isn't looking for accel data, then don't do anything. */
+    if (!(type & SENSOR_TYPE_ACCELEROMETER)) {
+        rc = SYS_EINVAL;
+        goto err;
+    }
+
+    lsm = (struct lsm303dlhc *) SENSOR_GET_DEVICE(sensor);
+
+    x = y = z = 0;
+    lsm303dlhc_read48(LSM303DLHC_ADDR_ACCEL,
```
microbuilder: The return code should be checked here :/

hw/drivers/sensors/lsm303dlhc/include/lsm303dlhc/lsm303dlhc.h
```diff
@@ -0,0 +1,90 @@
...
+    LSM303DLHC_MAG_RATE_15              = 0x04 << 2,  /* 15 Hz   */
+    LSM303DLHC_MAG_RATE_30              = 0x05 << 2,  /* 30 Hz   */
+    LSM303DLHC_MAG_RATE_75              = 0x06 << 2,  /* 75 Hz   */
+    LSM303DLHC_MAG_RATE_220             = 0x07 << 2   /* 220 Hz  */
+};
+
+struct lsm303dlhc_cfg {
+    enum lsm303dlhc_accel_range accel_range;
+    enum lsm303dlhc_accel_rate accel_rate;
+    uint16_t sample_itvl;
```
microbuilder: This should probably be removed as well since it has no purpose without BG data streaming

hw/drivers/sensors/lsm303dlhc/src/lsm303dlhc.c
```diff
@@ -0,0 +1,398 @@
...
+
+    lsm = (struct lsm303dlhc *) dev;
+
+    #if MYNEWT_VAL(LSM303DLHC_LOG)
+    log_register("lsm303dlhc", &_log, &log_console_handler, NULL, LOG_SYSLEVEL);
+    #endif
+
+    sensor = &lsm->sensor;
+
+    #if MYNEWT_VAL(LSM303DLHC_STATS)
```
sterlinghughes: Hi Kevin,  I'll change these after commit (no need to submit new PR), however, in the future our style guides say: 

#if BLA
    log_register()
#endif 

and not 

    #if BLA
    log_register()
    #endif

sterlinghughes: err, the GH style highlight made that confusing.  however, the point i was making if the 'if' condition is not also indented, only the code block, as it makes it easier to scan for #if conditions without reading the code.

microbuilder: Sorry, I originally had them that way and should have scanned the style guide again. I'm always on the fence about this. Having them left-aligned also seriously hurts code readability in my opinion since you lose the flow inside functions, but the norm is more left-aligned. In the future I'll stick to full left-alignment, thanks for the heads up.

PR #150, opened 2016-12-29
Fix gdb shell command on Windows

simonratner:
Since `gdb` executes shell commands using system interpreter, it can't run the shell command to launch `JLinkGDBServer` as currently written. 

Changes:
* Reference `sh` without explicit path, so any `sh` or `sh.exe` in user's environment path will work
* Replace quotes; since `gdb` uses system interpreter to execute shell commands, the entire command has to be passed to `sh`, including job control
* Replace `JLinkGDBServer` with `JLinkGDBServerCL` command-line version

### DIFF
--- a/hw/scripts/jlink.sh
+++ b/hw/scripts/jlink.sh
@@ -45,7 +45,7 @@ jlink_load () {
     # downloading somewhere in the flash. So need to figure out how to tell it
     # not to do that, or report failure if gdb fails to write this file
     #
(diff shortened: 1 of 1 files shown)


mkiiskila:
Otherwise would be ok, but JLink tools I've installed don't have JLinkGDBServerCL.

simonratner:
No problem, I've reverted that bit. Not essential, was just a convenience to not lose terminal input focus.
(Fyi, the version of JLink tools I have is v6.12a)

PR #152, opened 2016-12-29
Add magnetometer support with LSM303DLHC test driver

microbuilder:
This pull requests adds magnetometer support to the sensor API. It also adds magnetometer data to the LSM303DLHC driver.


utzig:
LGTM

PR #153, opened 2016-12-29
Fix gdb server command when NO_GDB is set

simonratner:
Missed in #150.


simonratner:
Merged by @mkiiskila, closing.

PR #154, opened 2016-12-30
Fix GATT after PTS tests

michal-narajowski:


ccollins476ad:
Awesome, thanks Michał!

PR #155, opened 2017-01-04
Nimble: proper encryption state handling in ATT

andrzej-kaczmarek:
This adds proper support for encryption state in ATT code, i.e.
- attribute can have minimum encryption key size requirement
- "Insufficient Encryption" error code returned when LTK is present but encryption is not enabled
- "Insufficient Encryption Key Size" error code returned when encryption key size does not minimum requirement for given attribute

...and other changes which were required to run and test this code.


ccollins476ad:
Looks great!

PR #156, opened 2017-01-04
support for rejecting SM commands when pairing is disabled

sjanc:
support for rejecting SM commands when pairing is disabled (this is mandated by spec)  + some minor cleanup


ccollins476ad:
Looks good.  Thanks, Szymon.

PR #160, opened 2017-01-17
nimble: Refactor UUID handling (ATT/GATT)

andrzej-kaczmarek:
This patch introduces dedicated type for UUIDs in order to handle them
consistently everywhere. It not only makes handling of UUIDs easier,
but also allows for some flash and RAM savings due to some extra code
being removed and less memory required to store UUIDs in common cases
(especially with 16-bit UUIDs).

*IMPORTANT NOTES*
- ble_uuid_t is a "stub" type which shall be always used as a pointer;
  to access actual UUID value other helpers and types shall be used
- when registering services using new API, only UUID pointer is stored
  for all attributes (there is *no* copy of value made) thus it is not
  allowed to pass UUIDs created e.g. on stack
- it is illegal to have 128-bit UUID declared which has its 16-bit UUID
  equivalent - there is no implicit conversion from 128-bit to 16-bit
  UUID done inside Nimble
- ...but this means you can rely on UUID type received from Nimble and
  128-bit UUID will be always non-Bluetooth one

Below are few examples how to use new UUIDs API:

// 16-bit UUID value with initialization
ble_uuid16_t uuid16 = BLE_UUID16_INIT(0x2801);
// 128-bit UUID value with initialization
ble_uuid128_t uuid128 = BLE_UUID128_INIT(0x00, 0x11, ... 0xFF);

// "stub" type is used to pass any type of UUID via API
ble_uuid_t *uuid1 = &uuid16.u;
ble_uuid_t *uuid2 = &uuid128.u;

// comparing UUIDs is really easy now
if (ble_uuid_cmp(uuid1, uuid2) == 0) { /* equal */ )

// since in most cases we use 16-bit UUIDs, it's convenient to easily
// retrieve 16-bit value (or 0 if UUID is not 16-bit)
uint16_t u16 = ble_uuid_u16(uuid1);

// the above is equivalent of
u16 = uuid1->type == BLE_UUID_TYPE_16 ? BLE_UUID16(uuid1)->value : 0;

// we can easily define UUID inline as well, e.g. when defining service
static const struct ble_gatt_svc_def ble_svc_gap_defs[] = {
    {
        .type = BLE_GATT_SVC_TYPE_PRIMARY,
        .uuid = BLE_UUID16_DECLARE(BLE_SVC_GAP_UUID16),
(...)

// finally, if we need to store any UUID there is an union defined to help
ble_uuid_any_t uuid;
uuid.u16 = uuid16;


ccollins476ad:
Thanks, Andrzej.  This is great.

PR #162, opened 2017-01-19
Support for AT45DB SPI flash

utzig:
WIP - do not merge!


utzig:
The latest commit should update all hal_flash usage by MCUs and enable them to keep working. Still trying to settle on some design issue with hal_flash subclassing in at45db_dev. Not very happy with the current solution but didn't find one that is both better and easier yet. I think this PR is OK for merging and the design issue I can fix in the following days (when I settle on a final design).

ccollins476ad:
Hi Fabio,

I'm getting a compilation error in `sys/flash_map/src/flash_map.c`:

```
sys/flash_map/src/flash_map.c: In function 'flash_read_sector':
sys/flash_map/src/flash_map.c:72:33: error: passing argument 1 of 'hf->hf_itf->hff_sector_info' makes pointer from integer without a cast [-Werror=int-conversion]
     hf->hf_itf->hff_sector_info(sector_idx, &start, &size);
                                 ^
sys/flash_map/src/flash_map.c:72:33: note: expected 'const struct hal_flash *' but argument is of type 'int'
```

Aside from that, I think the changes are good.  However, I would like to make sure Marko doesn't see any issues with the HAL changes (unless you have already talked to him).  Like I said, I don't have a problem with any of the changes; I'm just being cautious :).

utzig:
The change to `flash_map` is part of the PR already:

https://github.com/apache/incubator-mynewt-core/pull/162/files#diff-c8f38113c19dc0921abb4c8849795aceL83

Not sure what is going on but I'll try doing a rebase, because develop seems to have changed significantly in the last few days...

utzig:
Rebased on `develop` and fixed the compilation error.

mkiiskila:
This looks good to me. But we should have the matching set of changes ready for arduino repository, in order to not break the build for that MCU.

utzig:
Done: https://github.com/runtimeinc/mynewt_arduino_zero/pull/3

utzig:
This was merged on main repo. Closing!

PR #163, opened 2017-01-19
Avoid hard-coding absolute compiler path

ilg-ul:
Hard-coding the absolute path for cc & as is not only non-consistent
with the other definitions (none use absolute paths), but it is also
wrong, gcc-5 can be installed in any location (via homebrew or
otherwise).


ilg-ul:
Thank you!

PR #164, opened 2017-01-24
nimble/gatt: unify discovery api

michal-narajowski:
This patch unifies api to make it consistent across all discovery
methods. Always pass true range of attributes to search (start-handle,
end-handle).


ccollins476ad:
Thanks, Michał.

PR #171, opened 2017-02-01
L2CAP LE COC signaling part plus some additional fixes

rymanluk:
In this pull LE COC signaling part has been implemented together with extensions to Bletiny app needed to test it. It is tested against BlueZ and PTS and works fine.

Note: This pull will not enable LE COC,  BLE_L2CAP_COC_MAX_NUM is 0

As you will notice I change a way of creating and sending signal commands. It is now done similar to what BLE_SM does. I plan to do same for l2cap update command.

Please have a look on API as I change it a bit comparing to what we discussed on mailing list. I found it more useful when writing Bletiny extensions.

In addition in this pull you can find some fixes and cleaning.

### DIFF
--- a/kernel/os/include/os/queue.h
+++ b/kernel/os/include/os/queue.h
@@ -158,10 +158,11 @@ struct {                                                \
     }                                                       \
     else {                                                  \
         struct type *curelm = SLIST_FIRST((head));          \
-        while (SLIST_NEXT(curelm, field) != (elm))          \
-            curelm = SLIST_NEXT(curelm, field);             \
-        SLIST_NEXT(curelm, field) =                         \
-            SLIST_NEXT(SLIST_NEXT(curelm, field), field);   \
+        while (SLIST_NEXT(curelm, field) &&                 \
+               SLIST_NEXT(curelm, field) != (elm)) {        \
+              curelm = SLIST_NEXT(curelm, field);           \
+        }                                                   \
+        SLIST_NEXT(curelm, field) = SLIST_NEXT(elm, field); \
     }                                                       \
 } while (0)
 
--- a/net/nimble/host/src/ble_l2cap_coc.c
+++ b/net/nimble/host/src/ble_l2cap_coc.c
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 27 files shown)


rymanluk:
ah...just run newt test @apache-mynewt-core/net/nimble/host/test - so yes, I need to fix some test cases

rymanluk:
Tests fixed

net/nimble/host/src/ble_l2cap_coc.c
```diff
@@ -0,0 +1,135 @@
...
+#include "ble_l2cap_coc_priv.h"
+#include "ble_l2cap_sig_priv.h"
+
+#if MYNEWT_VAL(BLE_L2CAP_COC_MAX_NUM) != 0
+
+STAILQ_HEAD(ble_l2cap_coc_srv_list, ble_l2cap_coc_srv);
+
+static struct ble_l2cap_coc_srv_list ble_l2cap_coc_srvs;
+
+typedef int ble_l2cap_coc_rx_fn(uint16_t conn_handle,
```
rymanluk: hmm, that is not needed, let me fix that.

rymanluk: fixed

kernel/os/include/os/queue.h
```diff
@@ -158,10 +158,11 @@ struct {                                                \
...
         struct type *curelm = SLIST_FIRST((head));          \
-        while (SLIST_NEXT(curelm, field) != (elm))          \
-            curelm = SLIST_NEXT(curelm, field);             \
-        SLIST_NEXT(curelm, field) =                         \
-            SLIST_NEXT(SLIST_NEXT(curelm, field), field);   \
+        while (SLIST_NEXT(curelm, field) &&                 \
+               SLIST_NEXT(curelm, field) != (elm)) {        \
+              curelm = SLIST_NEXT(curelm, field);           \
+        }                                                   \
+        SLIST_NEXT(curelm, field) = SLIST_NEXT(elm, field); \
```
ccollins476ad: I'm a little nervous about this change.  A lot of the queue.h macros are awful, but they are a de facto standard from BSD from way back.  I think a lot of users would be surprised if our queue.h has a different API from the BSD version.

ccollins476ad: Hi Łukasz,

Aside from my comment above, I think It looks great.  Thanks for fixing a lot of those nagging issues (e.g., renaming ble_l2cap_chan_mtu to ble_att_chan_mtu).

sterlinghughes: yes, we should not change queue.h

rymanluk: OK, I understand that we don't want to touch BSD code.
Just need to remember that API assumes that element is on the list otherwise macro goes outside list memory and can do crazy things.

Will remove this patch.

PR #173, opened 2017-02-01
Nimble: BD address handling

andrzej-kaczmarek:
- ble_addr_t type provided instead of type/address pair (for now used in all public APIs)
- clean up usage of address type and own address type symbols (they were mixed in different parts of code)
- fix pairing with privacy enabled (i.e. we use RPA)
- add identity changed event (i.e. identity address received after pairing)

### DIFF
--- a/apps/blecent/src/main.c
+++ b/apps/blecent/src/main.c
@@ -243,7 +243,7 @@ blecent_scan(void)
     disc_params.filter_policy = 0;
     disc_params.limited = 0;
 
-    rc = ble_gap_disc(BLE_ADDR_TYPE_PUBLIC, BLE_HS_FOREVER, &disc_params,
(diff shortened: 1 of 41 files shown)


ccollins476ad:
Hi Andrzej,

Thanks, it looks good.  However, I am getting some compiler errors when I run "newt test net/nimble/host".  Could you please take a look?

Thanks,
Chris

andrzej-kaczmarek:
Yes, absolutely. I forgot to check unit tests on final version...

andrzej-kaczmarek:
I updated test cases and also fixed few small issues I did not see when testing manually.

There's one problem though: test vectors for SM SC test cases with privacy have to be recalculated since they were using invalid address types. I just commented them out for now, will try to update them in upcoming days...

ccollins476ad:
It looks good.  Thanks for the late night, Andrzej!

PR #175, opened 2017-02-08
MYNEWT-292; Don't allow two tasks to have the same priority

utzig:
Add an assert checking that same prio is not used by two different
tasks.

### DIFF
--- a/hw/mcu/native/pkg.yml
+++ b/hw/mcu/native/pkg.yml
@@ -28,6 +28,6 @@ pkg.keywords:
 pkg.deps:
     - hw/hal
     - compiler/sim
-    
+
 pkg.req_apis:
     - console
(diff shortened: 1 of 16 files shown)


mkiiskila:
Cool. Can you also update the unit tests for kernel/os and sys/flash_map? They start failing after I pull this in. I assume the fault is within the tests, not with the code.

utzig:
Ok, should be fixed now!

There were basically two issues:

1) Duplicated calls of `os_init` and `sysinit`. Since `tu_init` already calls those functions, I removed them from some of the testcases which were creating duplicated tasks.

2) Some tasks where being created with same priorities as tasks that already existed. At least one I did not find where the original task was created (with prio=2) but I changed the tests to start at some other initial prio.

The calls to `os_start` I also had to remove because those tasks functions don't have an infinite loop in them and there is an assert that checks that started tasks never finish! Someone should do some careful review of this particular change (if deemed necessary).

PR #176, opened 2017-02-13
MYNEWT-335

utzig:
- fcb_offset_last_n() changed to return a struct fcb_entry
- Added a test case
- Now properly returns error when there is no item at requested offset

### DIFF
--- a/fs/fcb/src/fcb.c
+++ b/fs/fcb/src/fcb.c
@@ -205,14 +205,15 @@ fcb_sector_hdr_read(struct fcb *fcb, struct flash_area *fap,
 }
 
 /**
- * Finds n-th element offset
+ * Finds the last n-th element fcb_entry (0 index means last element!)
  * @param0 ptr to fcb
  * @param1 n number of entries to calculate offset before
- * @param2 ptr to the offset before to be returned
+ * @param2 ptr to the fcb_entry to be returned
  * @return 0 on success; non-zero on failure
  */
 int
-fcb_offset_last_n(struct fcb *fcb, uint8_t entries, uint32_t *last_n_off)
+fcb_offset_last_n(struct fcb *fcb, uint8_t entries,
+        struct fcb_entry *last_n_entry)
 {
     struct fcb_entry loc;
     struct fcb_entry start;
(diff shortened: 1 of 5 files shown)


utzig:
@mkiiskila @vrahane This pull request implements a change to `fcb_offset_last_n` to return a `fcb_entry`. Since you both have worked on this code, would be nice to get a review. Something that I might have involuntarily changed might be a "offset by one" error. But I added a test now and I assumed that offset 0 refers to the last element...

fs/fcb/src/fcb.c
```diff
@@ -205,14 +205,15 @@ fcb_sector_hdr_read(struct fcb *fcb, struct flash_area *fap,
...
+ * @param2 ptr to the fcb_entry to be returned
  * @return 0 on success; non-zero on failure
  */
 int
-fcb_offset_last_n(struct fcb *fcb, uint8_t entries, uint32_t *last_n_off)
+fcb_offset_last_n(struct fcb *fcb, uint8_t entries,
+        struct fcb_entry *last_n_entry)
 {
     struct fcb_entry loc;
     struct fcb_entry start;
```
mkiiskila: start is not needed. you can just use last_n_entry directly.

fs/fcb/src/fcb.c
```diff
@@ -223,18 +224,17 @@ fcb_offset_last_n(struct fcb *fcb, uint8_t entries, uint32_t *last_n_off)
...
+        else if (i > entries) {
             fcb_getnext(fcb, &start);
-            *last_n_off = start.fe_elem_off;
+            *last_n_entry = start;
         }
         i++;
     }
 
-    return 0;
+    return (entries + 1) > i ? OS_ENOENT : 0;
```
mkiiskila: This should return 0 even if fcb has less than number of entries asked. User of this interface is trying to copy up to 'entries' number of fcb elements.

utzig:
Updated to return error only if there are no fcbs available and number of entries is considered as 1-indexed (was previously offset by one). Documentation was updated to explain it better (I hope!).

mkiiskila:
Thanks for the unittest for this.

vrahane:
Looks great utzig. Thank you for the unit tests for this.

PR #177, opened 2017-02-14
nimble/gap: Fix validation of conn parameters

rymanluk:
This fix obvious incorrect check for parameter validation.


sterlinghughes:
are we ready to merge this?

rymanluk:
I think that one also has some problem with unit test. I scheduled this work for tomorrow. Will put comment when ready.

rymanluk:
ok, had to fixed supervision timeout in unit test so now it is according to spec. 
it's ready to merge

rymanluk:
Added some more unit tests. still ready to merge

ccollins476ad:
Thanks, Łukasz.

PR #182, opened 2017-02-21
Couple of fixes after upf56

rymanluk:


net/nimble/host/src/ble_att.c
```diff
@@ -452,7 +452,7 @@ ble_att_chan_mtu(const struct ble_l2cap_chan *chan)
 
         mtu = BLE_ATT_MTU_DFLT;
     } else {
-        mtu = min(chan->my_mtu, chan->peer_mtu);
+        mtu = min(ble_att_preferred_mtu_val, chan->peer_mtu);
```
ccollins476ad: Hi Łukasz,

Sorry if I am just being slow, but could you explain why this change (and the one below) is necessary?  I am envisioning this sequence:

1. We exchange MTU with peer: we specify 527, peer specifies 800; connection MTU is now 527.
2. Application sets preferred MTU to 100.

At this point, I would expect the connection MTU to be 527.  I believe your change causes the host to behave as if the MTU is 100.  We need to be prepared to receive 527-byte packets, since that is what we told the peer we support, so it seems the ble_att_chan_mtu() function should return 527.

But maybe I am misreading the code, or I forget how this works.

Thanks,
Chris

rymanluk: Hi Chris,

Ahh I missed that preferred MTU can be changed by user any time and that it actually set chan->my_mtu which is not what I wanted.
My goal is to make sure chan->my_mtu is actuall MTU as negotiated over the air.
Main reason for this is to make sure channel does not get more data on the channel than it is allowed.

Anyway, I need to keep track on MTU sent over the air to remote device. I will use proc from gattc as exchange procedure is attached to that.  BTW I would move it to ble_att but this is topic for later maybe.
When it comes to  unit tests, I did some helper function for exchange mtu procedure as now we cannot just set chan->my_mtu without exchanging with peer.
I think that should work like this.

Thanks
Łukasz

rymanluk: Now that branch should be update according to my comment.

ccollins476ad:
I guess I don't understand why this needs to be changed at all.  Was something broken prior to the change?  If I'm not mistaken, `ble_att_chan_mtu()` would always return the correct number (lesser of two exchanged values, or 23 if the exchange never happened).

rymanluk:
Reason is that I'm starting to check chan->my_mtu on data receiving in order to block invalid data coming on the channel.  Now I see  that for ATT CID I should do check against ble_att_chan_mtu() instead of chan->my_mtu, right? Otherwise for ATT CID we are accept 527 bytes even before exchange procedure and that is wrong.

Please have a look at this patch:
https://github.com/apache/incubator-mynewt-core/pull/182/commits/a69a1893aa27a3dc7e6cfa3d6964cca9d2e8c539

Hmm, as a quick fix, if you really don't like this one, I could create some helper function to retrieve MTU from the channel and there we could hide different handling for ATT. This helper function would be a part of  fix I mentioned and in theory this patch should not be needed.

As already mentioned, with this patch chan->my_mtu is always the one used over the air, which for me is cleaner.

Hope that makes things clear. What do you think, which way we should go?

ccollins476ad:
I'm just trying to understand why it is better than the original code.  It's an honest question: was there a bug in the old code?  Sorry if you have already answered this and I missed it.

This looks like extra code that potentially introduces bugs, so I'm just trying to understand why it's necessary.  I do see one issue here:

```
/* Get MTU which was sent on MTU Exchange request */
int
ble_gattc_rx_mtu_get_sent_mtu(uint16_t conn_handle, uint16_t *sent_mtu)
{
    struct ble_gattc_proc *proc;

    proc = STAILQ_FIRST(&ble_gattc_procs);
    while (proc != NULL) {
        if (proc->conn_handle == conn_handle && proc->op == BLE_GATT_OP_MTU) {
            *sent_mtu = proc->mtu.sent_mtu;
            return 0;
        }
        proc = STAILQ_NEXT(proc, next);
    }

    return BLE_HS_ENOENT;
}
```

The ble_hs_mutex needs to be locked while the GATT procedure list is iterated.  But going back to my original point: why is this function necessary?  What was wrong with the original approach?  I understand why we would want to drop L2CAP frames that are larger than the MTU, but why does that require this change?

rymanluk:
"I'm just trying to understand why it is better than the original code. It's an honest question: was there a bug in the old code? "

Maybe I was not clear. Story is as follow:
I did patch https://github.com/apache/incubator-mynewt-core/pull/182/commits/a69a1893aa27a3dc7e6cfa3d6964cca9d2e8c539 in first place and it turns out it does not work on ATT CID as I wanted. As we conclude on previous comment, it is because for ATT CID I should use ble_att_chan_mtu() to get actual MTU. Back then I just missed that,  so I came up with this patch which makes sure that chan->my_mtu is the actual MTU not the preferred one. My idea was like this: let's make chan->my_mtu to have always is actual MTU for all the possible channels. Just make them consistent. But If I would not missed ble_att_chan_mtu() in first place, I would probably not work on this patch at all.

To conclude, this patch is not really due to some obvious bug in current code but rather due to my misunderstanding of how should I use chan->my_mtu from ble_att. I'm really OK to just skip that patch and do change as mentioned in previous comment. I will check if that work fine.

You asked about 
int
ble_gattc_rx_mtu_get_sent_mtu()
In this approach we need to keep track on MTU we send in exchange MTU request, because as you point out in very fist comment, preferred MTU can be changed by application user any time. So the idea was to keep MTU we sent in the gattc proc (in progress procedure). Later when handling response, we are checking what MTU we sent and take min from this and peer mtu. 

Hope that's clear now.

Chris, I'm going to change the approach and improve https://github.com/apache/incubator-mynewt-core/pull/182/commits/a69a1893aa27a3dc7e6cfa3d6964cca9d2e8c539 that. If that works as it expected I will just skip this patch.

rymanluk:
Alright, now it should be OK.

ccollins476ad:
Looks good, thanks!  I will merge it.

I just have one comment: there are a few places in ble_l2cap_sig.c where we drop an incoming request if we fail to allocate a buffer for the response.  I think it would be good to arrange for this never to happen.  If the ATT code fails to allocate a response, it reuses the request mbuf for an error response.  Anyway, maybe something for the future.

By the way, I *finally* understand the MTU thing.  I was confused because I didn't realize there is a slight difference in semantics between the ATT channel and other L2CAP channels.  In ATT, both sides use the lower of the two MTU values.  In other L2CAP channels, each side tells the other what it can receive, so the MTU varies per direction.  Anyway, I am sure you already realized this, but I wanted to explain my prior confusion!

PR #185, opened 2017-02-24
Sensors branch

vrahane:


apps/bletiny/src/gatt_svr.c
```diff
@@ -366,7 +366,7 @@ static int
 gatt_svr_access_test(uint16_t conn_handle, uint16_t attr_handle,
                      struct ble_gatt_access_ctxt *ctxt,
                      void *arg)
-{
+
```
devlware: need to keep {

utzig: This looks more like a GH quirk. This was a complicated merge but the resulting file looks ok to me.

vrahane: I agree, it was indeed a complicated merge. Thank you for the review  @utzig and @devlware. I fixed it in a subsequent commit.

PR #192, opened 2017-03-04
Fix invalid memory accesses in ble_uuid_cmp

simonratner:
When the two uuid values differ in type, one of two things can happen:

1. Access to unallocated or uninitialised memory
2. Unaligned access to 16/32-bit values

Both of these cause crashes, so always make sure we are comparing like types.


ccollins476ad:
Thanks, Simon.

PR #194, opened 2017-03-05
Fix ctrl c win

mkiiskila:


mkiiskila:
Argh, I already forgot about the ctrl-c with another terminal. Let me give this another go.

mkiiskila:
And that does not work for me. I get the following.

'"start"' is not recognized as an internal or external command,
operable program or batch file.

The following kind of escaping would be ok for me as well.
"$COMSPEC /C" start "$COMSPEC /C" arm-none-eabi-gdb -x $GDB_CMD_FILE $FILE_NAME

I'm tempted to go all batch with these.

PR #195, opened 2017-03-06
Fix printing not found image and tabulation

utzig:
- This fixes printing the name of an image when it is not found (eg user forgot to run `create-image`).
- I changed the mix of tabs/spaces to space only which is responsible for the somewhat messy diff.

### DIFF
--- a/hw/scripts/common.sh
+++ b/hw/scripts/common.sh
@@ -21,11 +21,10 @@
 #
 common_file_to_load () {
     if [ "$MFG_IMAGE" ]; then
-	FILE_NAME=$BIN_BASENAME.bin
+        FILE_NAME=$BIN_BASENAME.bin
     elif [ "$BOOT_LOADER" ]; then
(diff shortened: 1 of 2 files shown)


mkiiskila:
Hi,

I actually like seeing the full path name in this message, as then I can cut'n'paste the directory name when I start poking further.

You have a different opinion?

utzig:
I do copy/paste the relative paths without issues, not sure what's your problem. But I would not mind removing the code that removes the path prefix. The main problem was that it was not printing anything before, because it was using a variable `FILE` instead of `FILE_NAME`. Everything else is "noise".

mkiiskila:
Ah, I see now. The actual change was lost in the noise of the other lines changing. I'm fine with this kind of name stripping. I'm bringing this change in along with the same update to jlink script.

utzig:
This was merged but not automatically closed.

PR #196, opened 2017-03-07
nimble/gap: Fix bad calculation on connection parameters validation

rymanluk:


ccollins476ad:
Thanks, Łukasz.  I wonder where that 6 came from...

rymanluk:
Don't ask me where that 6 came from. It must be some error in the Matrix. I
really could not remind

PR #197, opened 2017-03-07
[RFC] Shell improvements

michal-narajowski:
Hi,

for the last few weeks, I've been working on improving the shell functionality.
I prepared a separate app so that it's easy to work on it separately and play 
around with it.

Here are some features of the new shell:

1. First of all, the new shell improves the concept of modules. Modules group
commands under a namespace. So like in bletiny there is a module named 'b' 
and contains commands like 'adv', 'conn', 'term', etc. The module name is so 
short because you have to type it before every command: 'b adv ...'. 

With the new shell, you can select a module to be the default module and just type in
the command names from this module. To select a module type: 'select [module]'. 
To go back to the main shell type 'select'. When you select a new default module, 
the prompt changes:

Default:
'shell> '
selected:
'os> '

The newshell app has two example modules: 'os' and 'prompt'. 'os' contains
system commands from the old shell. 'prompt' contains 'ticks' command.
It can toggle a dynamic shell. A dynamic shell could be used in Bluetooth app
to display name or address of a connected device.

2. Next major improvement is tab completion. You can complete modules,
commands, and params. When there is more than one option with a specific prefix 
they are printed to the console followed by the current line so
that you can continue to type the name.

```
help
tasks
mpool
date
shell> os <tab>
```

3. Improved help messages
Each command can have additional information associated with it. There is
a summary string, which provides a brief description of the command. It is displayed
when listing all possible commands in a specific context. Usage string usually 
contains more info and is displayed when the user invokes help command or when
the user mistyped some parameter for this command. Moreover, a list
of parameters can be passed with the command for tab completion of parameters
and displaying short descriptions of params.

The help api is designed, so the app can easily compile out the additional
help info to save flash.

4. I separated UART from the shell and added RTT as an optional backend.
RTT is a very high-speed memory-mapped I/O. More info here: 

https://www.segger.com/jlink-rtt.html

So now shell can be used with either RTT or UART. It frees up UART so it 
can be used for some other purposes while still having shell functionality.

Comments are welcome. 

Best regards,
Michał


jacobrosenthal:
Slightly Off topic, thoughts on newtmgr transports?

sjanc:
general idea is to be able to run shell on rtt and newtmgr on uart

mkiiskila:
Really do appreciate the effort.

Here's some feedback, in random order when I'm trying this out:
1. I suggest you don't modify baselibc for this; instead of using __stdout_hook there, move that to live inside console.c. This way this can be built to native platform when using their libc. Also console_printf() is currently dependent on having baselibc underneath. Take a look at what was done in console/full for this (which does preconstruct the buffer before calling console_write; this is not the greatest either, less stack use would be better).
2. I would like to keep uart_console.c still asynchronous. I.e. TX is fed using tx interrupts like it was before. This can be achieved by bringing back the small ringbuffer for it.
3. Does shell need it's own task? If it could be made event driven like the current shell, it could be used with more projects (less tasks -> less stacks).
4. syscfg.h should be included before other mynewt headers get included (libc stuff is fine); as those header files might have conditional stuff depending on syscfg settings
5. rtt_console.h is missing apache license
6. I had to include <string.h> from shell_prompt.c to get it to compile for sim platform.
7. I really want to continue having timestamps in my console output. Not for the lines starting with prompt, but for stuff which is 'logging' output from packages.
8. For some reason, my backspace was not having desired effect. I was using uart as my output, and minicom as my terminal.
9. Segger is not used with most of the MCU targets. We should default the console to UART.
10. You could investigate use of os_cputime for polling with Segger (I assume interrupts are not possible?), and avoid the dedicated task for polling.

Hope this helps,
M

jacobrosenthal:
Sort of off topic again, but Segger have released a jlink firmware for microbit that includes RTT! So theres a free target out there for anyone to check out.
https://www.segger.com/bbc-micro-bit.html

Note: I reported the lack of CDC serial support in the original firmware, and theres a new release as of today which adds that.

michal-narajowski:
Hi Marko,

Thanks for a great review! Really appreciate it. Here are my answers:

> 1. I suggest you don't modify baselibc for this; instead of using __stdout_hook there, move that to live inside console.c. This way this can be built to native platform when using their libc.

That's probably a good idea. I changed baselibc because I thought that Mynewt specific stuff for baselibc was in mynewt.c.

> Also console_printf() is currently dependent on having baselibc underneath. Take a look at what was done in console/full for this (which does preconstruct the buffer before calling console_write; this is not the greatest either, less stack use would be better).

It was dependent earlier too, the string was formatted into a buffer before calling console_write, like you said. Maybe if we returned to asynchronous (2.) uart console we could avoid the console_printf buffer and just format into the uart ringbuffer.

> 2. I would like to keep uart_console.c still asynchronous. I.e. TX is fed using tx interrupts like it was before. This can be achieved by bringing back the small ringbuffer for it.

Ok, I will try to make it asynchronous.

> 3. Does shell need it's own task? If it could be made event driven like the current shell, it could be used with more projects (less tasks -> less stacks).

Ok, I will try to fix that. 

> 4. syscfg.h should be included before other mynewt headers get included (libc stuff is fine); as those header files might have conditional stuff depending on syscfg settings

Will fix that.

> 5. rtt_console.h is missing apache license

Same

> 6. I had to include <string.h> from shell_prompt.c to get it to compile for sim platform.

Same

> 7. I really want to continue having timestamps in my console output. Not for the lines starting with prompt, but for stuff which is 'logging' output from packages.

Ok, I will try to keep the same behavior (we could make it optional).

> 8. For some reason, my backspace was not having desired effect. I was using uart as my output, and minicom as my terminal.

In minicom you have to change "Backspace sends" to DEL. In PuTTY set Backspace key to Control-? (127). I don't know if it should be like this. Comments on how to solve it are welcome. At worst case, I can document it in README.

> 9. Segger is not used with most of the MCU targets. We should default the console to UART.

I already changed that. It should be on my branch "migrate-shell" https://github.com/michal-narajowski/incubator-mynewt-core/tree/migrate-shell, which is still work in progress.

> 10. You could investigate use of os_cputime for polling with Segger (I assume interrupts are not possible?), and avoid the dedicated task for polling.

That's a great idea. I will look into that. With RTT we need to pull, unfortunately.

Best regards,
Michał

michal-narajowski:
Hi,
I applied suggestions from all the points except for 2. I don't know how to solve it yet, with the current abstraction level in the console. I force-pushed the changes on `rtt-console`. `migrate-shell` is now on the same commit as `rtt-console`. 

Best regards,
Michał

jacobrosenthal:
Ping mynewt crew :) 
Im really excited about getting rtt shell support upstreamed too.

michal-narajowski:
Hi everyone,
I'm splitting this big pull request into smaller ones that will be easier to digest. I submitted a pull request (https://github.com/apache/incubator-mynewt-core/pull/230) with console improvements that are still compatible with existing apps, so the merge should be painless. You can use the console with UART or RTT backend now. New shell will be added later. BTW async uart console is back :) 
Closing this one.
Best regards,
Michał

PR #199, opened 2017-03-08
Add missing include for the AT45DB flash driver

utzig:
PR #162 which added the driver for AT45DB was missing the include
file.


utzig:
This was already merged but have not been automatically closed.

PR #201, opened 2017-03-09
Adgbranch

gemcore:
There are 2 different ports included in this PR:
(1) The  'hw/bsp/nina-b1' port configures the nRF52832 UART console pins to match a standard 'ublox NINA-B112 BLE module'. Not much has been done so far. The idea is to replace the supplied ublox 'AT command' and bootloader firmware, which relies on the Nordic Soft Device stack with MyNewt OS and NimBLE stack.
(2) The 'hw/bsp/sensorhub' port is for a board that uses the STM32F427xx MCU where the serial console pins are configured to UART4. I have a custom board with a lot of sensors including an audio I/O interface running a FFT to differentiate various sounds that I would like to use. I can't release proprietary information, however many are I2C interfaces, so that would be something I'd like to be able to support on this port.
ALan


mkiiskila:
Very nice.

Looks like there's 2 accidentally checked in files, 'diff.txt' and 'hw/bsp/sensorhub/f407.cfg'. Could you remove these?

Also, don't modify repository.yml. This only has effect if it's done in the master branch, and mods done in develop will make merges harder. But this is not that big of an issue.

After those 2 changes, I think this is ready to be merged in.

gemcore:
Sure no problem. I'm not sure how the file repository.yml got modified
because I certainly didn't intend to do that. The 'sensorhub' name isn't
particularly descriptive. Do you think this should be changed? 

  _____

mkiiskila:
Now that you point it out, sensorhub is not the best of names :) I can't deduce which board this is for. Would I be able to buy it from somewhere?

gemcore:
Yeah unfortunately it's not a board that I'm able to sell directly. I
suppose I could design something to sell through my own company, but hadn't
planned on that. Perhaps I'll look around and see if someone has a board out
there that is based on this processor. 
 
ALan
 
 

  _____

aditihilbert:
Hi Alan, thanks for the sensorhub! Do you mind publishing the schematic of your board in this directory?

gemcore:
Hi Aditi,

Unfortunately I can't release the schematic of the custom Sensor Hub board I'm using because it belongs to the company Delta Controls​ which I'm now working for. I had mentioned to Marko that the closest open source board I could find out there was this nice one from Adafruit:
http://www.netduino.com/netduino3/

There is a problem also that Adafruit has discontinued the board. It's still listed as available through Mouser for $39. BTW I also thought about making a similar sensor board that could be offered through my company Gemcore, but I'm not sure it would be worth the effort. What do you think?

Regards,
ALan

PR #204, opened 2017-03-11
Fix stm32f429 nvic num vectors

grapherd:
From st datasheet page 24:

    The devices embed a nested vectored interrupt controller
    able to manage 16 priority levels, and handle up to 91
    maskable interrupt channels plus the 16 interrupt lines
    of the Cortex®-M4 with FPU core.


utzig:
LGTM

PR #206, opened 2017-03-14
Fixes for L2CAP LE CoC found when doing unit tests for it

rymanluk:
Some fixes for issues found during writing unit tests + initial unit tests + pts project file + pts instructions

BTW. Those patches shows that we need reference counter for L2CAP Channel. I've created JIRA for that #710


ccollins476ad:
Thanks, Łukasz.

PR #208, opened 2017-03-16
Fix timtest timer index

louisom:
Change timtest timer index start at 0

### DIFF
--- a/apps/timtest/src/main.c
+++ b/apps/timtest/src/main.c
@@ -32,7 +32,7 @@
 #define TASK1_STACK_SIZE    OS_STACK_ALIGN(64)
 struct os_task task1;
 
-#define TASK1_TIMER_NUM     (1)
+#define TASK1_TIMER_NUM     (0)
 #define TASK1_TIMER_FREQ    (4000000)
 
 /* Task 2 */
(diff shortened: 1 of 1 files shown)


utzig:
LGTM

wes3:
I do not think we want to make this change. The timtest application is a bit of a tricky one as it sort of depends on bsp/mcu for timer support. If the mcu does not have enough timers this application will not work. I think pretty much every mcu will have a few timers so was not terribly worried about this.

Another issue with this change is that TIMER_1 and TIMER_2 syscfg values are set to 1 in syscfg.yml for this app. That is why main.c uses TIMER_1 and TIMER_2.

A better way to change this might be to determine if the timer is actually enabled in main.c and use timers that actually exist without having to really on syscfg.yml (which may not apply to the bsp).

One final note: timtest should not use a timer that the os is using for its timing. So that is another thing to worry about.

grapherd:
@wes3 Or maybe we can move these setting to `pkg.yml`? since I saw that not every bsp have the same setting of timer.

aditihilbert:
@lulouie, this is best handled by the specific app. Coming up with something generic that works for all different bsp/mcu is too tricky. We will be closing this PR.

skiiskila:
@lulouie, could you please close the PR (based on the history and comments).

PR #213, opened 2017-03-23
Expose I2C0 on STM32F4 Discovery board

antoinealb:
This commit expose the I2C0 interface on the STM32F4 discovery board. I am not very familiar with Mynewt yet, so sorry if this is a weird way to add a peripheral to a board. I took most of the code from the Olimex E407 and adapted it to this board.


mkiiskila:
No worries. This looks fine.

PR #215, opened 2017-03-24
ATT size reduction and endianess API cleanup

sjanc:
Hi,

This turned out a bit bigger than I originally anticipated so sorry for size of this PR.
In short this makes ATT use packed structs for received and send data. But instead
of using packed stack variables it cast those to mbuf memory. This allows to construct
packed directly in mbuf while sending and parsing incoming data without extra memcpy.

Besides that it also gets rid of awkward TOFROM16() endianness swap API in favor of
using os/endian.h

In general it saves around 1.1-1.3kB of flash memory depending on arch.
Size reduction measured for net_nimble_host.a when build for bletiny application:
on nRF52     40130 -> 39026
on nRF51    37732  -> 36420


rymanluk:
Looks good to me

ccollins476ad:
Looks good to me, except for one thing: there is a coding standard prohibition against declaring / defining variables in the middle of a function:

> * Place all function-local variable definitions at the top of the function body, before any statements.

For better or worse, that's the designated style, and it is best to try to stay consistent.

sjanc:
Hi Chris,

Yes, Will mentioned this to me already.  I'll fix that while rebasing to resolve conflicts.

Thanks!

sjanc:
Rebased, conflicts solved and Chris' comment addressed.  Should be ready to go :)

PR #217, opened 2017-03-28
Basic pin mapping for RedBear Blend2

pctj101:
Based mostly on BSP for rb-nano2.

Updated pin mapping for UART0
- LED Pin remapped
- SPI0 remapped/enabled
- I2C-0 remapped/enabled
- UART1/TWI-1/SPIM-1  - remapped/disabled due to pin conflicts

### DIFF
--- a/hw/bsp/rb-blend2/include/bsp/boards.h
+++ b/hw/bsp/rb-blend2/include/bsp/boards.h
@@ -1,0 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
--- a/hw/bsp/rb-blend2/boot-nrf52xxaa.ld
+++ b/hw/bsp/rb-blend2/boot-nrf52xxaa.ld
@@ -1,0 +1,25 @@
+/* Linker script for Nordic Semiconductor nRF5 devices
+ *
+ * Version: Sourcery G++ 4.5-1
+ * Support: https://support.codesourcery.com/GNUToolchain/
+ *
+ * Copyright (c) 2007, 2008, 2009, 2010 CodeSourcery, Inc.
+ *
+ * The authors hereby grant permission to use, copy, modify, distribute,
+ * and license this software and its documentation for any purpose, provided
+ * that existing copyright notices are retained in all copies and that this
+ * notice is included verbatim in any distributions.  No written agreement,
+ * license, or royalty fee is required for any of the authorized uses.
+ * Modifications to this software may be copyrighted by their authors
+ * and need not follow the licensing terms described here, provided that
+ * the new terms are clearly indicated on the first page of each file where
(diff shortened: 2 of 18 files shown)


wes3:
The pull request looks good except for one item: it appears you have mapped the LED_BLINK_PIN (P25) to both the LED and SPI0 SCK pin. I took a look at the schematic and P25 does seem to be the LED pin (although not sure what that MCP6002T chip is) so you will need to map SCK to some other pin. Note that I do not have a blend2 board with which to test and admittedly I did not go through every line of the .ld files and scripts so am taking some of this on faith :-)


To answer your questions:
1) Best way to make UART1 optional is to go into the syscfg.yml file and change UART_1 to value: 0
2) Looking at the schematic it appears that there are two sets of pins for i2c: SDA1/SCL1 (pins 26/27) and SDA/SCL (pins 13/12). So I would map i2c to those pins. None of those pins appear mapped already in your syscfg or hal_bsp.
3) I would not map SPI1 personally. There are not a ton of GPIO that go out to the connectors. If folks need another SPI they should just map it in their BSP.

pctj101:
Believe it or not... P25 is "officially" shared with SPI0_SCK
https://github.com/redbear/nRF5x#blend-2

Should we follow the "official" image, or... do you think we should make a command decision to make something more... "useful"?

Looking into the other items...

pctj101:
Wanted to post an update in regards to your feedback.  

I added I2C_0 to syscfg/hal_bsp

I left UART_1 configured as value:1 because "most arduino shaped boards like the Blend2" have serial on those pins.

I'll post another note when I have time to test the new pin mappings.

pctj101:
Regarding the .ld/script files - since the rb-nano2 and rb-blend2 share the same "Redbear MB-N2" module, it's quite likely that the rb-nano2 settings can simply be used for the rb-blend2.

wes3:
I would not share the LED with SCK personally. There are other pins to use and you can use any pin for any peripheral so keeping them separate makes sense to me

hw/bsp/rb-blend2/include/bsp/boards.h
```diff
@@ -0,0 +1,19 @@
+/*
```
utzig: Is this file supposed to be only a license header?

pctj101: Ah I blindly copied it from rb-nano2... so I meant nothing other than the license comment already included in this file.

pctj101:
@wes3 You know what's crazy? In the Arduino Uno pinout, SCK and LED are actually officially shared.

Does MyNewt have any intention of following the Arduino Uno pinout for boards like the Blend2 which are based on the Arduino Uno pinout?

wes3:
I cannot speak for all of Mynewt obviously but I think folks generally try to allocate the pins such that there is one i2c interface, one spi interface, and the LEDs are not shared. I see no reason to commit a bsp that shares any of these unless there is not enough gpio. I do not think that is the case with any of the boards you mention (although I could be wrong).

pctj101:
@wes3 Here's some more info for both sides of the coin:

http://playground.arduino.cc/Main/ShieldPinUsage
Any correctly designed shield will connect the SPI lines(MISO, MOSI, SCK) via the ICSP header instead of the standard Arduino I/O pin headers because the SPI is found on different pins depending on the board.

But.... the Blend2 board (in the pinning for the Arduino Uno) doesn't have an ICSP header, so.......

Would it be better to at least have the Arduino Uno standard available? (not a super strong argument in light of even Arduino saying "don't do it")

Anyways, this is the only reason I can find to keep SCK and LED on the same pin.

Otherwise, I totally agree with you.

pctj101:
I'm totally out of good reasons to support keeping with the Arduino Uno SPI standard especially when Arduino says don't do it.  

So I picked the next nearest pin which is unassigned.

Is there a suggested place to document this "deviation" from the official "RedBear Blend2 pinout diagram" so that new users aren't surprised?

wes3:
You could submit a tutorial here that describes getting the blend2 up and running: http://mynewt.incubator.apache.org/latest/os/introduction/

Otherwise a comment in the code is fine :-)

wes3:
It looks good; commit has been merged.

utzig:
@pctj101 Please close this issue, the PR was already merged but the issue was not automatically closed. Thanks!

PR #218, opened 2017-03-30
PIC32 port

IMGJulian:
PIC32 port including BSPs targeting the Digilent [Wi-Fire board](http://store.digilentinc.com/chipkit-wi-fire-wifi-enabled-mz-microcontroller-board/) and the [6LoWPAN Clicker board](https://docs.creatordev.io/clicker/guides/quick-start-guide/). Kernel mode only, soft floating point only.

### DIFF
--- a/compiler/xc32/compiler.yml
+++ b/compiler/xc32/compiler.yml
@@ -1,0 +1,36 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+compiler.path.cc: "xc32-gcc"
+compiler.path.as: "xc32-gcc"
+compiler.path.archive: "xc32-ar"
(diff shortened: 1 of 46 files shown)


compiler/xc32/compiler.yml
```diff
@@ -0,0 +1,36 @@
...
+#
+
+compiler.path.cc: "xc32-gcc"
+compiler.path.as: "xc32-gcc"
+compiler.path.archive: "xc32-ar"
+compiler.path.objdump: "xc32-objdump"
+compiler.path.objsize: "xc32-size"
+compiler.path.objcopy: "xc32-objcopy"
+
+compiler.flags.base: -I/opt/microchip/xc32/v1.42/pic32mx/include -std=gnu11
```
utzig: Is this a microchip SDK path? Apart from this hardcoded path, if I decided to build pic32 gcc myself would this port work or it's limited to microchip's own sdk?

utzig:
Looks good to be merged.

PS: What would be the difference between pic32mx and pic32mz? (Their respective arch files look very similar)

compiler/xc32/compiler.yml
IMGJulian: The hardcoded path was something I meant to ask about and it slipped my mind... I am not sure how to improve this under your newt tool. It would be nice if the user could specify a path in a more intuitive manner should there be a non-default install of XC32 or a version change. Any suggestions would be appreciated.

Currently our GCC implementation does not fully support the PIC External interrupt controller. The choice was made to use Microchip's XC32 in order to make use of multiple interrupt vectors (making it significantly easier for the user to add their own ISRs).

As for the similarities between the MX and MZ kernel ports, I belive they are currently the same. The MZ port still has some functionality to drop in so to minimise the changes that would need to be made in the future I made 2 copies, Thinking about it now, it might be better to use preprocessor options to conditionally compile from one file. Again, any suggestions would be appreciated.

IMGJulian:
I belive they are currently the same. The MZ port still has some functionality to drop in so to minimise the changes that would need to be made in the future I made 2 copies, Thinking about it now, it might be better to use preprocessor options to conditionally compile from one architecture. Again, any suggestions would be appreciated.

PR #219, opened 2017-03-31
Start work on BT 5 features

rymanluk:


rymanluk:
It went to bluetooth5 branch

PR #221, opened 2017-04-04
net/nimble; add HCI transport which has Linux bluetooth socket,

mkiiskila:
or TCP socket as it's target.

### DIFF
--- a/net/nimble/transport/socket/src/ble_hci_socket.c
+++ b/net/nimble/transport/socket/src/ble_hci_socket.c
@@ -794,4 +794,9 @@ ble_hci_sock_init(void)
 
     rc = ble_hci_sock_config();
     SYSINIT_PANIC_ASSERT_MSG(rc == 0, "Failure configuring socket HCI");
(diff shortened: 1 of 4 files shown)


net/nimble/transport/socket/src/ble_hci_socket.c
```diff
@@ -0,0 +1,797 @@
...
+
+    rc = bind(s, (struct sockaddr *)&shci, sizeof(shci));
+    if (rc) {
+        dprintf(1, "bind() failed %d\n", errno);
+        goto err;
+    }
+
+    rc = 1;
+
+    rc = ioctl(s, FIONBIO, (char *)&rc);
```
sjanc: why not use fcntl() for manipulating fd instead of ioctl?

mkiiskila: I've been preferring ioctl() over fcntl() with this use because it's one system call vs two :)

PR #222, opened 2017-04-05
Add Privacy modes and other basic stuff

rymanluk:


apps/bletiny/src/cmd.c
```diff
@@ -2464,6 +2464,46 @@ static struct kv_pair cmd_set_addr_types[] = {
...
+    if (rc != 0) {
+        console_printf("invalid 'addr' parameter\n");
+        help_cmd_byte_stream_exact_length("addr", 6);
+        return rc;
+    }
+
+    priv_mode = parse_arg_uint8("mode", &rc);
+    if (rc != 0) {
+        console_printf("missing mode\n");
+        return -1;
```
ccollins476ad: Certainly not a big deal, but should this be `return rc` rather than `return -1`?

rymanluk: ah correct, thanks!

rymanluk:
a) yes, will add check for valid mode
b) will add additional patch on top with ble_ll_is_controller_busy(), I think that should work. please review.
c) yes, I forgot to do the check if resolving list is enabled. Actually it does not break anything because if it is not enabled we will basically not find the IRK on the list. However I will add it

PR #223, opened 2017-04-05
Log testsuite name and keep valid json when truncating

spoonofpower:


ccollins476ad:
This looks good to me, with one small caveat: strncpy() doesn't necessarily null-terminate its destination parameter, so there is a chance a non-string gets passed to the log macro.  strncpy() is a pretty lame function.  It zero-fills the entire unwritten part of the destination buffer for some reason, and if the entire buffer gets written by the input string, no '\0' gets written.

A simple solution is to use strncat() rather than strncpy().

ccollins476ad:
Sorry, I misread the code.  There is no bug here.  Looks good to me.

spoonofpower:
I moved the PR from develop to master.

PR #224, opened 2017-04-06
controller: Support for LE Channel Selection Algorithm #2

sjanc:
This adds implementation of Channel Selection Algorithm #2 as specified in Core Specification 5.0 Vol 6 Part B 4.6.14.

Includes required HCI event, changes to supported features masks and unit tests based on sample data from Core Specification.

### DIFF
--- a/net/nimble/controller/src/ble_ll_conn.c
+++ b/net/nimble/controller/src/ble_ll_conn.c
@@ -599,6 +599,7 @@ ble_ll_conn_remapped_channel(uint8_t remap_index, const uint8_t *chanmap)
     return 0;
 }
 
+#if (MYNEWT_VAL(BLE_LL_CFG_FEAT_LE_CSA2) == 1)
 static uint16_t
 ble_ll_conn_csa2_perm(uint16_t in)
 {
@@ -663,6 +664,7 @@ ble_ll_conn_calc_dci_csa2(struct ble_ll_conn_sm *conn)
 
     return ble_ll_conn_remapped_channel(remap_index, conn->chanmap);
 }
+#endif
 
 static uint8_t
 ble_ll_conn_calc_dci_csa1(struct ble_ll_conn_sm *conn)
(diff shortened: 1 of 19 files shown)


net/nimble/controller/src/ble_ll_conn.c
```diff
@@ -1858,10 +1945,17 @@ ble_ll_conn_next_event(struct ble_ll_conn_sm *connsm)
     }
 
     /* Calculate data channel index of next connection event */
-    while (latency > 0) {
-        connsm->last_unmapped_chan = connsm->unmapped_chan;
-        connsm->data_chan_index = ble_ll_conn_calc_dci(connsm);
-        --latency;
+    if (CONN_F_CSA2_SUPP(connsm)) {
+        /* TODO handle this better */
+        connsm->event_cntr += latency;
```
cvinayak: Why are you doing this? hasn't this been done in line 1845?

sjanc: Hi,

yeap, this is invalid.  Thanks for review:)

sjanc:
Hi Will,

I dropped couple of patches already after I rebased my branch related to building native BLE.

As for configuration of BT5 vs BT4.2 I'll send email to dev mailing list. There is question of granularity of this configuration eg CSA#2 is quite fundamental feature for BT5 and should probably be always enabled with BT5, but other things like Ext/Periodic Advertising, 2Mbit, LE CodedPHY etc could most likely be configurable. Anyway, I'll send email about this.

About CSA#2 correctness there is UT (as you noticed) based on sample data from spec. I was also able to successfully  IOP Mynewt implementation against initial draft implementation from Zephyr. So I think this is good enough for such fresh feature :)

wes3:
I agree on your BT5/4.2 config comment which is why I thought it was worth a larger discussion.

As for CSA#2 completeness, I was just trying to indemnify myself from any future issues :-) I agree that what you did is perfectly fine.

sjanc:
All right,  as discussed on ML we will go with always enabling BT5 trivials and then have opt-in for other optional feature.  Last few patches add this for CSA #2 feature.

wes3:
Looks good.

PR #228, opened 2017-04-11
Switch to mcuboot

utzig:
After this is merged, users need to `newt install -v` to fetch mcuboot.

Still needs updates to docs.


utzig:
Bye!

PR #229, opened 2017-04-12
Host: Support for read/set/set_default LE PHY

rymanluk:
Bletiny got commands for testing.
b phy read conn=1
b phy set conn=1 tx_phys_mask=1 rx_phys_mask=5 phy_opts=1
b phy set_def tx_phys_mask=7 rx_phys_mask=7

### DIFF
--- a/apps/bletiny/src/cmd.c
+++ b/apps/bletiny/src/cmd.c
@@ -3551,6 +3551,180 @@ cmd_svcchg(int argc, char **argv)
     return 0;
 }
 
+static const struct cmd_entry cmd_phy_entries[];
+
+static int
+cmd_phy_help(int argc, char **argv)
+{
+    int i;
+
+    console_printf("Available PHY commands:\n");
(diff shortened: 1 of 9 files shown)


ccollins476ad:
I just have a few minor comments:

1. In ble_gap_read_le_phy():
```
    /* First two octets is conn_handle. We can ignore it */
```

It looks like only the first octet gets ignored.  Maybe the comment is incorrect?


2. Just a question:
```
ble_gap_read_le_phy
ble_gap_set_prefered_default_le_phy
ble_gap_set_prefered_le_phy
```

Why do these functions declare the `dst` pointer and point it at the `buf` array?  It might be simpler to just use the array directly.

rymanluk:
1. Comment is good. Thanks for catching it.
2. This is indeed odd. I don't know why I did that. for a moment I was sure that everybody does it there, but just checked code and nobody does it. It's odd :) fixed.

ccollins476ad:
Thanks.  Looks good to me.

PR #230, opened 2017-04-13
Console improvements

michal-narajowski:
This patch brings improved console functionality.
Some of the features:
* support for different console backends; UART and RTT for now
* improved ANSI codes handling
* backward compatible API, so it doesn't break existing apps although the old API is deprecated and will be deleted in the future
* support for tab completion, which will be added later
* ability to compile out input handling code to save flash


rymanluk:
Looks good to me

mkiiskila:
This breaks builds for native targets, e.g. my_blinky_sim.

mkiiskila:
Also, newtmgr and OIC over shell breaks. Which might, or might not, be ok. Could you ask how much folks care about that in the mailing list? I myself use it occasionally, but I can change how I treat my targets. In the mailing list because not everyone gets these messages in the pull requests :)

I did not try serial update over bootloader; it uses 'console/minimal'. Have you had a chance to try that one out? We can fix it later if it breaks, but it would be nice to know that we'll need to do it.

Thanks,
M

microbuilder:
For us (Adafruit) breaking newtmgr over shell would be a big deal breaker for us and require a HW redesign, and we're at the production stage with a single hard-wired UART port into a USB serial bridge chip, with the assumption that shell, bootloader and newtmgr over serial would all work over that since USB connection.

WayneKeenan:
A -1 on breaking newtmgr over shell, from me personally and on behalf some other projects.

michal-narajowski:
Hi @mkiiskila,
could you give me some info on how to reproduce the newtmgr and OIC over shell bugs?
Best regards,
Michał

mkiiskila:
I use the app 'slinky' for this purpose; it has newtmgr over shell package enabled.
We do have a tutorial which talks about how to get newtmgr to do your bidding:
http://mynewt.apache.org/latest/os/tutorials/project-nrf52-slinky/

With your patch in, I enabled the console over uart setting.

Once that's done, start with 'echo' command. Also the 'image' family of commands is useful to test, this because it'll end up using bigger frames.

Another thing which is important to test the bootloader with newtmgr support in it.
Here's my target for it on nrf52:
targets/boot_nrf52_serial
    app=@apache-mynewt-core/apps/boot
    bsp=@apache-mynewt-core/hw/bsp/nrf52dk
    build_profile=optimized
    syscfg=BOOT_SERIAL=1

This target stops at bootloader to await for newtmgr commands, if you reboot the board while pressing the 'user' button on the board (I don't have a board with me now to say if it had more than
one button). Once it stops, you should be able to upload a new image to board using newtmgr.

Use 'newtmgr -lDEBUG' to get more debug info; specifically it'll print out bytes received/transmitted. Which will probably be useful when debugging this.

Hope this helps! You can also unicast me if you have further questions/need help.

mkiiskila:
Also, once you get the native target running, you can test this without actual hardware.
Here's my target for that.
targets/slinky
    app=@apache-mynewt-core/apps/slinky
    bsp=@apache-mynewt-core/hw/bsp/native
    build_profile=debug

It'll print out the pty name it created for console when it starts up. Then you can create a newtmgr connection profile which uses that as the UART device, and you should be able to send commands.
I find debugging things in simulator is easier. But YMMV.

Use 'newt run' when running under simulator, as it includes hw/bsp/native/sim.gdb automatically. Without that, single-stepping through code is arduous and painful.

PR #234, opened 2017-04-17
MYNEWT-729 SensorAPI: Send sensor data over OIC

vrahane:
Link: https://issues.apache.org/jira/browse/MYNEWT-729

This commit adds basic support for sending sensor data over OIC using the sensor framework. It is upto the app to create the gatt server and define a device address and device name. There is a sysconfig setting SENSOR_BLE_OIC using which you can specify if you would like to enable the functionality. You can find an example of the same in the sensors_test app.


hw/sensor/pkg.yml
```diff
@@ -23,6 +23,12 @@ pkg.author: "Apache Mynewt <dev@mynewt.incubator.apache.org>"
 pkg.homepage: "http://mynewt.apache.org/"
 pkg.keywords:
 
+pkg.deps:
+    - sys/shell
+    - sys/console/full
+    - sys/log/full
+    - sys/stats/full
```
mkiiskila: Instead of depending on specific implementation of stats, you should use pkg.req_apis.
This allows app builder to either include full stats, or the stub version. The same applies to log, and console.

hw/sensor/syscfg.yml
```diff
@@ -26,3 +26,16 @@ syscfg.defs:
...
+        value: 1
+        restrictions:
+            # Disable central and observer roles.
+            - BLE_ROLE_BROADCASTER
+            - '!BLE_ROLE_CENTRAL'
+            - '!BLE_ROLE_OBSERVER'
+            - BLE_ROLE_PERIPHERAL
+            # OC server, with bluetooth transport.
+            - OC_SERVER
+            - OC_TRANSPORT_GATT
```
mkiiskila: Instead of SENSOR_BLE_OIC, I recommend changing the knob to be SENSOR_OIC. I don't see anything bluetooth specific in the code, and you probably want people to use use hw/sensor OIC interface with Wi-Fi and/or ethernet.

hw/sensor/pkg.yml
vrahane: This was a remnant from the earlier implementation where the sensor library itself used logs for the oic implementation. It doesn't use that anymore so, it can be improved. However, I think keeping shell and console makes sense as they should be depended on by the sensor package. So, I am thinking of doing this instead:

pkg.deps.SENSOR_CLI:
   - sys/shell
   - sys/console/full

hw/sensor/syscfg.yml
vrahane: I  agree, this was another remnant of moving gatt server from the sensor library to the app. I will keep only the OC_SERVER syscfg as a restriction. Rest would be moved to the app.

PR #235, opened 2017-04-17
Lwip updates

mkiiskila:
Add driver for STM32 ethernet.
Make socket abstraction functional with LwIP.
Update LwIP version to 2.0.2


aditihilbert:
Passes all tests in sim and in test bench. This PR is good to be merged. 
Thanks for this addition!

PR #236, opened 2017-04-18
hw/bsp: MYNEWT-730 Add BSP for Ruuvi tag

wes3:
Add BSP for Ruuvi tag revision B3/B4. Not sure if this same BSP
will work on other Ruuvi Tag B revisions. Note that the UART
only uses tx/rx and goes to test points on the board (the
test points are TP10 and TP11).


aditihilbert:
Looks good. 
Thanks!

PR #237, opened 2017-04-19
MYNEWT-732: Fix ble_gattc_disc_svc_by_uuid

simonratner:
Broken in a refactor in 191acab73c9c77c157a806d8871b653f3c64cbc9, merged as part of https://github.com/apache/incubator-mynewt-core/pull/215.

### DIFF
--- a/net/nimble/host/src/ble_att_clt.c
+++ b/net/nimble/host/src/ble_att_clt.c
@@ -295,7 +295,7 @@ ble_att_clt_tx_find_type_value(uint16_t conn_handle, uint16_t start_handle,
         return BLE_HS_EINVAL;
     }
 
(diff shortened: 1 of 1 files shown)


ccollins476ad:
Thanks Simon.  If the host unit tests aren't catching this bug, then they are totally broken.  We should fix the tests before committing this fix.

simonratner:
With you on that. Commit 191acab73c9c77c157a806d8871b653f3c64cbc9 didn't touch any tests, so I assume there simply isn't one for this command. I would totally add one, but I still have not managed to get unittest to build on Windows :|

simonratner:
Might be worth looking at which commits from the `att` merge didn't come with test changes. Those changes were fairly well split up into individual commits per tx/rx command, so each one should in theory have a corresponding test change.

PR #239, opened 2017-04-19
MYNEWT-735 tinyprintf: negative field width spec.

ccollins476ad:
From the printf(3) man page on my OS X:

 `-'          A negative field width flag; the converted value is to
              be left adjusted on the field boundary.  Except for n
              conversions, the converted value is padded on the right
              with blanks, rather than on the left with blanks or
              zeros.  A - overrides a 0 if both are given.


aditihilbert:
Looks good. Passes all test bench tests.

PR #240, opened 2017-04-19
MYNEWT-734 console_printf() return # printed chrs.

ccollins476ad:
The *printf family of functions do this. It is useful for aligning text.


aditihilbert:
Looks good. Passes all CI tests.

PR #243, opened 2017-04-21
nimble/transport; MYNEWT-737: Make ble_hci_uart_tx_char more efficient

wes3:
Modified the code in ble_hci_uart_tx_char to be faster. This
added a bit more code (should not have increased ram usage) and
should in general be faster than the old method and should also
free the mbufs slightly faster (each one is freed as it is
used).

### DIFF
--- a/net/nimble/transport/uart/src/ble_hci_uart.c
+++ b/net/nimble/transport/uart/src/ble_hci_uart.c
@@ -338,6 +338,7 @@ ble_hci_uart_tx_char(void *arg)
                 }
                 /* NOTE: om should never be NULL! What to do? */
                 if (om == NULL) {
+                    assert(0);
                     ble_hci_uart_state.tx_type = BLE_HCI_UART_H4_NONE;
                 } else {
                     ble_hci_uart_state.tx_pkt.dptr = om->om_data;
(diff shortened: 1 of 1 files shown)


net/nimble/transport/uart/src/ble_hci_uart.c
```diff
@@ -297,17 +321,36 @@ ble_hci_uart_tx_char(void *arg)
...
+            om = ble_hci_uart_state.tx_pkt.tx_acl;
+            --om->om_len;
+            if (om->om_len == 0) {
+                /* Remove and free any zero mbufs */
+                while ((om != NULL) && (om->om_len == 0)) {
+                    ble_hci_uart_state.tx_pkt.tx_acl = SLIST_NEXT(om, om_next);
+                    os_mbuf_free(om);
+                    om = ble_hci_uart_state.tx_pkt.tx_acl;
+                }
+                /* NOTE: om should never be NULL! What to do? */
```
sjanc: if om should never be NULL then maybe change while() loop above to  do{...}while(om->om_len) and have assert for this?  this will make it easier to debug issues and also check in loop is simpler

wes3: Yeah I was wondering about the assert. Was not sure if I should let the host deal with this or simply assert and reset the controller. I will add the assert.

ccollins476ad:
Looks (and tests) good to me.

PR #245, opened 2017-04-24
MYNEWT-738: Fix parameter array errors

wes3:
Fix number of completed packet events in both controller and host.


mkiiskila:
seems ok to me

PR #248, opened 2017-04-25
Mynewt port of Semtech's LoRaMac-node library

ccollins476ad:
Note: This PR depends on another PR: #247.  247 should be merged before this one.


hw/drivers/lora/node/board/nrf52/syscfg.yml
```diff
@@ -0,0 +1,27 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+    LORA_NODE_BOARD_SPI_IDX:
+        description:
+        value: 0
```
mkiiskila: I would default this value to '-1', and have it set with BSP which has it.

hw/drivers/lora/node/board/nrf52/src/lora_nrf52.c
```diff
@@ -0,0 +1,44 @@
...
+
+    spi_settings.data_order = HAL_SPI_MSB_FIRST;
+    spi_settings.data_mode = HAL_SPI_MODE0;
+    spi_settings.baudrate = MYNEWT_VAL(LORA_NODE_BOARD_SPI_BAUDRATE);
+    spi_settings.word_size = HAL_SPI_WORD_SIZE_8BIT;
+
+    rc = hal_spi_config(RADIO_SPI_IDX, &spi_settings);
+    SYSINIT_PANIC_ASSERT_MSG(rc == 0, "Failed to configure LoRa SPI");
+
+    rc = hal_spi_enable(0);
```
ccollins476ad: Yes, good catch.

hw/drivers/lora/node/board/nrf52/pkg.yml
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: hw/drivers/lora/node/board/nrf52
+pkg.description: LoRaMac-node board driver for nRF52 systems.
```
ccollins476ad: I tend to agree.  The source library (LoRaMac-node) has "radio" drivers and "board" drivers.  I tried to maintain this distinction, but after "Mynewting" the source, the distinction ceased to make sense.  I guess we should just forget about trying to maintain the LoRaMac-node design and do what makes sense for Mynewt.

hw/drivers/lora/node/board/nrf52/syscfg.yml
ccollins476ad: Are you saying this setting should be defined in the BSP's syscfg?  I would worry that the BSP config would get cluttered with settings that only get used by a single package.  On the other hand, it is easier to see which SPI port is being used by what if these settings are all in the same place.

mkiiskila: What I meant to say was that the default value could be -1.
I'd prefer to have build/boot fail, if I include lora to my BSP, and I have not told what SPI port the hardware is at.

apps/loraping/src/main.c
```diff
@@ -0,0 +1,279 @@
...
+#include "syscfg/syscfg.h"
+#include "hal/hal_gpio.h"
+#include "hal/hal_spi.h"
+#include "bsp/bsp.h"
+#include "os/os.h"
+#include "board/board.h"
+#include "node/radio.h"
+#include "loraping.h"
+
+#define LORAPING_SPI_BAUDRATE               500
```
mkiiskila: This #define could be removed.

ccollins476ad: Thanks, you are correct.

mkiiskila:
Looks ok to me.

PR #249, opened 2017-04-26
Nimble host running against different controllers

mkiiskila:
Changes to allow host to start when doing HCI against 2 different controllers.

### DIFF
--- a/net/nimble/host/src/ble_hs_startup.c
+++ b/net/nimble/host/src/ble_hs_startup.c
@@ -189,7 +189,7 @@ ble_hs_startup_set_evmask_tx(void)
     ble_hs_hci_cmd_build_set_event_mask2(0x0000000000800000, buf, sizeof buf);
     rc = ble_hs_hci_cmd_tx_empty_ack(buf);
     if (rc != 0) {
-        return rc;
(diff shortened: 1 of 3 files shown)


net/nimble/host/src/ble_hs_startup.c
```diff
@@ -189,7 +189,7 @@ ble_hs_startup_set_evmask_tx(void)
     ble_hs_hci_cmd_build_set_event_mask2(0x0000000000800000, buf, sizeof buf);
     rc = ble_hs_hci_cmd_tx_empty_ack(buf);
     if (rc != 0) {
-        return rc;
+        BLE_HS_LOG(WARN, "ble_hs_startup_set_evmask_tx() failed\n");
```
sjanc: I think that host should check for supported commands first and then issue this only when command is supported.

ccollins476ad: Checking the controller for supported commands might be the best thing to do, though I wonder if it is really worth the extra code.  If we are just going to ignore failures anyway, querying the controller might not be necessary.

I am going to merge this PR, and we can decide if we want to query the controller for supported features as a separate issue.

PR #253, opened 2017-04-28
Updated PIC32 to call main through _start

IMGJulian:
Now includes non-default startup code


IMGJulian:
Fixed the `console_read` function that was ignoring the `cnt` argument. 

Please be aware: The meaning of the cnt input could be misinterpreted to mean "Maximum number of characters to write" rather than "Maximum number of characters to copy" and cause people to write over the end of their buffer, (they have a difference of 1 when the length of the line > `cnt`).

I would like to suggest that we rename the `str` argument to something like `buf` and dont null terminate it if the lenght of line > `cnt` (similar to the functionality of `strncpy`). Either this or make the docs very clear that this is not the case.

mkiiskila:
Yup, I can see console_read() is broken.

Renaming argument names would make sense to me. NULL-termination is optional, but I would keep it to reduce possibility of user errors.

sys/console/full/src/console.c
```diff
@@ -116,7 +116,11 @@ console_read(char *str, int cnt, int *newline)
     }
     cmd = ev->ev_arg;
     len = strlen(cmd->line);
-    strncpy(str, cmd->line, len+1);
+    if (cnt < len) {
+        len = cnt;
```
mkiiskila: Maybe this would be better?
if (cnt -1 < len) {
   len = cnt - 1;
}

Thanks for noticing this, BTW.

mkiiskila:
Other than that, this looks good to me as well.

sys/console/full/src/console.c
IMGJulian: That would prevent the unintended buffer overrun, but then, if we are being pedants, the docs require a change as cnt is no longer the number of chars *coppied* when `cnt - 1 < len`. If we just change the word *coppied* to *written* in the docs that might work? Thanks!

mkiiskila: We have to fix the buffer overrun one way or another :)
If I remember correctly, before new console code, it was copying up to cnt - 1 characters, and null terminating
the string. My opinion would be to keep it like that, and update documentation like you suggested.

mkiiskila:
> On May 2, 2017, at 2:04 AM, IMGJulian <notifications@github.com> wrote:
> 
> @IMGJulian commented on this pull request.
> 
> In sys/console/full/src/console.c <https://github.com/apache/incubator-mynewt-core/pull/253#discussion_r114275898>:
> 
> > @@ -116,7 +116,11 @@ console_read(char *str, int cnt, int *newline)
>      }
>      cmd = ev->ev_arg;
>      len = strlen(cmd->line);
> -    strncpy(str, cmd->line, len+1);
> +    if (cnt < len) {
> +        len = cnt;
> That would prevent the unintended buffer overrun, but then, if we are being pedants, the docs require a change as cnt is no longer the number of chars coppied when cnt - 1 < len. If we just change the word coppied to written in the docs that might work? Thanks!
> 
> 
We have to fix the buffer overrun one way or another :)
If I remember correctly, before new console code, it was copying up to cnt - 1 characters, and null terminating
the string. We should update documentation like you suggested.

PR #256, opened 2017-05-03
Shell improvements

michal-narajowski:
Hi,

Here are some features of the new shell:

1. Modules

First of all, the new shell brings the concept of modules. Modules group
commands under a namespace. So like in `bletiny` there is a module named 'b' 
and contains commands like `adv`, `conn`, `term`, etc. The module name is so short because you have to type it before every command: `b adv ...`. 

With the new shell, you can select a module to be the default module and just type in
the command names from this module. To select a module type: 'select [module]'. 
To go back to the main shell type 'select'. When you select a new default module, 
the prompt changes:

Default:
`shell> `
selected:
`os> `

Two example modules were added: `os` and `prompt`. `os` contains
system commands from the old shell. `prompt` contains `ticks` command.
It can toggle a dynamic shell. A dynamic shell could be used in Bluetooth app
to display name or address of a connected device.

2. Tab completion

Next major improvement is tab completion. You can complete modules,
commands, and params. When there is more than one option with a specific prefix they are printed to the console followed by the current line so
that you can continue to type the name.

```
help
tasks
mpool
date
shell> os <tab>
```

3. Improved help messages

Each command can have additional information associated with it. There is
a summary string, which provides a brief description of the command. It is displayed
when listing all possible commands in a specific context. Usage string usually contains more info and is displayed when the user invokes help command or when
the user mistyped some parameter for this command. Moreover, a list
of parameters can be passed with the command for tab completion of parameters
and for displaying short descriptions of params.

The help API is designed, so the app can easily compile out the additional
help info to save flash.

4. Compatible API

The new shell is fully compatible with the old API. However old shell was used to register separate commands instead of modules. So to make this work, when the old API is used, a `compat` module is registered and all the commands that use old API are added to this module. The commands can be invoked in exactly the same way as before. You will just see a `compat> ` prompt at the beginning of the line. 

5. Newtmgr over shell

It works with the new shell too!

That's it. I hope it will be useful.

Best regards,
Michał


sys/shell/src/shell.c
```diff
@@ -449,6 +449,33 @@ shell_register(const char *module_name, const struct shell_cmd *commands)
...
+#define SHELL_COMPAT_MODULE_NAME "compat"
+static struct shell_cmd compat_commands[MYNEWT_VAL(SHELL_MAX_COMPAT_COMMANDS)];
+static int num_compat_commands;
+static int module_registered = 0;
+
+int
+shell_cmd_register(struct shell_cmd *sc)
+{
+    if (num_compat_commands >= MYNEWT_VAL(SHELL_MAX_COMPAT_COMMANDS)) {
+        console_printf("Max number of compat commands reached");
```
sjanc: missing "\n"

sys/shell/include/shell/shell.h
```diff
@@ -101,6 +101,10 @@ void shell_register_prompt_handler(shell_prompt_function_t handler);
  */
 void shell_register_default_module(const char *name);
 
+#if MYNEWT_VAL(SHELL_COMPAT)
+int shell_cmd_register(struct shell_cmd *sc);
```
sjanc: can sc be const?

sys/shell/src/shell.c
```diff
@@ -449,6 +449,33 @@ shell_register(const char *module_name, const struct shell_cmd *commands)
     return 0;
 }
 
+#if MYNEWT_VAL(SHELL_COMPAT)
+#define SHELL_COMPAT_MODULE_NAME "compat"
+static struct shell_cmd compat_commands[MYNEWT_VAL(SHELL_MAX_COMPAT_COMMANDS)];
+static int num_compat_commands;
```
sjanc: be consistent,  either we always initialize statics or never,  please check coding style for this

sys/shell/src/shell.c
```diff
@@ -449,6 +449,33 @@ shell_register(const char *module_name, const struct shell_cmd *commands)
...
+{
+    if (num_compat_commands >= MYNEWT_VAL(SHELL_MAX_COMPAT_COMMANDS)) {
+        console_printf("Max number of compat commands reached");
+        assert(0);
+    }
+
+    if (!module_registered) {
+        shell_register(SHELL_COMPAT_MODULE_NAME, compat_commands);
+        set_default_module(SHELL_COMPAT_MODULE_NAME);
+        module_registered = 0;
```
sjanc: should this be set to 1?

sys/shell/include/shell/shell.h
```diff
@@ -16,40 +16,90 @@
...
+struct shell_cmd_help {
+    const char *summary;
+    const char *usage;
+    const struct shell_param *params;
+};
 
-typedef int (*shell_cmd_func_t)(int argc, char **argv);
 struct shell_cmd {
-    char *sc_cmd;
+    const char *sc_cmd;
```
sjanc: just name fields name, func, help    those already are part of struct shell_cmd so this is clear

sys/shell/include/shell/shell.h
```diff
@@ -16,40 +16,90 @@
...
     shell_cmd_func_t sc_cmd_func;
-    STAILQ_ENTRY(shell_cmd) sc_next;
+    const struct shell_cmd_help *help;
+};
+
+struct shell_module {
+    const char *module_name;
```
sjanc: name

sys/shell/include/shell/shell.h
```diff
@@ -16,40 +16,90 @@
...
 
-int shell_cmd_register(struct shell_cmd *sc);
+/** @brief Callback called when registering module
+ *
+ *  @param module_name Name of the module
+ *  @param commands Array of shell_cmd structs
+ *
+ * @return 0 in case of success or negative value in case of error.
+ */
+typedef int (*shell_register_function_t)(const char *module_name,
```
sjanc: maybe add this is commit that acctually starts using it

sys/shell/src/shell.c
```diff
@@ -21,543 +21,432 @@
...
+
+static struct os_eventq avail_queue;
+static struct os_event shell_console_ev[MYNEWT_VAL(SHELL_MAX_CMD_QUEUED)];
+
+static const char *
+get_prompt(void)
 {
-    int rc;
+    if (app_prompt_handler) {
+        const char *str;
```
sjanc: coding style:  variables should be declared add beginning of function, not block

sys/shell/src/shell.c
```diff
@@ -401,6 +401,407 @@ shell(struct os_event *ev)
...
+    const struct shell_cmd *command;
+
+    command = &shell_modules[module_idx].commands[command_idx];
+
+    if (!(command->help && command->help->params)) {
+        return 0;
+    }
+
+    for (i = 0; command->help->params[i].param_name; i++) {
+        int j;
```
sjanc: coding style: move to the top of function body

sys/shell/src/shell.c
```diff
@@ -401,6 +401,407 @@ shell(struct os_event *ev)
...
+                 int command_len, int module_idx)
+{
+    const char *first_match = NULL;
+    int i, common_chars = -1, space = 0;
+    const struct shell_module *module;
+
+    module = &shell_modules[module_idx];
+
+    for (i = 0; module->commands[i].sc_cmd; i++) {
+        int j;
```
sjanc: coding style: move to the top of function body

sys/shell/src/shell.c
```diff
@@ -401,6 +401,407 @@ shell(struct os_event *ev)
...
+        for (i = 0; i < num_of_shell_entities; i++) {
+            console_printf("%s\n", shell_modules[i].module_name);
+        }
+        console_printf("%s", get_prompt());
+        console_printf("%s", line);
+        return 0;
+    }
+
+    for (i = 0; i < num_of_shell_entities; i++) {
+        int j;
```
sjanc: coding style: move to the top of function body

sys/shell/src/shell.c
```diff
@@ -401,6 +401,407 @@ shell(struct os_event *ev)
...
+    }
+
+    return common_chars - module_len + space;
+}
+
+static int
+complete_select(char *line, int len, char *cur, int tok_len)
+{
+    int null_terminated = 0;
+    cur += tok_len + 1;
```
sjanc: add empty line

sys/shell/src/shell.c
```diff
@@ -401,6 +401,407 @@ shell(struct os_event *ev)
...
+    char *cur;
+    int tok_len;
+    int module, command;
+    int null_terminated = 0;
+
+    /*
+     * line to completion is not ended by '\0' as the line that gets from
+     * os_eventq_get function
+     */
+    line[len] = '\0';
```
sjanc: just make sure you don't overflow line here

sys/shell/src/shell.c
```diff
@@ -401,6 +401,407 @@ shell(struct os_event *ev)
...
+            console_printf("%s", get_prompt());
+            console_printf("%s", line);
+            return 0;
+        }
+
+        if (null_terminated) {
+            return complete_command(line, len, cur, tok_len, module);
+        }
+    }
+
```
sjanc: bogus extra lines

PR #257, opened 2017-05-05
MYNEWT-526 + few Nimble/controller fixes

andrzej-kaczmarek:
This provides solution for MYNEWT-526 (connection timeout with Polar H7). This is not a fix, since there is no problem on our side - it's just an configuration option to adjust initial connection event scheduling in a way that broken TI CC254x (which Polar H7 is based on) can work properly. See commit message for details.

...and few minor fixes for LL in Nimble controller.

### DIFF
--- a/net/nimble/controller/src/ble_ll_ctrl.c
+++ b/net/nimble/controller/src/ble_ll_ctrl.c
@@ -1283,6 +1283,7 @@ ble_ll_ctrl_proc_init(struct ble_ll_conn_sm *connsm, int ctrl_proc)
             } else {
                 opcode = BLE_LL_CTRL_SLAVE_FEATURE_REQ;
             }
+            memset(ctrdata, 0, BLE_LL_CTRL_FEATURE_LEN);
             ctrdata[0] = ble_ll_read_supp_features();
             break;
         case BLE_LL_CTRL_PROC_VERSION_XCHG:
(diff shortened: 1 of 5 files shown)


net/nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -1283,6 +1283,7 @@ ble_ll_ctrl_proc_init(struct ble_ll_conn_sm *connsm, int ctrl_proc)
             } else {
                 opcode = BLE_LL_CTRL_SLAVE_FEATURE_REQ;
             }
+            memset(ctrdata, 0, 8);
```
rymanluk: Maybe we could use define BLE_LL_CTRL_FEATURE_LEN instead of 8?

rymanluk:
Indeed in the code common_features = "our" & "their" in case of remote sending us LL_FEATURE_REQ or LL_SLAVE_FEATURE_REQ.

In case of LL_FEATURE_RSP  common_features is equal to what remote respond:
 ( connsm->common_features = dptr[0];) btw I believe get_le64 is missing there.

The tricky thing here is that what remote respond is not "their" & "our" features. According to spec only first octet contains "our" & "their" and rest is just "their". Below is part of spec saying that.

2.4.2.10 LL_FEATURE_RSP

The LL_FEATURE_RSP CtrData consists of one field:
• FeatureSet[0] shall contain a set of features supported by the Link Layers of
both the master and slave.
• FeatureSet[1-7] shall contain a set of features supported by the Link Layer
that transmits this PDU

andrzej-kaczmarek:
@wes3 indeed, connsm->common_features is what we need here so the check with our features is redundant.

@rymanluk common features are the first octet of FeatureSet (as you quoted from the spec) and only these features are useful for us to check capabilities when starting LL procedures so it's all good here - at least for the time being

I'll fix comments and push new version

rymanluk:
Ah ok. I'm looking at bluetooth5 branch where common_feature is uint32_t already.

PR #258, opened 2017-05-06
MYNEWT-744 SensorAPI: improvements

vrahane:
- Change data structure to SLIST instead of TAILQ so that it can get initialized statically. This is needed so that sysinit could be used for initializing the sensors.
- Remove dummy functions in the bsp and add the sensor device initialization functions instead.
- Adding syscfg based configuration for onboard Vs offboard sensors
- Adding sensor creator package for creating external sensor devices if they are not on the board itself. This is just for convenience sake. The right way to do this would be from the bsp. As a result, removed sensor device creation from nrf52dk and nrf52840pdk bsps.
- Add sensor interface struct so that it can get used by the individual drivers so that we can talk to multiple sensors using the same driver using same or different interfaces.
- For TSL2561, apart from the above changes, various fixes were made:
    1. Making a few changes to this driver along with sensor_itf changes
    2. Gain, integration time and enabled flag was a global instead of reading every time gain is set or integration time is set. Changing it so that we read the timing register before setting either gain or integration time.
    3. Also changing a few functions in the driver to handle errors more gracefully. Changing shell to work with new functions.
    4. Making changes so that it goes in synch with the drivers.
- Add a mask field to the sensor structure so that differentiating between supported Vs configured sensors is possible
- Change the sensor_mgr_match_bytype() function to check w.r.t mask and make it a public API.
- Adding config variable for mask for different drivers in their respective structures.
- Adding an API to set the sensor types based on the mask.
- Moving config_sensor() out of the app into the creator package and the bsp. Objective is to show how a developer would initialize and configure the sensor in the bsp which is normally the practice.
- Changing sensor_shell command "list" to show only configured types.
- sensor_shell cmd "type" shows supported types set by the driver.
- In the sensor_oic source file changing code to check type only while allocating resources.
- Changing device name in the sensors_test app to be "sensy" instead of "pi"
- For the LSM303DLHC driver the address field in the "sensor_itf" structure is not enough since it talks to the sensor using two addresses: one for accelerometer and one for magnetometer. The best way to deal with this is to make the addresses config variables in the sensor driver structure. For any such sensor devices which have multiple addresses, a similar approach should be followed.
- Have changes the drivers to work with all the fixes mentioned above.
- Gyroscope data structure was not defined, defined that and adding support for shell to read it.


apps/sensors_test/src/main.c
```diff
@@ -413,10 +410,10 @@ config_sensor(void)
...
     dev = (struct os_dev *) os_dev_open("light0", OS_TIMEOUT_NEVER, NULL);
     assert(dev != NULL);
-    rc = tsl2561_init(dev, NULL);
-    if (rc) {
-        os_dev_close(dev);
-        goto err;
+
+    if (dev->od_flags & OS_DEV_F_STATUS_READY) {
+        rc= SYS_ENODEV;
+        console_printf("config_sensor %s init failed\n", dev->od_name);
```
jacobrosenthal: Should these not still have the goto err; ?

vrahane: That is an intentional change so that even if initialization on one device fails the code goes ahead and tries to configure the same one and tries to initialize others.

rymanluk: Maybe it is worth to mention that in commit message then.

BTW. There are two patches with same title. Would it be possible to change title for them?  That would really helps later when you search something in the git tree e.g. using "git log --oneline"  
What we try to do is to keep MYNEWT Jira as a prefix and then high level Title. If we do work without JIRA ticket we put submodule name as a prefix. In commit message we usually put what we change and more important why we change it.

vrahane: There would be more changes to come. I will explain them in detail together. git follows a 50/72 limit and hence fitting everything in 50 characters is not possible. Adding a high level title is exactly what I have done. If you expand the title you will see additional comments below.

apps/sensors_test/src/main.c
```diff
@@ -182,6 +189,7 @@ sensor_oic_advertise(void)
         return;
     }
 
+#if 0
```
mkiiskila: If it's #if 0, maybe you should consider removing the code block altogether?

hw/bsp/ruuvi_tag_revb2/src/hal_bsp.c
```diff
@@ -187,71 +164,62 @@ hal_bsp_get_nvic_priority(int irq_num, uint32_t pri)
...
+#if MYNEWT_VAL(BME280_ONB)
+    rc = os_dev_create((struct os_dev *) &bme280, "bme280",
+      OS_DEV_INIT_PRIMARY, 0, bme280_init, (void *)&spi_0_itf_bme);
     assert(rc == 0);
-#endif
 
-#if MYNEWT_VAL(BME280_PRESENT)
-    rc = os_dev_create((struct os_dev *) &bme280, "bme280",
-      OS_DEV_INIT_PRIMARY, 0, press_init, NULL);
+    rc = config_bme280_sensor();
```
mkiiskila: Sensor config is not done by the user of the sensor?

vrahane: The consensus was, there would be an initial config that happens at the bsp level. The user of the sensor can then go and re-config the sensor in the app if needed.

apps/sensors_test/src/main.c
vrahane: Removed, thanks for pointing it out.

PR #259, opened 2017-05-07
hw: mcu: pic32mz2048efg100: Implement hal_gpio driver

francois-berder:
This has been tested on a Wi-Fire board.

Signed-off-by: Francois Berder <fberder@outlook.fr>


francois-berder:
@IMGJulian Can you have a look ?

PR #268, opened 2017-05-09
Add back the 2nd explict COMSPEC /C.

cwanda:
Simon's msys2 environment doesn't work without it.
My msys2 works with and without the 2nd explicit COMPSEC /C/
Looks like the only fix that was needed is to move the opening "


aditihilbert:
Has been tested. Thanks!

PR #269, opened 2017-05-10
MYNEWT-745 Sim - deadlock involving system calls

ccollins476ad:
This commit splits sim into two separate implementations:
    * "signals"
    * "no-signals"

The user chooses which implementation to use via the
MCU_NATIVE_USE_SIGNALS syscfg setting (defined in hw/mcu/native).  The
two implementations are described below:

signals:
    More correctness; less stability.  The OS tick timer will
    cause a high-priority task to preempt a low-priority task.
    This causes stability issues because a task can be preempted
    while it is in the middle of a system call, potentially
    causing deadlock or memory corruption.

no-signals:
    Less correctness; more stability.  The OS tick timer only
    runs while the idle task is active.  Therefore, a sleeping
    high-priority task will not preempt a low-priority task due
    to a timing event (e.g., delay or callout expired).
    However, this version of sim does not suffer from the
    stability issues that affect the "signals" implementation.

### DIFF
--- a/hw/mcu/native/syscfg.yml
+++ b/hw/mcu/native/syscfg.yml
@@ -24,3 +24,24 @@ syscfg.defs:
             Specifies the required alignment for internal flash writes.
             Used internally by the newt tool.
         value: 1
+
+    MCU_NATIVE_USE_SIGNALS:
+        description: >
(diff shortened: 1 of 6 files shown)


rymanluk:
Looks/works good to me.

PR #272, opened 2017-05-10
PIC32: Moved the contents os_bsp.c to hal_bsp.c

IMGJulian:
Moved the contents os_bsp.c to hal_bsp.c for the pic32 based BSPs to fit in with the rest of the boards.


kernel/os/include/os/arch/pic32/os/os_arch.h
```diff
@@ -59,7 +59,7 @@ typedef uint32_t os_stack_t;
 /* Exit a critical section, restore processor state and unblock interrupts */
 #define OS_EXIT_CRITICAL(__os_sr) __builtin_set_isr_state(__os_sr)
 /*  This is not the only way interrupts can be disabled */
-#define OS_IS_CRITICAL() ((__builtin_get_isr_state() & 1) == 0)
+#define OS_IS_CRITICAL() ((_CP0_GET_STATUS() & 1) == 0)
```
francois-berder: Could you put this change in another commit ? From my understanding, the only difference is that ```__builtin_get_isr_state``` adds an extra ```& 0xF``` to CP0 status register while you only want to check EI bit. Is that correct ?

IMGJulian: It does this:

mfc0   $3, $12, 0
srl    $2,$3,10
ins    $2,$3,3,1
andi   $2,$2,0xf
sw     $2,0($fp)

At least 3 out of those 5 instructions are redundant, not a huge thing I know, but IMO its also clearer using _CP0_GET_STATUS(). Yes I can put it in a separate commit if you like.

PR #274, opened 2017-05-11
various fixes to sensors

jacobrosenthal:


apps/sensors_test/pkg.yml
```diff
@@ -42,7 +42,7 @@ pkg.deps:
     - sys/stats/full
     - boot/split
     - mgmt/imgmgr
-    #- mgmt/newtmgr
```
vrahane: Thank you for the changes Jacob
- either mgmt/newtmgr or mgmt/oicmgr should be specified as a dependency. Both of them cannot be specified. In syscfg.yml, SENSOR_OIC is set to 1 which pulls in oicmgr here. Hence, newtmgr was commented out. Apart from that everything looks good.

jacobrosenthal: reverted

PR #277, opened 2017-05-12
BLE Host - Store management API additions

ccollins476ad:
This pull request adds the following three functions to the host store
API:

```
/**
 * Deletes all entries from the store that match the specified key.
 *
 * @param type                  The type of store entry to delete.
 * @param key                   Entries matching this key get deleted.
 *
 * @return                      0 on success;
 *                              Other nonzero on error.
 */
int
ble_store_util_delete_all(int type, const union ble_store_key *key)
```
```
/**
 * Retrieves the set of peer addresses for which a bond has been
 * established.
 *
 * @param out_peer_id_addrs     On success, the set of bonded peer
 *                                  addresses gets written here.
 * @param out_num_peers         On success, the number of bonds gets
 *                                  written here.
 * @param max_peers             The capacity of the destination buffer.
 *
 * @return                      0 on success;
 *                              BLE_HS_ENOMEM if the destination buffer
 *                                  is too small;
 *                              Other nonzero on error.
 */
int
ble_store_util_bonded_peers(ble_addr_t *out_peer_id_addrs,
                            int *out_num_peers, int max_peers)
```
```
/**
 * Deletes all entries from the store that are attached to the specified
 * peer address.  This function deletes security entries and CCCD
 * records.
 *
 * @param peer_id_addr          Entries with this peer address get
 *                                  deleted.
 *
 * @return                      0 on success;
 *                              Other nonzero on error.
 */
int
ble_store_util_delete_peer(const ble_addr_t *peer_id_addr)
```
It also contains some miscellaneous fixes.

### DIFF
--- a/net/nimble/host/src/ble_store_util.c
+++ b/net/nimble/host/src/ble_store_util.c
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "host/ble_store.h"
+#include "ble_hs_priv.h"
+
+struct ble_store_util_peer_set {
+    ble_addr_t *peer_id_addrs;
+    int num_peers;
+    int max_peers;
+    int status;
+};
+
+static int
+ble_store_util_iter_unique_peer(int obj_type,
+                                union ble_store_value *val,
+                                void *arg)
+{
+    struct ble_store_util_peer_set *set;
+    int i;
+
+    BLE_HS_DBG_ASSERT(obj_type == BLE_STORE_OBJ_TYPE_OUR_SEC ||
+                      obj_type == BLE_STORE_OBJ_TYPE_PEER_SEC);
+
+    set = arg;
+
+    /* Do nothing if this peer is a duplicate. */
+    for (i = 0; i < set->num_peers; i++) {
+        if (ble_addr_cmp(set->peer_id_addrs + i, &val->sec.peer_addr) == 0) {
+            return 0;
+        }
+    }
+
+    if (set->num_peers >= set->max_peers) {
+        /* Overflow; abort the iterate procedure. */
+        set->status = BLE_HS_ENOMEM;
+        return 1;
+    }
+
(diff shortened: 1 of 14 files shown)


net/nimble/host/store/ram/src/ble_store_ram.c
```diff
@@ -322,6 +322,28 @@ ble_store_ram_find_cccd(const struct ble_store_key_cccd *key)
...
+    if (idx == -1) {
+        return BLE_HS_ENOENT;
+    }
+
+    rc = ble_store_ram_delete_obj(ble_store_ram_cccds,
+                                  sizeof *ble_store_ram_cccds,
+                                  idx,
+                                  &ble_store_ram_num_cccds);
+    if (rc != 0) {
+        return rc;
```
rymanluk: return rc should be just fine since below is return 0;

net/nimble/host/src/ble_store_util.c
```diff
@@ -0,0 +1,164 @@
...
+    ble_addr_t *peer_id_addrs;
+    int num_peers;
+    int max_peers;
+    int status;
+};
+
+static int
+ble_store_util_iter_unique_peer(int obj_type,
+                                union ble_store_value *val,
+                                void *arg)
```
ccollins476ad: This function gets passed to ble_store_iterate(), which takes an `int (*fn)(int, union ble_store_value *, void *)` parameter.

net/nimble/host/src/ble_store_util.c
```diff
@@ -0,0 +1,164 @@
...
+        .num_peers = 0,
+        .max_peers = max_peers,
+        .status = 0,
+    };
+    int rc;
+
+    rc = ble_store_iterate(BLE_STORE_OBJ_TYPE_OUR_SEC,
+                           ble_store_util_iter_unique_peer,
+                           &set);
+    if (rc != 0) {
```
ccollins476ad: Not quite - ble_store_iterate() returns 0 in this case.  ble_store_iterate() only returns nonzero if there is a store read error.

net/nimble/host/src/ble_store_util.c
```diff
@@ -0,0 +1,164 @@
...
+    rc = ble_store_util_delete_all(BLE_STORE_OBJ_TYPE_PEER_SEC, &key);
+    if (rc != 0) {
+        return rc;
+    }
+        
+    memset(&key, 0, sizeof key);
+    key.cccd.peer_addr = *peer_id_addr;
+
+    rc = ble_store_util_delete_all(BLE_STORE_OBJ_TYPE_CCCD, &key);
+    if (rc != 0) {
```
ccollins476ad: That is true.  I guess it is a style issue, but I prefer to `return 0;` at the bottom of the function.  This allows for safe additions to the end of a function.  This is similar to the golang style of returning an error as soon as it is detected, and returning nil at the end of a function.

net/nimble/host/src/ble_store_util.c
rymanluk: ah you are right.

net/nimble/host/src/ble_store_util.c
rymanluk: I see now, ble_store_iterate just stop iterate when callback returns non zero and then it returns 0 up to the user. user then need to check status. got it.

net/nimble/host/src/ble_store_util.c
rymanluk: :) Ok

ccollins476ad:
Thanks for taking a look :).

PR #278, opened 2017-05-12
boot: Tickle watchdog during image swap and verify

simonratner:
Watchdog timer that was enabled by the app is still running after a soft reset, but without a scheduler the bootloader invariably trips it whenever it has to swap or verify images. The bootloader actually does an incredible job recovering; the only shreds of evidence are longer boot time and a reset reason value of `WATCHDOG` saved in `NRF_POWER->RESETREAS` of the nRF51; which is possibly why this has gone unnoticed.

This is somewhat related to https://issues.apache.org/jira/browse/MYNEWT-618.

### DIFF
--- a/boot/bootutil/src/image_validate.c
+++ b/boot/bootutil/src/image_validate.c
@@ -24,6 +24,7 @@
 
 #include "syscfg/syscfg.h"
 #include "hal/hal_flash.h"
+#include "hal/hal_watchdog.h"
 #include "flash_map/flash_map.h"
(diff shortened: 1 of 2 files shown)


mkiiskila:
Indeed, I assumed that watchdog resets at bootloader were happening, so I did add a tickle just before managed reset via newtmgr. Of course, this might not work when reset is done elsewhere, or if watchdog tickle requirement frequency changes.
I think we should merge this.

PR #279, opened 2017-05-12
BLE Host - Policy for SM key overflow

ccollins476ad:
This pull request addresses the issue identified in MYNEWT-751 (https://issues.apache.org/jira/browse/MYNEWT-751).

This PR incorporates changes from another PR: 277 (https://github.com/apache/incubator-mynewt-core/pull/277).  Therefore, this PR should only be merged after 277.

### DIFF
--- a/net/nimble/host/include/host/ble_store.h
+++ b/net/nimble/host/include/host/ble_store.h
@@ -243,7 +243,8 @@ int ble_store_read(int obj_type, const union ble_store_key *key,
                    union ble_store_value *val);
 int ble_store_write(int obj_type, const union ble_store_value *val);
 int ble_store_delete(int obj_type, const union ble_store_key *key);
-int ble_store_status(struct ble_store_status_event *event);
+int ble_store_overflow_event(int obj_type, const union ble_store_value *value);
+int ble_store_overflow_next_event(int obj_type, uint16_t conn_handle);
 
 int ble_store_read_our_sec(const struct ble_store_key_sec *key_sec,
                            struct ble_store_value_sec *value_sec);
--- a/net/nimble/host/include/host/ble_store.h
+++ b/net/nimble/host/include/host/ble_store.h
@@ -31,6 +31,12 @@ extern "C" {
 #define BLE_STORE_OBJ_TYPE_PEER_SEC     2
 #define BLE_STORE_OBJ_TYPE_CCCD         3
 
+/** Failed to persist record; insufficient storage capacity. */
+#define BLE_STORE_EVENT_OVERFLOW        1
+
+/** About to execute a procedure that may fail due to overflow. */
+#define BLE_STORE_EVENT_OVERFLOW_NEXT   2
+
 /**
  * Used as a key for lookups of security material.  This struct corresponds to
  * the following store object types:
@@ -132,6 +138,40 @@ union ble_store_value {
     struct ble_store_value_cccd cccd;
 };
 
+struct ble_store_status_event {
+    /**
+     * The type of object that failed to persist; one of the
+     * BLE_STORE_OBJ_TYPE_[...] codes.
+     */
+    int obj_type;
+
+    /**
+     * The type of event being reported; one of the BLE_STORE_EVENT_TYPE_[...]
+     * codes.
+     */
+    int event_code;
+
+    /**
+     * Additional data related to the event; the valid field is inferred from
+     * the obj_type,event_code pair.
+     */
+    union {
+        /**
+         * The record that failed to be written.  Valid for the following event
+         * types:
+         *     o BLE_STORE_EVENT_OVERFLOW
+         */
+        const union ble_store_value *value;
+
+        /**
+         * The connection handle of the peer.  Valid for the following event
+         * types:
+         *     o BLE_STORE_EVENT_OVERFLOW_NEXT
+         */
+        uint16_t conn_handle;
+    };
+};
+
 /**
  * Searches the store for an object matching the specified criteria.  If a
  * match is found, it is read from the store and the dst parameter is populated
@@ -182,10 +222,28 @@ typedef int ble_store_write_fn(int obj_type, const union ble_store_value *val);
  */
 typedef int ble_store_delete_fn(int obj_type, const union ble_store_key *key);
 
+/**
+ * Indicates an inability to perform a store operation.  This callback should
+ * do one of two things:
+ *     o Address the problem and return 0, indicating that the store operation
+ *       should proceed.
+ *     o Return nonzero to indicate that the store operation should be aborted.
+ *
+ * @param event                 Describes the store event being reported.
+ * @param arg                   Optional user argument.
+ *
+ * @return                      0 if the store operation should proceed;
+ *                              nonzero if the store operation should be
+ *                                  aborted.
+ */
+typedef int ble_store_status_fn(struct ble_store_status_event *event,
+                                void *arg);
+
 int ble_store_read(int obj_type, const union ble_store_key *key,
                    union ble_store_value *val);
 int ble_store_write(int obj_type, const union ble_store_value *val);
 int ble_store_delete(int obj_type, const union ble_store_key *key);
+int ble_store_status(struct ble_store_status_event *event);
 
 int ble_store_read_our_sec(const struct ble_store_key_sec *key_sec,
                            struct ble_store_value_sec *value_sec);
@@ -225,6 +283,9 @@ int ble_store_util_bonded_peers(ble_addr_t *out_peer_id_addrs,
                                 int max_peers);
 int ble_store_util_delete_all(int type, const union ble_store_key *key);
 int ble_store_util_delete_peer(const ble_addr_t *peer_id_addr); 
+int ble_store_util_delete_oldest_peer(void);
+int ble_store_util_count(int type, int *out_count);
+int ble_store_util_status_rr(struct ble_store_status_event *event, void *arg);
 
 #ifdef __cplusplus
 }
--- a/net/nimble/host/include/host/ble_store.h
+++ b/net/nimble/host/include/host/ble_store.h
@@ -35,7 +35,7 @@ extern "C" {
 #define BLE_STORE_EVENT_OVERFLOW        1
 
 /** About to execute a procedure that may fail due to overflow. */
-#define BLE_STORE_EVENT_OVERFLOW_NEXT   2
+#define BLE_STORE_EVENT_FULL            2
 
 /**
  * Used as a key for lookups of security material.  This struct corresponds to
@@ -140,12 +140,6 @@ union ble_store_value {
 
 struct ble_store_status_event {
     /**
-     * The type of object that failed to persist; one of the
-     * BLE_STORE_OBJ_TYPE_[...] codes.
-     */
-    int obj_type;
-
-    /**
      * The type of event being reported; one of the BLE_STORE_EVENT_TYPE_[...]
      * codes.
      */
@@ -157,18 +151,30 @@ struct ble_store_status_event {
      */
     union {
         /**
-         * The record that failed to be written.  Valid for the following event
-         * types:
+         * Represents a write that failed due to storage exhaustion.  Valid for
+         * the following event types:
          *     o BLE_STORE_EVENT_OVERFLOW
          */
-        const union ble_store_value *value;
+        struct {
+            /** The type of object that failed to be written. */
+            int obj_type;
+
+            /** The object that failed to be written. */
+            const union ble_store_value *value;
+        } overflow;
 
         /**
-         * The connection handle of the peer.  Valid for the following event
-         * types:
-         *     o BLE_STORE_EVENT_OVERFLOW_NEXT
+         * Represents the possiblity that a scheduled write will fail due to
+         * storage exhaustion.  Valid for the following event types:
+         *     o BLE_STORE_EVENT_FULL
          */
-        uint16_t conn_handle;
+        struct {
+            /** The type of object that may fail to be written. */
+            int obj_type;
+
+            /** The handle of the connection which prompted the write. */
+            uint16_t conn_handle;
+        } full;
     };
 };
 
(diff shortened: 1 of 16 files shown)


net/nimble/host/include/host/ble_store.h
```diff
@@ -31,6 +31,12 @@ extern "C" {
 #define BLE_STORE_OBJ_TYPE_PEER_SEC     2
 #define BLE_STORE_OBJ_TYPE_CCCD         3
 
+/** Failed to persist record; insufficient storage capacity. */
+#define BLE_STORE_EVENT_OVERFLOW        1
+
+/** About to execute a procedure that may fail due to overflow. */
+#define BLE_STORE_EVENT_OVERFLOW_NEXT   2
```
rymanluk: since for BLE_STORE_EVENT_OVERFLOW_NEXT we specify conn_handle because it is used during pairing,  why not to name type BLE_STORE_EVENT_OVERFLOW_BONDING or similar? That would make clear for application what is going wrong and what to do i.e. remove old bonded devices.

net/nimble/host/src/ble_sm.c
```diff
@@ -892,6 +895,72 @@ ble_sm_key_dist(struct ble_sm_proc *proc,
...
+    }
+
+    rc = ble_store_util_count(obj_type, &count);
+    if (rc != 0) {
+        return rc;
+    }
+
+    /* Pessimistically assume all active procs will persist bonds. */
+    ble_hs_lock();
+    count += ble_sm_num_procs();
```
rymanluk: count is not initialized

net/nimble/host/src/ble_sm.c
```diff
@@ -892,6 +895,72 @@ ble_sm_key_dist(struct ble_sm_proc *proc,
     }
 }
 
+static int
+ble_sm_chk_store_overflow_once(int is_our_sec, uint16_t conn_handle)
```
rymanluk: maybe instead of is_our_sec we could provide obj_type? 

But actually maybe it would be better to create helper ble_store_overflow_event(obj_type, event_code, void *event_data) and keep this function to basically check if there is a sufficient place.

I see that helper could be used at least in three places in this code.

net/nimble/host/src/ble_sm.c
```diff
@@ -1610,6 +1693,7 @@ ble_sm_pair_rsp_rx(uint16_t conn_handle, struct os_mbuf **om,
     struct ble_sm_pair_cmd *rsp;
     struct ble_sm_proc *proc;
     uint8_t ioact;
+    int rc;
```
rymanluk: This seems to be not related to this patch

net/nimble/host/src/ble_store.c
```diff
@@ -28,49 +28,98 @@ ble_store_read(int obj_type, const union ble_store_key *key,
 {
     int rc;
 
+    ble_hs_lock();
```
rymanluk: Is this related to this patch? BTW why we need it in ble_store?

net/nimble/host/src/ble_store.c
```diff
@@ -28,49 +28,98 @@ ble_store_read(int obj_type, const union ble_store_key *key,
...
+        }
+    }
 }
 
 int
 ble_store_delete(int obj_type, const union ble_store_key *key)
 {
     int rc;
 
+    ble_hs_lock();
```
rymanluk: same as above

net/nimble/host/include/host/ble_store.h
ccollins476ad: I wasn't sure if this event would be used for other record types as well (not just bonds).  Since the first one (OVERFLOW) is used for multiple record types, I opted to keep both generic.

I do agree BONDING is more clear.  However, my only concern is that we'll have to rename it again (change the API) if we want to use it for other record types in the future.

net/nimble/host/src/ble_sm.c
ccollins476ad: It is initialized by the prior call to `ble_store_util_count()`.

net/nimble/host/src/ble_store.c
ccollins476ad: You're right - I snuck it into this PR :).  I do think we need these locks here to prevent concurrent reads and writes.  For example, if code is iterating the RAM store while a record is being written, a garbage record could be retrieved.

net/nimble/host/src/ble_sm.c
ccollins476ad: OK, I can do that.  Where are the other places where such a helper could be used?  This is the only place I can see.

net/nimble/host/include/host/ble_store.h
rymanluk: I have feeling that for other stuff we will basically have new type.
My understanding is that callback implementation looks first into event type and then it know what union field is related to this.

net/nimble/host/src/ble_sm.c
rymanluk: two times in ble_sm_chk_store_overflow():
1) if check for our_keys fails 
2) if check for their_keys fails

Third place is in "ble_store_write" 

Just to make sure I was clear in my first comment :) Idea is to keep ble_sm_chk_store_overflow_once(obj_type)  - maybe change name to ble_sm_chk_store_overflow_by_type() and this function just check if there is required space.
And second helper  ble_store_overflow_event(obj_type, event_code, void *event_data) which just sends event.

net/nimble/host/src/ble_sm.c
rymanluk: ah I see

net/nimble/host/src/ble_store.c
rymanluk: Ah ok, for some reason I thought that ble_hs_lock() protects only conn object.

net/nimble/host/src/ble_sm.c
ccollins476ad: Thanks, Łukasz.  I will make this change.

net/nimble/host/include/host/ble_store.h
ccollins476ad: This event has an `obj_type` field which indicates the type of record that didn't fit.  Do you think we should remove this field, and create separate events for each record type?

rymanluk: good point. somehow I forgot about that obj_type. So callback first checks event_type then looks into obj_type and then specific data.  You are right.. in this case it makes sense to have generic event_type.

So maybe we can try to improve name: e.g.  BLE_STORE_EVENT_OVERFLOW_WARNING, BLE_STORE_EVENT_OVERFLOW_IS_COMING, BLE_STORE_EVENT_OVERFLOW_AHEAD, BLE_STORE_EVENT_POSSIBLE_OVERFLOW ? Not sure if any of this is better :)

ccollins476ad: I don't think BLE_STORE_EVENT_OVERFLOW_NEXT is particularly unclear.  It is so named because the next store operation will overflow.  The user may confuse this event type with the other (BLE_STORE_EVENT_OVERFLOW) because the names are similar, but I think this is going to be hard to prevent.  After all, the meanings of the events are similar!

If there is a specific name you like, I'm happy to use it.  Personally, I am OK with the current one, or several of the suggestions.

net/nimble/host/include/host/ble_store.h
```diff
@@ -131,6 +137,40 @@ union ble_store_value {
...
+
+    /**
+     * Additional data related to the event; the valid field is inferred from
+     * the obj_type,event_code pair.
+     */
+    union {
+        /**
+         * The record that failed to be written.  Valid for the following event
+         * types:
+         *     o BLE_STORE_EVENT_OVERFLOW
```
rymanluk: Since a key for union data is pair (obj_type, event_code pair) could we extend comment here and say that value is valid for event type "BLE_STORE_EVENT_OVERFLOW" and all possible obj_type.

Similar we could do for description to conn_handle, but int this case we have only couple of obj_types valid.

net/nimble/host/include/host/ble_store.h
rymanluk: Well I don't have very strong opinion on this, so I suggest to leave it as it is.  In the end I think that my confusion was created due to comments on the fields in union. Please check  comment over there.

net/nimble/host/src/ble_sm.c
```diff
@@ -892,6 +895,72 @@ ble_sm_key_dist(struct ble_sm_proc *proc,
...
+    if (count < MYNEWT_VAL(BLE_STORE_MAX_BONDS)) {
+        /* There is sufficient capacity for another bond. */
+        return 0;
+    }
+
+    /* No capacity for an additional bond.  Tell the application to make
+     * room.
+     */
+    memset(&event, 0, sizeof event);
+    event.obj_type = obj_type;
```
rymanluk: I was hoping to extract that part of code as a helper which sends overflow event, but maybe we can do it when see need to construct same events in more places.

ccollins476ad: Isn't that what ble_store_status() does?  Or were you thinking this function would take parameters and fill the event object in itself?

net/nimble/host/include/host/ble_store.h
ccollins476ad: I didn't think that information is necessary for the reader to understand the code.  The obj_type field isn't part of the union, so is valid for all events.  The union fields, on the other hand, are not always valid; the user needs to know how to determine which one to use.

I am a bit confused about what you are saying here.  Which of these are you saying I should do:

1. Valid for all possible obj_types

2. Valid for the following obj_types:
* BLE_STORE_OBJ_TYPE_OUR_SEC
* BLE_STORE_OBJ_TYPE_PEER_SEC
* BLE_STORE_OBJ_TYPE_CCCD

rymanluk: yes it is it

net/nimble/host/src/ble_sm.c
rymanluk: yes, function which take parameter. I think I mentioned before something like that:
ble_store_overflow_event(obj_type, event_code, void *event_data)

net/nimble/host/include/host/ble_store.h
ccollins476ad: I was asking which of those two you would like to see :).

rymanluk: Ok, please see below what I have in mind.

   union {
        /**
         * The record that failed to be written. Valid for all the obj_type and following event
         * types:
         *     o BLE_STORE_EVENT_OVERFLOW
         */
        const union ble_store_value *value;

        /**
         * The connection handle of the peer.  Valid for the following event
         * types:
         *     o BLE_STORE_EVENT_OVERFLOW_NEXT
         *
         *   and following obj_types:
         *  
         *     o BLE_STORE_OBJ_TYPE_OUR_SEC
         *     o BLE_STORE_OBJ_TYPE_PEER_SEC
         */
        uint16_t conn_handle;
    };

In this way, whenever we need to handle object which also uses BLE_STORE_EVENT_OVERFLOW_NEXT but  has different data than conn_handle, we can easly add it here.

ccollins476ad: I really do want to make changes that you suggest, but, subjectively, I'm just not sure its an improvement in the signal-to-noise of the code.  In my mind, the concept of "validity" doesn't really apply here.  This is a discriminated union, and the user needs to know which field can be validly read.  That selection is made via the "event_code" field (i.e., that is the descriminator).  The "obj_type" field has no effect on the validity of each member of the union.

Also, I'm not sure we'll never use this event for the CCCD object type.  I think we should use it for all object types where possible.  The current CCCD persistence code doesn't do a good job of handling all the error cases, so hopefully we can improve it going forward.

net/nimble/host/src/ble_sm.c
ccollins476ad: OK, I understand.  Thanks, that is a good idea.  I have added two functions:

```
int ble_store_overflow_event(int obj_type, const union ble_store_value *value);
int ble_store_overflow_next_event(int obj_type, uint16_t conn_handle);
```

and made `ble_store_status` static.

net/nimble/host/include/host/ble_store.h
rymanluk: Hmm, I'm sorry but for me it still not clear from API point of view.

So maybe we should do something similar to e.g. struct ble_gap_event

Event type always defines valid union type.

What would you say about that?

> struct ble_store_status_event {
>     /**
>      * The type of event being reported; one of the BLE_STORE_EVENT_TYPE_[...]
>      * codes.
>      */
>     int event_code;
> 
>     /**
>      * Additional data related to the event; the valid field is inferred from
>      * the obj_type,event_code pair.
>      */
>     union {
>         /**
>          * Represent storage object which could not be written. Valid for the following event
>          * types:
>          *     o BLE_STORE_EVENT_OVERFLOW
>          */
>         struct {
>            /* Object type which failed to be written */
>             int obj_type;
>            
>             /* The record that failed to be written.*/
>             const union ble_store_value *value;
>         } overflow ;
> 
>         /**
>          * Represents element which might be written but storage is full.  Valid for the following event
>          * types:
>          *     o BLE_STORE_EVENT_STORAGE_FULL
>          */
> 
>         struct {
>            /* Object type which could not be written due to storage full*/
>             int obj_type;
> 
>             /* conn handle used for:
>              * BLE_STORE_OBJ_TYPE_OUR_SEC
>              * BLE_STORE_OBJ_TYPE_PEER_SEC
>              */
>             uint16_t conn_handle;
> 
>             // We can add fieds for other types or create union if needed.
>         } storage_full;
>     };
> };

ccollins476ad: Thanks, I have made that change.

rymanluk: We can also remove now `int obj_type;` from  `struct ble_store_status_event`, so we have there only  `int event_code;` and `union`

ccollins476ad: Yikes, I can't believe I missed that.  Thanks for catching this.  I have fixed it.

PR #280, opened 2017-05-13
BLE Host - Persist bonding material to sys/config

ccollins476ad:
This pull request adds a new package: ble/host/store/config.  This package holds security material and CCCDs in RAM, and optionally saves them to sys/config.  Because this package is a superset of store/ram, it deprecates that one.

This PR incorporates changes from other PRs:
* 277 (#277)
* 279 (#279)

Therefore, this PR should only be merged after those.

### DIFF
--- a/net/nimble/host/store/config/src/ble_store_config.c
+++ b/net/nimble/host/store/config/src/ble_store_config.c
@@ -1,0 +1,533 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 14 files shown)


mkiiskila:
I could review this. Do you mind resolving the conflicts first, given that the other PRs have been merged? That would make the patch easier to read :)

ccollins476ad:
Thanks, Marko.  I didn't realize so many conflicts had sprouted.  They should be resolved now.

net/nimble/host/store/config/src/ble_store_config.c
```diff
@@ -0,0 +1,533 @@
...
+    uint8_t *src;
+    int move_count;
+
+    (*num_values)--;
+    if (idx < *num_values) {
+        dst = values;
+        dst += idx * value_size;
+        src = dst + value_size;
+
+        move_count = *num_values - idx;
```
mkiiskila: Should this be multiplied by value_size?

ccollins476ad: Yes, good catch.  I will fix that and see if I can add some unit tests for this.

ccollins476ad: It turns out there already were unit tests, but they were using the old deprecated store package (net/nimble/host/store/ram)!

In addition to the above change, I also had to reduce some setting defaults:
* BLE_STORE_MAX_BONDS: 3 (was: 4)
* BLE_STORE_MAX_CCCDS: 8 (was: 16)

The problem is that sys/config limits a value to <= 256 bytes.  We should probably make this configurable, but even if we do, we don't have a good way to ensure these settings are in sync.  The best we can do is raise an error with `#if`/`#error`.

PR #281, opened 2017-05-16
imgmgr: add erase command, remove erase from upload command

jacobrosenthal:
closes https://issues.apache.org/jira/browse/MYNEWT-747

Im guessing this needs code review. I just stripped what looks like what was needed for erase command and put it in a new command.

### DIFF
--- a/mgmt/imgmgr/src/imgmgr.c
+++ b/mgmt/imgmgr/src/imgmgr.c
@@ -271,6 +271,8 @@ imgr_erase(struct mgmt_cbuf *cb)
         }
         rc = flash_area_erase(imgr_state.upload.fa, 0,
           imgr_state.upload.fa->fa_size);
+        flash_area_close(imgr_state.upload.fa);
+        imgr_state.upload.fa = NULL;
     } else {
         /*
          * No slot where to erase!
--- a/mgmt/imgmgr/src/imgmgr.c
+++ b/mgmt/imgmgr/src/imgmgr.c
@@ -35,6 +35,7 @@
 #include "imgmgr_priv.h"
 
 static int imgr_upload(struct mgmt_cbuf *);
+static int imgr_erase(struct mgmt_cbuf *);
 
 static const struct mgmt_handler imgr_nmgr_handlers[] = {
     [IMGMGR_NMGR_ID_STATE] = {
@@ -44,6 +45,10 @@ static const struct mgmt_handler imgr_nmgr_handlers[] = {
     [IMGMGR_NMGR_ID_UPLOAD] = {
         .mh_read = NULL,
         .mh_write = imgr_upload
+    },
+    [IMGMGR_NMGR_ID_ERASE] = {
+        .mh_read = NULL,
+        .mh_write = imgr_erase
     },
     [IMGMGR_NMGR_ID_CORELIST] = {
 #if MYNEWT_VAL(IMGMGR_COREDUMP)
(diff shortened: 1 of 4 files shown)


ccollins476ad:
I don't think we should remove the erase operation from the upload command handler.  Doing so breaks compatibility with clients who don't erase the slot first.

I think the code should read the entire slot to determine if there are any written bytes (i.e., any bytes that aren't 0xff).  If there are, the erase should be performed.  Otherwise, the erase can be bypassed.

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -223,6 +228,70 @@ imgr_find_by_hash(uint8_t *find, struct image_version *ver)
...
+        if (imgr_state.upload.fa) {
+            flash_area_close(imgr_state.upload.fa);
+            imgr_state.upload.fa = NULL;
+        }
+        rc = flash_area_open(area_id, &imgr_state.upload.fa);
+        if (rc) {
+            return MGMT_ERR_EINVAL;
+        }
+        rc = flash_area_erase(imgr_state.upload.fa, 0,
+          imgr_state.upload.fa->fa_size);
```
mkiiskila: Could call flash_area_close() for completeness.

sys/flash_map/include/flash_map/flash_map.h
```diff
@@ -41,6 +41,7 @@ extern "C" {
  * match the linker scripts when platform executes from flash,
  * and match the target offset specified in download script.
  */
+#include <stdlib.h>
```
mkiiskila: Do we need this inclusion?

sys/flash_map/src/flash_map.c
```diff
@@ -21,6 +21,7 @@
 #include <string.h>
 #include <assert.h>
 
+#include <os/os.h>
```
mkiiskila: Do we need this inclusion?

sys/flash_map/src/flash_map.c
```diff
@@ -131,6 +132,40 @@ flash_area_align(const struct flash_area *fa)
...
+flash_area_is_empty(const struct flash_area *fa, bool *empty)
+{
+    uint32_t data[64 >> 2];
+    uint32_t data_off = 0;
+    int8_t bytes_to_read;
+    uint8_t i;
+    int rc;
+
+    while(data_off < fa->fa_size)
+    {
```
mkiiskila: Can you fix the coding style?
See https://github.com/apache/incubator-mynewt-core/blob/master/CODING_STANDARDS.md

sys/flash_map/src/flash_map.c
```diff
@@ -131,6 +132,40 @@ flash_area_align(const struct flash_area *fa)
...
+        for (i = 0; i < bytes_to_read >> 2; i++){
+          if (data[i] != (uint32_t) -1) {
+            goto not_empty;
+          }
+        }
+        for (i = i << 2; i < bytes_to_read; i++) {
+          if (*(((uint8_t *) data) + i) != (uint8_t) -1) {
+            goto not_empty;
+          }
+        }
```
mkiiskila: I don't think this 2nd for loop is needed. I have not encountered a flash sector which would not be a multiple of minimally 4 bytes.

sys/flash_map/src/flash_map.c
jacobrosenthal: yes for min, though that could be duplicated so os isnt needed, or put somewhere else

sys/flash_map/include/flash_map/flash_map.h
jacobrosenthal: Yes for bool

ccollins476ad: For bool, you should include <stdbool.h>

jacobrosenthal: Ah yes. Roger

mgmt/imgmgr/src/imgmgr.c
jacobrosenthal: Done

jacobrosenthal:
BTW hold on this until I get this last discussion finished on why nrf51 radio doesnt appear to come back up cleanly and advertise after a flash_erase disconnect. 

Its interesting though if I break and step a bit it does come back, so it seems like a timing issue?

sys/flash_map/src/flash_map.c
mkiiskila: Sounds good enough reason to me.

mkiiskila:
Single stepping sometimes leads to things taking off on their own; I have not been able to pinpoint when and why it does. However, my guess would be that it is unrelated to reconnect issue. Explicit breakpoints are more reliable, I use those when single-stepping runs wild.

jacobrosenthal:
Oh im saying the opposite. breaking/stepping in the advertise function fixes it. Otherwise the radio doesnt come back after erase. I think its not ready yet and breaking gives it a millisecond to catch up.

jacobrosenthal:
OK. Turns out I had low power crystal support syscfg flags turned on for my bsp and it doesnt have a low power crystal, removing those removed this odd issue. 
Id say this is ready for squash and merge whenever you all decide.

PR #283, opened 2017-05-17
Coded PHY support

andrzej-kaczmarek:
This adds basic support for Coded PHY, i.e. connection can be switched to Coded PHY and should work as long as data can fit in Coded PHY timings since constraints for Coded PHY are not enforced yet.


sterlinghughes:
+1 -- good man!

PR #287, opened 2017-05-17
Repeat pairing

ccollins476ad:
This pull request addresses the issue identified in MYNEWT-750 (https://issues.apache.org/jira/browse/MYNEWT-750).

This PR incorporates changes from another PR: 277 (#277). Therefore, this PR should only be merged after 277.

### DIFF
--- a/apps/blecent/src/main.c
+++ b/apps/blecent/src/main.c
@@ -435,6 +435,22 @@ blecent_gap_event(struct ble_gap_event *event, void *arg)
                     event->mtu.value);
         return 0;
 
+    case BLE_GAP_EVENT_REPEAT_PAIRING:
+        /* We already have a bond with the peer, but it is attempting to
+         * establish a new secure link.  This app sacrifices security for
+         * convenience: just throw away the old bond and accept the new link.
+         */
+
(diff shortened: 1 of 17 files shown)


sjanc:
Hi,
does this cover scenario for security level increase? from unauthenticated to authenticated

ccollins476ad:
Yes, this does apply to a security level increase.  The application has to do a little more work, though.  Specifically, the application needs to compare the properties of the current bond to those of the imminent secure link.  E.g.,

```
if (event.repeat_pairing.cur_authenticated && !event.repeat_pairing.new_authenticated) {
    return BLE_GAP_REPEAT_PAIRING_IGNORE;
}

if (event.repeat_pairing.cur_sc && !event.repeat_pairing.new_sc) {
    return BLE_GAP_REPEAT_PAIRING_IGNORE;
}

/* Delete existing bond using ble_store_delete_peer(). */

return BLE_GAP_REPEAT_PAIRING_RETRY;
```

I do think the application needs this level of control, though.  If there is a good default behavior, we can supply it as a library function.

limal:
I'm having the same issue with pairing. When is this PR expected to be merged?

PR #288, opened 2017-05-18
MYNEWT-742 Initialize OIC automatically.

ccollins476ad:
Prior to this commit, the OIC library requires some manual
initiliazation.  E.g.,

    static const oc_handler_t omgr_oc_handler = {
        .init = omgr_app_init,
    };

    int
    main(int argc, char **argv)
    {
        /* ... */
        oc_main_init((oc_handler_t *)&omgr_oc_handler);
        oc_ble_coap_gatt_srv_init();
    }

Now, the library gets initialized automatically.  The application can
optionally configure the library in main() by writing to the global
oc_cfg object.  E.g.,

    oc_cfg.main_handler.init = omgr_app_init;

### DIFF
--- a/apps/bleprph_oic/src/main.c
+++ b/apps/bleprph_oic/src/main.c
@@ -32,6 +32,7 @@
 #include <oic/oc_api.h>
 #include <oic/oc_gatt.h>
 #include <oic/oc_log.h>
+#include <oic/oc_mynewt.h>
 #include <cborattr/cborattr.h>
 
 /* BLE */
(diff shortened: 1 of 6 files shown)


mkiiskila:
I would hold off from this change for now.

The reason is that this would break the existing platforms where OIC runs over IP. When OIC initializes, it joins multicast groups on all existing interfaces. If interfaces come up later, they will not be enabled.
What we should have before this commit is merged is a way for net/oic to notice that interface state has changed (or periodically poll for new interfaces). At the moment we don't have this facility.

mkiiskila:
Rest of the changes are ok, BTW. I would just remove oc_ev_start/oc_event_start().

PR #289, opened 2017-05-19
Shell fixes from review #256

michal-narajowski:
These patches were meant to be sent along with #256, but that PR was merged earlier.


michal-narajowski:
test this please

PR #290, opened 2017-05-23
nimble/controller: Fix setting parameters for scanning

andrzej-kaczmarek:
Parameters set by HCI LE Set Scan Parameters should be stored and then
applied each time scan is enabled using HCI LE Set Scan Enable. This way
scan uses proper parameters as set by host.

Without this fix scan parameters in SM will be overwritten when
initiating new connection. Since also scan type is set to 'initiating'
in such case, no advertising reports will be sent to host until scan
parameters are set once more by the host.


wes3:
This is a good catch. I did not realize that the host could issue a scan enable with issuing a set scan parameters first. Changes look good.

PR #293, opened 2017-05-24
apps/btshell: add bletiny clone that uses new shell

michal-narajowski:
Features improved command naming and tab completion functionality.
In the future it could replace bletiny. Later some commands
implementations could me placed inside respective packages and added
to shell as separate modules, e.g. l2cap, gatt, gap, etc.

### DIFF
--- a/apps/btshell/src/cmd.c
+++ b/apps/btshell/src/cmd.c
@@ -71,7 +71,7 @@ cmd_parse_conn_start_end(uint16_t *out_conn, uint16_t *out_start,
     return 0;
 }
 
-static struct kv_pair cmd_own_addr_types[] = {
+static const struct kv_pair cmd_own_addr_types[] = {
     { "public",     BLE_OWN_ADDR_PUBLIC },
     { "random",     BLE_OWN_ADDR_RANDOM },
     { "rpa_pub",    BLE_OWN_ADDR_RPA_PUBLIC_DEFAULT },
@@ -79,7 +79,7 @@ static struct kv_pair cmd_own_addr_types[] = {
     { NULL }
 };
 
-static struct kv_pair cmd_peer_addr_types[] = {
+static const struct kv_pair cmd_peer_addr_types[] = {
     { "public",     BLE_ADDR_PUBLIC },
     { "random",     BLE_ADDR_RANDOM },
     { "public_id",  BLE_ADDR_PUBLIC_ID },
@@ -87,7 +87,7 @@ static struct kv_pair cmd_peer_addr_types[] = {
     { NULL }
 };
 
-static struct kv_pair cmd_addr_type[] = {
+static const struct kv_pair cmd_addr_type[] = {
     { "public",     BLE_ADDR_PUBLIC },
     { "random",     BLE_ADDR_RANDOM },
     { NULL }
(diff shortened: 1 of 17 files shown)


net/nimble/host/src/ble_gatts.c
```diff
@@ -2080,6 +2080,141 @@ ble_gatts_count_cfg(const struct ble_gatt_svc_def *defs)
     return 0;
 }
 
+static const char* ble_gatt_chr_f_names[] = {
```
ccollins476ad: The pointed-to strings are const, but the pointers themselves aren't.  The pointers should probably be made const so that this array can be placed in .text rather than .data.

E.g.,

`static const char * const ble_gatt_chr_f_names[]`

The same comment applies to the ble_gatt_dsc_f_names array below.

apps/btshell/src/cmd.c
```diff
@@ -0,0 +1,2515 @@
...
+};
+
+static const struct shell_cmd_help gatt_service_changed_help = {
+    .summary = "send service changed indication",
+    .usage = NULL,
+    .params = gatt_service_changed_params,
+};
+
+/*****************************************************************************
+
```
ccollins476ad: I noticed a lot of the command names have changed from bletiny.  I'm personally OK with this if we think the new names are better, but just beware that there is some documentation that will need to change if bt_shell replaces bletiny (e.g., http://mynewt.apache.org/latest/network/ble/bletiny/bletiny_api/)

apps/btshell/src/parse.c
```diff
@@ -0,0 +1,611 @@
+/*
```
ccollins476ad: Btw, a lot of this file (parse.c) has been extracted from bletiny into a new package: util/parse.  I wouldn't expect it to be part of this pull request, but at some point we should think about leveraging that package and removing the duplicate code in these apps.

apps/btshell/src/cmd.c
```diff
@@ -0,0 +1,2515 @@
...
+    *out_end = parse_arg_uint16("end", &rc);
+    if (rc != 0) {
+        return rc;
+    }
+
+    return 0;
+}
+
+static struct kv_pair cmd_own_addr_types[] = {
```
ccollins476ad: These arrays should all be const so that they can be placed in .text (yes, I know I am the one who screwed this up in bletiny!)

net/nimble/host/src/ble_gatts.c
```diff
@@ -2080,6 +2080,141 @@ ble_gatts_count_cfg(const struct ble_gatt_svc_def *defs)
...
+    return buf;
+}
+
+
+#define STRINGIFY(X) #X
+#define FIELD_NAME_LEN STRINGIFY(12)
+#define FIELD_INDENT STRINGIFY(2)
+
+static void
+ble_gatt_show_local_chr(const struct ble_gatt_svc_def *svc,
```
ccollins476ad: This may verge on nit-picking, but I am not sure about putting code to print the registered resources in this file (ble_gatts.c).  This file is already pretty big and ugly, and this new functionality feels different from what is already there.

I would suggest one of the following:

1. 
* Move ble_gatt_show_local_chr() to a new file.
* Create ble_gatt_show_local_svc() in the new file.
* Modify ble_gatts_show_local() such that it just iterates the list of services, calling ble_gatt_show_local_svc() on each one.

This removes most of the new code from ble_gatts.c

2. Add accessor functions for the service list; move all new code into a new file.  ble_gatts_show_local() calls the new accessor function and prints each service.

3. Add a generic ble_gatts_lcl_svc_foreach() function to ble_gatts.c.  This function takes a function pointer which gets called for each registered service.  The new print code could go in a new file and be implemented with a call to the new foreach function.

michal-narajowski:
Hi, Chris. Thanks for the review. Fixed the things you mentioned. 

Of course, the documentation will need to change if btshell replaces bletiny. For now, here is the mapping between bletiny and btshell commands:
```
adv     -> advertise
chrup   -> gatt-notify
conn    -> connect
datalen -> conn-datalen
disc    -> gatt-discover-{characteristic,descriptor,service,full}
find    -> gatt-find-included-services
l2cap   -> l2cap-{update,connect,disconnect,create-server}
mtu     -> gatt-exchange-mtu
passkey -> auth-passkey
read    -> gatt-read
rssi    -> conn-rssi
scan    -> scan
sec     -> security-{pair,start,encryption,set-data}
set     -> set{-adv-data}
show    -> gatt-show{-local,-conn,-coc,-addr}
store   -> keystore-{add,del,show}
svcchg  -> gatt-service-changed
term    -> disconnect
tx      -> test-tx
update  -> conn-update-params
wl      -> white-list
write   -> gatt-write
```

PR #296, opened 2017-05-27
Gap log

jacobrosenthal:
ble_gap_log_adv  is referencing globals that have not been set yet, instead of what was passed to it. Secondarily, there are ble_gap_log_adv struct members that arent actually used? 

Several ways to fix this. Could log after the globals are set ( and keep them even though theyre unused, maybe someone has a future plan here...) Or reference the passed object only and clean house as I have.


rymanluk:
Looks good to me. I see that advertising scan response data we set directly to controller so no need to keep it here. I will give other guys to comment on this for 1 day or so and if no objections I'm going to merge it.

rymanluk:
Thanks Jacob, meged

PR #297, opened 2017-05-27
Corrected two issues in I2C clear bus function. 

EarthLord:
1st the incorrect initialization of the SCL and SDA GPIOs and 2nd, the typo in generation SCL clock.


EarthLord:
Sorry, should have read the contributing doc properly. Now it's ready to be merged?

PR #306, opened 2017-06-05
Nosig armv7

ccollins476ad:
The rationale for this change is summarized in https://issues.apache.org/jira/browse/MYNEWT-770

Regarding implementation, I decided on the following:
* Create a new package: `@apache-mynewt-core/kernel/sim`.  This package contains the "signals" and "no-signals" implementations of the simulator.
* Create two new BSPs: `native-armv7` and `native-mips`
* BSPs which use the simulator depend on the kernel/sim package.  This set currently consists of three BSPs: native, native-armv7, and native-mips.

This solution comes with a few annoyances:
* Duplicate code and scripts in the new native BSPs.
* Wrapper functions in all the `kernel/os/src/arch/sim.../os_arch.c` files.  These functions just call into functions defined in the `kernel/sim` package.

I still think this is the right approach.  Conceptually, each of these platforms is indeed a different BSP.  The wrapper functions are a bit annoying, but I think it is good to keep the sim implementation as a separate module rather than finding a way to force shared code into architecture-specific source directories.  However, I am definitely open to comments or suggestions.


hw/bsp/native-armv7/include/bsp/bsp.h
```diff
@@ -16,24 +16,32 @@
...
+#define sec_bss_nz_core
+
+/* More convenient section placement macros. */
+#define bssnz_t
+
+/* LED pins */
+#define LED_BLINK_PIN   (0x1)
+
+/* UART info */
+#define CONSOLE_UART    	"uart0"
```
mkiiskila: Dont' forget to remove this line, and the other one in native-mips bsp when pull request #314 gets merged.

PR #312, opened 2017-06-06
MYNEWT-664 add length to coredownload rsp.

ccollins476ad:
This change is discussed in https://issues.apache.org/jira/browse/MYNEWT-664


vrahane:
Changes look good. I do have a question, should there be a corresponding change in newtmgr client code to look at the length and compare it against the length of the data that it receives ?

PR #314, opened 2017-06-07
MYNEWT_689: Make CONSOLE_UART a syscfg variable.

wes3:
Removed CONSOLE_UART from bsp.h and made it a syscfg variable.
The variable is called CONSOLE_UART_DEV and is defined in the
sys/console directories. This can be overridden by the bsp, app,
or target. Some bsps used different settings for the console uart
and those bsps override the default uart (which is uart0).


utzig:
I just merged `stm32f7discovery` which will probably require those changes applied too.

mkiiskila:
Is there going to be another PR against BSPs in arduino zero repo as well?

wes3:
Yes, there will be PRs for the other repos. Thanks for reminding me.

PR #318, opened 2017-06-07
MYNEWT-681 rb-nano2 remove flash write protection in download scripts 

vrahane:
- removing write protection in download script by setting the write
  protection bits in the BPROT config registers. Also, setting the
  DISABLEINDEBUG register.

### DIFF
--- a/hw/bsp/rb-nano2/rb-nano2_download.sh
+++ b/hw/bsp/rb-nano2/rb-nano2_download.sh
@@ -37,6 +37,14 @@ if [ "$MFG_IMAGE" ]; then
     FLASH_OFFSET=0
 fi
 
+# We write the config registers for BPROT so that NVMC write
+# protection gets disabled for all blocks of the flash
+# We also write the DISABLEINDEBUG register so that, write
+# protection gets disabled in debug register by default
+CFG_POST_INIT="mww 0x40000600 0;mww 0x40000604 0;mww 0x40000608 1;\
+               mww 0x40000610 0;mww 0x40000614 0;
+              "
+
 common_file_to_load
 openocd_load
 openocd_reset_run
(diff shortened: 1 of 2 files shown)


hw/bsp/rb-nano2/rb-nano2_download.sh
```diff
@@ -37,6 +37,15 @@ if [ "$MFG_IMAGE" ]; then
     FLASH_OFFSET=0
 fi
 
+# We write the config registers for BPROT so that NVMC write
+# protection gets disabled for all blocks of the flash
+# We also write the DISABLEINDEBUG register so that, write
+# protection gets disabled in debug register by default
+EXTRA_GDB_CMDS="set {unsigned long}0x40000600=0;set {unsigned long}0x40000604=0;\
```
mkiiskila: Did you mean CFG_POST_INIT instead of EXTRA_GDB_CMDS?

vrahane: Sorry, my bad. I assumed there was a EXTRA_GDB_CMDS in the openocd script. CFG_POST_INIT goes to the OCD_CMD_FILE which is more of ocd commands than gdb commands. I think what is missing in the openocd script is EXTRA_GDB_CMDS which goes to the GDB_CMD_FILE.

vrahane:
Changed it to use openocd commands instead of gdb commands. Also added the EXTRA_GDB_CMDS variable for future use.

hw/scripts/openocd.sh
```diff
@@ -101,7 +101,9 @@ openocd_debug () {
         if [ ! -z "$RESET" ]; then
             echo "mon reset halt" >> $GDB_CMD_FILE
         fi
-	echo "FILENAME" $FILE_NAME >>out
+  echo "$EXTRA_GDB_CMDS" >> $GDB_CMD_FILE
```
mkiiskila: Indentation is wrong on this line. Are you using tab instead of space?

hw/scripts/openocd.sh
```diff
@@ -101,7 +101,9 @@ openocd_debug () {
...
+  echo "FILENAME" $FILE_NAME >>out
```
mkiiskila: I don't see a reason for this line. Why is it here? Could it be removed as part of this commit?

PR #324, opened 2017-06-09
Ensure Redbear Blend2 has i2c headers

pctj101:
This code at the end of the rb-blend2/src/hal_bsp.c files requires the hal_i2c.h headers.

```
#if MYNEWT_VAL(I2C_0)
    rc = hal_i2c_init(0, (void *)&hal_i2c_cfg);
    assert(rc == 0);
#endif
```

### DIFF
--- a/hw/bsp/rb-blend2/src/hal_bsp.c
+++ b/hw/bsp/rb-blend2/src/hal_bsp.c
@@ -29,6 +29,7 @@
 #include "flash_map/flash_map.h"
 #include "hal/hal_flash.h"
 #include "hal/hal_spi.h"
+#include "hal/hal_i2c.h"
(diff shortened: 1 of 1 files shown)


utzig:
The BSP for nano2 doesn't seem to use any `i2c` stuff. Why is the include required?

pctj101:
Aha, you're right.  While I was testing the raw module I didn't notice the Nano didn't default to exposing i2c.  Thanks for the feedback.

pctj101:
As a note, redbear does mark sda0 and scl0 in their pinout as an alternate function:
https://github.com/redbear/nRF5x#ble-nano-2

If you suggest, in the future, I could look into implementing some code for that especially since it doesn't disable the primary TX/RX serial pins.  Let me know what you think.

PR #329, opened 2017-06-12
Wi-Fire: Implement hal_timer

francois-berder:


hw/mcu/microchip/pic32mz2048efg100/src/hal_timer.c
```diff
@@ -0,0 +1,548 @@
...
+}
+
+void
+__attribute__((interrupt(IPL3AUTO), vector(_TIMER_2_VECTOR)))
+timer2_isr(void)
+{
+    ++timers[0].interrupt_count;
+
+    if (!(TAILQ_EMPTY(&timers[0].hal_timer_queue)) &&
+        timers[0].expiry <= timers[0].interrupt_count) {
```
mkiiskila: Is there a register you can read for the time instead of relying on interrupt count? If frequency of the timer is set to be high, there's a fair chance that interrupts are going to be missed when someone else is blocking interrupts for some amount of time.

It would be useful to look at other hal_timer implementations to see how they deal with this issue.

hw/mcu/microchip/pic32mz2048efg100/src/hal_timer.c
```diff
@@ -0,0 +1,548 @@
...
+            ++timers_to_remove;
+        } else {
+            break;
+        }
+    }
+
+    for (i = 0; i < timers_to_remove; ++i) {
+        TAILQ_REMOVE(&timers[timer_num].hal_timer_queue,
+                     TAILQ_FIRST(&timers[timer_num].hal_timer_queue), link);
+    }
```
francois-berder: OK, I will modify the function.

hw/mcu/microchip/pic32mz2048efg100/src/hal_timer.c
francois-berder: I tested my code by setting a timer at 500kHz and I was already hitting some issues because the ```hal_timer_read``` took too much time to execute in the interrupt handler, that is why I used interrupt count to make ```hal_timer_read``` as fast as possible.

> Is there a register you can read for the time instead of relying on interrupt count?

Yes I can use ```PRx``` and ```TMRx```. 

I was looking at the nrf51xxx hal_timer implementation and I found out two things:
- The tick does not seem to be related to ```OS_TICKS_PER_SEC```. Having a 64-bit counter that increases at a fixed rate would be the current tick of the timer.
- The ```hal_config``` only seems to configure prescaler as the compare register is changed when a soft timer is added/removed in the list to generate as few interrupts as possible. From my understanding, the parameter ```freq_hz``` in ```hal_config``` does not specify the frequency of  interrupts generated by the timer module, but instead specifies the speed of increase of the ```TMRx``` register.

Are these two assumptions correct ?

wes3: The hal timer code is not related to OS time at all. As you mention with hal_config the timer gets configured with a frequency and that is the rate at which the underlying timer counter counts (each tick of the counter takes 1/freq). The rate at which interrupts are generated are completely independent of this.

mkiiskila: Yes, hal_timer is independent of os tick. These are meant to be used to do things when user needs to do things with a finer granularity of timing than what OS ticker provides. E.g. low-layer radio stuff, bitbanging output etc.

Yes for the 2nd statement as well. It is better to take less number of interrupts! If you have compare register available, then you could use those. Also note that you might need to deal with overflows, if your TMRx/PRx are 32-bit registers. Hardware most likely already provides you an interrupt source for that. If there is no such capability, you can use compare register for that same purpose.

Hope this helps!

francois-berder: Thanks! I will do the appropriate changes.

PR #334, opened 2017-06-13
MYNEWT-699; re-add sample linker scripts to Olimex BSP.

mkiiskila:


hw/bsp/olimex_stm32-e407_devboard/run_from_loader.ld
```diff
@@ -0,0 +1,31 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/* Linker script for STM32F407 when running from flash and using the bootloader */
+
+/* Linker script to configure memory regions. */
+MEMORY
+{
+  FLASH (rx) :  ORIGIN = 0x08020000, LENGTH = 384K /* First image slot. */
```
ccollins476ad: Should this be 256K (rather than 384K)?  The bsp.yml file specifies a slot size of 256kB.

mkiiskila: Looks like linker script and bsp.yml are out of sync in this platform.

PR #335, opened 2017-06-14
Mynewt 529

mkiiskila:
Moved i2c scanner from sensor cli to it's own package.
Changed flash_test package to use sysinit.
Added backspace support to console.


vrahane:
Can you please also add the i2c_scan dependency to app/sensors_test.

mkiiskila:
i2c_scan package added to sensors_test app

PR #338, opened 2017-06-16
Increase MCU clock speed on STM32F7-Discovery to maximum 168 MHz.

thejpster:
Function generated using Cube MX, with the STM32F7-DISCOVERY board
definition. Uses internal oscillator at 16 MHz.


utzig:
Is there a reason to use 168MHz (as for the F4 family) instead of the max 216MHz support by most parts in the F7 family?

thejpster:
That's what CubeMX offered when I selected the F7 Discovery.

utzig:
What is the exact MCU model?

utzig:
Btw, it's ok to merge IMO, but if the MCU goes to 216MHz (I believe it does!) we can update it later. I'll test clock updates on the nucleo stm32f767.

PR #341, opened 2017-06-20
MYNEWT-785: Remove support for 1MHz clock in nimble

andrzej-kaczmarek:
This removes controller code which uses 1MHz clock so we always use code for 32.768kHz clock.


rymanluk:
Looks good to me!

PR #343, opened 2017-06-22
Enable i2c_scan only when there is I2C peripheral

utzig:
The slinky app includes i2c_scan, but this causes a build error
on HAL ports which don't include an i2c driver. This change only
includes i2c_scan when some I2C_x switch is enabled.

### DIFF
--- a/apps/slinky/pkg.yml
+++ b/apps/slinky/pkg.yml
@@ -26,7 +26,6 @@ pkg.keywords:
 
 pkg.deps:
     - test/flash_test
-    - test/i2c_scan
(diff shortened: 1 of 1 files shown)


utzig:
If `i2c_scan` was actually used by other apps apart from slinky, maybe it would have been better to protect its code with something like:
```
#if MYNEWT_VAL(I2C_0) || MYNEWT_VAL(I2C_1) || MYNEWT_VAL(I2C_2)
// i2c_scan's code...
#endif
```
But doesn't seem necessary for now.

PR #345, opened 2017-06-22
Add support for the NRF52 thingy, add a button driver

sterlinghughes:


sterlinghughes:
after discussion on slack, closing and resubmitting as just a BSP for the thingy.  We're going to make the button driver a bit more general purpose.

PR #349, opened 2017-06-25
Wi-Fire: Add I2C driver 

francois-berder:
This pull request implements a driver for the I2C peripheral.
It has been tested using a MPU6050 and reading the product ID of this device using a wi-fire.

### DIFF
--- a/hw/mcu/microchip/pic32mz2048efg100/src/hal_i2c.c
+++ b/hw/mcu/microchip/pic32mz2048efg100/src/hal_i2c.c
@@ -1,0 +1,276 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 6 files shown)


hw/mcu/microchip/pic32mz2048efg100/src/hal_i2c.c
```diff
@@ -0,0 +1,276 @@
...
+ * under the License.
+ */
+
+#include <xc.h>
+#include "bsp/bsp.h"
+#include <hal/hal_gpio.h>
+#include <hal/hal_i2c.h>
+#include <mcu/p32mz2048efg100.h>
+#include "mcu/mips_hal.h"
+#include "syscfg/syscfg.h"
```
mkiiskila: You probably want to move include for syscfg.h higher. I recommend including it before any of the other Mynewt headers.

PR #351, opened 2017-06-26
Bt5 extended advertising - host part

rymanluk:
This pull request contains host part of extended advertising feature.
Note: This is experimental API.

### DIFF
--- a/net/nimble/host/src/ble_gap.c
+++ b/net/nimble/host/src/ble_gap.c
@@ -109,6 +109,7 @@ struct ble_gap_master_state {
 
         struct {
             uint8_t limited:1;
+            uint8_t extended:1;
         } disc;
     };
 };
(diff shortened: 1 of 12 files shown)


rymanluk:
Thanks for review.
1) your right however in practice it works. I did it because I was remember that patch

433aab55c nimble host test; change the way empty struct is initialized to get this to compile on linux gcc.
But I see now that problem there was probably due to struct inside struct.

Anyway, fixed.

2) Fixed
3) Fixed
4) Fixed
5) BLE_LL_CFG_FEAT_LL_EXT_ADV is there to keep consistency with other features which are present in supported features mask.  Actually with new patchset I move BLE_EXT_ADV into 'nimble/syscfg.yml'. Idea is that user application will set only this flag to configure host and controller. E.g. also for blehci
So maybe indeed BLE_LL_CFG_FEAT_LL_EXT_ADV could be just  #define inside controller  code which would depend on BLE_EXT_ADV. In general current configuration looks good to me. The only problem we could have with it is when somebody manually set BLE_LL_CFG_FEAT_LL_EXT_ADV and not BLE_EXT_ADV. In this case some build errors shows up in hci code I guess.

Lets wait for Will comment on this.
Also that can be fixed in separate patch once we came to the conclusion.

6) ups I missed that somehow. Should be fixed now.

wes3:
I am not sure I have a strong opinion on the BLE_LL_CFG_FEAT_LL_EXT_ADV and BLE_EXT_ADV. I see why we want to keep BLE_LL_CFG_FEAT_LL_EXT_ADV; that makes sense. What I think we should do is restrict LL_EXT_ADV so that it basically mirrors BLE_EXT_ADV. Chris says you can do this:

    BLE_LL_CFG_FEAT_LL_EXT_ADV:
        description: >
            This option is used to enable/disable support for Extended
            Advertising Feature. That means extended scanner, advertiser
            and connect.
        value: 0
        restrictions:
            - 'BLE_EXT_ADV if 1'
            - '!BLE_EXT_ADV if 0'


BTW: I reviewed some of the other code and it all looks good.

net/nimble/host/src/ble_gap.c
```diff
@@ -2154,6 +2155,66 @@ ble_gap_disc_tx_params(uint8_t own_addr_type,
...
+#if MYNEWT_VAL(BLE_EXT_ADV)
+static int
+ble_gap_ext_disc_tx_params(uint8_t own_addr_type, uint8_t filter_policy,
+                       const struct ble_hs_hci_ext_scan_param *uncoded_params,
+                       const struct ble_hs_hci_ext_scan_param *coded_params)
+{
+    uint8_t buf[BLE_HCI_CMD_HDR_LEN + BLE_HCI_LE_EXT_SCAN_BASE_LEN +
+                                2 * BLE_HCI_LE_EXT_SCAN_SINGLE_PARAM_LEN];
+    uint8_t phy_mask = 0;
+    struct ble_hs_hci_ext_scan_param param[2] = {0};
```
ccollins476ad: I'm sorry to cause so much trouble with that nitpick, but I don't think this will compile.  Since it's an array, you need two initializers, e.g., `{{0}}`.

ccollins476ad: Aside from the syntax thing, it all looks good to me.  Thanks.

rymanluk: @ccollins476ad 
Actually it compiles. I think that it would not compile if there would be anything else than 0.
Anyway, I can do suggested fix

@wes3  Thanks will try that

rymanluk:
@wes3 OK it works. In addition I set default value to MYNEWT_VAL_BLE_EXT_ADV. Like this:

`BLE_LL_CFG_FEAT_LL_EXT_ADV:.
    value:  MYNEWT_VAL_BLE_EXT_ADV`

In this way, if anybody will try to change manually BLE_LL_CFG_FEAT_LL_EXT_ADV without  changing BLE_EXT_ADV than newt  will throw 'restriction' error. Otherwise both flags will have always valid values.

net/nimble/host/src/ble_gap.c
ccollins476ad: You're right - `{0}` is valid and should be accepted by the compiler. It appears older versions of gcc complain about this case.

ccollins476ad:
It seems like a lot of YAML code just to create a setting that can't be changed!   I already spoke my piece on this, so I am not going to get too hung up on it, but now I really don't understand the point of the `BLE_LL_CFG_FEAT_LL_EXT_ADV` setting.  If I understand correctly, its default value is equal to BLE_EXT_ADV, and any attempt to change it to something else will result in a newt error (restriction violation).  Wouldn't it all be simpler to just remove this setting and use BLE_EXT_ADV everywhere instead?

I understand the need to maintain consistency, but I am not convinced that argument applies here.  It would maintain consistency to provide an option that the user can actually change.  Providing an option that elicits an error if you try to change it just strikes me as confusing.

Anyway, that is just my opinion.  If you and Will prefer the current implementation, then I am fine with it being merged.

wes3:
I do not have a strong opinion but I did think it was confusing to have every other LL feature variable in the controller and have one outside of it. I do understand your point though Chris; does seem a bit useless and in the end you have to change the BLE_EXT_ADV setting to turn it on/off.

rymanluk:
The argument for current implementation is that by looking into single syscfg file, user gets idea how looks its mask of supported features :) but really, I'm open to change that if it is an issue.

net/nimble/host/src/ble_gap.c
rymanluk: Ok, will fix it, but cannot verify it on macOS

rymanluk:
updated initializer according to Chris comment.

ccollins476ad:
Thanks, it looks good to me.

PR #352, opened 2017-06-26
Make BLE sample apps more robust

ccollins476ad:
There are several scenarios in which the sample BLE apps fail to run:

_bleprph and bleprph_oic_
* *Low MTU.*  Several devices use a BLE ATT MTU of 104 (e.g., Mac OS).  With this MTU, some newtmgr responses failed to send.  The fix is to increase the number of mbufs and decrease their size

* *Image upgrade.*  Image upgrade requires a fairly large main stack.  The default values for both bleprph apps was too low, resulting a crash during processing of the upload request.

_bletiny_

Bletiny had grown too large to fit on an nRF51.  This PR reduces bletiny's code size in the following ways:

* Remove support for newtmgr-over-shell.  Newtmgr isn't included in bletiny anyway, so this change removes unused functionality.

* Make bletiny help optional via the BLETINY_HELP syscfg setting.

* Make support for the alert notification service ooptional via the BLETINY_ANS syscfg setting.

### DIFF
--- a/apps/bleprph/syscfg.yml
+++ b/apps/bleprph/syscfg.yml
@@ -19,10 +19,6 @@
 # Package: apps/bleprph
 
 syscfg.vals:
-    # Use INFO log level to reduce code size.  DEBUG is too large for nRF51.
-    LOG_LEVEL: 1
-    CONSOLE_INPUT: 0
-
     # Disable central and observer roles.
     BLE_ROLE_BROADCASTER: 1
     BLE_ROLE_CENTRAL: 0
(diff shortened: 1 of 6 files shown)


jacobrosenthal:
I think we should set blesplit OS_MAIN_STACK_SIZE as well so when those are used together they don't 
```Syscfg ambiguities detected```

ccollins476ad:
Good catch, thanks.  I have updated blesplit to have a default OS_MAIN_STACK_SIZE of 468 (same as bleprph).

PR #361, opened 2017-06-28
Shell improvements before 1.1

michal-narajowski:
This patch introduces simplifications and improvements suggested by @cwanda.

Changes include:
* bringing back shell_evq_set
* directly using shell_register api in internal shell modules
* asserting when max number of modules exceeded 
* removing module name length limit
* adding prompt suffix option to syscfg
* some whitespace fixes

### DIFF
--- a/sys/shell/include/shell/shell.h
+++ b/sys/shell/include/shell/shell.h
@@ -24,6 +24,8 @@
 extern "C" {
 #endif
 
+struct os_eventq;
+
 /** @brief Callback called when command is entered.
  *
  *  @param argc Number of parameters passed.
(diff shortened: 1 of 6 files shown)


cwanda:
Thanks, Michal for the updates. They look good and address my questions.

sjanc:
Looks good to me.

cwanda:
Please add conditional include #if MYNEWT_VAL(SHELL_OS_MODULE) around the shell_os.c code  #if MYNEWT_VAL(SHELL_PROMPT_MODULE) around the shell_prompt.c code. We have a tutorial that recommends users to disable SHELL features as a way to reduce application code size.

michal-narajowski:
This is not necessary, because when there is no call to any function defined in a file then the file is not included in the final executable. You can check that with `newt size <target> -F`.

cwanda:
Sorry about that. I weren't sure how the linker would handle the static data for the command and help structures etc and forgot about the newt size command to check.

PR #362, opened 2017-06-29
nimble: add monitor protocol/interface + improve logs

andrzej-kaczmarek:
monitor protocol core is now available in separate PR: https://github.com/apache/mynewt-core/pull/372
this will be for logging part only

See here for details: http://mail-archives.apache.org/mod_mbox/incubator-mynewt-dev/201706.mbox/%3CCADAkoF4LQOdqH%3DLh%3D5GgBPC9nEaXV95TV_uQPXXWuhrpCXfbtQ%40mail.gmail.com%3E

### DIFF
--- a/net/nimble/host/src/ble_hs_log.c
+++ b/net/nimble/host/src/ble_hs_log.c
@@ -17,8 +17,14 @@
  * under the License.
  */
 
+#include <stdio.h>
 #include "os/os.h"
 #include "host/ble_hs.h"
+
+#define HEXBUF_LENGTH   128
+#define HEXBUF_COUNT    4
+
+static const char hex[] = "0123456789abcdef";
 
 struct log ble_hs_log;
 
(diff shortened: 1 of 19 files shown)


net/nimble/host/src/ble_hs_log.c
```diff
@@ -45,3 +51,97 @@ ble_hs_log_flat_buf(const void *data, int len)
...
+            addr->val[5], addr->val[4], addr->val[3],
+            addr->val[2], addr->val[1], addr->val[0], type);
+
+    return str;
+}
+
+static char *
+get_buf(void)
+{
+    static char hexbufs[HEXBUF_COUNT][HEXBUF_LENGTH + 1];
```
ccollins476ad: This allocates a fairly large buffer, even when logging is disabled entirely.  The problem is that the log macros always evaluate their arguments even if output is disabled.

I am not sure what the right solution is, but I don't think we want to allocate such a large buffer unconditionally.  The old method of repeated calls to `ble_hs_log_flat_buf()` was clunky, but it did avoid allocating buffers upfront.

One possible solution is to use something like `#if MYNEWT_VAL(LOG_LEVEL) <= LOG_LEVEL_CRITICAL` to stub out this function when logging is disabled.  That is kind of a hack since it presumes this function would only be used for logging, and we still allocate the buffer when the log level is set to a low level.

Maybe another syscfg setting controlling whether this gets enabled?  If disabled, it could return "<ble hex dump disabled>" or something like that.  That might be a pain in the neck for the user, but it least it gives him full control.

andrzej-kaczmarek: I don't like this either and default sizes of these buffers are way too big - I just set them like this at the very beginning... so this has to be changed for sure.

An additional idea here: what if we add % formatter to printf implementation which would output bt_addr_t structure and hexdumps? This would require virtually zero extra memory and should not be hard to implement I guess.

simonratner: Big 👍 on that idea, i feel like it will save some code space since every project rebuilds its own. 

Glibc has a way to register custom printf handlers: https://www.gnu.org/software/libc/manual/html_node/Customizing-Printf.html#Customizing-Printf. I don't think we need to go that far, better a lightweight implementation that just does a couple of common things - `ble_addr_t`, `ble_uuid_t`, space-separated hex with a max width modifier - but the precedent is certainly there.

mkiiskila: My 2 cents: I would not like to see custom printf formatter. I still want to use this with other libc implementations besides baselibc. For some of those, it would probably be difficult to add custom printf formatting (thinking native MacOS/Windows here).

andrzej-kaczmarek: We could extract tinyprintf to separate package and let nimble (or anyone) use this for outputting logs - create package libc/tinyprintf and make it required by libc/baselibc and net/nimble/host. This means we can have our mynewt_printf with whatever printf specifiers may come in handy later and you can use whatever libc you want at the same time.

andrzej-kaczmarek:
will push upload modified logging stuff in another PR

PR #365, opened 2017-06-30
nimble/controller: Remove not working restrictions

rymanluk:
When BLE_EXT_ADV is set to 0 then we got:

Error: Syscfg restriction violations detected:
    BLE_LL_CFG_FEAT_LL_EXT_ADV=MYNEWT_VAL_BLE_EXT_ADV requires
BLE_EXT_ADV be set, but BLE_EXT_ADV=0

Setting history (newest -> oldest):
    BLE_EXT_ADV: [net/nimble:0]
    BLE_LL_CFG_FEAT_LL_EXT_ADV:
[net/nimble/controller:MYNEWT_VAL_BLE_EXT_ADV]

This is because BLE_LL_CFG_FEAT_LL_EXT_ADV value is other flag which is
not resolved yet.

This patch remove restrictions. It should still work fine until user use
BLE_EXT_ADV to configure extended advertising - will put it in
documentation

### DIFF
--- a/net/nimble/controller/syscfg.yml
+++ b/net/nimble/controller/syscfg.yml
@@ -241,9 +241,6 @@ syscfg.defs:
             Advertising Feature. That means extended scanner, advertiser
             and connect.
         value: MYNEWT_VAL_BLE_EXT_ADV
-        restrictions:
-            - 'BLE_EXT_ADV if 1'
-            - '!BLE_EXT_ADV if 0'
 
     BLE_PUBLIC_DEV_ADDR:
         description: >


ccollins476ad:
I know this has already been merged, but I just wanted to mention that you can still keep the restrictions.  You would do this as follows:

```
syscfg.defs:
    BLE_LL_CFG_FEAT_LL_EXT_ADV:
        description: >
            This option is used to enable/disable support for Extended
            Advertising Feature. That means extended scanner, advertiser
            and connect.
        value: 0
        restrictions:
            - 'BLE_EXT_ADV if 1'
            - '!BLE_EXT_ADV if 0'

syscfg.vals.BLE_EXT_ADV:
    BLE_LL_CFG_FEAT_LL_EXT_ADV: 1
```

I think using "MYNEWT_VAL..." as a setting's default value is not desirable.  It takes advantage of knowledge of how newt generates setting names, and if you get the name wrong (or if the setting never gets defined), you'll get strange compiler errors that may be difficult to debug.

Anyway, I don't think it is necessary to change this now; this is more of an "fyi" for future changes.

I think the real fix is to improve the syscfg system to allow full expressions.  I started doing some work on this, but we won't get it into 1.1.  So, we are stuck with several lame solutions for now :).

PR #366, opened 2017-06-30
bsp: Fix nRF5x BSP configurations for BLE_LP_CLOCK

andrzej-kaczmarek:
This is needed so other nRF5x BSPs can work with nimble controller using 32768Hz clock.


jacobrosenthal:
The microbit (and maybe others here?) actually dont have 32k 
https://github.com/bbcmicrobit/hardware/blob/master/SCH_BBC-Microbit_V1.3B.pdf

simonratner:
The bmd200 too, I believe.

wes3:
Some comments:

* Looking at the bmd200 syscfg.yml and microbit I see this in the bmd200:
    XTAL_32768:
        description: 'External 32k oscillator available.'
        value: 0
        restrictions:
            - "!XTAL_32768_SYNTH"

    XTAL_32768_SYNTH:
        description: 'Synthesize 32k clock.'
        value: 1
        restrictions:
            - "!XTAL_32768"

In the microbit syscfg.yml I do not see XTAL_32768 set to 0 (or that syscfg variable at all). Hopefully I am looking at the latest.

* The code can still use the low power clock without an external crystal; it just needs to use the synthesized crystal. I have not looked at this in a bit but I think when the synthesized crystal is used the accuracy of the clock (SCA) is affected and I am not sure this is accounted for in the code. I think this needs to be reviewed before merging this pull request.

mkiiskila:
BBC microbit does not have 32k crystal, it needs to be synthesized. So far it seems that setting XTAL_32768_SYNTH to 1 has been sufficient, doesn't look like the BSP sets that other syscfg variable.

andrzej-kaczmarek:
So as I understand both BBC microbit and bmd200 do not have 32k crystal but are configured to synthesize it and this should work, correct? I don't have hardware to check this (will buy one anyway).

Regarding SCA, I think we need this patch anyway since without proper configuration nimble controller won't work on any BSP except for "standard" nRF5x so perhaps we should not block this PR. I created Jira ticket to track the remaining issue: https://issues.apache.org/jira/browse/MYNEWT-798 so will take care of this once I have HW to test.

wes3:
I agree; we should not block this PR.

wes3:
I forgot to comment on the microbit: yes, they should work with the synthesized crystal (at least in theory).

PR #367, opened 2017-06-30
nimble: Couple of fixes and improvements to bluetooth 5

rymanluk:
Controller and host part


wes3:
I reviewed the controller changes. They seem ok.

PR #368, opened 2017-06-30
nimble/controller: Allow to configure usage of hardware while list 

rymanluk:


wes3:
This is a good code change. Looks good.

PR #370, opened 2017-07-01
BLE Host - Don't persist security material if either key distribution field is 0

ccollins476ad:
This issue is documented in: https://issues.apache.org/jira/browse/MYNEWT-749

This commit implements the following behavior:

If either peer indicates a key distribution flags value of 0, then don't persist security material when pairing completes.

It isn't clear from the spec what should happen when both peers indicate bonding support, but one indicates a 0 key distribution field.  I think proceding with pairing and just not persisting makes the most sense. This is also what Lightblue (or CoreBluetooth, I guess) seems to do.


ccollins476ad:
Thanks, Szymon.  That is a good catch.  I have updated the original commit with a fix.

PR #372, opened 2017-07-03
nimble: add monitor protocol/interface

andrzej-kaczmarek:
See here for details: http://mail-archives.apache.org/mod_mbox/incubator-mynewt-dev/201706.mbox/%3CCADAkoF4LQOdqH%3DLh%3D5GgBPC9nEaXV95TV_uQPXXWuhrpCXfbtQ%40mail.gmail.com%3E

This is core monitor protocol functionality extracted from https://github.com/apache/mynewt-core/pull/362 so logging stuff does not block it.

### DIFF
--- a/net/nimble/host/src/ble_monitor.c
+++ b/net/nimble/host/src/ble_monitor.c
@@ -307,7 +307,7 @@ ble_monitor_log(int level, const char *fmt, ...)
     int len;
 
     va_start(va, fmt);
-    len = vsprintf(NULL, fmt, va);
+    len = vsnprintf(NULL, 0, fmt, va);
     va_end(va);
 
     switch (level) {
--- a/net/nimble/host/src/ble_monitor.c
+++ b/net/nimble/host/src/ble_monitor.c
@@ -94,6 +94,9 @@ monitor_uart_queue_char(uint8_t ch)
             tx_ringbuf_tail) {
         uart_start_tx(uart);
         OS_EXIT_CRITICAL(sr);
+        if (os_started()) {
+            os_time_delay(1);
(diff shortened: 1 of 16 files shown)


rymanluk:
Changes look good

net/nimble/host/src/ble_monitor.c
```diff
@@ -243,4 +257,58 @@ ble_monitor_new_index(uint8_t bus, uint8_t *addr, const char *name)
     return 0;
 }
 
+int
+ble_monitor_log(int level, const char *fmt, ...)
+{
+    struct ble_monitor_hdr hdr;
+    struct ble_monitor_user_logging ulog;
+    const char id[] = "nimble";
```
sjanc: could be static const

andrzej-kaczmarek:
@ccollins476ad 
regarding vsnprintf(): C99 guarantees that with size=0 nothing is written to buffer and we can pass buffer as NULL so I used it this way. This however required fixing of tinyprintf's implementation of vsnprintf() which I also included here.

ccollins476ad:
Thanks, Andrzej.  It looks good.  Also, thanks for clarifying the snprintf() semantics.  I should have checked the spec.  Just for the record, from n1256:

```
7.19.6.5 The snprintf function

Synopsis

    #include <stdio.h>
    int snprintf(char * restrict s, size_t n,
         const char * restrict format, ...);

Description

The snprintf function is equivalent to fprintf, except that the output is written into
an array (specified by argument s) rather than to a stream. If n is zero, nothing is written,
and s may be a null pointer.
```

PR #375, opened 2017-07-04
Nimble: improvements on controller/peer features handling

andrzej-kaczmarek:
This introduces multiple improvements in the way we handle local and remote features in both host and controller.


andrzej-kaczmarek:
@wes3 there's no need for callout - I couldn't find any better way to send the response at that time, but what you propose seems like a much better idea. I'll change it this way.

PR #378, opened 2017-07-05
nimble/controller: Bluetooth 5 improvements

rymanluk:


rymanluk:
Ok now it should be good. Just need to do some regression test (unfortunately cannot test chain adv itself)

rymanluk:
For extended advertising it seems to be sizeof(struct ble_ll_ext_adv) + 254. For legacy we could create struct for advertising event and also calculate it. Is that what you  have in mind?

Anyway, we still have big blocks in the mempool. Probably using os_mbufs would help...need to think about that.

wes3:
Chaining buffers would indeed help I would think (as most events are probably small) but would create a bit more code. There could also be a large and small pool or other options; we should just file a new ticket with an improvement/modification to the event buffer size if we feel like we should.

As far as calculating size, what I meant was to go through the code to determine the maximum event buffer required by the controller. Not sure if it is advertising reports or a different command (get features or some such; Chris had figured it out a ways back) that requires a contiguous buffer. Once we know what it is we can at least put the calculation in the code when we build with/without extended advertising.s

rymanluk:
I think that advertising reports (legacy and extended) are the biggest possible events controller can send.
Here is Jira for that. 

https://issues.apache.org/jira/browse/MYNEWT-799

PR #379, opened 2017-07-05
Some improvements for Extended Advertiser

sjanc:


wes3:
As you mentioned in your comments the code does not support mbuf chaining when creating advertising PDU's or scan response PDU's. This means that the maximum amount of advertising data you can send is based on the mbuf size. Since the user can change the mbuf size and amount of EXT_ADV_DATA I think there should be some check added to the code. Maybe a #error for now until we decide if we want to address this issue in a different way (the #error would check if the mbuf size is too small to create a maximum size advertising pdu).

sjanc:
Hi Will,

Chaining here was about PDUs chaining over the radio :) That's why I limited data to 238 bytes. But yes, I plan to introduce support for mbuf chaining in advertising code too:)

For now I'll just add #error as you suggested

wes3:
Oh, you were talking about radio chaining :-) Now I get the fragmentation comment. Well, I guess my comment still makes sense. And regarding mbuf chaining for advertising: I think we should discuss how to manage buffers for advertising. Should the advertising state machine have a separate mbuf pool or should that memory be allocated within the advertising state machine? That is the subject of another ticket.

sjanc:
as discussed offline,  will merge it as is and we can have #error thing (sizeof is not allowed in preprocessor macros) figure out in separate PR

PR #387, opened 2017-07-07
nimble/controller: Improve validation on extended connect 

rymanluk:
This fixed build error on nrf51 when optmized is on

### DIFF
--- a/net/nimble/controller/src/ble_ll_conn_hci.c
+++ b/net/nimble/controller/src/ble_ll_conn_hci.c
@@ -42,6 +42,19 @@
 static uint32_t g_ble_ll_last_num_comp_pkt_evt;
 extern uint8_t *g_ble_ll_conn_comp_ev;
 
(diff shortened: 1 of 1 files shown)


net/nimble/controller/src/ble_ll_conn_hci.c
```diff
@@ -724,7 +724,12 @@ ble_ll_ext_conn_create(uint8_t *cmdbuf)
         return BLE_ERR_CONN_LIMIT;
     }
 
-    ble_ll_conn_init_hcc_params(hcc, valid_param_idx);
+    /* Check if we should init some initial parameters which host did not
+     * specified
+     */
+    if (valid_param_idx >= 0) {
```
andrzej-kaczmarek: This should be non-zero here anyway. I think the reason why GCC does not catch this properly (btw, for me it builds fine) is because you do initial phy masks validation as if all phys are enabled - perhaps if the initial validation includes phys that are enabled only, it would compile for you.

PR #388, opened 2017-07-07
nimble/controller: Fix for resolving RPA from extended advertising packets

rymanluk:


sjanc:
Hi Will,

regarding 1.  The thing is that rxinfo.flags are now uint16_t and without it one has to use uint16_t instead of uint8_t (and only for flags on 9th bit and above) for storing value.  I'll check if this increase code size though.   This change came from compilation error but if code size is increased I think we can leave with using uint16_t for result :)

andrzej-kaczmarek:
@wes3 
I created new PR with cleanup for ADDRPTR (can't modify this one) - https://github.com/apache/mynewt-core/pull/401

PR #394, opened 2017-07-08
MYNEWT-789 Add RuuviTag polling LIS2DH12 3-axis accel SPI driver

vrahane:
- Adding basic SPI driver for LIS2DH12 to work with the Ruuvitag
- Changing the frequency for SPI to 4Mhz.
- Change Ruuvi sample rate for LIS2DH12

### DIFF
--- a/hw/drivers/sensors/bme280/src/bme280.c
+++ b/hw/drivers/sensors/bme280/src/bme280.c
@@ -53,7 +53,7 @@ double NEG_INF = -1.0/0.0;
 static struct hal_spi_settings spi_bme280_settings = {
     .data_order = HAL_SPI_MSB_FIRST,
     .data_mode  = HAL_SPI_MODE0,
-    .baudrate   = 500,
+    .baudrate   = 4000,
     .word_size  = HAL_SPI_WORD_SIZE_8BIT,
 };
 
--- a/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
+++ b/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
@@ -22,9 +22,9 @@
 #include <errno.h>
 #include <string.h>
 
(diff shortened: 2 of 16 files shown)


hw/drivers/sensors/bme280/src/bme280.c
```diff
@@ -53,7 +53,7 @@ double NEG_INF = -1.0/0.0;
 static struct hal_spi_settings spi_bme280_settings = {
     .data_order = HAL_SPI_MSB_FIRST,
     .data_mode  = HAL_SPI_MODE0,
-    .baudrate   = 500,
+    .baudrate   = 4000,
```
mkiiskila: Did you mean to bump the SPI datarate for this other sensor too?

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -0,0 +1,866 @@
...
+ */
+
+#include <assert.h>
+#include <stdio.h>
+#include <errno.h>
+#include <string.h>
+
+#include "defs/error.h"
+#include "os/os.h"
+#include "sysinit/sysinit.h"
```
mkiiskila: Maybe include sysinit before os.h? I can't remember if there were any MYNEWT_VAL() checks in those header files, but it is quite possible.

hw/drivers/sensors/bme280/src/bme280.c
vrahane: Yes, it was intentional. I tested it as well.

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
vrahane: done

PR #396, opened 2017-07-10
Support for HiFive1 board

kasjer:
Support for HiFive1 risc-v board from SiFive.

- openocd modifications for non ARM base chips
- load and debug with newt using openocd and gdb
- system clock 1-320 MHz with external/internal oscillator and pll
- boot-loader and application support
- interrupts (internal and external)
 - sys tick on 32 K oscillator
 - uart/gpio/timers
 - context switch on internal software interrupt

- hal support for
  - uart
  - flash
  - watchdog
  - timers
  - gpio

current limitations:
- image size 420 KB
- hal spi not done
- uart at 115200 requires clock 16 MHz and up


hw/bsp/hifive1/hifive1.ld
```diff
@@ -0,0 +1,161 @@
+OUTPUT_ARCH( "riscv" )
```
mkiiskila: We'll need Apache license even in linker script files.

hw/bsp/hifive1/riscv_openocd.cfg
```diff
@@ -0,0 +1,34 @@
+adapter_khz     10000
```
mkiiskila: The same comment regarding license blurb.

Thanks for doing this as ftdi openocd device. It's great to have this as an example in one of the BSPs.
Do you mind including the wiring info in text form either in this file, or README.md within this directory?

hw/scripts/openocd.sh
```diff
@@ -17,6 +17,9 @@
 #
 . $CORE_PATH/hw/scripts/common.sh
 
+CFG_RESET="reset halt"
+GDB=arm-none-eabi-gdb
```
mkiiskila: Ah, I didn't think of this. gdb has to be platform specific, if we're to use this script to debug other cpu architectures.

hw/mcu/sifive/src/ext/freedom-e-sdk_3235929/bsp/env/encoding.h
```diff
@@ -0,0 +1,1313 @@
+// See LICENSE for license details.
```
mkiiskila: Do you have a pointer to the LICENSE file which is referenced here? I think we'll want that as part of this SDK.

hw/scripts/openocd.sh
mkiiskila: Maybe we should add it to compiler.yml file along with other tool related links, and then pass as environment variable to this script. I'll file a Jira ticket about that.

hw/bsp/hifive1/riscv_openocd.cfg
kasjer: I wish it was my doing. Script is mostly taken from freedom-e-sdk.
I modified it to use same script for upload and debugging, part of sdk script was moved to bash script and later passed to already existing openocd.sh.

PR #408, opened 2017-07-12
MYNEWT-811 Console compatibility broken

vrahane:
fixing the following issues:

- Non-printable characters except for the newtmgr control characters being
  input to the console were ignored.
- console_read() was replacing single characters by '\0'

### DIFF
--- a/sys/console/full/src/console.c
+++ b/sys/console/full/src/console.c
@@ -117,11 +117,17 @@ console_read(char *str, int cnt, int *newline)
     }
     cmd = ev->ev_arg;
     len = strlen(cmd->line);
+
     if ((cnt - 1) < len) {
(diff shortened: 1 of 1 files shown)


sys/console/full/src/console.c
```diff
@@ -467,6 +473,8 @@ console_handle_char(uint8_t byte)
         case ESC:
             esc_state |= ESC_ESC;
             break;
+        default:
+            insert_char(&input->line[cur], byte, end);
```
sjanc: if this fall through is intentional please add a comment (GCC7 will emit warning without it)

vrahane: @sjanc Can you please approve the PR now that your concern is taken care of ?

PR #410, opened 2017-07-13
bsp: Update configurations for BLE_LP_CLOCK

andrzej-kaczmarek:


jacobrosenthal:
Successfully ran microbit with bleprph example

PR #413, opened 2017-07-14
native_sockets - Add syscfg settings.

ccollins476ad:
The motivation for this change was to allow the polling interval to be changed.  The default value, 200ms, was making blehostd rather sluggish during image upgrade.

### DIFF
--- a/net/ip/native_sockets/syscfg.yml
+++ b/net/ip/native_sockets/syscfg.yml
@@ -1,0 +1,36 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


net/ip/native_sockets/syscfg.yml
```diff
@@ -0,0 +1,36 @@
...
+        description: 'The number of allocated sockets.'
+        value: 8
+    NATIVE_SOCKETS_MAX_UDP:
+        description: 'The maximum UDP datagram size (send and receive).'
+        value: 2048
+    NATIVE_SOCKETS_POLL_ITVL:
+        description: >
+            The frequency at which to poll for received data.  Units
+            are OS ticks.
+        value: 1
```
mkiiskila: Would it be ok to keep the default as it was before, and override this in the blehostd? Although I could be convinced to set it to 1; just means a lot of polling when nothing is going on :)

ccollins476ad: Yikes, I didn't mean to make the default 1!  I have changed it back to OS_TICKS_PER_SEC / 5.

PR #415, opened 2017-07-14
MYNEWT-659 Rerun PTS on Bluetooth5 branch (regression testing)

michal-narajowski:
Rerun GAP, GATT, L2CAP, SM test cases. Most tests are green.
We should look into the following test cases because I couldn't get them to pass with current instructions:
TC_BROB_BCST_BV_03_C
TC_SEC_AUT_BV_14_C
GATT/SR/GAC/BV-01-C
GATT/SR/GAW/BV-08-C


michal-narajowski:
Updated dates and missing tests. Included fixes for one test.

PR #418, opened 2017-07-15
MYNEWT-808 Adding support for I2C LIS2DH12 driver

vrahane:
- Also, cleaning up a bit

### DIFF
--- a/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
+++ b/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
@@ -26,6 +26,7 @@
 #include "defs/error.h"
 #include "os/os.h"
 #include "hal/hal_spi.h"
+#include "hal/hal_i2c.h"
 #include "sensor/sensor.h"
 #include "sensor/accel.h"
 #include "lis2dh12/lis2dh12.h"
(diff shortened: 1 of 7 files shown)


jacobrosenthal:
I didnt extensively test, but I was able to read accell values from shell from lis2dh, and values seemed to invert when I inverted the thingy.

Looks good to me.

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -860,6 +1089,8 @@ lis2dh12_config(struct lis2dh12 *lis2dh12, struct lis2dh12_cfg *cfg)
 
     lis2dh12->cfg.lc_s_mask = cfg->lc_s_mask;
 
+    os_time_delay((OS_TICKS_PER_SEC * 110)/1000);
```
mkiiskila: If this delay line looks surprising. If it is meant to be here, should comment why.

vrahane: Sorry, I was trying different configs in a desperate attempt to get it working. This delay is needed for any initial interrupts that might occur but for the polling config it really doesn't matter. Removing it.

PR #419, opened 2017-07-15
Adding missing LED pin macros to nRF52dk bsp.h

mlaz:
LED's 1, 3 and 4 are not defined on hw/bsp/nrf52dk/include/bsp/bsp.h
This commit adds the missing macro definitions.


mlaz:
Ok, I can add some more pins, in fact all of them. Sorry for PRing without it.

PR #420, opened 2017-07-17
sys/console: fix console compatibility

michal-narajowski:
The previous fix (#408, MYNEWT-811) introduced a regression in handling CR/LF characters.
When CR/LF was used, the LF character was added to the buffer and handled by the shell resulting in "Unrecognized command" error.
Additionally, this patch also fixes these issues in console/minimal.

### DIFF
--- a/sys/console/full/src/console.c
+++ b/sys/console/full/src/console.c
@@ -385,6 +385,7 @@ console_handle_char(uint8_t byte)
 
     static struct os_event *ev;
     static struct console_input *input;
+    static char prev_endl = '\0';
 
     if (!avail_queue || !lines_queue) {
         return 0;
@@ -473,10 +474,15 @@ console_handle_char(uint8_t byte)
         case ESC:
             esc_state |= ESC_ESC;
             break;
-        default:
-            insert_char(&input->line[cur], byte, end);
-            /* Falls through. */
         case '\r':
+            /* Falls through. */
+        case '\n':
+            if (byte == '\n' && prev_endl == '\r') {
+                /* handle double end line chars */
+                prev_endl = byte;
+                break;
+            }
+            prev_endl = byte;
             input->line[cur + end] = '\0';
             console_out('\r');
             console_out('\n');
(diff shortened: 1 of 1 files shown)


sys/console/minimal/src/console.c
```diff
@@ -86,11 +86,17 @@ console_read(char *str, int cnt, int *newline)
...
     len = strlen(cmd->line);
+
     if ((cnt - 1) < len) {
         len = cnt - 1;
+        if (len > 0) {
+            memcpy(str, cmd->line, len);
+            str[len] = '\0';
+        } else {
+            str[len] = cmd->line[0];
+        }
```
mkiiskila: This is broken. The common case, where cnt > len, would copy no bytes to str.

Also, I don't understand line 96 here. Why are you setting last character in str to be
first character of input?

vrahane: @michal-narajowski, I agree with @mkiiskila.  Can you remove changes to this file for the time being as it is going to break boot-serial. I will fix it in console/full and console/minimal and make sure it works with boot-serial as well. I will open a new PR with a fix to both.

sys/console/minimal/src/console.c
```diff
@@ -222,6 +229,13 @@ console_handle_char(uint8_t byte)
             nlip_state |= NLIP_DATA_START1;
             break;
         case '\r':
```
vrahane: /* Falls through. */ comment is needed here as GCC7 gives an error without an explicit commented fallthrough.

sys/console/full/src/console.c
```diff
@@ -473,10 +474,14 @@ console_handle_char(uint8_t byte)
         case ESC:
             esc_state |= ESC_ESC;
             break;
-        default:
-            insert_char(&input->line[cur], byte, end);
-            /* Falls through. */
         case '\r':
```
vrahane: /* Falls through. */ comment is needed here as GCC7 gives an error without an explicit commented fallthrough.

michal-narajowski:
Updated.

PR #425, opened 2017-07-17
MYNEWT-811 Console compatibility broken

vrahane:
#408 and #420 add a code snippet in console_read() which breaks compatibility for boot serial. It is fixed by this PR. As mentioned in #420, code changes to console/minimal should be reverted from PR #408 before merging this.


vrahane:
I agree with the above comments. However, I might not be the right person to make a lot of changes here as I am not completely aware of the design decisions. We should perhaps consider opening a clean up ticket.

PR #428, opened 2017-07-18
ext advertising support for host

sjanc:


net/nimble/host/src/ble_gap.c
```diff
@@ -1679,10 +1685,24 @@ ble_gap_wl_set(const ble_addr_t *addrs, uint8_t white_list_count)
 static int
 ble_gap_adv_enable_tx(int enable)
 {
+#if MYNEWT_VAL(BLE_EXT_ADV)
+    uint8_t buf[BLE_HCI_CMD_HDR_LEN + 6];
+    static const struct hci_ext_adv_set set = {0, 0 , 0};
```
andrzej-kaczmarek: extra space before comma :)

net/nimble/host/src/ble_gap.c
```diff
@@ -1803,6 +1855,64 @@ ble_gap_adv_params_tx(uint8_t own_addr_type, const ble_addr_t *peer_addr,
...
+        hci_adv_params.chan_map = adv_params->channel_map;
+    }
+
+    /* Zero is the default value for filter policy and high duty cycle */
+    hci_adv_params.filter_policy = adv_params->filter_policy;
+    hci_adv_params.tx_power = ext_adv_tx_pwr;
+
+    /* if phy was not set this means legacy advertising PDUs */
+    if (ext_adv_pri_phy == 0) {
+        hci_adv_params.properties = ble_gap_adv_type(adv_params);
```
andrzej-kaczmarek: I don't think this does what it should: ble_gap_adv_type() returns ordinal value 0..4 while properties for legacy pdus shall be one of bitmasks defined in spec

net/nimble/host/src/ble_gap.c
```diff
@@ -1772,6 +1792,38 @@ ble_gap_adv_type(const struct ble_gap_adv_params *adv_params)
...
+        break;
+    case BLE_GAP_CONN_MODE_UND:
+        prop |= BLE_HCI_LE_SET_EXT_ADV_PROP_CONNECTABLE;
+        break;
+    case BLE_GAP_CONN_MODE_DIR:
+        prop |= BLE_HCI_LE_SET_EXT_ADV_PROP_CONNECTABLE;
+        prop |= BLE_HCI_LE_SET_EXT_ADV_PROP_DIRECTED;
+
+        if (adv_params->high_duty_cycle) {
+            prop = BLE_HCI_LE_SET_EXT_ADV_PROP_HD_DIRECTED;
```
andrzej-kaczmarek: this bit should not be used if extended advertising pdu types are used (and as I understand this function is used to convert properties only in this case)

net/nimble/host/src/ble_gap.c
```diff
@@ -2182,6 +2322,47 @@ ble_gap_adv_active(void)
...
+{
+    ble_hs_lock();
+
+    if (ble_gap_adv_active()) {
+        ble_hs_unlock();
+        return BLE_HS_EBUSY;
+    }
+
+    if (primary_phy > BLE_HCI_LE_PHY_CODED ||
+        secondary_phy > BLE_HCI_LE_PHY_CODED) {
```
andrzej-kaczmarek: so perhaps we could also check if both are non-zero and primary is not 2M?
I would skip this in case this function calls HCI command which returns an error immediately, but in this case it just stores both values so can be a bit hard to guess that error in other call may be due to invalid values stored here.

PR #430, opened 2017-07-18
Enables newt resigned images

utzig:
This removes use of key-id when looking for a sign key and tries to
validate images using every available key in the bootloader, adding
compatibility with "newt resign-image" command which accepts whatever
key-id was supplied by the user.


mkiiskila:
Looks good to me.

PR #432, opened 2017-07-19
Adding  LED, button and pin values to bsp.h

mlaz:
hw/bsp/nrf52dk/include/bsp/bsp.h was lacking on some LED, button and pin value definitions.
This commit adds the missing macro definitions.

### DIFF
--- a/hw/bsp/nrf51dk/include/bsp/bsp.h
+++ b/hw/bsp/nrf51dk/include/bsp/bsp.h
@@ -37,8 +37,44 @@ extern uint8_t _ram_start;
 #define RAM_SIZE        0x8000
 
 /* LED pins */
-#define LED_BLINK_PIN   (21)
-#define LED_2           (22)
+#define LED_1           (17)
+#define LED_2           (18)
+#define LED_3           (19)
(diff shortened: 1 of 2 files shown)


hw/bsp/nrf51dk/include/bsp/bsp.h
```diff
@@ -37,8 +37,44 @@ extern uint8_t _ram_start;
 #define RAM_SIZE        0x8000
 
 /* LED pins */
-#define LED_BLINK_PIN   (21)
-#define LED_2           (22)
+#define LED_1           (17)
+#define LED_2           (18)
+#define LED_3           (19)
+#define LED_4           (20)
+#define LED_BLINK_PIN   (LED_1)
```
vrahane: I did not cross verify everything with the schematic but LED_BLINK_PIN was 21 earlier and now it's getting changed to 17. Is this intentional ?

mlaz: No, it was not I missed the directory, I was aiming at nrf52dk, not nrf51dk.

PR #439, opened 2017-07-19
os_mempool_is_sane() debug function.

ccollins476ad:
I used this function to debug a memory corruption issue recently.  I thought it might come in handy again at some point.

```
/**
 * Performs an integrity check of the specified mempool.  This function
 * attempts to detect memory corruption in the specified memory pool.
 *
 * @param mp                    The mempool to check.
 *
 * @return                      true if the memory pool passes the integrity
 *                                  check;
 *                              false if the memory pool is corrupt.
 */
bool
os_mempool_is_sane(const struct os_mempool *mp)
```

### DIFF
--- a/kernel/os/include/os/os_mempool.h
+++ b/kernel/os/include/os/os_mempool.h
@@ -88,8 +88,11 @@ typedef uint64_t os_membuf_t;
 os_error_t os_mempool_init(struct os_mempool *mp, int blocks, int block_size,
                            void *membuf, char *name);
 
+/* Performs an integrity check of the specified mempool. */
+bool os_mempool_is_sane(const struct os_mempool *mp);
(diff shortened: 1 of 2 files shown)


wes3:
I should also comment that the name of the function is well-named. Using the word "is", to me, means that this is a "boolean" kind of function. If we keep that paradigm that is fine too.

ccollins476ad:
I hear what you are saying.  I agree that since this function is logically boolean, it should return a bool.  I tend to shy away from bool because... well, I don't have a great answer for that one :).  I think it is time to change that; I have updated the function so that it returns bool.

sterlinghughes:
it would be great to see this tie into the sanity task/syscfg as well.

PR #440, opened 2017-07-19
bsp: fix nrf52 thingy pins

jacobrosenthal:
Please reference https://nordicsemiconductor.github.io/Nordic-Thingy52-FW/documentation/pca20020_8h_source.html

### DIFF
--- a/hw/bsp/nrf52-thingy/include/bsp/bsp.h
+++ b/hw/bsp/nrf52-thingy/include/bsp/bsp.h
@@ -40,11 +40,11 @@ extern uint8_t _ram_start;
 #define RAM_SIZE        0x10000
 
 /* LED pins */
-#define LED_BLINK_PIN   (17)
-#define LED_2           (18)
+#define LED_BLINK_PIN   (-1)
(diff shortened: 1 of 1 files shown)


vrahane:
From the schematics it looks like the pins to the LEDs go through an IO expander: SX1509BIULTRT. This IO expander needs a driver as it works over I2C. So, the definition of the pins mentioned in this PR is not correct as it is not directly connected to the LEDs.
@sterlinghughes, @sjanc, @wes3, @mkiiskila  what do you think ?

jacobrosenthal:
Ah hell, I guess thats what SX_ stands for.
Ill test and try to figure out if by any chance that pin is a default output that would 'just work'

This is a similar problem to microbit then, in that youve got no direct led to blink. For what its worth we use -1 there https://github.com/apache/mynewt-core/blob/master/hw/bsp/bbc_microbit/include/bsp/bsp.h#L39

Button is direct connect it looks like though.

jacobrosenthal:
Removed led/blink support.

PR #442, opened 2017-07-19
Enabeling PWM1 and PWM2 on nrf52dk bsp

mlaz:
Only PWM0 is the only PWM device enabled on the bsp, is there any reason for this? 
This enables PWM1 and PWM2 on nrf52dk bsp.


hw/bsp/nrf52dk/pkg.yml
```diff
@@ -39,8 +39,8 @@ pkg.cflags:
     - '-DPDM_ENABLED=0'
     - '-DPERIPHERAL_RESOURCE_SHARING_ENABLED=1'
     - '-DPWM0_ENABLED=1'
-    - '-DPWM1_ENABLED=0'
-    - '-DPWM2_ENABLED=0'
+    - '-DPWM1_ENABLED=1'
+    - '-DPWM2_ENABLED=1'
```
vrahane: @mlaz The default bsp in mynewt does not need to enable all PWMs. One is enabled and we can keep it at that to be used as an example. Any specific reason why you want to enable all by default ?

mlaz:
This is being addressed on issue https://github.com/apache/mynewt-core/pull/448.

PR #448, opened 2017-07-20
PWM high level API

mlaz:
This defines the PWM high level API.


apps/pwm_test/src/main.c
```diff
@@ -0,0 +1,71 @@
...
+int arg = 0;
+struct pwm_dev *pwm;
+static int value = 10000;
+
+int
+main(int argc, char **argv)
+{
+    sysinit();
+
+    struct nrf52_pwm_chan_cfg chan_conf = {
```
mkiiskila: Coding style thing; declare variables in the beginning of function, before other statements. I.e. before call to sysinit().

apps/pwm_test/pkg.yml
```diff
@@ -0,0 +1,35 @@
...
+
+pkg.deps:
+    - kernel/os
+   # - hw/hal
+   # - hw/cmsis-core
+    - sys/console/full
+   # - hw/bsp/nrf52dk
+    - hw/drivers/pwm
+   # - "@mynewt_nordic/hw/mcu/nordic_sdk"
+    - "@mynewt_nordic/hw/drivers/pwm/pwm_nrf52"
```
mkiiskila: I would remove the dependency on pwm_nrf52, and move it to hw/bsp/<your_bsp>/pkg.yml instead.
Also I would do the os_dev_create() within that BSPs hal_bsp_init(). Make the creation of the device to be a syscfg within the BSP. That would allow you to use this same app with BSPs which use some other MCU (once PWM drivers for those architectures appear :).

apps/pwm_test/src/main.c
```diff
@@ -0,0 +1,71 @@
...
+        .pin = LED_1,
+        .inverted = true
+    };
+
+    os_dev_create(&dev,
```
mkiiskila: I would move this device creation to happen under hw/bsp/<your_bsp>, and use name "pwm0" for the first one.

apps/pwm_test/src/main.c
```diff
@@ -0,0 +1,71 @@
...
+    pwm = (struct pwm_dev *) os_dev_open("pwm", 0, NULL);
+
+    pwm_chan_config(pwm, 0, &chan_conf);
+    pwm_enable_duty_cycle(pwm, 0, 10000);
+
+    chan_conf.pin = LED_2;
+    pwm_chan_config(pwm, 1, &chan_conf);
+    pwm_enable_duty_cycle(pwm, 1, value/10);
+
+    //changing frequency while playing
```
mkiiskila: Also c++ comments are not allowed according to our coding style. See CODING_STANDARDS.md

hw/bsp/nrf52dk/pkg.yml
```diff
@@ -39,8 +39,8 @@ pkg.cflags:
     - '-DPDM_ENABLED=0'
     - '-DPERIPHERAL_RESOURCE_SHARING_ENABLED=1'
     - '-DPWM0_ENABLED=1'
-    - '-DPWM1_ENABLED=0'
-    - '-DPWM2_ENABLED=0'
+    - '-DPWM1_ENABLED=1'
+    - '-DPWM2_ENABLED=1'
```
mkiiskila: I wonder if we could make these SDK flags conditional, such that PWMx_ENABLED would be set only if the matching os_dev for that PWM is enabled with that BSP.
I.e. in hw/bsp/nrf52dk/pkg.yml, you would have:
pkg.cflags.PWM0_ENABLED: -DPWM0_ENABLED
and in this file you'd have pkg.cflags: -DPWM0_ENABLED.

It *should* work, as newt orders the cflags passed to command line, although it might be a bit too hacky. I.e. depends on newt doing the cflag ordering. Others might have comments regarding this approach?

hw/drivers/pwm/src/pwm.c
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
```
mkiiskila: I would recommend including <syscfg/syscfg.h> here before other #include's. In case there'll be syscfg settings later which affect this file.

apps/pwm_test/src/main.c
mlaz: Changed as requested.

apps/pwm_test/src/main.c
mlaz: Changed as requested.

hw/drivers/pwm/src/pwm.c
mlaz: Changed as requested.

apps/pwm_test/pkg.yml
mlaz: Changed as requested.

apps/pwm_test/src/main.c
mlaz: Changed as requested.

hw/bsp/nrf52dk/pkg.yml
mlaz: Partially changed as requested. Only changed PWM1_ENABLED and PWM2_ENABLED values to 0. Waiting for further instructions/requests on this.

mkiiskila:
This looks pretty good. Is there more work that you want to get in before we merge this?

mlaz:
I believe, according to the feedback I had both on the developers mailing list and slack, for now there is nothing to add.

PR #449, opened 2017-07-20
[MYNEWT-285] nimble/phy: Fix CRC checking

andrzej-kaczmarek:
The CRCSTATUS register is not cleared automatically by hardware so it
may not reflect current state of CRC check in some cases. The better way
to check CRC is to clear EVENT_CRCOK when starting to receive a packet
and then check its state when done.


andrzej-kaczmarek:
This was not yet really tested (as described in Jira) but I'm using this for some time now and did not see any connection problems...

andrzej-kaczmarek:
I forgot about nRF51, will upload similar change in a moment...

andrzej-kaczmarek:
Well, as it turns out nRF51 does not have this event so there's nothing to change there.

wes3:
Hmm.... maybe this is why I coded it this way in the first place :-) All good then.

PR #451, opened 2017-07-20
MYNEWT-816 SensorAPI: Add API for sensor poller & fix the poller callout

vrahane:
- Adding API for setting sensor poller in ms
- Also fixing the sensor manager poller which was trying to reset the
  callout to absolute time instead of relative time
- Changing the syscfg for SENSOR_MGR_WAKEUP_RATE to 2000 ms

### DIFF
--- a/hw/sensor/include/sensor/sensor.h
+++ b/hw/sensor/include/sensor/sensor.h
@@ -414,7 +414,7 @@ sensor_get_config(struct sensor *sensor, sensor_type_t type,
  * be triggered instead.
  */
 #define SENSOR_MGR_WAKEUP_TICKS (MYNEWT_VAL(SENSOR_MGR_WAKEUP_RATE) * \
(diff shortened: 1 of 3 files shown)


hw/sensor/include/sensor/sensor.h
```diff
@@ -482,6 +482,15 @@ struct sensor *sensor_mgr_find_next_bydevname(char *, struct sensor *);
...
 int sensor_mgr_match_bytype(struct sensor *sensor, void *arg);
 
+/**
+ * Set the sensor poll rate
+ *
+ * @param The devname
+ * @param The poll rate in milli seconds
+ */
+int
+sensor_set_poll_rate_ms(char *devname, uint32_t poll_rate);
```
sjanc: other functions take struct sensor instead of name,  is there a reason this one is special?

vrahane: changed, done.

PR #452, opened 2017-07-21
nimble/controller: Fix legacy scanning

andrzej-kaczmarek:
In legacy scanning next_phy is unconfigured so we should not try to
configure channels for next_phy as it will corrupt some random
memory.

Also add assert to make sure cur_phy is always configured, because it
should be.


sjanc:
test this please

PR #457, opened 2017-07-24
nimble: Fix BCC initialization on LE Coded PHY

andrzej-kaczmarek:


hw/drivers/nimble/nrf52/src/ble_phy.c
```diff
@@ -544,8 +545,19 @@ ble_phy_rx_xcvr_setup(void)
     g_ble_phy_data.phy_rx_started = 0;
     g_ble_phy_data.phy_state = BLE_PHY_STATE_RX;
 
+    /*
+     * On Coded PHY there are CI and TERM1 fields before PDU starts so we need
+     * to take this into account when setting up BCC.
+     */
+    if (g_ble_phy_data.phy_cur_phy_mode == BLE_PHY_MODE_CODED_125KBPS ||
```
rymanluk: I think this should be guarded with BLE_LL_BT5_PHY_SUPPORTED flag.

PR #459, opened 2017-07-24
nimble: Fix CCM configuration

andrzej-kaczmarek:


andrzej-kaczmarek:
It does not work on 2M and Coded. I honestly don't know why as I didn't look too much into it - my guess is that key-stream is initialized too late, but would need to check this. The way this is done now is the same as in controller code from Zephyr and works fine on all phys.

PR #461, opened 2017-07-25
nimble/controller: Improve scanner

rymanluk:


net/nimble/controller/src/ble_ll_conn.c
```diff
@@ -808,6 +808,22 @@ ble_ll_conn_reset_pending_aux_conn_rsp(void)
...
+    return false;
+#endif
+    struct ble_ll_conn_sm *connsm;
+
+    connsm = g_ble_ll_conn_create_sm;
+    if (!connsm) {
+        return false;
+    }
+
+    return !!CONN_F_AUX_CONN_REQ(connsm);
```
andrzej-kaczmarek: no need for !!

net/nimble/controller/src/ble_ll_conn_priv.h
```diff
@@ -139,7 +139,7 @@ uint32_t ble_ll_conn_get_ce_end_time(void);
 void ble_ll_conn_event_halt(void);
 uint8_t ble_ll_conn_calc_used_chans(uint8_t *chmap);
 void ble_ll_conn_reset_pending_aux_conn_rsp(void);
-
+bool ble_ll_conn_init_aux_conn_is_sent(void);
```
andrzej-kaczmarek: ble_ll_conn_init_is_aux_conn_req_sent would be better
or we already have 'pending_aux_conn_rsp' as above so let's keep it one way or another for both

net/nimble/controller/src/ble_ll_scan.c
```diff
@@ -151,13 +151,28 @@ static int
...
+    /* Check if there is no aux connect sent. If so drop the sched item */
+    if (lls == BLE_LL_STATE_INITIATING && ble_ll_conn_init_aux_conn_is_sent()) {
+        ble_ll_scan_aux_data_free(sch->cb_arg);
+        goto done;
+    }
+
+    if (lls != BLE_LL_STATE_STANDBY) {
+        ble_phy_disable();
+        ble_ll_wfr_disable();
+        ble_ll_state_set(BLE_LL_STATE_STANDBY);
```
andrzej-kaczmarek: this will break something for sure - what if we're in the middle of connection or advertising event?

net/nimble/controller/src/ble_ll_scan.c
```diff
@@ -344,6 +359,26 @@ ble_ll_scan_find_dup_adv(uint8_t *addr, uint8_t txadd)
 }
 
 /**
+ * Do scan machine clean up on PHY disabled
+ *
+ */
+void
+ble_ll_scan_clean_on_phy_disabled(void)
```
andrzej-kaczmarek: I'd just call it ble_ll_scan_clean_aux_data or sth and use wherever appropriate - I saw this cleanup in few other places in the code iirc...

net/nimble/controller/src/ble_ll_sched.c
```diff
@@ -1107,30 +1107,43 @@ ble_ll_sched_execute_item(struct ble_ll_sched_item *sch)
     int rc;
     uint8_t lls;
 
+    lls = ble_ll_state_get();
+    if (lls == BLE_LL_STATE_STANDBY) {
+        goto sched;
+    }
+
+    if (sch->sched_type == BLE_LL_SCHED_TYPE_AUX_SCAN) {
+        if (lls == BLE_LL_STATE_INITIATING || lls == BLE_LL_STATE_SCANNING) {
```
andrzej-kaczmarek: priority for init state over aux scan does make sense, but do we need to prioritize regular scan as well? I'd rather think of priority for connection over aux scan.

also some comment would be nice here

net/nimble/controller/src/ble_ll_conn_priv.h
rymanluk: ble_ll_conn_init_pending_aux_conn_rsp looks nice.

net/nimble/controller/src/ble_ll_scan.c
rymanluk: no because we are here only in 3 states:
* BLE_LL_STATE_STANDBY
* BLE_LL_STATE_INITIATING
* BLE_LL_STATE_SCANNING

I will put comment about that.

net/nimble/controller/src/ble_ll_sched.c
rymanluk: Scanner is running in INITIATING and SCANNING state. When aux scan is scheduled during those two states, we call scheduler callback in the scanner and let him decide what to do instead of disabling phy here.

WIll put comment on it

rymanluk:
test this please

PR #464, opened 2017-07-25
nimble/phy: Fix resolving address

andrzej-kaczmarek:
g_ble_phy_rx_buf is declared as uint32_t[] so [3] and [5] does not
really point where I expected...


sjanc:
yeah, we could use #pragma for this but using intermediate pointer is fine too I guess

PR #467, opened 2017-07-26
[MYNEWT-819] nimble/phy: Fix resolving address on nRF51

andrzej-kaczmarek:
This is the same as 2d34c653a but for nRF51 which I forgot to change
there...


sjanc:
test this please

PR #469, opened 2017-07-27
sys/shell: fix shell_evq_set not updating console queues

michal-narajowski:
PR #361 added shell_evq_set function to shell API. This PR fixes a bug introduced in that patch. When shell_evq_set was used it didn't reset console queues, which led to still using the main eventq. This patch fixes this issue. When shell_evq_set is used both shell and console use the specified eventq.

Moreover, this patch renames a local function to a more appropriate name.

### DIFF
--- a/sys/shell/src/shell.c
+++ b/sys/shell/src/shell.c
@@ -50,6 +50,7 @@ void
 shell_evq_set(struct os_eventq *evq)
 {
     os_eventq_designate(&shell_evq, evq, NULL);
+    console_set_queues(&avail_queue, shell_evq);
 }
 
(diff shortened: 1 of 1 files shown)


mkiiskila:
I would recommend naming symbols such that it's easy to identify which package they belong to.
I.e. avail_queue_init() -> shell_avail_queue_init().
This makes it possible to identify owner of a symbol e.g. while in a debugger, or when inspecting output of nm and such.

michal-narajowski:
Thanks, Marko. I renamed it to shell_avail_queue_init(). I will merge it now.

PR #470, opened 2017-07-27
btshell: port changes from bletiny from the last few months

michal-narajowski:


apps/btshell/src/cmd.c
```diff
@@ -596,7 +597,6 @@ static const struct shell_param connect_params[] = {
     {"timeout", "usage: =[UINT16], default: 0x0100"},
     {"min_conn_event_len", "usage: =[UINT16], default: 0x0010"},
     {"max_conn_event_len", "usage: =[UINT16], default: 0x0300"},
-    {"duration", "usage: =[1-INT32_MAX], default: 0"},
```
rymanluk: this change looks good, but it would be nice to have comment why it is needed as it does not change functionality.

michal-narajowski:
Thanks @rymanluk. I fixed these issues and will merge the PR.

PR #472, opened 2017-07-27
Sensor evq

vrahane:
- Make sensor_mgr_evq_set() static
- Make sensor_mgr_evq_set() arg a syscfg with default picking up default evq.

### DIFF
--- a/hw/sensor/syscfg.yml
+++ b/hw/sensor/syscfg.yml
@@ -43,4 +43,4 @@ syscfg.defs:
 
     SENSOR_MGR_EVQ:
         description: 'Specify the eventq to be used by sensor mgr'
-        value: 0
+        value:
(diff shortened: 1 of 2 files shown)


hw/sensor/syscfg.yml
```diff
@@ -40,3 +40,7 @@ syscfg.defs:
     SENSOR_OIC_OBS_RATE:
         description: 'Set OIC server observation rate in milli seconds'
         value: 1000
+
+    SENSOR_MGR_EVQ:
+        description: 'Specify the eventq to be used by sensor mgr'
+        value: 0
```
ccollins476ad: Should the default value of `SENSOR_MGR_EVQ` be blank, rather than `0`?  The code uses `#ifdef` to check this setting, and a value of 0 is considered defined.  You can specify an empty value like this:
```
    SENSOR_MGR_EVQ:
        description: 'Specify the eventq to be used by sensor mgr'
        value:
```

vrahane: Thank you for pointing it out Chris. Will fix it.

PR #475, opened 2017-08-01
merge 1_1_0_dev into master

sjanc:
1.1.0 release branch has some fixes we also want in master.


rymanluk:
I looks good. I went ahead and merged it since sjanc is on vacation.

PR #480, opened 2017-08-08
Allow GATT services to be registered after startup

ccollins476ad:
Prior to this change, supported GATT services could only be registered once.  This change allows the set of supported servces to be replaced "on the fly".

```
/**
 * Resets the GATT server to its initial state.  On success, this function
 * removes all supported services, characteristics, and descriptors.  This
 * function requires that:
 *     o No peers are connected, and
 *     o No GAP operations are active (advertise, discover, or connect).
 *
 * @return                      0 on success;
 *                              BLE_HS_EBUSY if the GATT server could not be
 *                                  reset due to existing connections or active
 *                                  GAP procedures.
 */
int
ble_gatts_reset(void)

/**
 * Makes all registered services available to peers.  This function gets called
 * automatically by the NimBLE host on startup; manual calls are only necessary
 * for replacing the set of supported services with a new one.  This function
 * requires that:
 *     o No peers are connected, and
 *     o No GAP operations are active (advertise, discover, or connect).
 *
 * @return                      0 on success;
 *                              A BLE host core return code on unexpected
 *                                  error.
 */
int
ble_gatts_start(void)
```

### DIFF
--- a/net/nimble/host/include/host/ble_gatt.h
+++ b/net/nimble/host/include/host/ble_gatt.h
@@ -235,7 +235,7 @@ struct ble_gatt_chr_def {
      * At registration time, this is filled in with the characteristic's value
      * attribute handle.
      */
-    uint16_t * const val_handle;
+    uint16_t *val_handle;
 };
 
 struct ble_gatt_svc_def {
--- a/net/nimble/host/include/host/ble_gatt.h
+++ b/net/nimble/host/include/host/ble_gatt.h
@@ -430,6 +430,8 @@ typedef void (*ble_gatt_svc_foreach_fn)(const struct ble_gatt_svc_def *svc,
                                         uint16_t handle,
                                         uint16_t end_group_handle);
 void ble_gatts_show_local(void);
+int ble_gatts_reset(void);
+int ble_gatts_start(void);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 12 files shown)


sjanc:
Hi Chris,

so I was wondering if we could have   ble_gatts_add/remove_service() instead and keep tthose internally on list, not array.

Rationale is that now each application has to provide GATT db from 'central' point which make it hard to have library like components registering own services. One example could be BT mesh  but this would be useful in general as we could provide sample services as libraries.

ccollins476ad:
You're right - that would be better.  I will take a look.

ccollins476ad:
I may have misunderstood your comment initially.  Just to clarify: different packages can register their own services individually.  For example, the `net/nimble/host/services/` packages do this.  When `ble_gatts_start()` is called, all registered services become available.  The sequence might look something like this:

* Package 1 registers services A, B, and C.
* Package 2 registers service D.
* App registers services E and F.
* App calls `ble_gatts_start()`

At the end of this sequence, all the services (A, B, C, D, E, and F) are supported and available to peers.

This is how service registration has always worked in the nimble host.

This PR allows the app to perform the above sequence a second time, with a different set of services.  The second sequence would look like this:

* Call `ble_gatts_reset()`
* Register services G and H.
* Register services I and J.
* Call `ble_gatts_start()`

Now, services G, H, I, and J are available to peers.

The point I'm trying to make is: not all services need to be registered at the same time in a single array.  I am not sure if you thought that was the case.

Anyway, when I first read your comment, I thought you were referring to the need to call `ble_gatts_start()` to make all the registered services available.  This is somewhat of a burden, because it prevents a package from registering additional services on the fly without cooperation from the application.

I probably should have explained the motivation for this PR.  The motivation is: blehostd needs to allow the user to register GATT services.  The user interfaces with blehostd via JSON messages sent over a unix domain socket.  By time the user is able to communicate with blehostd, blehostd has already initialized and started the nimble host.  To allow the user to register services, blehostd has to provide the ability to reset its GATT state and accept new service registration calls.

PR #483, opened 2017-08-10
BLE Host - Automatic own-addr-type resolution

ccollins476ad:
This commit adds a new function to the NimBLE host API:

```
/**
 * Determines the best address type to use for automatic address type
 * resolution.  Calculation of the best address type is done as follows:
 *
 * if privacy requested:
 *     if we have a random static address:
 *          --> RPA with static random ID
 *     else
 *          --> RPA with public ID
 *     end
 * else
 *     if we have a random static address:
 *          --> random static address
 *     else
 *          --> public address
 *     end
 * end
 *
 * @param privacy               (0/1) Whether to use a private address.
 * @param out_addr_type         On success, the "own addr type" code gets
 *                                  written here.
 *
 * @return                      0 if an address type was successfully inferred.
 *                              BLE_HS_ENOADDR if the device does not have a
 *                                  suitable address.
 *                              Other BLE host core code on error.
 */
int
ble_hs_id_infer_auto(int privacy, uint8_t *out_addr_type)
```

The motivation for this change is to allow BLE packages to be more self-contained.  If a package doesn't care what type of address it uses, it can just use whatever is available.


rymanluk:
it looks good to me

PR #488, opened 2017-08-12
Delay start events until end of main()

ccollins476ad:
Packages currently have two initialization phases: 1) init, and 2) start:

*init:* Runs automatically during sysinit.
*start:* A start event can optionally be put on the package's event queue; it runs when the event is processed.

The second phase is useful when the package has some settings that must be configure at runtime, but before the package starts.

The problem arises when a package is assigned a dedicated task and event queue.  If the task is higher priority than the default task, it will preempt the default task with the package's start event, before the application has had a chance to finish configuring the package.

This PR makes it so that start events always run in the default task, even if the package has been configured to run in a different one.  This ensures the start events won't run until the end of main() when the default event queue is processed.  Since start events always run in the default task, it is important that they not do much more than configure their package.

### DIFF
--- a/hw/sensor/src/sensor.c
+++ b/hw/sensor/src/sensor.c
@@ -325,7 +325,7 @@ sensor_mgr_evq_get(void)
 static void
 sensor_mgr_evq_set(struct os_eventq *evq)
 {
-    os_eventq_designate(&sensor_mgr.mgr_eventq, evq, NULL);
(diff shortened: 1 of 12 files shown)


ccollins476ad:
Thanks, Marko.  I have re-added `os_eventq_designate()` and labelled it deprecated.

PR #490, opened 2017-08-16
nimble/host: Refactor HCI commands sending functions

sjanc:
This refactors the way HCI commands are constructed and pass to
controller. Instead of embedding opcode and len in command parameters
buffer just pass it along with it. This has few adventages:
 - opcode OCF/OGF are contructed once on build time
 - buffer is always pass along with len
 - code is less confusing for static analysis tools (0 len buffer is
   always pass as NULL so no dangling pointers)
 - it is easier to follow (and debug)

Nice bonus is ROM size reduction of net_nimble_host.a (54304->53710).

This is also first step to making use of packed structures for HCI
command instead of builders functions. This should reduce code size
even further.

### DIFF
--- a/net/nimble/host/src/ble_gap.c
+++ b/net/nimble/host/src/ble_gap.c
@@ -442,7 +442,7 @@ int
 ble_gap_read_le_phy(uint16_t conn_handle, uint8_t *tx_phy, uint8_t *rx_phy)
 {
     struct ble_hs_conn *conn;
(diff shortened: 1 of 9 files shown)


sjanc:
ahh yes,  c'n'p typo :)  will update shortly

sjanc:
udpated (with few additional cleanup)

ccollins476ad:
Looks good!

PR #491, opened 2017-08-16
MYNEWT-801: Lora API

wes3:
Add first cut at lora api for application devlopers and first cut at MAC/higher layer API.


sterlinghughes:
tremendous!

PR #494, opened 2017-08-18
nimble: Allow to hide/restore services from clients in runtime

andrzej-kaczmarek:
This is substitute for registering and unregistering services in runtime.

Instead of the above, we allow already registered service to be hidden from clients (its attributes are removed from attributes list in ATT server). This can be done at any time and does not need any in terms of GATT services handling - we just move attributes from "default" list to "hidden" list so they are not picked by any ATT API.

So basically, if there are services A and B which cannot exists in ATT DB at the same time (like in Mesh), application should register both of them and then just toggle visibility in runtime.

### DIFF
--- a/net/nimble/host/src/ble_att_priv.h
+++ b/net/nimble/host/src/ble_att_priv.h
@@ -219,6 +219,8 @@ int ble_att_svr_read_handle(uint16_t conn_handle, uint16_t attr_handle,
 void ble_att_svr_reset(void);
 int ble_att_svr_init(void);
 
+void ble_att_svr_hide_range(uint16_t start_handle, uint16_t end_handle);
+void ble_att_svr_restore_range(uint16_t start_handle, uint16_t end_handle);
 
 /*** $clt */
 
--- a/net/nimble/host/src/ble_att_svr.c
+++ b/net/nimble/host/src/ble_att_svr.c
@@ -46,6 +46,7 @@
 
 STAILQ_HEAD(ble_att_svr_entry_list, ble_att_svr_entry);
 static struct ble_att_svr_entry_list ble_att_svr_list;
+static struct ble_att_svr_entry_list ble_att_svr_hidden_list;
 
 static uint16_t ble_att_svr_id;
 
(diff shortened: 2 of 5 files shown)


net/nimble/host/src/ble_att_priv.h
```diff
@@ -219,6 +219,8 @@ int ble_att_svr_read_handle(uint16_t conn_handle, uint16_t attr_handle,
 void ble_att_svr_reset(void);
 int ble_att_svr_init(void);
 
+void ble_att_svr_remove_range(uint16_t start_handle, uint16_t end_handle);
```
rymanluk: maybe better to name it ble_att_svr_hide_range() ?

net/nimble/host/src/ble_att_svr.c
```diff
@@ -2654,6 +2655,80 @@ ble_att_svr_rx_indicate(uint16_t conn_handle, struct os_mbuf **rxom)
...
+    /* Find first matching element to move */
+    remove = NULL;
+    entry = STAILQ_FIRST(src);
+    while (entry && entry->ha_handle_id < start_handle) {
+        remove = entry;
+        entry = STAILQ_NEXT(entry, ha_next);
+    }
+
+    /* Nothing to remove? */
+    if (!entry || entry->ha_handle_id < start_handle) {
```
ccollins476ad: It looks great to me, too.  My only comment is: do you need the check for `entry->ha_handle_id < start_handle` here?  I think the above loop takes care of that.

andrzej-kaczmarek: this is just in case there is only single entry on src list and its handle is lower than start_handle - loop does not cover this. it's of course highly unlikely to happen, but possible.

ccollins476ad: I might be wrong, but I still think the condition is not necessary.  The loop continues until either:
* entry is NULL, or
* `entry->ha_handle_id >= start_handle`

I don't see how entry can be non-null but still have a handle ID less than start_handle.

andrzej-kaczmarek: yes, you're obviously right. in my mind loop was not executed in case I described, so entry was still not null - but after reevaluating this it now makes much more sense :-)

I'll fix this.

PR #498, opened 2017-08-24
nordic: LF clock modifications

wes3:
The following changes were made to the code regarding the LF
clock for nordic platforms. The initial impetus for this change
was that there was no support for using a synthesized 32kHz clock
for nrf52 platforms. LF clock management was different between
the nrf51 and nrf52 as well and this commit is intended to make
them consistent. Here is a list of the changes made:

1) Always turn on the LF clock for nordic bsps/platforms.
2) Move the LF clock source syscfg definitions from the BSP into
the mcu.
3) BSP now sets the proper LF clock based on the BSP hw (either
external, synthesized or internal RC oscillator).
4) The OS will now always use the 32768 crystal for OS time. All
code that dealt with it using a different clock source has been
removed.
5) Both the nrf51 and nrf52 now use the same RTC time resources.

### DIFF
--- a/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
+++ b/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
@@ -26,6 +26,7 @@
 #include "sysflash/sysflash.h"
 #include "flash_map/flash_map.h"
 #include "hal/hal_bsp.h"
+#include "hal/hal_system.h"
 #include "hal/hal_flash.h"
 #include "hal/hal_spi.h"
 #include "hal/hal_watchdog.h"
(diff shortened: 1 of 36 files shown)


wes3:
Just an FYI: I tried to build a fair amount of targets but there are many that I did not build. I also did very limited testing of this fix so was hoping the community would help test :-)

wes3:
Regarding Marko's comment: I have to say I really do not like the name/location of hal_system_clock_start(). I did not want to make that change as well here (not quite sure why other than not making too many changes all at once). I guess also I was not sure what to do. I was thinking that the code to turn on the clock would exist only in hal_bsp.c as a static function to be honest. This way we could remove the proto from hw/hal and not have it in mcu at all. I do agree with you though Marko: the current code should get modified to remove hal_system_clock_start() for hw/hal.

Another possible issue with this change is that the code now always uses the LFCLK. I do not think this a big deal as generally I suspect folks will use nordic platforms w/BLE but I could envision cases where an application does not care about power consumption and wants to use the HFCLK for everything. This is currently not possible and if folks want to do that they are going to have to modify the code to handle this.

PR #499, opened 2017-08-25
Convert some defines in `bsp.h` to syscfg settings

ccollins476ad:
This PR converts a number of `#define`s in each BSP's `bsp.h` file to syscfg settings.  The motivation for this change was to simplify the `bsp.h` files for the purposes of documentation.  I think this change is also desirable for these reasons:

1. Reduce code duplication: several of these settings were defined with the same value in each BSP.  Now these settings are defined once in one place.

2. Better modularity: each of these settings are used by a single package.  It is better to define the setting in the package that needs it and allow the BSP to override it.

3. More opportunities to override: syscfg settings can be overridden by the BSP, app, and / or target, unlike `#define`s.

The following syscfg settings are added:

```
    NFFS_NUM_AREAS:
        description: >
            Number of areas to allocate in the NFFS disk.  A smaller number is
            used if the flash hardware cannot support this value.
        value: 8

    CONFIG_FCB_NUM_AREAS:
        description: >
            Number of areas to allocate in the config FCB.  A smaller number is
            used if the flash hardware cannot support this value.
        value: 8

    BOOT_SERIAL_DETECT_PIN:
        description: >
            Start the serial boot loader if this pin is asserted at boot time.
        value:
        restrictions:
            - '$notnull'

    BOOT_SERIAL_DETECT_PIN_CFG:
        description: >
            GPIO configuration for the serial boot loader detect pin.
        value: 'HAL_GPIO_PULL_UP'

    BOOT_SERIAL_DETECT_PIN_VAL:
        description: >
            The value the detect pin must be set to for the serial boot loader
            to start.
        value: 0

    BOOT_SERIAL_REPORT_PIN:
        description: >
            The GPIO to toggle while the serial boot loader is running.  Set to
            -1 to disable reporting.
        value: 'LED_BLINK_PIN'

    BOOT_SERIAL_REPORT_FREQ:
        description: >
            The toggle rate, in Hz, of the serial boot loader report pin.
        value: 4
```


ccollins476ad:
This PR breaks some code.  FCB also uses this `#define`, and the replacement syscfg setting isn't present unless NFFS is part of the build.

PR #501, opened 2017-08-26
Fix warning for clicker and wifire boards

francois-berder:


hw/bsp/pic32mz2048_wi-fire/src/hal_bsp.c
```diff
@@ -212,7 +212,7 @@ hal_bsp_init(void)
 
     #if MYNEWT_VAL(UART_3)
         rc = os_dev_create((struct os_dev *) &os_bsp_uart3, "uart3",
-            OS_DEV_INIT_PRIMARY, 0, uart_hal_init, &uart4_cfg);
+            OS_DEV_INIT_PRIMARY, 0, uart_hal_init, (void *)&uart4_cfg);
```
mkiiskila: These void * typecasts of arguments are not needed for os_dev_create()/hal_spi_init(), as the function prototype already has declared that argument as void *.
I recommend not adding them, as they're not needed.

francois-berder: Actually you have to. Otherwise you get warnings such as:
```
repos/apache-mynewt-core/hw/bsp/pic32mz2048_wi-fire/src/hal_bsp.c:223:13: warning: passing argument 6 of 'os_dev_create' discards 'const' qualifier from pointer target type [enabled by default]
```

mkiiskila: Ah, missed the const in the declaration.

PR #506, opened 2017-08-28
shell - Don't register NULL command.

ccollins476ad:
As pointed out by Chris Merck, a command entry with all-fields-null was getting registered.  This made subsequently registered commands inaccessible.


ccollins476ad:
Thanks, I will remove  SHELL_OS_NUM_CMDS.

PR #507, opened 2017-08-28
boot_serial; access uart driver directly instead of going through console.

mkiiskila:
This is for space savings. Before the change:
objsize
   text	   data	    bss	    dec	    hex	filename
  16248	     84	   5352	  21684	   54b4	/Users/marko/src2/incubator-mynewt-blinky/bin/targets/boot_nrf52_serial/app/apps/boot/boot.elf

After the change:
   text	   data	    bss	    dec	    hex	filename
  14816	     76	   4996	  19888	   4db0	/Users/marko/src2/incubator-mynewt-blinky/bin/targets/boot_nrf52_serial/app/apps/boot/boot.elf

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -37,8 +37,6 @@
 #include <os/os_malloc.h>
 #include <os/os_cputime.h>
 
-#include <console/console.h>
-
 #include <tinycbor/cbor.h>
 #include <tinycbor/cbor_buf_reader.h>
 #include <base64/base64.h>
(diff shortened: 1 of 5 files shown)


mkiiskila:
The serial bootloader opens the UART, and uses syscfg settings to figure out which one to open, baudrate etc. I could've placed these syscfg definitions to under boot/boot_serial instead, but it felt more natural to have them under console/stub.

I wanted to use the same syscfg variable names (as these might be overwritten in bsp). console/stub gets used when building bootloader, so it sort of works naturally. Although the reason to have them there is a bit unnatural.

PR #510, opened 2017-08-30
Revert update of TinyCrypt library

sjanc:
TinyCrypt 0.2.7 changed ECC API in backward incompatible way. While SMP code was updated this was causing issues with some external components (eg mcuboot).


rymanluk:
agree that it might be better to do update of TinyCrypt after 1.2 is released as there is couple of dependencies we should solved first.

PR #511, opened 2017-08-30
nimble: Port Bluetooth Mesh from Zephyr to Mynewt

rymanluk:
This is a port of Bluetooth Mesh. We tried to keep it as close to Zephyr as possible so both OSes can leverage from each other.


rymanluk:
Should be merged after TinyCrypto revert.

ccollins476ad:
Looks really good.  Wow, that is a lot of code! :)

I have some review comments.  I am not very familiar with the Bluetooth mesh spec, so I focused most of my attention on the Mynewt-specific parts of this PR.  Also, I understand that a goal is to modify this code as little as possible to make it easier to accept upstream changes.  I agree that this is a good idea.  If some of my comments oppose this goal, then you should feel free to ignore them.

Anyway, here are my comments:

---

`net/nimble/host/mesh/include/mesh/glue.h:NET_BUF_SIMPLE()`

This macro may yield NULL, so the resulting mbuf should be checked by the caller.  If it is not practical to modify all the instances where this macro gets invoked, I would add an `assert()` to the macro.

---

`net/nimble/host/mesh/src/glue.c:bt_gatt_get_mtu()`

Is this function needed?  If you remove it, then you can also remove the include of private header "../src/ble_hs_conn_priv.h"

---

`net/nimble/host/mesh/src/glue.c:net_buf_simple_pull()`
`net/nimble/host/mesh/src/glue.c:net_buf_simple_pull_le16()`
`net/nimble/host/mesh/src/glue.c:net_buf_simple_pull_be16()`
`net/nimble/host/mesh/src/glue.c:net_buf_simple_pull_be32()`

These might not be safe.  I see two possible issues:

1. source mbuf is too small
2. data is split over multiple mbufs in a chain.

In both cases, these function read past the end of the mbuf buffer.  I suggest calling `os_mbuf_pullup()` at the start of these functions.

---

`net/nimble/host/mesh/src/glue.c:net_buf_simple_push()`
`net/nimble/host/mesh/src/glue.c:net_buf_simple_push_u8()`
`net/nimble/host/mesh/src/glue.c:net_buf_simple_push_le16()`
`net/nimble/host/mesh/src/glue.c:net_buf_simple_push_be16()`

These functions seem to duplicate the functionality of `os_mbuf_prepend()`.  I suggest calling `os_mbuf_prepend()` within these functions.

---

`net/nimble/host/mesh/src/glue.c:bt_pub_key_gen()`

There is some dead code inside an `if (0)` at the top.

---

`net/nimble/host/mesh/src/glue.c:net_buf_reserve()`

I would use os_mbuf_adj() here instead of adjusting the data pointer manually, just to be sure the resulting mbuf is valid.

---

`net/nimble/host/mesh/src/adv.h:BT_MESH_ADV()`

This macro looks like a copy of `OS_MBUF_USRHDR()` with a cast.  It might be better if this macro calls `OS_MBUF_USRHDR()` rather than duplicating the code.

---

`net/nimble/host/mesh/src/beacon.c:secure_beacon_recv()`

You might want to call `os_mbuf_pullup(buf, 21)` at the start of this function.

---

`net/nimble/host/mesh/src/cfg.c:get_model()`

Consider calling `os_mbuf_pullup(buf, 4)` at the start of this function.

---

`net/nimble/host/mesh/src/mesh_priv.h`

the `struct bt_mesh_net;` forward declaration seems unnecessary.

---

`net/nimble/host/mesh/src/prov.c:bt_mesh_input_string()`

Is `strncpy()` the correct function to use here?  If `strlen(str) >= prov->output_size`, then `link.auth` won't be null-terminated.

---

`net/nimble/host/mesh/src/prov.c:prov_data()`

Does `msg` leak if `bt_mesh_session_key()` fails?

rymanluk:
Thanks for review. Here are replies:

1) good point , assert() added
2) Fixed

3 / 4) Operations on os_mbuf was one of the main issues during porting I had to overcome. Basically we don't what any chaining for the moment, because it would required much more code changes. E.x. using os_mbuf_prepend() could result in new buf pointer which is something we don't want as user has no way to get new buffer pointer if we want to keep current  net_buf API. Frankly speaking, I tried that approach at the beginning but then I run into problems.

The thing is that Mesh code always allocates big enough buffers for its messages. Having this in mind, and choosing big enough mempool blocks we can avoid chaining. The only thing left is to reserve headroom,  and since os_mbuf packet header has been used for user_data, we came up  with headroom space just after packet header ( see net_buf_reserve). You can see that for all foo_push() functions there is and assert if there is not enough headroom space. For foo_pull() functions I can add pullup() and add assert() in case buffer has changed as it shall not happen.

I was thinking that next step would be to use all the power of mbuf, but this is bigger chunk of work. I will create Jira for that.

5) OMG - that was for testing when I had fixed public key - thanks for catching it.

6) We don't want to change len of the buffer here. Here is our headroom I mentioned before. I would leave it for now

7) You are right. Fixed.

8/9 This buf is allocated in ble_adv_gap_mesh_cb() and we are sure that block size is big enough. So there is no need for pullup

10) you are right. fixed

11) Well, link.auth is zero-ed so there should not be a problem. But thanks to this comment I found other issue as there should be prov->input_size instead of output_size.

12) Fixed

rymanluk:
Chris, I merged that in order to start tagging. If we need other fixes we can do it on top.

PR #512, opened 2017-08-30
native; support setting hwid for native target via command line argum…

mkiiskila:
…ent.

### DIFF
--- a/hw/mcu/native/src/hal_hw_id.c
+++ b/hw/mcu/native/src/hal_hw_id.c
@@ -22,9 +22,14 @@
 
 #include "hal/hal_bsp.h"
 
+#include "hal_native_priv.h"
+
 #ifndef min
 #define min(a, b) ((a)<(b)?(a):(b))
 #endif
+
+static uint8_t hal_hw_id[HAL_BSP_MAX_ID_LEN];
+static uint8_t hal_hw_id_len;
 
 /*
  * This can be used as the unique hardware identifier for the platform, as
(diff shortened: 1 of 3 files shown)


hw/mcu/native/src/hal_hw_id.c
```diff
@@ -22,20 +22,39 @@
...
+
 /*
  * This can be used as the unique hardware identifier for the platform, as
  * it's supposed to be unique for this particular MCU.
  */
 int
 hal_bsp_hw_id(uint8_t *id, int max_len)
 {
+    if (hal_hw_id_len) {
+        memcpy(id, hal_hw_id, hal_hw_id_len);
```
ccollins476ad: Should this function make sure it doesn't copy more than `max_len` bytes here?

hw/mcu/native/src/hal_hw_id.c
```diff
@@ -22,20 +22,39 @@
...
+    }
     if (max_len > HAL_BSP_MAX_ID_LEN) {
         max_len = HAL_BSP_MAX_ID_LEN;
     }
     memset(id, 0x42, max_len);
     return max_len;
 }
+
+void
+hal_bsp_set_hw_id(uint8_t *id, int len)
```
ccollins476ad: Just a suggestion: you could make `id` pointer to const.

hw/mcu/native/src/hal_hw_id.c
mkiiskila: Yes, it should. Lemme fix that.

PR #513, opened 2017-08-30
Lora improv

wes3:


sjanc:
something is not right with second commit message :)

wes3:
I think I see what you mean regarding second commit message. There are additional changes to be made so will update this PR...

wes3:
I know what you mean re: some of the exported symbols. I can change them prior to release if folks think it is important.

PR #534, opened 2017-09-06
nimble/sm: Use privacy device mode as default

rymanluk:
In order to not brake existing applications lets use privacy
device mode instead of network mode which is default in the
Controller accorting to BT 5 specification.


rymanluk:
test please

PR #535, opened 2017-09-06
nimble/controller: Use privacy device mode on host before BT 5

rymanluk:
Controller does not know what Bluetooth version is supported by Host.
In order to avoid issues with reconnecting device due to
PRIVACY_MODE_NETWORK used as default on the controller, it tries to
figure out if Host is < BT5. If so PRIVACY_MODE_DEVICE is used


rymanluk:
I don't like it

PR #536, opened 2017-09-06
Works with new image format.

utzig:
This adds support for the new image header format (done by @mkiiskila).


ccollins476ad:
Where does `struct image_tlv_info` get defined?  I couldn't find it anywhere.

utzig:
@ccollins476ad That code is #ifdefs out for our bundled bootloader. It's a new TLV that is only used by mcuboot to support the new image format.

PR #543, opened 2017-09-12
sys/shell: add condition to newtmgr dependencies

michal-narajowski:
These packages are used only by newtmgr part of shell code.


michal-narajowski:
test this please

PR #545, opened 2017-09-12
Enable the external antenna for the SX1276 on the EE-02 board.

gordonklaus:
The internal one is of no use.  We always want to use the external.


gordonklaus:
That's correct about the switching.

I'm actually not sure that the internal antenna is useless, I shouldn't have said that.  As I recall, it was very difficult to tune and it did behave poorly, and so it's actually been removed in a more recent version of the board.

wes3:
All sounds good. I will go ahead and merge this. Thanks for changing this.

PR #548, opened 2017-09-14
os_task: Moved the initialisation of t_stacktop and t_stacksize.

IMGJulian:
So they can be modified by the architecture dependent stack init.

As discussed [here](https://github.com/apache/mynewt-core/pull/474#discussion_r131878603)

### DIFF
--- a/kernel/os/src/os_task.c
+++ b/kernel/os/src/os_task.c
@@ -129,10 +129,10 @@ os_task_init(struct os_task *t, const char *name, os_task_func_t func,
     }
 
     _clear_stack(stack_bottom, stack_size);
-    t->t_stackptr = os_arch_task_stack_init(t, &stack_bottom[stack_size],
-            stack_size);
     t->t_stacktop = &stack_bottom[stack_size];
     t->t_stacksize = stack_size;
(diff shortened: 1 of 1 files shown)


kernel/os/src/os_task.c
```diff
@@ -129,10 +129,10 @@ os_task_init(struct os_task *t, const char *name, os_task_func_t func,
     }
 
     _clear_stack(stack_bottom, stack_size);
-    t->t_stackptr = os_arch_task_stack_init(t, &stack_bottom[stack_size],
-            stack_size);
     t->t_stacktop = &stack_bottom[stack_size];
     t->t_stacksize = stack_size;
+    t->t_stackptr = os_arch_task_stack_init(t, &stack_bottom[stack_size],
+            stack_size);
```
rymanluk: maybe we could do os_arch_task_stack_init(t, t->stacktop, t->stacksize) as those are already set and imho it is cleaner when reading code

Next step would be to end up only with os_arch_task_stack_init(t), but that requires more changes :) so it is ok to leave it for now.

rymanluk:
Looks good to me

PR #552, opened 2017-09-15
native-armv7; stack initialization 

mkiiskila:
…for tasks was violating ARM call rules.

Task stack init routine departed with $sp pointing to value which was not aligned to 8 byte boundary.

### DIFF
--- a/kernel/os/src/arch/sim-armv7/os_arch_stack_frame.s
+++ b/kernel/os/src/arch/sim-armv7/os_arch_stack_frame.s
@@ -29,7 +29,7 @@
 os_arch_frame_init:
     mov	    r1, sp
     mov	    sp, r0		/* stack for the task starts from sf */
-    sub     sp, sp, #20
+    sub     sp, sp, #24	/* stack must be aligned by 8 */
     str	    r1, [sp, #12]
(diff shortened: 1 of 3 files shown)


kernel/os/src/arch/sim-armv7/os_arch_stack_frame.s
```diff
@@ -40,6 +40,7 @@ os_arch_frame_init:
     beq     end
     mov	    r1, r0
     ldr     r0, [sp, #4]
+    and     sp, sp, #0xfffffff0
```
ccollins476ad: I might have misunderstood, so feel free to ignore this comment.  Should this be `0xfffffff8` instead of `0xfffffff0`?  Currently, this ensures the stack pointer is a multiple of 16, but I thought the intent was to ensure it is a multiple of eight.

mkiiskila: Indeed, multiple of 8 is sufficient. I confused myself by looking at OS_STACK_ALIGNMENT define in os_arch.h. That number for armv7 was inherited from x86, where that might've been a restriction set by the underlying OS (I can't remember!). And is most likely not quite right.

PR #553, opened 2017-09-17
Conform the sample light resource to binary switch

bgiori:
This conforms the sample light resource in the bleprph_oic
sample app to the binary switch resource type as defined
in the OCF Resource Type Spec:

https://openconnectivity.org/draftspecs/OCF_Resource_Type_Specification_v1.0.0.pdf

### DIFF
--- a/apps/bleprph_oic/src/main.c
+++ b/apps/bleprph_oic/src/main.c
@@ -345,6 +345,7 @@ omgr_app_init(void)
     oc_resource_set_periodic_observable(res, 1);
     oc_resource_set_request_handler(res, OC_GET, app_get_light);
     oc_resource_set_request_handler(res, OC_PUT, app_set_light);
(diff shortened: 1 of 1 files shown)


bgiori:
This will allow people to use the android sensors_test app with a device running bleprph_oic without having to modify the code.

mkiiskila:
This should be ok.
Do you also want to change the write handler to use POST instead of current PUT?

bgiori:
That would conform better but the sensors test app uses put, and uploading a new app to the store for such a small change is not worth the effort involved. My concern is more that the two work out of the box rather than conformity.

mkiiskila:
How about support both?
i.e. add
oc_resource_set_request_handler(res, OC_POST, app_set_light);
after setting the handler for OC_PUT

PR #556, opened 2017-09-18
Add compat. rule to require newt 1.1.0+.

ccollins476ad:
Core 1.1.0+ requires newt 1.1.0+ (feature: self-overrides).  This change enforces that requirement through a compatibility rule.

### DIFF
--- a/repository.yml
+++ b/repository.yml
@@ -40,3 +40,11 @@ repo.versions:
     "1.0-latest": "1.0.0"  # 1.0.0
     "1.1-latest": "1.1.0"  # 1.1.0
     "1.2-latest": "1.2.0"  # 1.2.0
+
+repo.newt_compatibility:
+    # Core 1.1.0+ requires newt 1.1.0+ (feature: self-overrides).
+    1.2.0:
+        1.1.0: good
+    1.1.0:
+        1.1.0: good
+


sjanc:
so it defines minimal version of newt tool for specified release of mynewt-core?  (just trying to get my head around this)

ccollins476ad:
Yes, that's exactly right.  Here is a brief description of the compatibility table format:

```
    This is done with a separate map "repo.newt_compatibility" in the repo's
    repository.yml file.

    Example:

    repo.newt_compatibility:
        1.0.0:
            1.1.0: error
            1.0.0: good
            0.9.99: warn
            0.9.9: error

    This says that this version 1.0.0 of this repo is compatible with newt
    versions [1.0.0, 1.1.0). Newt versions in the range [0.9.99, 1.0.0)
    elicit a warning message. All other newt versions generate an error.
```

This PR adds the following table:

```
repo.newt_compatibility:
    # Core 1.1.0+ requires newt 1.1.0+ (feature: self-overrides).
    1.2.0:
        1.1.0: good
    1.1.0:
        1.1.0: good
```

This table contains two rules:
* core 1.2.0
* core 1.1.0

In both cases, newt 1.1.0+ is "good".  Any version of newt older than 1.1.0 will elicit this error:

```
[ccollins@ccollins-mac:~/tmp/myproj]$ newt test all
Error: This version of newt (1.0.0) is incompatible with your version of the apache-mynewt-core repo (1.2.0); Please upgrade your newt tool to version 1.1.0
```

The table format is a bit unwieldy, and it is unfortunate that we have to add a new rule for each version of the repo.  Perhaps that is something we can improve in an upcoming release.

ccollins476ad:
The example sets an upper bound on the "good" newt versions.  The top subrule (`1.1.0: error`) indicates that newt 1.1.0+ is incompatible with 1.0.0 of this repo.  I guess we might add a rule like this if we modify newt in a way that breaks backwards compatibility.

PR #560, opened 2017-09-19
Set explicit default values for LSM303DLHC mag

microbuilder:
This PR avoids repetitive warnings in the shell when the LSM303DLHC is used.

The default values corresponds to the IC values coming out of reset (+/-1.3 gauss and 15Hz sample rate on the magnetometer).

### DIFF
--- a/hw/drivers/sensors/lsm303dlhc/src/lsm303dlhc.c
+++ b/hw/drivers/sensors/lsm303dlhc/src/lsm303dlhc.c
@@ -290,7 +290,7 @@ lsm303dlhc_init(struct os_dev *dev, void *arg)
     }
 
     /* Add the accelerometer/magnetometer driver */
-    rc = sensor_set_driver(sensor, SENSOR_TYPE_LINEAR_ACCEL |
+    rc = sensor_set_driver(sensor, SENSOR_TYPE_ACCELEROMETER |
             SENSOR_TYPE_MAGNETIC_FIELD,
             (struct sensor_driver *) &g_lsm303dlhc_sensor_driver);
     if (rc != 0) {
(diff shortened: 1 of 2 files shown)


microbuilder:
There is also an issue with the accelerometer `sensor_type_t` value, which is currently defined as `SENSOR_TYPE_LINEAR_ACCEL`.

Linear Acceleration is defined as `/* Linear Accelerometer (Without Gravity) */` (sensor.h), and requires some signal processing to remove the influence of gravity. The low cost LSM303DLHC only provides standard accelerometer data, which includes the influence of gravity, and should be defined as `SENSOR_TYPE_ACCELEROMETER`.

microbuilder:
I don't know if there is a standard/formal distinction between 'acceleration' and 'linear acceleration', but that's how we've designed things interally, for lack of better labels. 

For me, the LSM303DLHC doesn't provide linear acceleration data in the sense that I understand the term, whereas a more complex chip like the BNO055 does at the expense of additioanl internal calculation based on the raw acceleration data (which is also available). 

It's true that the LSM **can** be a linear accelerometer, but it's like marketting a bare LED as a lightbulb only because it can be one with additional components. :) Maybe the distinction in terms aren't as universal as I thought, though! For me:

- Acceleration = the raw values including the effect of gravity, meaning you can measure orientation.
- Linear Acceleration = the values with gravity removed, meaning that you are measuring acceleration in a linear fashion on each axis to estimate rate of fall, speed moving forward, impact strength, etc.

This follows a similar distinction on Android, for example: https://developer.android.com/guide/topics/sensors/sensors_motion.html#sensors-motion-linear

You can see both sensor data types in the Android API here: https://developer.android.com/guide/topics/sensors/sensors_overview.html

PR #563, opened 2017-09-20
Mesh qualification improvements

michal-narajowski:


net/nimble/host/mesh/src/net.c
```diff
@@ -789,26 +789,37 @@ int bt_mesh_net_send(struct bt_mesh_net_tx *tx, struct os_mbuf *buf,
...
+	net_buf_put(&bt_mesh.local_queue, net_buf_ref(buf));
+	if (cb) {
+		cb(buf, 0);
+	}
+	k_work_submit(&bt_mesh.local_work);
+
+local_done:
+
+#elif
+    bt_mesh_adv_send(buf, cb);
```
sjanc: if (bt_mesh_fixed_group_match(tx->ctx->addr)) {
            bt_mesh_adv_send(buf, cb);
        } else if (!bt_mesh_elem_find(tx->ctx->addr)) {
                bt_mesh_adv_send(buf, cb);
                goto done;
            }
        }

	net_buf_put(&bt_mesh.local_queue, net_buf_ref(buf));
	if (cb) {
		cb(buf, 0);
	}
	k_work_submit(&bt_mesh.local_work);

done:



this requires only 1 label,  you may also use local variable   is_local and set it properly in if-elseif and then use it to decide if cb should be called

net/nimble/host/mesh/src/net.c
```diff
@@ -1081,7 +1081,7 @@ int bt_mesh_net_decode(struct os_mbuf *data, enum bt_mesh_net_if net_if,
 		       struct bt_mesh_net_rx *rx, struct os_mbuf *buf,
 		       struct net_buf_simple_state *state)
 {
-	if (data->om_len < 18) {
+	if (data->om_len < 14) {
```
rymanluk: As it turns out this patch is incorrect. In calculations in commit message assumes NetMIC is 32. According to spec, NetMIC can be 32 only when CTL is 0 and when it happens then "Access Message" is included into msg which is 32 bits. That means 18 is OK. We need PTS issue instead of this patch.

"3.4.4.3
CTL
The CTL field is a 1-bit value that is used to determine if the message is part of a Control Message or an
Access Message, as illustrated in Table 3.8.
If the CTL field is set to 0, the NetMIC is a 32-bit value and the Lower Transport PDU contains an Access
Message.
If the CTL field is set to 1, the NetMIC is a 64-bit value and the Lower Transport PDU contains a Control
Message.
"

michal-narajowski:
Thanks for the review. I fixed the issues and updated the pull request.

michal-narajowski:
Added commit msg to "mesh: fix handling group send with local interface enabled" to provide more information and fixed a typo in the same commit.

PR #564, opened 2017-09-20
Do not allow both SPI 0 and I2C 0 to be enabled on the EE-02 board.

gordonklaus:
On nRF52, SPI 0 shares resources with I2C 0, so they cannot both be enabled at the same time.

Also add I2C_1 (with analogous restrictions with SPI 1).  Don’t add SPI_1 yet as we had trouble getting it to work.

More ideally, it seems like we ought to be able to use both SPI 0 and I2C 0, not simultaneously but by disabling them at runtime while they are not in use.  This is surely a much larger change.

### DIFF
--- a/hw/bsp/telee02/syscfg.yml
+++ b/hw/bsp/telee02/syscfg.yml
@@ -54,6 +54,7 @@ syscfg.defs:
         value:  1
         restrictions:
             - "!SPI_0_SLAVE"
+            - "!I2C_0"
     SPI_0_MASTER_SS_PIN:
         description: 'SPI 0 (master) SS pin number.'
         value:  22
@@ -62,6 +63,7 @@ syscfg.defs:
         value:  0
         restrictions:
             - "!SPI_0_MASTER"
+            - "!I2C_0"
 
     TIMER_0:
         description: 'NRF52 Timer 0'
@@ -85,6 +87,15 @@ syscfg.defs:
     I2C_0:
         description: 'NRF52 I2C (TWI) interface 0'
         value:  0
+        restrictions:
+            - "!SPI_0_MASTER"
+            - "!SPI_0_SLAVE"
+    I2C_1:
+        description: 'NRF52 I2C (TWI) interface 1'
+        value:  0
+        restrictions:
+            - "!SPI_1_MASTER"
+            - "!SPI_1_SLAVE"
 
 syscfg.defs.BLE_LP_CLOCK:
     TIMER_0:


gordonklaus:
@wes3 Thanks for the explanation, it's nice to get some perspective 😄   I do see now that this is documented under Porting to your Platform – oops, I hadn't read that yet.

Is the idea then that we should use a BSP and adapt it by adding configuration to our apps/targets?  Or do you suggest copying the BSP and modifying it to suit our needs?  The latter would be unfortunate.

The reason I was thinking of adding SPI1 was just so that the restriction between it and I2C1 would be in place to prevent future users from having to rediscover the conflict.  Maybe it would be better to implement these checks in the MCU.  They are, after all, relevant for all nRF52 based boards.  What do you think?

Also, what do you think about my suggestion that these configurations should not be static but that SPI and I2C should only enable themselves as necessary, allowing e.g. both SPI0 and I2C0 to be used in the same program?

simonratner:
@we3, to add a personal perspective, while BSPs are not meant to be generic, maintaining a modified one can be onerous. If the changes are restricted to target config -- great -- but if there are code changes, such as in `src/hal_bsp.c`, then you need to fork the BSP and keep up with any upstream updates. 

There have been several cases of a change in MCU package or `kernel/os` requiring a matching change in the BSP hal or startup code, and failing to merge that change into your fork can result in anything from the board not booting to undefined runtime behaviour. Even syscfg changes have had this problem (like the LP clock settings), although those are significantly easier to spot and port.

I don't know what a good answer is for maintaining a custom BSP. Perhaps `gcc_startup.s` can move into the MCU package and allow the BSP to optionally replace it by providing its own, since in the majority of cases the startup code doesn't change. Similar with `hal_bsp.c`, the MCU package could initialise all on-die devices (switchable with config vars), and provide a hook or two to initialise additional peripherals in the BSP. Or alternatively, some sort of BSP layering?

wes3:
I will admit I do not have a good answer for either of you. I would hope that bsp changes would be infrequent so merging in changes to your custom bsp would not be terribly difficult but I do understand it is a pain. A couple of things I do agree with: the gcc_startup code probably does not belong in the bsp. I would hope that would be generic enough to use linker script definitions and not have to be customized per BSP. The linker scripts are another issue although they are probably more "bsp-ish" than the startup code. Seems like a good conversation for the dev list.

wes3:
@gordonklaus So I am not sure about using both spi0 and i2c0 and disabling one then using the other. That does not seem to me to be a very common use case. There are a fair amount of i2c and spi peripherals and not alot of gpio so I would think that most folks would dedicate them. Do you have a use case for doing that? I think it fine that the syscfg file does not allow to enable both; seems like what folks would actually want but that is just my opinion.

As to copying a bsp and modifying it to suit your purposes or specifying app/target configuration, well, a decision was made a while back to put the pin configurations in a c source file and not make them syscfg definitions (for example, specifying which pins to use for a spi peripheral). To me this implies, dare I say dictates, the copy and modify approach. Not saying it was the correct decision mind you.

Lastly, I was considering mentioning that the checks you added to the bsp syscfg file might be better located in the mcu dir. I agree; that is mcu and not bsp specific and seems to me to be the proper place for checking that. Not sure why I did not mention it; maybe I did not want to sound too critical or something. :-)

gordonklaus:
@wes3 I don't have a current need for using both spi0 and i2c0 but maybe it doesn't seem far fetched given that there are only 3 spi instances and 2 i2c instances on the nrf52?  I don't know, the suggestion actually came from one of my teammates and I don't have any personal experience to shed on the idea.

That is perhaps unfortunate to have to copy and modify a bsp, but time will tell.

Thanks for being gentle and all but do bring the criticism on :-)  I'll look at reverting this change and doing it in the mcu instead, since we agree.

gordonklaus:
@wes3 By the way, what prompted this PR in the first place was that we were reading from a sensor via I2C0 and when we tried to use LoRa at the same time things stopped working.  There seems to be no option but to set RADIO_SPI_IDX to zero for LoRa.  Do you plan to make this actually configurable?

gordonklaus:
I am now wondering wondering why I2C_0 and _1 and SPI_0_MASTER are defined in the BSPs, whereas their primary use seems to be in the MCU.  Should these be moved?  It would certainly reduce duplication, and it would let us define restrictions in the syscfg instead of in code.

wes3:
@gordonklaus I think we should probably move some of this conversation to either the mynewt slack channel or to the dev list as I certainly do not have all the answers and others may have reasons why things are the way they are. The bsp/mcu "distinctions" have always been a bit "confused" and we have definitely moved things from bsp to mcu.

As to the spi used by the radio I had not thought of making it configurable but it certainly seems reasonable to me.

gordonklaus:
@wes3 Good idea, I'll move the bsp/mcu stuff to a broader forum.  😄 

The RADIO_SPI_IDX configurability isn't urgent.  But it could be nice to be able to use SPI2 since there's no contention with I2C for it on nrf52.

PR #565, opened 2017-09-21
mesh: Use privacy for mesh node

rymanluk:


sjanc:
I think mesh app should always use NRPA.

rymanluk:
@sjanc, agree will change it

PR #566, opened 2017-09-21
Remove I2C0 initialization code from the EE-02 BSP.

gordonklaus:
This code assumed that pins 26 and 28 would be used for I2C, but this is not dictated by the EE-02 design.  It does not belong in the generic BSP.


gordonklaus:
Furthermore, prior to removing it herein, the call to hal_i2c_init seems to have had no effect.  It was still necessary to call hal_i2c_init from application code.  Out of curiosity, can this be explained?

PR #573, opened 2017-09-26
Move SPI and I2C config definitions from BSPs to MCUs.

gordonklaus:
Also add restrictions between I2C and SPI on nRF5x boards, as was already done for the EE-02.

This is primarily a refactoring, but it could cause some builds to fail.  In particular, any BSP not the core that defines these settings will see “setting XXX redefined” errors.  And any nRF5x app or target that had both I2C_0 and SPI_0 enabled (or _1) will run into the new restrictions.

Otherwise, all settings (interfaces enabled, pin numbers) are unchanged, with one exception:  The VBLUno boards had enabled I2C and SPI 0 and 1.  To resolve the conflict, I disabled SPI, as at least I2C_0 seems like it ought to be enabled by default for the on-board OLED.


andrzej-kaczmarek:
correct me if I'm wrong but SPI_0_MASTER_SS_PIN is a slave select pin so basically it can be virtually any pin which application controls anyway. so it should not be defined at mcu level as it does not make much sense there and honestly I'd remove this definition entirely. I see this syscfg val is only used now by lora driver, but in this case this driver should have its own syscfg val for defining SS pin number.

gordonklaus:
@andrzej-kaczmarek That's correct.  I was a bit uncertain about where this should be defined.  You are right, it makes sense that it should be defined by the package that uses it.  I will move the definition of SPI_0_MASTER_SS_PIN into sx1276.  I will also rename it to LORA_SPI_CS_PIN (the SX1276 datasheet refers to it as Chip Select).

gordonklaus:
@andrzej-kaczmarek There was one other usage of SPI_0_MASTER_SS_PIN: apps/spitest.  It's a bit odd that it now references LORA_SPI_CS_PIN.  Suggestions?

gordonklaus:
@wes3 That sounds sensible.

What about LORA_SPI_CS_PIN, should this be in any of the BSPs other than telee02?

wes3:
Gordon: based on the conversation you had with andrzej, I think the CS_PIN definition belongs in hw/drivers/lora/sx1276/syscfg.yml. I had thought you had moved it there but I did not see it in the commits (might have missed it). Overriding it in the BSP makes sense to me, or we can leave it to the target, but I think it fine if their is an override for it in the BSP. It can always be overridden in the target for folks who want to change the pin.

As a minor nit, I think renaming this to be SX1276_CS_PIN might be a bit better as this chip could be used for non-lora purposes. Fine by me if you do not want to change the name. Well, I guess I would only do this if that definition is agreed to be moved into the sx1276 dir.

andrzej-kaczmarek:
@wes3 yes, this is exactly what I meant. this pin should be defined by a package which "owns" it so in our cases it's either lora or spitest and the name of syscfg val should be selected by this package.

also I agree in case of lora pin this can be overriden in bsp which has it connected to some well known pin.

gordonklaus:
Sounds good.  I'll make the changes when I'm back in the office tomorrow.

gordonklaus:
Done.  Please take another look.

gordonklaus:
@wes3 Ok, I submitted PRs for those repos.

PR #576, opened 2017-09-26
[MYNEWT-783] nimble/phy: Fix T_IFS on coded phy

andrzej-kaczmarek:
This changes T_IFS control from hardware to software. This is required in order to have proper T_IFS on LE Coded PHY since now we're ~16 usecs over the limit which means we can't maintain connection with other devices (except for another Mynewt).

Basically what's need to be done is to schedule TXEN/RXEN using TIMER0 exactly 150 usecs after end of TX/RX which is stored in CC[2]. It's a bit more complicated though as we need to take into account various delays in TX/RX pipeline which are not specified in nRF documentation so they were measured instead - code used to do this is also included as I think it can be useful for debugging purposes later.

The result is the same T_IFS on all phys: 150 +/- 1 usecs.

In addition, code to calculate wfr timer is upgraded to be more precise by including pipeline delays.

### DIFF
--- a/hw/drivers/nimble/nrf52/src/ble_phy.c
+++ b/hw/drivers/nimble/nrf52/src/ble_phy.c
@@ -24,6 +24,7 @@
 #include "os/os.h"
 #include "ble/xcvr.h"
 #include "bsp/cmsis_nvic.h"
+#include "hal/hal_gpio.h"
 #include "nimble/ble.h"
 #include "nimble/nimble_opt.h"
 #include "controller/ble_phy.h"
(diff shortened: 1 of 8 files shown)


wes3:
Looked through the changes and it seems good to me. Nothing that has to change so you can go ahead and merge it if you like.

andrzej-kaczmarek:
updated commit messages...

hw/drivers/nimble/nrf52/src/ble_phy.c
```diff
@@ -1009,6 +1010,59 @@ ble_phy_isr(void)
...
+                        ((pin & 0x1F) << GPIOTE_CONFIG_PSEL_Pos) |
+                        ((pin > 31) << GPIOTE_CONFIG_PORT_Pos);
+}
+#endif
+
+static void
+ble_phy_dbg_time_setup(void)
+{
+#if MYNEWT_VAL(BLE_PHY_DBG_TIME_TXRXEN_READY_PIN) >= 0
+    ble_phy_dbg_time_setup_gpiote(0,
```
rymanluk: It is disabled for now but as discussed instead of hardcoded `index`,  hal_gpio_irq_init should be used in order to not overwrite application

andrzej-kaczmarek: I agree, but the problem here is that while hal_gpio_irq_init() can solve GPIOTE allocation problem, I can't get config index to setup PPI. so as an intermediate solution I'll just use last 3 configs (instead of first 3) so it will be easier to avoid collision. and will add an appropriate note.

later we can consider addin some private gpio api to resolve this in a proper way.

andrzej-kaczmarek:
I will merge this PR now and the remaining open points will be fixed in separate PR(s).

PR #577, opened 2017-09-27
MYNEWT-840 bmp280 driver and shell and sensor framework support

vrahane:
- BMP280 Sensor driver
- BMP280 Sensor specific shell
- Sensor framework support


wes3:
Approving this PR

PR #578, opened 2017-09-27
Update TinyCrypt to 0.2.8

sjanc:


sjanc:
yeah, I simply implemented default_CSPRNG() inside ble_sm_alg.c but maybe calling uECC_set_rng() is in fact better.  I'll send PR for this.

As for g_rng_function, I think this should be first done in upstream TC and then we can backport.

mkiiskila:
I opened a PR against tinycrypt for g_rng_function consolidation.

PR #581, opened 2017-09-27
Mesh fixes related to local interface

rymanluk:


rymanluk:
@jhedberg  FYI

jhedberg:
Yep, looks good to me as well

PR #583, opened 2017-09-28
Mesh fixes and improvements

michal-narajowski:


michal-narajowski:
@jhedberg FYI

net/nimble/host/mesh/src/cfg.c
```diff
@@ -2913,6 +2913,8 @@ static void heartbeat_sub_set(struct bt_mesh_model *model,
 
 	if (period_ms) {
 		cfg->hb_sub.expiry = k_uptime_get() + period_ms;
+	} else {
+		cfg->hb_sub.expiry = k_uptime_get();
 	}
```
sjanc: you could just do
cfg->hb_sub.expiry = k_uptime_get() + period_ms;

without if-else

rymanluk: shouldn't we set cfg->hb_sub.expiry = 0; instead?
@jhedberg  fyi

apps/blemesh/src/main.c
```diff
@@ -90,7 +92,7 @@ fault_get_reg(struct bt_mesh_model *model,
     *test_id = 0x00;
 
     if (has_reg_fault) {
-        uint8_t reg_faults[24] = { [0 ... 23] = 0xff };
+        uint8_t reg_faults[FAULT_ARR_SIZE] = { [0 ... FAULT_ARR_SIZE-1] = 0xff };
```
rymanluk: Can we squash this patch with previous patch adding this into blemesh?

net/nimble/host/mesh/src/health.c
```diff
@@ -99,6 +99,7 @@ static size_t health_get_current(struct bt_mesh_model *mod,
 			fault_count = 0;
 		} else {
 			sys_put_le16(company_id, company_ptr);
+			net_buf_simple_add(msg, fault_count);
```
rymanluk: Could you please extend commit msg by explanation why we need that?
if i undertstand it correctly fault_get_cur() gives application a pointer to buffer where faults are written. If it succeed we need to update our os_mbuf with with len of faults array written by application.

By the way, hope application does not exceed maximum faults array size.
Maybe it would be better to have os_mbuf (net_buf) in the callback instead of *u8_t?
@jhedberg  ^^

rymanluk: Ah I see that &fault_count in the callback contains maximum bytes which can be written to the buffer but still, worth to consider passing os_mbuf in the callback.

jhedberg: I think I didn't go with a net_buf since net_buf_tailroom() wouldn't actually give you how many faults you can add to the buffer, since there needs to be space for more mesh-internal data after that (in particular the MIC).

jhedberg:
As a general comment, could you try to keep app changes separate from stack changes. That makes it easier to port the fixes back to Zephyr.

net/nimble/host/mesh/src/cfg.c
jhedberg: Yes, expiry = 0 would be more appropriate

jhedberg: Also note that this is not about invalid parameters. Messages with invalid parameters should not yield any reaction at all by the implementation. Instead, this particular code path is about disabling the subscription.

net/nimble/host/mesh/src/cfg.c
```diff
@@ -2551,14 +2551,22 @@ static void lpn_timeout_get(struct bt_mesh_model *model,
 	BT_DBG("net_idx 0x%04x app_idx 0x%04x src 0x%04x lpn_addr 0x%02x",
 	       ctx->net_idx, ctx->app_idx, ctx->addr, lpn_addr);
 
+	/* check if it's the address of the Low Power Node? */
+	if (!BT_MESH_ADDR_IS_UNICAST(lpn_addr)) {
+		BT_WARN("invalid LPNAddress; ignoring msg");
+		goto done;
+	}
```
jhedberg: Friend support is far from complete, which is why this was never implemented. Once the support is in place then we should be doing an actual LPN lookup here. Checking first against non-unicast is probably still fine since it means we don't try an unnecessary lookup, and as the commit message states helps pass one more CFG test case.

michal-narajowski:
> Do I understand right, that "mesh: fix setting period div right after receiving it" fixes a regression that was introduced by your earlier commit that moved the publish callback later in mod_publish()? I.e. that patch fixed on thing (accurate period timing) but broke another (updating the period).

Yes, I should have stated that in the commit message. I will fix that and break that patch into two separate.

net/nimble/host/mesh/src/health.c
```diff
@@ -319,6 +319,7 @@ static void health_period_set_unrel(struct bt_mesh_model *model,
 	BT_DBG("period %u", period);
 
 	srv->period = period;
+	model->pub->period_div = srv->period;
```
jhedberg: I went with a different approach for this: the health server context period divider tracking (srv->period) is in practice redundant since the same information is already stored in the generic model publication context (model->pub->period_div). Removing the health-model specific value and just using the generic one solves the problem equally well. I'll provide a reference to my Zephyr patch here as soon as I upload it.

jhedberg:
https://github.com/zephyrproject-rtos/zephyr/pull/4102

michal-narajowski:
Updated the PR. I synchronized it with zephyrproject-rtos/zephyr#4102

PR #585, opened 2017-09-28
nimble/host: Allow setting adv data when advertising

andrzej-kaczmarek:
Not sure why this was not allowed, but it is perfectly ok to do so and
controller will update data while advertising. And even if for some
reason it won't, there will be proper error returned.


andrzej-kaczmarek:
@ccollins476ad I see that you added these checks in 3890be32 so let me know if there was any issue with this as updating advertising data while advertising works perfectly fine for me.

PR #586, opened 2017-09-28
MYNEWT-844 ms5837-02ba01 driver and sensor framework support

vrahane:
- Add sensor framework support and driver  for ms5837-02ba01


wes3:
Approving this PR

PR #605, opened 2017-10-06
nimble/ll: Fix scheduling of extended advertising events

andrzej-kaczmarek:
see commit message for details

### DIFF
--- a/net/nimble/controller/src/ble_ll_adv.c
+++ b/net/nimble/controller/src/ble_ll_adv.c
@@ -103,6 +103,7 @@ struct ble_ll_adv_sm
 #if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)
     uint32_t adv_secondary_start_time;
     struct ble_ll_sched_item adv_secondary_sch;
+    struct os_event adv_sec_txdone_ev;
     uint16_t duration;
     uint16_t adi;
     uint8_t adv_secondary_chan;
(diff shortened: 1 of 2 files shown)


net/nimble/controller/src/ble_ll_adv.c
```diff
@@ -118,13 +119,27 @@ struct ble_ll_adv_sm
 #define BLE_LL_ADV_SM_FLAG_RX_ADD           0x02
 #define BLE_LL_ADV_SM_FLAG_SCAN_REQ_NOTIF   0x04
 #define BLE_LL_ADV_SM_FLAG_CONN_RSP_TXD     0x08
-#define BLE_LL_ADV_SM_FLAG_SECONDARY        0x10
+#define BLE_LL_ADV_SM_FLAG_ACTIVE_CHANSET   0x30 /* use helpers to get/set */
```
sjanc: This flag is 2bit wide so put comment here :)  or maybe just name it mask

net/nimble/controller/src/ble_ll_adv.c
```diff
@@ -118,13 +119,27 @@ struct ble_ll_adv_sm
...
+
+/* active channels set is 1 for primary, 2 for secondary and 0 for none */
+static inline int
+ble_ll_adv_active_chanset_get(struct ble_ll_adv_sm *advsm)
```
sjanc: if you have separate helper for testing primary and secondary you could avoid magic 1 and 2 in code

net/nimble/controller/src/ble_ll_adv.c
```diff
@@ -1230,11 +1282,8 @@ ble_ll_adv_sm_stop(struct ble_ll_adv_sm *advsm)
     if (advsm->adv_enabled) {
         /* Remove any scheduled advertising items */
         ble_ll_sched_rmv_elem(&advsm->adv_sch);
-
 #if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)
-        if (!(advsm->props & BLE_HCI_LE_SET_EXT_ADV_PROP_LEGACY)) {
-            ble_ll_sched_rmv_elem(&advsm->adv_secondary_sch);
-        }
+        ble_ll_sched_rmv_elem(&advsm->adv_secondary_sch);
```
sjanc: isn't this check needed?

andrzej-kaczmarek: for legacy we won't have anything scheduled on aux so this call won't do anything - the result will be the same with and without flag so just removed it since I use similar "logic" in other places

net/nimble/controller/src/ble_ll_adv.c
andrzej-kaczmarek: makes sense, will do

net/nimble/controller/src/ble_ll_adv.c
andrzej-kaczmarek: makes sense, will do

PR #609, opened 2017-10-10
Mesh improvements

michal-narajowski:


sjanc:
@jhedberg FYI

net/nimble/host/mesh/src/cfg.c
```diff
@@ -1742,11 +1742,11 @@ static void mod_sub_va_overwrite(struct bt_mesh_model *model,
 	elem_addr = net_buf_simple_pull_le16(buf);
 	label_uuid = buf->om_data;
 	net_buf_simple_pull(buf, 16);
-
-	BT_DBG("elem_addr 0x%04x", elem_addr);
-
 	mod_id = buf->om_data;
 
+	BT_DBG("elem_addr 0x%04x, addr %s, mod_id 0x%x", elem_addr,
+	       bt_hex(label_uuid, 16), bt_hex(mod_id, buf->om_len));
```
jhedberg: It seems the mod_id gets printed in the wrong endianess here. The access/model layer is little endian, but the first byte after '0x' should be the most significant one. If you don't want to bother with conversions and dealing with the two possible lengths, you could just omit the 0x and have a dump of the raw bytes, I suppose.

jhedberg: Actually, even the format specifier %x is wrong for bt_hex()

michal-narajowski:
Fixed the formatting issue

PR #613, opened 2017-10-11
BLE Host - Honor controller's flow control

ccollins476ad:
Prior to this commit, the host ignored any "Number of Completed Packets" events from the controller.  If the host had any data to send, it would always send it, even if the controller had no free buffers.

Now, the host only sends an ACL packet if the controller has a buffer to hold it:
* When the controller cannot accept more data, the host enqueues outgoing data to the appropriate connection's transmit queue.

* If the host has queued data and the controller indicates that it has freed up buffers, the host attempts to send all its queued data.
* If a disconnect occurs, all outstanding packets associated with the connection are considered flushed.

### DIFF
--- a/kernel/os/include/os/os_mbuf.h
+++ b/kernel/os/include/os/os_mbuf.h
@@ -285,8 +285,7 @@ int os_mbuf_free(struct os_mbuf *mb);
 int os_mbuf_free_chain(struct os_mbuf *om);
 
 void os_mbuf_adj(struct os_mbuf *mp, int req_len);
-int os_mbuf_cmpf(const struct os_mbuf *om, int off, const void *data,
-                   int len);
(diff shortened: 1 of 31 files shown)


andrzej-kaczmarek:
overall new code looks good to me, however there is one more change required for this to work properly with any controller: hci_le_read_buffer_size may return 0 for either max data length or max num of packets and in this case proper value can be read using hci_read_buffer_size command (basically it can happen for dual mode controllers). this also implies that packet counters should be uint16_t since this parameter is 2 octets long in cc for hci_read_buffer_size .

ccollins476ad:
Thanks Andrzej, that is a good point.  I have added code to read the BR/EDR buffer size if the LE buffer size is 0.

PR #618, opened 2017-10-15
Removing use of undefined TASKS_START behavior

seryoni:
Tested using the bleprph app. Seems to be working the same as before. Happy to help testing further!

The problem was that the behavior of the RNG TASKS_START is defined to be write only. Reading from it is undefined and though it might have worked up until now doesn't mean it'll constantly work. I was only able to catch on to this using an emulator.


andrzej-kaczmarek:
lgtm.

seems like we also have the same issue in nrf51 - could you please also make the same change in nrf51 code so both are fixed?

PR #622, opened 2017-10-17
nimble: Some fixes for scheduling aux tx/rx

andrzej-kaczmarek:
We were few usecs off when sheduling aux on both tx and rx - now it is almost perfect.


hw/drivers/nimble/nrf52/src/ble_phy.c
```diff
@@ -918,12 +918,10 @@ ble_phy_rx_start_isr(void)
...
-    /*
-     * Calculate receive start time.
-     *
-     * XXX: possibly use other routine with remainder!
-     */
-    usecs = NRF_TIMER0->CC[1] - ble_phy_mode_pdu_start_off(g_ble_phy_data.phy_cur_phy_mode);
+    /* Calculate accurate packets start time (with remainder) */
+    usecs = NRF_TIMER0->CC[1];
+    usecs -= ble_phy_mode_pdu_start_off(ble_hdr->rxinfo.phy_mode);
+    usecs -= g_ble_phy_t_rxaddrdelay[ble_hdr->rxinfo.phy_mode];
```
sjanc: just wondering if this will not wrap around when subtracting.

rymanluk: imho even so, code in the controller will handle it correctly.

rymanluk:
looks good to me!

PR #623, opened 2017-10-17
 Removing use of undefined TASKS_START behavior

seryoni:
Redoing this pull request to add the nRF51 as well.

Tested using the bleprph app. The problem was that the behavior of the RNG TASKS_START is defined to be write only. Reading from it is undefined and though it might have worked up until now doesn't mean it'll constantly work. I was only able to catch on to this using an emulator.

Please note I don't have an nRF51, but i tested it on an emulator.


andrzej-kaczmarek:
Note: I'll merge this squashed so there is no additional merge commit.

Just a hint for next contributions:
If your PR needs updates, you can just add/modify existing commits on your local branch and do a force push to github branch - github will update PR with new contents. Also there is no need to do PR in your local repository in order to update code to latest master - just do rebase before pushing to github.

PR #630, opened 2017-10-18
LP5523: LED Driver

vsyn:
I tried to stick to the data sheet nomenclature, looking back, I'm not sure that was the best way to go as it isn't very consistent, any thoughts?

### DIFF
--- a/hw/drivers/lp5523/include/lp5523/lp5523.h
+++ b/hw/drivers/lp5523/include/lp5523/lp5523.h
@@ -1,0 +1,1256 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef __LP5523_H__
+#define __LP5523_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <sensor/sensor.h>
+
+#define LP5523_I2C_BASE_ADDR (0x32)
+
+/* Engine control mask */
+#define LP5523_ENGINE3_MASK (0x03)
+#define LP5523_ENGINE2_MASK (0x0c)
+#define LP5523_ENGINE1_MASK (0x30)
+
+/* Engine IDs */
+#define LP5523_ENGINE3 (3)
+#define LP5523_ENGINE2 (2)
+#define LP5523_ENGINE1 (1)
+
+/* LED IDs */
+#define LP5523_LED9 (9)
+#define LP5523_LED8 (8)
+#define LP5523_LED7 (7)
+#define LP5523_LED6 (6)
+#define LP5523_LED5 (5)
+#define LP5523_LED4 (4)
+#define LP5523_LED3 (3)
+#define LP5523_LED2 (2)
+#define LP5523_LED1 (1)
+
+struct lp5523_cfg {
+    /* The 2 LSBs of this represent ASEL1 and ASEL0 */
+    uint8_t asel;
+    uint8_t clk_det_en;
+    uint8_t int_clk_en;
+    uint8_t cp_mode;
+    uint8_t int_conf;
+};
+
+struct lp5523 {
+    struct os_dev dev;
+    struct sensor sensor;
+    struct lp5523_cfg cfg;
+};
+
+/* register addresses */
+
+enum lp5523_bitfield_registers {
+    LP5523_OUTPUT_RATIOMETRIC = 0x02,
+    LP5523_OUTPUT_CONTROL = 0x04,
+    LP5523_ENG_MAPPING = 0x70
+};
+
+enum lp5523_output_registers {
+    LP5523_CONTROL = 0x06,
+    LP5523_PWM = 0x16,
+    LP5523_CURRENT_CONTROL = 0x26
+};
+
+enum lp5523_engine_registers {
+    LP5523_ENGINE_PC = 0x37,
+    LP5523_ENGINE_VARIABLE_A = 0x45,
+    LP5523_MASTER_FADER = 0x48,
+    LP5523_ENG_PROG_START_ADDR = 0x4c
+};
+
+enum lp5523_engine_control_registers {
+    LP5523_ENGINE_CNTRL1 = 0x00,
+    LP5523_ENGINE_CNTRL2 = 0x01
+};
+
+enum lp5523_registers {
+    LP5523_ENABLE = 0x00,
+    LP5523_MISC = 0x36,
+    LP5523_STATUS = 0x3a,
+    LP5523_INTERRUPT = 0x3a,
+    LP5523_INT = 0x3b,
+    LP5523_VARIABLE = 0x3c,
+    LP5523_RESET = 0x3d,
+    LP5523_TEMP_ADC_CONTROL = 0x3e,
+    LP5523_TEMPERATURE_READ = 0x3f,
+    LP5523_TEMPERATURE_WRITE = 0x40,
+    LP5523_LED_TEST_CONTROL = 0x41,
+    LP5523_LED_TEST_ADC = 0x42,
+    LP5523_PROG_MEM_PAGE_SEL = 0x4f,
+    LP5523_PROGRAM_MEMORY = 0x50,
+    LP5523_GAIN_CHANGE_CTRL = 0x76
+};
+
+/* registers */
+struct lp5523_register_value {
+    uint8_t reg;
+    uint8_t pos;
+    uint8_t mask;
+};
+
+#define LP5523_REGISTER_VALUE(r, n, p, m) \
+static const struct lp5523_register_value n = { .reg = r, .pos = p, .mask = m }
+
+/* control1 */
+LP5523_REGISTER_VALUE(LP5523_ENABLE, LP5523_CHIP_EN, 6, 0x40);
+
+#define LP5523_ENGINES_HOLD (0x00)
+#define LP5523_ENGINES_STEP (0x15)
+#define LP5523_ENGINES_FREE_RUN (0x2a)
+#define LP5523_ENGINES_EXECUTE_ONCE (0x3f)
+
+/* control2 */
+#define LP5523_ENGINES_DISABLED (0x00)
+#define LP5523_ENGINES_LOAD_PROGRAM (0x15)
+#define LP5523_ENGINES_RUN_PROGRAM (0x2a)
+#define LP5523_ENGINES_HALT (0x3f)
+
+/* output control */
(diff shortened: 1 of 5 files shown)


hw/drivers/lp5523/include/lp5523/lp5523.h
```diff
@@ -0,0 +1,426 @@
...
+#define LP5523_INS_SUB_IM(target_variable, value) \
+    (0x9200 | \
+    ((target_variable & 0x03) << 10) | \
+    (value & 0xff))
+
+#define LP5523_INS_SUB(target_variable, variable1, variable2) \
+    (0x9310 | \
+    ((target_variable & 0x03) << 10) | \
+    ((variable1 & 0x03) << 2) | \
+    (variable2 & 0x03))
```
vrahane: For the macros above, can you please add #defines for individual bits and register addresses. It would make these macros easier to understand. Also, can you please add some comments for what the macros do. 

For example: 
```
#define LP5523_INS_ADD_IM(target_variable, value) \
    (0x9100 | \
    ((target_variable & 0x03) << 10) | \
    (value & 0xff)) 
```

would be 

```
#define LP5523_INS_ADD_IM(target_variable, value) \
    ( LP5523_OP_ADD_IM | \
    ((target_variable & LP5523_VAR_TARGET) << LP5523_VAR_TARGET_POS) | \
    (value & 0xff))
```

So, you would basically have defines for variables: LP5523_VAR_1, LP5523_VAR_2,  LP5523_VAR_1_POS and LP5523_VAR_2_POS. 

Also, for LED map instructions you could use 0x9c00 as a bitmap : LP5523_INS_LED_MAP, so that the class of the instruction is known. I know generalising for all of them is not possible but LED_MAP and DATA_ARITH seem generalizable.

hw/drivers/lp5523/src/lp5523.c
```diff
@@ -0,0 +1,1144 @@
...
+ *
+ * @param The sensor interface
+ * @param The register address to write to (the MSB register)
+ * @param Value to write
+ *
+ * @return 0 on success, non-zero error on failure.
+ */
+int
+lp5523_set_output_reg(struct sensor_itf *itf, enum lp5523_output_registers addr,
+    uint8_t op, uint8_t value)
```
vrahane: One argument is not commented for here: "op". What if op is 0 ? Is that possible ?

hw/drivers/lp5523/src/lp5523.c
```diff
@@ -0,0 +1,1144 @@
...
+ * @param The register address to read from (the MSB register)
+ * @param Pointer to where the register values should be written
+ *
+ * @return 0 on success, non-zero error on failure.
+ */
+int
+lp5523_get_outout_reg(struct sensor_itf *itf, enum lp5523_output_registers addr,
+    uint8_t op, uint8_t *value)
+{
+    --op;
```
vrahane: One argument is not commented for here: "op". What if op is 0 ? Is that possible ?

hw/drivers/lp5523/src/lp5523.c
```diff
@@ -0,0 +1,1144 @@
...
+ * @param The register address to write to (the MSB register)
+ * @param Value to write
+ *
+ * @return 0 on success, non-zero error on failure.
+ */
+int
+lp5523_set_engine_reg(struct sensor_itf *itf, enum lp5523_engine_registers addr,
+    uint8_t engine, uint8_t value)
+{
+    --engine;
```
vrahane: One argument is not commented for here: "engine". What if engine is 0 ? Is that possible ?

hw/drivers/lp5523/src/lp5523.c
```diff
@@ -0,0 +1,1144 @@
...
+ *
+ * @param The sensor interface
+ * @param The register address to read from (the MSB register)
+ * @param Pointer to where the register values should be written
+ *
+ * @return 0 on success, non-zero error on failure.
+ */
+int
+lp5523_get_engine_reg(struct sensor_itf *itf, enum lp5523_engine_registers addr,
+    uint8_t engine, uint8_t *value)
```
vrahane: One argument is not commented for here: "engine". What if engine is 0 ? Is that possible ?

hw/drivers/lp5523/src/lp5523_shell.c
```diff
@@ -0,0 +1,1017 @@
...
+#if MYNEWT_VAL(LP5523_CLI)
+
+static int lp5523_shell_cmd(int argc, char **argv);
+
+static struct shell_cmd lp5523_shell_cmd_struct = {
+    .sc_cmd = "lp5523",
+    .sc_cmd_func = lp5523_shell_cmd
+};
+
+static struct sensor_itf *lp5523_itf;
```
vrahane: lp5523_itf should be defined here, something like below in hw/drivers/sensors/bno055/src/bno055_shell.c:

```
   45 static struct sensor_itf g_sensor_itf = {
   46     .si_type = MYNEWT_VAL(BNO055_SHELL_ITF_TYPE),
   47     .si_num = MYNEWT_VAL(BNO055_SHELL_ITF_NUM),
   48     .si_addr = MYNEWT_VAL(BNO055_SHELL_ITF_ADDR),
   49 };
```
where BNO055_SHELL_ITF_TYPE, BNO055_SHELL_ITF_NUM and BNO055_SHELL_ITF_ADDR would be syscfgs defined in the drivers syscfg.yml file.

hw/drivers/lp5523/src/lp5523.c
```diff
@@ -0,0 +1,1144 @@
...
+ * @param The sensor interface struct
+ * @param Output number 1-9
+ * @param Mapping 0-3
+ *
+ * @return 0 on success, non-zero error on failure.
+ */
+int
+lp5523_set_op_mapping(struct sensor_itf *itf, uint8_t op, uint8_t mapping)
+{
+    --op;
```
vrahane: What if op is 0 ? Is that possible ? There are multiple places in the code where op is decremented before getting used. It would be nice to cap it at 0 or add #defines for specific values OP would have, it's because these APIs seem to be public.

vrahane:
Which nomenclature are you talking about exactly, the instructions ?

hw/drivers/lp5523/pkg.yml
```diff
@@ -0,0 +1,34 @@
...
+pkg.author: "Julian Ingram <julian@apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+    - led
+    - texas
+    - lp5523
+        
+pkg.deps:
+    - hw/sensor
+
```
vrahane: Here you  should also add : 

    - "@apache-mynewt-core/kernel/os"
    - "@apache-mynewt-core/hw/hal"
    - "@apache-mynewt-core/hw/sensor"

pkg.deps.LP5523_CLI:
    - "@apache-mynewt-core/util/parse"

hw/drivers/lp5523/src/lp5523.c
vsyn: It is possible, but invalid (uint8_t)(0 - 1) > 8 so the function would return -1. The arg is supposed to represent an output number in the datasheet so 1-9 are valid values. Must have missed the comment, will add it in.

hw/drivers/lp5523/src/lp5523.c
vsyn: Same as above but the arg is supposed to represent an engine number in the datasheet so 1-3 are valid values.

hw/drivers/lp5523/src/lp5523.c
vsyn: I am not sure what you mean by "cap it at 0"? I can make them 0-relative if you like, or add defines as you suggested (or both).

vsyn:
The nomenclature I was referring to is some places with CNTRL is used and others where CTRL or CONTROL are used, ENG and ENGINE, TEMP and TEMPERATURE, PROG and PROGRAM etc. You also brought up the 1-relative indexing for engines and outputs above, which is in line with the datasheet, but with the shell module needing to be 1-relative at runtime, the compiler would probably not optimise this to 0-relative for the internal API. It might also be more confusing for some?

hw/drivers/lp5523/src/lp5523_shell.c
vsyn: I agree that I should follow the convention here, but it seems odd that one should have to have 2 different structures in memory with the same constant value (one to pass into `os_dev_create` and another that is declared static in the `*_shell.c` file). Is it the intention to include the `*_shell_init` functions in sysinit at some point in the future?

I would love to see values pulled from the config similar to what you posted above in `sensor_creator.c`. It would them be simple to unify the 2 `sensor_itf` structs and bring the shell initialisation into sysinit while allowing the user to pick their own interface without changing any of the code in the repo, something like:

```
MYNEWT_VAL(BNO055_OFB)
static struct sensor_itf creator_itf_bno = {
    .si_type = MYNEWT_VAL(BNO055_CREATOR_ITF_TYPE),
    .si_num  = MYNEWT_VAL(BNO055_CREATOR_ITF_NUM),
    /* HW I2C address for the BNO055 */
    .si_addr = MYNEWT_VAL(BNO055_CREATOR_ITF_ADDR),
};
#endif

#if MYNEWT_VAL(BNO055_OFB)
    rc = os_dev_create((struct os_dev *) &bno055, "bno055_0",
      OS_DEV_INIT_PRIMARY, 0, bno055_init, (void *)&creator_itf_bno);
    assert(rc == 0);

    rc = config_bno055_sensor();
    assert(rc == 0);

#if MYNEWT_VAL(BNO055_CLI)
    rc = bno055_shell_init(&creator_itf_bno);
    assert(rc == 0);
#endif
#endif
```

hw/drivers/lp5523/src/lp5523.c
vrahane: That sounds fine. Can you please define some valid define values for them. Thank you.

hw/drivers/lp5523/src/lp5523.c
vrahane: Can you please define some valid define values for them.

hw/drivers/lp5523/src/lp5523.c
vrahane: Adding defines sounds like the right approach. By cap it at 0, I meant make 0 look like an invalid value explicitly.

hw/drivers/lp5523/src/lp5523_shell.c
vrahane: The suggestion is great and probably we want to able to do that at some point but sensor device shell can't really be initialized using sysinit. Reason is we we cannot specify arguments in sysinit and that extent of code generation is not our intent. Also, we want the app to initialize the shell only for one instance of the sensor and it's only for experimentation purposes. It is not going to be present in most products.
so, the expected flow of development would be:
- Bring up one sensor using the driver
- Use the shell to talk to it and experiment
- Test all sensor features
- Once tested, plug it into the sensor framework
- Test the sensor using the sensor shell. 
- If you see the data coming out as expected, the driver is good.
The shell is optional and it is not needed most of the times.
Hope I answered your question.

vrahane:
Ah! I see. I don't think that should be an issue as these are driver specific #defines and anything you use is probably fine enough :-). About the outputs and engines, it would be nice if we had some specific #defines.

vsyn:
I think I managed to cover most of your requests, I did not implement a lp5523_priv.h file b/c I can't think of much that I would want to remove from the public API. I also went a little bit further with the #defines than you suggested, I hope that's OK.

hw/drivers/lp5523/include/lp5523/lp5523.h
```diff
@@ -0,0 +1,1255 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <sensor/sensor.h>
+
+#define LP5523_I2C_BASE_ADDR (0x32)
+
+/* Engine control */
```
vsyn: This definitely needs a better name.

PR #631, opened 2017-10-18
removing illegal dependency on external repository

mlaz:


sjanc:
This kind of commits should not be merge without _any_ explanation in commit message.

mlaz:
I am sorry, I thought the original commit message was replicated on this commit's message.

mlaz:
On some previous pull request this two lines got added mistakenly. 
These are not legal due to the fact that they make this BSP depend on mynewt_nordic repository.

PR #632, opened 2017-10-19
BLE Host - Remove BLE_GAP_EVENT_CONN_CANCEL

ccollins476ad:
Cancelled connect procedures are reported as:
* event.type: `BLE_GAP_EVENT_CONNECT`
* event.connect.status: `BLE_HS_EAPP`

Discussion regarding this change can be found here:
https://lists.apache.org/thread.html/6575cbf8faef793381845aaa99bea7c4c053926af43928c82f62e236@%3Cdev.mynewt.apache.org%3E

### DIFF
--- a/net/nimble/host/include/host/ble_gap.h
+++ b/net/nimble/host/include/host/ble_gap.h
@@ -98,7 +98,7 @@ struct hci_conn_update;
 
 #define BLE_GAP_EVENT_CONNECT               0
 #define BLE_GAP_EVENT_DISCONNECT            1
-#define BLE_GAP_EVENT_CONN_CANCEL           2
+/* Reserved                                 2 */
 #define BLE_GAP_EVENT_CONN_UPDATE           3
 #define BLE_GAP_EVENT_CONN_UPDATE_REQ       4
 #define BLE_GAP_EVENT_L2CAP_UPDATE_REQ      5
--- a/apps/bletiny/src/main.c
+++ b/apps/bletiny/src/main.c
@@ -1031,10 +1031,6 @@ bletiny_gap_event(struct ble_gap_event *event, void *arg)
         console_printf("advertising complete.\n");
         return 0;
 
-    case BLE_GAP_EVENT_CONN_CANCEL:
-        console_printf("connection procedure cancelled.\n");
-        return 0;
-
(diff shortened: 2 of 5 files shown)


net/nimble/host/include/host/ble_gap.h
```diff
@@ -98,7 +98,7 @@ struct hci_conn_update;
 
 #define BLE_GAP_EVENT_CONNECT               0
 #define BLE_GAP_EVENT_DISCONNECT            1
-#define BLE_GAP_EVENT_CONN_CANCEL           2
+#define BLE_GAP_EVENT_CONN_CANCEL           2 /* XXX: Deprecated. */
```
andrzej-kaczmarek: We should remove this symbol as well (or rename it to reserved as discussed) since code which uses this event will complie, but won't receive this event anymore which is confusing. Also corresponding structure form GAP event structure should be removed.

Another option is to keep this for one release but we should send both old and new events to retain compatibility.
I'd go with removal and we'll put note about this in release notes. It's not a big change and won't require major changes in existing apps.

ccollins476ad:
Thanks, Andrzej.

I did want to implement the full change, but I chickened out because of the "Backwards Compatibility" section of the release processs: https://cwiki.apache.org/confluence/display/MYNEWT/Release+and+Support+Policy#ReleaseandSupportPolicy-BackwardsCompatibility.  That page indicates that code using the deprecated feature should successfully build for at least six months after the feature was marked as deprecated.

I am thinking you are right - there is not much benefit in keeping the old symbols around.  I don't think it is possible to slowly drop support for this feature.  Any application which actually depends on the `BLE_GAP_EVENT_CONN_CANCEL` event would benefit much more from a build error rather than silently breaking.

I would like to send an email about this to the dev list to see what others think.  I am sure this is a scenario that will come up again, so we should make the appropriate change to the release policy.

ccollins476ad:
I have removed the `BLE_GAP_EVENT_CONN_CANCEL` code entirely now.  I don't believe there is any data in `struct ble_gap_event` that is specific to this event type, so I did not change the struct.

ccollins476ad:
Thanks for reviewing, Andrzej.

PR #637, opened 2017-10-25
LoRa Datarate changes

wes3:
Please read the individual commit messages for more information about the changes. The main change was to remove setting the data rate in the port.

### DIFF
--- a/apps/lora_app_shell/src/las_cmd.c
+++ b/apps/lora_app_shell/src/las_cmd.c
@@ -713,7 +713,6 @@ las_cmd_app_port(int argc, char **argv)
 {
     int rc;
     uint8_t port;
-    uint8_t dr;
     uint8_t retries;
 
     if (argc < 3) {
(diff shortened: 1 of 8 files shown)


apps/lora_app_shell/src/las_cmd.c
```diff
@@ -743,27 +742,21 @@ las_cmd_app_port(int argc, char **argv)
...
             console_printf("Invalid # of arguments.\n");
             goto cmd_app_port_err;
         }
-        dr = parse_ull_bounds(argv[2], 0, LORAMAC_TX_MAX_DATARATE, &rc);
-        if (rc) {
-            console_printf("Invalid data rate. Must be between 0 and "
-                           "%d (inclusve)\n", LORAMAC_TX_MAX_DATARATE);
-            return 0;
-        }
         retries = parse_ull_bounds(argv[2], 1, MAX_ACK_RETRIES, &rc);
```
mkiiskila: I suspect you meant argv[3] here, not argv[2]?

wes3: Yes indeed. Good catch. I will fix it directly.

PR #639, opened 2017-10-31
BSP MIPS Ci40: brought in the MIPS HAL, fixes startup

IMGJulian:
Updates the MIPS port to the newer style startup method.


hw/bsp/ci40/src/arch/mips/abiflags.S
```diff
@@ -0,0 +1,82 @@
+/*
```
mkiiskila: Do you have license blurb for this file?

IMGJulian: Fixed to be the same as the others, thanks.

PR #641, opened 2017-10-31
mesh: Fix and clean up iv update procedure

michal-narajowski:
This patch fixes issue when receiving iv index greater than current
index + 42 in update mode. According to Specification when node is in update
state it should only accept iv index equal to the current iv index. When
node is in normal mode it should ignore index that is greater than current
index + 42.

This allows to pass MESH/NODE/IVU/BI-02-C.

Also this patch cleans up the iv update procedure, to make it easier to
read.

### DIFF
--- a/net/nimble/host/mesh/src/net.c
+++ b/net/nimble/host/mesh/src/net.c
@@ -556,35 +556,50 @@ void bt_mesh_iv_update(u32_t iv_index, bool iv_update)
 {
 	int i;
 
-	if (iv_index < bt_mesh.iv_index || iv_index > bt_mesh.iv_index + 42) {
-		BT_ERR("IV Index completely out of sync: 0x%08x != 0x%08x",
-			iv_index, bt_mesh.iv_index);
-		return;
-	}
-
-	if (iv_index > bt_mesh.iv_index + 1) {
-		BT_WARN("Performing IV Index Recovery");
-		memset(bt_mesh.rpl, 0, sizeof(bt_mesh.rpl));
-		bt_mesh.iv_index = iv_index;
-		bt_mesh.seq = 0;
-		goto do_update;
-	}
-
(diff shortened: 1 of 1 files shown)


michal-narajowski:
@jhedberg FYI

net/nimble/host/mesh/src/net.c
```diff
@@ -556,40 +556,57 @@ void bt_mesh_iv_update(u32_t iv_index, bool iv_update)
 {
 	int i;
 
-	if (iv_index < bt_mesh.iv_index || iv_index > bt_mesh.iv_index + 42) {
-		BT_ERR("IV Index completely out of sync: 0x%08x != 0x%08x",
-			iv_index, bt_mesh.iv_index);
-		return;
-	}
+	/* If we are in normal*/
+	if (!bt_mesh.iv_update) {
```
jhedberg: Looks fine to me, but I'd reverse the if-statement to avoid having to do a negative, i.e. the first branch would be if (bt_mesh.iv_update), which is actually what the code already has from before. I've got another pending patch related to Friend support that modifies bt_mesh_iv_update() to return a bool to indicate if the state was changed or not, so I'll be doing the Zephyr port of this patch on top of my change.

net/nimble/host/mesh/src/net.c
```diff
@@ -556,40 +556,57 @@ void bt_mesh_iv_update(u32_t iv_index, bool iv_update)
...
```
jhedberg: I think you should add an initial check for (iv_index == bt_mesh.iv_index) here, since otherwise you'll get confusing errors for perfectly normal Secure Network Beacons which have the same IV Index as you.

michal-narajowski: Which errors are you referring to?

michal-narajowski:
Updates according to @jhedberg's comments.

PR #642, opened 2017-10-31
apps/blemesh: Add health fault handling and vendor model for PTS testing

michal-narajowski:


michal-narajowski:
Updated descriptions.

@jhedberg FYI

PR #650, opened 2017-11-08
sensors/bma253: Couple of improvements and fixes

rymanluk:


rymanluk:
Let's close it for now since we are working on other branch where more fixes will come to bma253

PR #653, opened 2017-11-09
Add new semaphore API + some nimble/os fixes

andrzej-kaczmarek:
Main addition is new os_sem_get_count() API - it is then used in nimble where applicable.

+ some random fixes inside nimble and os code.


andrzej-kaczmarek:
Thanks, Chris!

PR #655, opened 2017-11-10
Support for APOLLO 2 and emspi HCI transport

ccollins476ad:
This PR adds support for the Ambiq APOLLO 2 MCU and the apollo2_evb BSP.

In addition, it contains an implementation of a new HCI transport: emspi.  This is HCI over SPI, using EM's SPI protocol (as defined in the EM9304 datasheet: http://www.emmicroelectronic.com/sites/default/files/public/products/datasheets/9304-ds.pdf).

There are quite a few pieces still missing:
* Tickless idle task
* Additional timers
* Watchdog
* SPI slave mode
* I2C drivers
* Power management


apps/bleprph-emspi/pkg.yml
```diff
@@ -0,0 +1,44 @@
...
+    - kernel/os 
+    - mgmt/imgmgr
+    - mgmt/newtmgr
+    - mgmt/newtmgr/transport/ble
+    - net/nimble/host
+    - net/nimble/host/services/ans
+    - net/nimble/host/services/gap
+    - net/nimble/host/services/gatt
+    - net/nimble/host/store/config
+    - net/nimble/transport/emspi
```
rymanluk: Maybe instead of creating new bleperh app we could add dependency here so transport/emspi is used on apollo mcu ? Does it make sense?

utzig:
Very nice! Stupid question: you added a new app because the existing BLE apps would not work over this new HCI transport?

apps/bleprph-emspi/pkg.yml
andrzej-kaczmarek: I'd consider adding kind of metapackage for hci transport which only has syscfg settings to select hci transport. based on this it adds dependency to transport package with and actual implementation so basically we can run any app on any transport by proper target configuration.

net/nimble/transport/emspi/src/ble_hci_emspi.c
```diff
@@ -0,0 +1,935 @@
...
+    am_hal_gpio_int_clear(rdy_mask);
+
+    /* Deassert slave select. */
+    hal_gpio_write(MYNEWT_VAL(BLE_HCI_EMSPI_SS_PIN), 1);
+
+    OS_ENTER_CRITICAL(sr);
+    hal_gpio_irq_enable(MYNEWT_VAL(BLE_HCI_EMSPI_RDY_PIN));
+
+    if (hal_gpio_read(MYNEWT_VAL(BLE_HCI_EMSPI_RDY_PIN))) {
+        am_hal_gpio_int_set(rdy_mask);
```
andrzej-kaczmarek: I may be missing something here but I don't think you need to force interrupt here which means this does not need to have dependency on Apollo2 SDK (which imo should be avoided here since this is nice generic hci transport).

Basically, if rdy pin is asserted here you can simply post an event to queue just like you do in an interrupt handler. This means that hal_gpio_irq_enabled() would need to clear interrupt before it is enabled, but I believe this is what other bsp-s do (like nordic ones).

wes3: In defense of Chris, when we were discussing this I had mentioned that this would be one way to do it without realizing at the time that there was no generic way to set the interrupt pending flag (meaning no hal to do it). Given that the interrupt only posts an event I think your suggestion is a good one Andrzej; just post the event.

I do not understand why hal_gpio_irq_enabled() would need to clear the interrupt before it is enabled however. There would be no harm in that interrupt firing off (waste some time maybe). Personally, (just my opinion), I would prefer a hal that did not automatically clear a pending interrupt flag in the enable routine. The hal already provides a way to clear the interrupt. I can see cases where you would not want hal_gpio_irq_enabled() to automatically clear a pending interrupt. I just feel it is more flexible to have separate routines for enabling interrupts and clearing them (seems a bit more flexible to me). This way the developer can do what they want (clear it or not prior to enabling).

apps/bleprph-emspi/pkg.yml
ccollins476ad: I agree - it is kind of ludicrous to create an entirely new app just to change the transport.

I don't know if making the MCU dependent on the emspi transport would work.  The problem there is it prevents someone from using a different transport with this MCU (e.g., ram or uart).

The metapackage idea is not bad.  It would need to know about every transport, and therefore get updated whenever a transport is introduced or changed, but that is probably unavoidable regardless of how we solve this.

Another idea is to just leave it up to the target to bring in the transport dependency.  For combined host-controller, the target would depend on:
```
    - net/nimble/controller
    - net/nimble/transport/ram
```

For emspi, it would just depend on:
```
    - net/nimble/transport/emspi
```

...and so on.  This would put a lot more responsibility on the target.  I'm not sure if that is a good thing or not.  On one hand, it makes sense, because the target remains just a container for dependencies and settings.  On the other hand, it feels like we are just using the target as a convenient catch all for things like this.

The metapackage might make a lot more sense.  I probably need to think about it more or come up with an example.

PR #659, opened 2017-11-14
gdb macro os_wakeups, which lists scheduled os_callouts and sleeping tasks

mkiiskila:


compiler/gdbmacros/os.gdb
```diff
@@ -34,5 +34,40 @@ end
...
 usage: os_tasks
-
 Displays os tasks
+end
+
+define os_callouts
+	printf "Callouts:\n"
+	printf " %8s %10s %10s\n", "tick", "callout", "func"
+	while $c != 0
+		printf " %8d %10p %10p\n", $c->c_ticks, $c, $c->c_ev.ev_cb
```
andrzej-kaczmarek: It would be nice to print symbol name for ev_cb in addition to pointer value, but I don't know how and if this is possible.

mkiiskila: Indeed it would be nice. Alas, AFAIK there's no such gdb printf formatting character, so I'd have to interleave this with 'print', which wants to include additional things in the output.

So I opted for this for now, I left the address as the last thing in format, so we can change it easily later.

PR #661, opened 2017-11-16
OS minimal stack check at context switch time

mkiiskila:


kernel/os/syscfg.yml
```diff
@@ -39,6 +39,12 @@ syscfg.defs:
     OS_SCHEDULING:
         description: 'Whether OS will be started or not'
         value: 1
+    OS_CTX_SW_STACK_CHECK:
+        description: 'Whether to do stack sanity check during context switch'
+        value: 1
```
andrzej-kaczmarek: I'd make this disabled by default since it decreases effective stack size so existing apps may break.

mkiiskila:
Moved dup free check and memory block element checks behind a new syscfg variable.
Also added memory pool poisoning.
Fixed two of the issues revealed by unittests when executed with these settings set.

andrzej-kaczmarek:
@ccollins476ad I think before we enable this by default we should have some helper macro to calculate stack size including extra space for stack canaries and make applications use it. Then it makes sense to have it enabled by default for debugging since we'll have the same effective stack size regardless of configuration.

mkiiskila:
We do have a stack size computation macro, actually. And it is used almost everywhere.
I was going to enable a subset of these at least for unittests right away. Not all of them, because
MYNEWT-868 is still open. I think MYNEWT-869 is already fixed.

PR #667, opened 2017-11-19
Apollo2 - Tickless idle and HAL timer implementation

ccollins476ad:


net/nimble/transport/emspi/src/ble_hci_emspi.c
```diff
@@ -174,6 +174,9 @@ ble_hci_emspi_tx_chunk(const uint8_t *data, int len, int *out_bytes_txed)
     uint8_t buf_size;
     int rc;
 
+    /* Silence spurious "may be used uninitialized" warning. */
+    *out_bytes_txed = 0;
+
```
mkiiskila: Not important, but you could remove the line which assigns *out_bytes_txd to zero on line 188 now.

hw/mcu/ambiq/apollo2/src/hal_os_tick.c
```diff
@@ -17,31 +17,118 @@
...
+
+    /* Since the STIMER only uses relative scheduling, all ticks values are
+     * valid.  There is no need to check for wrap around.
+     */
+
+    /* Only set the timer for nonzero tick values.  For values of 0, just let
+     * the timer expire on the next tick, as scheduled earlier.
+     */
+    if (ticks > 0) {
+        apollo2_os_tick_set_timer(ticks);
```
mkiiskila: Should this be ticks * apollo2_os_tick_dur?

hw/mcu/ambiq/apollo2/src/hal_timer.c
```diff
@@ -0,0 +1,786 @@
...
+#endif
+
+    const struct apollo2_timer *bsp_timer;
+
+    bsp_timer = apollo2_timer_resolve(timer_num);
+    if (bsp_timer == NULL) {
+        return 0;
+    }
+
+    return 1000000000 / bsp_timer->freq_hz;
```
mkiiskila: Could check whether hal_timer is enabled here, otherwise this'll generate div0 trap.

hw/mcu/ambiq/apollo2/src/hal_os_tick.c
ccollins476ad: I could be wrong, but I don't think so.  `apollo2_os_tick_set_timer()` multiplies its parameter by `apollo2_os_tick_dur` before setting the output compare.

hw/mcu/ambiq/apollo2/src/hal_timer.c
ccollins476ad: Thanks - I have included this fix in the "Apollo2 - HAL timer implementation." commit.

net/nimble/transport/emspi/src/ble_hci_emspi.c
ccollins476ad: Thanks - I have included this fix in the "emspi transport - Fix compiler warning." commit.

hw/mcu/ambiq/apollo2/src/hal_os_tick.c
mkiiskila: You are right, indeed it does. In that case the argument to apollo2_os_tick_set_timer() at the end of apollo2_os_tick_handler() seems incorrect.

ccollins476ad: Thanks, that is a good catch.  I have change `apollo2_os_tick_handler()` as follows:

```
apollo2_os_tick_set_timer(1);
```

PR #671, opened 2017-11-21
Extended Advertising API

sjanc:
This adds host GAP support for all features provided by Extended Advertising:
 - multiple advertising instances
 - advertising TX power configuration
 - alternate PHYs configuration
 - large advertising data

Applications are now able to control multiple advertising instances and configure them separately.
btshell application is also extended with support for all new functionalities.

### DIFF
--- a/apps/bletiny/src/cmd.c
+++ b/apps/bletiny/src/cmd.c
@@ -412,16 +412,6 @@ bletiny_help_disabled(void)
 }
 #endif
 
-#if MYNEWT_VAL(BLE_EXT_ADV)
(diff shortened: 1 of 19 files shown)


sjanc:
Hi Chris,

thanks for review! I think I've fixed all the issues mentioned, and few other bugs as well that I found while working on your comments:-)

some notes:
in ble_hs_hci_evt_le_adv_set_terminated() I left ble_gap_rx_conn_complete without error check as we need to terminate instance regardless of it (TBH I was thinking on making all HCI event handlers return void as there is nothing we can do on such errors anyway). For clarity, I added comment that this check is omitted on purpose.

PR #686, opened 2017-11-30
BLE Host flow control

ccollins476ad:
(Jira ticket: https://issues.apache.org/jira/browse/MYNEWT-287)

## Overview

Controller to host flow control (hereon called "flow control") stops the controller from sending more ACL data packets than the host can accommodate.  Briefly, this feature is implemented as follows:

* On startup, host tells controller:
    * That flow control is being used.
    * About its ACL data buffer supply:
        * Total number of buffers.
        * Size of each buffer.

* After startup, host sends updates to controller containing: The number of ACL data buffers it has freed since the last update, and which connection handle each buffer was associated with.

* Host provides updates periodically (default: once per second).

* Host sends an immediate update if the number of buffers freed since the last update exceeds a threshold (default: `total_buffers - 2`).

* Host does not send an update if the number of buffers freed since the last update is 0.

* If a connection with unfreed buffers is terminated, the controller can assume that all buffers are freed (i.e., no host update required).

This feature prevents a buffer overrun as long as the controller keeps track of the host's free buffer count.

#### mbuf ownership

Flow control is quite simple as described above.  Unfortunately, there is one feature of the nimble host that complicates things.  When the host tells the application about an incoming data packet, such as a write-attribute command, the application can keep the mbuf for later processing.  If the application keeps the mbuf, then it is the application's responsibility to free the mbuf when it is done with it.  This is a problem because the flow control implementation needs to know when a buffer is freed to properly update the controller.

The solution I chose is to add a callback that gets called when a block is freed back to the ACL buffer pool.  The callback would increment the appropriate freed buffer count, and possibly schedule a host update.  Because the ACL buffer pool is an mbuf pool, there were two ways to implement this:

1. Add the callback to `struct os_mbuf_pool`.
2. Add the callback to the underlying `struct os_mempool`.

I thought this functionality might be useful to others, so I chose the more generic option 2.  First, here is how this is implemented:

```
struct os_mempool_ext;

/**
 * Block put callback function.  If configured, this callback gets executed
 * whenever a block is freed to the corresponding extended mempool.  Note: The
 * os_memblock_put() function calls this callback instead of freeing the block
 * itself.  Therefore, it is the callback's responsibility to free the block
 * via a call to os_memblock_put_no_ext().
 *
 * @param ome                   The extended mempool that a block is being
 *                                  freed back to.
 * @param data                  The block being freed.
 * @param arg                   Optional argument configured along with the
 *                                  callback.
 *
 * @return                      Indicates whether the block was successfully
 *                                  freed.  A non-zero value should only be
 *                                  returned if the block was not successfully
 *                                  released back to its pool.
 */
typedef os_error_t os_mempool_put_fn(struct os_mempool_ext *ome, void *data,
                                     void *arg);

struct os_mempool_ext {
    struct os_mempool mpe_mp;

    /* Callback that is executed immediately when a block is freed. */
    os_mempool_put_fn *mpe_put_cb;
    void *mpe_put_arg;
}
```

And a new `flags` field was added to `struct os_mempool`:

```
    struct os_mempool {
        // ...
        uint8_t mp_flags;           /* Bitmap of OS_MEMPOOL_F_[...] values. */
        // ...
    };

    /**
     * Indicates an extended mempool.  Address can be safely cast to
     * (struct os_mempool_ext *).
     */
    #define OS_MEMPOOL_F_EXT        0x01
```

When a block is freed back to a mempool, `os_memblock_put()` uses the flags field to determine if it should cast to `struct os_mempool_ext *` and call the callbacks.

#### API

Another annoyance is the update to the HCI transport API.  If flow control is enabled, the host needs to configure the transport to use a callback when ACL data mbufs are freed.  There was no way to fit this into the existing configuration API, so an additional function was added:

```
int ble_hci_trans_set_acl_free_cb(os_mempool_put_fn *cb, void *arg)
```

It is unfortunate that the host needs to call two functions to configure the transport instead of one.  Perhaps the transport API can be updated in a future release such that there is only one configuration function, and it accepts a struct of setting values.  (As an aside, I think the transport API should be split into two APIs: host-transport and controller-transport).

### DIFF
--- a/kernel/os/include/os/os_mempool.h
+++ b/kernel/os/include/os/os_mempool.h
@@ -44,10 +44,10 @@ struct os_memblock {
 
 /* Memory pool */
 struct os_mempool {
-    int mp_block_size;          /* Size of the memory blocks, in bytes. */
-    int mp_num_blocks;          /* The number of memory blocks. */
-    int mp_num_free;            /* The number of free blocks left */
-    int mp_min_free;            /* The lowest number of free blocks seen */
+    uint32_t mp_block_size;     /* Size of the memory blocks, in bytes. */
+    uint16_t mp_num_blocks;     /* The number of memory blocks. */
+    uint16_t mp_num_free;       /* The number of free blocks left */
+    uint16_t mp_min_free;       /* The lowest number of free blocks seen */
     uint32_t mp_membuf_addr;    /* Address of memory buffer used by pool */
     STAILQ_ENTRY(os_mempool) mp_list;
     SLIST_HEAD(,os_memblock);   /* Pointer to list of free blocks */
@@ -85,8 +85,8 @@ typedef uint64_t os_membuf_t;
     (sizeof (os_membuf_t) * OS_MEMPOOL_SIZE((n), (blksize)))
 
 /* Initialize a memory pool */
-os_error_t os_mempool_init(struct os_mempool *mp, int blocks, int block_size,
-                           void *membuf, char *name);
+os_error_t os_mempool_init(struct os_mempool *mp, uint16_t blocks,
+                           uint32_t block_size, void *membuf, char *name);
 
 /* Performs an integrity check of the specified mempool. */
 bool os_mempool_is_sane(const struct os_mempool *mp);
--- a/kernel/os/include/os/os_mempool.h
+++ b/kernel/os/include/os/os_mempool.h
@@ -48,10 +48,48 @@ struct os_mempool {
     uint16_t mp_num_blocks;     /* The number of memory blocks. */
     uint16_t mp_num_free;       /* The number of free blocks left */
     uint16_t mp_min_free;       /* The lowest number of free blocks seen */
+    uint8_t mp_flags;           /* Bitmap of OS_MEMPOOL_F_[...] values. */
     uint32_t mp_membuf_addr;    /* Address of memory buffer used by pool */
     STAILQ_ENTRY(os_mempool) mp_list;
     SLIST_HEAD(,os_memblock);   /* Pointer to list of free blocks */
     char *name;                 /* Name for memory block */
+};
+
+/**
+ * Indicates an extended mempool.  Address can be safely cast to
+ * (struct os_mempool_ext *).
+ */
+#define OS_MEMPOOL_F_EXT        0x01
+
+struct os_mempool_ext;
+
+/**
+ * Block put callback function.  If configured, this callback gets executed
+ * whenever a block is freed to the corresponding extended mempool.  Note: The
+ * os_memblock_put() function calls this callback instead of freeing the block
+ * itself.  Therefore, it is the callback's responsibility to free the block
+ * via a call to os_memblock_put_from_cb().
+ *
+ * @param ome                   The extended mempool that a block is being
+ *                                  freed back to.
+ * @param data                  The block being freed.
+ * @param arg                   Optional argument configured along with the
+ *                                  callback.
+ *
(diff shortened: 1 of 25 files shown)


ccollins476ad:
## Overview

Controller to host flow control (hereon called "flow control") stops the controller from sending more ACL data packets than the host can accommodate.  Briefly, this feature is implemented as follows:

* On startup, host tells controller:
    * That flow control is being used.
    * About its ACL data buffer supply:
        * Total number of buffers.
        * Size of each buffer.

* After startup, host sends updates to controller containing: The number of ACL data buffers it has freed since the last update, and which connection handle each buffer was associated with.

* Host provides updates periodically (default: once per second).

* Host sends an immediate update if the number of buffers freed since the last update exceeds a threshold (default: `total_buffers - 2`).

* Host does not send an update if the number of buffers freed since the last update is 0.

* If a connection with unfreed buffers is terminated, the controller can assume that all buffers are freed (i.e., no host update required).

This feature prevents a buffer overrun as long as the controller keeps track of the host's free buffer count.

## Complications

Flow control is quite simple as described above.  Unfortunately, there is one feature of the nimble host that really complicates things.  When the host tells the application about an incoming data packet, such as a write-attribute command, the application can keep the mbuf for later processing.  If the application keeps the mbuf, then it is the application's responsibility to free the mbuf when it is done with it.  This is a problem because the flow control implementation needs to know when a buffer is freed to properly update the controller.

The solution I chose is to add a callback that gets called when a block is freed back to the ACL buffer pool.  The callback would increment the appropriate freed buffer count, and possibly schedule a host update.  Because the ACL buffer pool is an mbuf pool, there were two ways to implement this:

    1. Add the callback to `struct os_mbuf_pool`.
    2. Add the callback to the underlying `struct os_mempool`.

I thought this functionality might be useful to others, so I chose the more generic option 2.  First, here is how this is implemented:

```
    struct os_mempool_ext;
    typedef void os_mempool_put_fn(struct os_mempool_ext *ome, void *data);

    struct os_mempool_ext {
        struct os_mempool mpe_mp;

        /* Callback that is executed immediately before a block is freed. */
        os_mempool_put_fn *mpe_pre_put_cb;

        /* Callback that is executed immediately after a block is freed.  The
         * contents of the block are indeterminate and should not be inspected.
         */
        os_mempool_put_fn *mpe_post_put_cb;
    };
```

And a new `flags` field was added to `struct os_mempool`:

```
    struct os_mempool {
        // ...
        uint8_t mp_flags;           /* Bitmap of OS_MEMPOOL_F_[...] values. */
        // ...
    };

    /**
     * Indicates an extended mempool.  Address can be safely cast to
     * (struct os_mempool_ext *).
     */
    #define OS_MEMPOOL_F_EXT        0x01
```

When a block is freed back to a mempool, the free function uses the flags field to determine if it should cast to `struct os_mempool_ext *` and call the callbacks.

What I don't like about this is the need for two callbacks.  We need two because the user might need to know two separate things:

* The contents of the block being freed, and
* That the block is fully freed and ready to be reallocated.

The contents of the block can only be inspected before it is freed, hence the need for two callbacks.  I just worry that this mechanism is too complicated for people to understand and use.

ccollins476ad:
Thanks for the detailed review, Andrzej.  I agree with all your comments and have pushed fixes to address them.  When you get a chance, could you please take another look?

I pushed the fixes as separate commits to make review easier.  However, I plan on combining them with the original commits before the merge.

andrzej-kaczmarek:
Looks good with fixes!

PR #687, opened 2017-11-30
add flash_area_getnext_sector(), and knob to select a style of emulated flash in sim

mkiiskila:


mkiiskila:
Yes, it should. Thanks for pointing that out.

PR #691, opened 2017-12-04
Fix license issues

utzig:


ccollins476ad:
Looks good to me.  The `apollo2.ld` file uses the same license as the other ld files.  Elsewhere in `LICENSE`, we call these out as a: "modified Tcl/Tk" license.  So I would just add `apollo2.ld` to this list of ld files and remove the TODO note about codesourcery.

utzig:
Fixed the `apollo2.ld` TODO.

PR #693, opened 2017-12-05
Complete Friend support ported from Zephyr

michal-narajowski:
Regression tests passed. I know this is way too big PR, but we need to be in sync with Zephyr code to be able to test further.


michal-narajowski:
Squashed 3 commits that I forgot to squash earlier

michal-narajowski:
Added one small fix and one cleanup for blemesh app.

PR #700, opened 2017-12-14
kernel/os_mempool: Fix for incorrect check on os_mempool_init

rymanluk:
This is regression from abd8c4598b as it was changing API incorretly.
Documentation says that 'block_size' parameter in os_mempool_init() is
"Minimum size of an individual element in pool"

It does not required from the user to take into account of 'block_size'
an internal overhead which is sizeof(struct os_memblock).

The only requirement on user is to make sure that provided 'membuf' has
enough bytes which can be calculated using OS_MEMPOOL_BYTES() macro.

Without this patch we have assert when creating mempool with block_size=4

### DIFF
--- a/kernel/os/src/os_mempool.c
+++ b/kernel/os/src/os_mempool.c
@@ -90,7 +90,7 @@ os_mempool_init(struct os_mempool *mp, int blocks, int block_size,
     struct os_memblock *block_ptr;
 
     /* Check for valid parameters */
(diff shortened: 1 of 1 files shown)


mkiiskila:
We should take into account linkage to freelist, therefore the size should not be smaller than a pointer. Correct fix would've been to change the block_size <= sizeof(struct os_memblock)
to block_size < sizeof(struct os_memblock). Opening PR for that.

PR #703, opened 2017-12-18
sensor: Add missing include

rymanluk:


joshgrob:
I chatted with @rymanluk about this and it's fine.  I wanted to keep it in my PR since not having was breaking my application build, but can update with this change as well.

PR #704, opened 2017-12-18
BLE Host - Don't crash if no store package.

ccollins476ad:
Prior to this change: the host asserted that IRK restoration succeeds at sync time.

After change: If IRK restoration at startup fails because there is no store package present, don't crash.

### DIFF
--- a/net/nimble/host/src/ble_hs.c
+++ b/net/nimble/host/src/ble_hs.c
@@ -313,7 +313,10 @@ ble_hs_sync(void)
 
     if (rc == 0) {
         rc = ble_hs_misc_restore_irks();
-        assert(rc == 0);
+        if (rc != 0) {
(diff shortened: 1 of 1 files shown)


net/nimble/host/src/ble_hs.c
```diff
@@ -313,7 +313,7 @@ ble_hs_sync(void)
 
     if (rc == 0) {
         rc = ble_hs_misc_restore_irks();
-        assert(rc == 0);
```
rymanluk: Maybe we could add some info log saying that storage is not present?

ccollins476ad: Thanks @rymanluk.  I agree that we should log something on failure.  I also think the `assert()` is a bit extreme, so I changed the code to log on failure but never crash.

PR #706, opened 2017-12-19
Continued conversion of mynewt-site docs to core with doxygen integration.

gj262:
Note: these changes aim to fix syntax from the change over and do not address content.


jacobrosenthal:
before these images are added I just wanted to check to make sure everyone is on board

every single newt install needs to pull core, and thus all the binaries (even if we delete them later)

Were going right now from 86.02 MiB to 103.07 MiB in just this PR, then as tutorials change, are added and removed.. that goes up forever from now on



Upstream:
```
$ git count-objects -v -H
count: 0
size: 0 bytes
in-pack: 75195
packs: 1
size-pack: 86.02 MiB
prune-packable: 0
garbage: 0
size-garbage: 0 bytes
```



Pull Request:
```
$ git count-objects -v -H
count: 0
size: 0 bytes
in-pack: 75402
packs: 1
size-pack: 103.07 MiB
prune-packable: 0
garbage: 0
size-garbage: 0 bytes
```

jacobrosenthal:
Related, newt install should probably clone depth 1 and Im not sure that it does right now. Submitted that question [here](https://issues.apache.org/jira/browse/MYNEWT-877)

jacobrosenthal:
Not to hammer this but heres a cleaner metric I think. Assuming newt now or later clones at depth 1 this one PR triples the size of the repo:

```
$ git clone --depth 1  --branch doxygen-fixes-1 git@github.com:gj262/mynewt-core.git
$ cd mynewt-core/
$ git count-objects -v -H
count: 0
size: 0 bytes
in-pack: 4935
packs: 1
size-pack: 27.14 MiB
prune-packable: 0
garbage: 0
size-garbage: 0 bytes
```
```
$ git clone --depth 1 git@github.com:gj262/mynewt-core.git
$ cd mynewt-core/
$ git count-objects -v -H
count: 0
size: 0 bytes
in-pack: 4872
packs: 1
size-pack: 10.13 MiB
prune-packable: 0
garbage: 0
size-garbage: 0 bytes
```

sterlinghughes:
HI @jacobrosenthal --- I definitely think we should clone at depth 1 as your previous PR points out.  We do need to manage upgrade & sync, and make sure we are properly handling the source control.  We avoided depth 1 just to be cautious around that, but with the repo size exploding we probably do need to manage that.

Personally, I think having the documentation inline with the source code makes a lot of sense, because the documentation then tracks the released version of source code, and can be generated.  This makes thing easier to maintain, especially as we break out individual projects.  There have been efforts to break out significant components: 

- https://github.com/runtimeco/mcuboot
- https://github.com/apache/mynewt-nimble
- https://github.com/apache/mynewt-mcumgr
- https://github.com/apache/mynewt-nffs

With the hope of making these independently released and documentable as well.  The idea has been to let those stabilize and work as independent projects before removing them from *-core (as well as improving newt so that i can truly work easily/well with multiple repos.) In my mind, this constitutes Mynewt 2.0.  

In order to move to separately releasable components, it would be nice if the documentation was contained within the source repo (and tracked the release trains.)  But we also need to make sure we evolve the documentation (& add a lot more doxygen for function descriptions) so that we can actually have this mostly working as well.  Thus the thought to begin moving the documentation into the source tree, and get it generating reliably/have it become the source of truth.

I think this may be a topic for dev@ if somebody can move it over.

jacobrosenthal:
I love the idea of documentation being directly at hand. Less so the binary
blobs of the pictures which may or may not be crucial being brought down by
every newt install that will never see them

sterlinghughes:
It's a good point, perhaps we can break out the tutorials / getting started, but keep the core of the documentation in *-core.

gj262:
Could we have `newt install` skip `docs`?

So users do not have to get unnecessary data and contributors still get the
convenience of co-location to maximize the likelihood that docs follow code.

Also, the images are mostly not optimized, so there could be a reduction in
size if we cared to go down that route. I was initially focused on
converting from site -> core without changes.

Cheers,
-Gavin

aditihilbert:
Having newt skip docs is a good idea. We could also pull the tutorials out of /docs but then, having it in the codebase is actually very convenient. 

We do need to pull in these changes for now. We want to serve up the doxygen based documentation to let people see the progress and encourage them to contribute. The hope is to be able to migrate by release 1.4 (or perhaps 1.5).

jacobrosenthal:
I really like the idea of tutorials (assuming that cover all the pictures) being left separate. I think people who download 30meg of sourcecode would love docs making contributions in place, but I dont think they expect theyre getting getting 20meg of tutorials 'fore free'

But again, I would caution against merge first, these are then stuck in history forever if you want to back out of this or convert them later.

sjanc:
I agree with Jacob that this grows too big with binary blobs. Maybe we should simply have dedicated mynewt-docs repo?

sterlinghughes:
i liked the idea of breaking out the tutorials, but leaving the introduction and section header documentation in *-core.  That way you get documentation just for the *-core repository (which works well as we break things out too), but the tutorials, which are great, but span repositories and not really what you'd see in formal technical docs aren't burdening *-core.

gj262:
Yeah, that should all be doable. And the tutorial/etc docs should be able
to use doxygen references to mynewt-core objects (and other repo objects)
if we do it right.

Adithi, can you create an apache mynewt-docs repo?

I'll just leave this PR open and apply it to the new repo once it is live.

Cheers,
-Gavin

aditihilbert:
Our PMC @sterlinghughes can submit the request to create the new repo:  https://gitbox.apache.org/setup/newrepo.html

gj262:
I've removed the pics part of this commit. I'd like to merge the rest of it to get a clean starting point to begin moving the un-core related docs to mynewt-documentation.

```
$ git count-objects -v -H
count: 424
size: 18.83 MiB
in-pack: 75361
packs: 2
size-pack: 86.12 MiB
prune-packable: 3
garbage: 0
size-garbage: 0 bytes
```

PR #711, opened 2017-12-21
Initial ARCv2 port of mynewt

wes3:
This is the initial port of mynewt to the ARCv2 architecture. This is not a complete port by a long shot. There are still quite a few things to be done to make it complete. These are:

* Better default interrupt routines which call assert or coredump.
* Peripheral implementation.
* Tickless timer implementation.
* Assumes kernel mode only
* Have not implemented automatic load/debug scripts.

I am sure there are other missing features but I wanted to commit this so folks could give it a look over. Note that using this port requires the installation of the arc-elf32-gcc compiler and tools


hw/bsp/embarc_emsk/arc_core.ld
```diff
@@ -0,0 +1,100 @@
+MEMORY
```
mkiiskila: Apache License?

hw/bsp/embarc_emsk/include/bsp/arc_core_config.h
```diff
@@ -0,0 +1,151 @@
+#ifndef __core_config_h
```
mkiiskila: License?

hw/bsp/embarc_emsk/src/sbrk.c
```diff
@@ -0,0 +1,59 @@
...
+
+#include <hal/hal_bsp.h>
+
+/* put these in the data section so they are not cleared by _start */
+static char *sbrkBase __attribute__ ((section (".data")));
+static char *sbrkLimit __attribute__ ((section (".data")));
+static char *brk __attribute__ ((section (".data")));
+
+void
+_sbrkInit(char *base, char *limit) {
```
mkiiskila: Coding style?
{ should go on a separate line.

hw/mcu/arc/snps/syscfg.yml
```diff
@@ -0,0 +1,46 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# Package: hw/mcu/arc/snps
+
+syscfg.defs:
+    I2C_0:
```
mkiiskila: I'd suggest removing syscfg definitions which are not valid.

wes3:
Thanks for the review Marko. Here are my comments.

I fixed the last two of your comments as they were simple. However, I am not sure what to do about the rest:

1) I am happy to remove any files that should be removed from the SDK. The ones you pointed out I can easily remove but not sure which others should be on the removal list. I am not even sure I would be able to recognize all the files which should be removed :-)

2) I really know nothing about licensing. The files without licenses come from their SDK and have been adapted for mynewt. What should I do here? Add the apache license to them only? Add the synopsys license? Add both?

3) The Synopsys license appears to be a standard 3-clause BSD license. I can add a section for Synopsys in the top-level license file easily enough for this.

mkiiskila:
1. Personally, I would remove the binary files, .cproject, .gitignore, .project, Makefiles, *.tcf (some XML code presumably for IDEs), *.ccarg (unless you want to keep these, they seem to be cflags for compiler). Seems like getting rid of ext/sdk/example ext/sdk/documetantaion directories would remove most of these?
2. I think what we've done is copied in the license blurb to such from files which are also part of that SDK, and have the same license.
3. That would be great.

PR #713, opened 2017-12-22
net/lora: display transmit power in dbm

wes3:
Got tired of seeing an index instead of actual transmit power in dbm.


wes3:
Nope; it is purely informational. It does change what normally comes back in the confirm but for it to be any use the upper layer would have to convert the index to actual transmit power anyway.

PR #715, opened 2018-01-02
os_mempool; block size for memory pool has to allow storage for free …

mkiiskila:
…list linkage.

### DIFF
--- a/kernel/os/src/os_mempool.c
+++ b/kernel/os/src/os_mempool.c
@@ -90,7 +90,7 @@ os_mempool_init(struct os_mempool *mp, int blocks, int block_size,
     struct os_memblock *block_ptr;
 
     /* Check for valid parameters */
-    if (!mp || (blocks < 0) || (block_size <= 0)) {
+    if (!mp || (blocks < 0) || (block_size < sizeof(struct os_memblock))) {
         return OS_INVALID_PARM;
     }
 


kernel/os/src/os_mempool.c
```diff
@@ -90,7 +90,7 @@ os_mempool_init(struct os_mempool *mp, int blocks, int block_size,
     struct os_memblock *block_ptr;
 
     /* Check for valid parameters */
-    if (!mp || (blocks < 0) || (block_size <= 0)) {
+    if (!mp || (blocks < 0) || (block_size < sizeof(struct os_memblock))) {
```
rymanluk: Hi Marco, I saw a comment on #700 but frankly speaking I don't understand why we need this change.
As I understand `void *membuf`already has already is big enough if OS_MEMPOOL_BYTES() was used to calculate needed memory. Maybe could you say where is the issue you are seeing? I must miss something.

mkiiskila: Hi! Issue would appear if the user did not use OS_MEMPOOL_BYTES(), and specified block_size of, say 2 bytes. Clearly the test against > 0 is not correct, and we should have at least a pointer's worth of memory on every block.
As you have seen from the code, we typecast the beginning of a block into struct os_memblock and link it through that struct.

I'd be ok with removing the argument checks altogether, in which case the responsibility of passing correct arguments is up to the user. But if we keep the checks, then they should be meaningful.

Hope this helps,
M

andrzej-kaczmarek: IMO block_size<=0 check is fine since this checks size of block as seen for application, not as it is required internally. We may add note to documentation about required minimum effective block size if someone decides not to use OS_MEMPOOL_BYTES() macro to get proper size.

But even without such note for  block_size<4 user needs to assume at least 4 bytes for each block anyway - documentation clearly states that effective block size is OS_ALIGN-ed so anything less than 4 bytes is aligned to 4 bytes for all archs now (since we have only 32-bit archs).

wes3:
I will throw in my two cents here. Well, call it 1c euro :-) I think the user should be able to pass in a value of 1, 2 or 3. I think a value of 0 should be thrown away (error). The only problem I can see is changing the size of os_memblock such that the minimum block size that we align the block to is < sizeof(os_memblock). So, this is what I would do:

1) The check in the code should be <= 0.
2) "Compile time" error (#error) if the sizeof(os_memblock) < alignment size.

Of course, we could modify the OS_MEMPOOL_BYTES macro to deal with this but why bother? I cannot see changing the size of os_memblock and alignment should always be to, at least, the min pointer size.

PR #718, opened 2018-01-06
MYNEWT-881: SensorAPI: Add sensor_clear_high/low_thresh() API and LIS2DH12 set threshold cb fix

vrahane:
- There are two issues for calling lis2dh12_sensor_set_trigger_thresh() multiple times:

1. gpio irq was not being released before doing a re-initialization
2. srec was malloced and calling the function multiple times would have
malloced additional memory without freeing it.

Solution:
1. Release the gpio irq
2. Remove the malloc and add a ptr for the sensor in each type trait so
that a sensor read could be performed in the irq handler.

- Add sensor_clear_low_thresh() and sensor_clear_high_thresh() APIs

### DIFF
--- a/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
+++ b/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
@@ -1297,7 +1297,6 @@ lis2dh12_sensor_set_trigger_thresh(struct sensor *sensor,
     struct sensor_itf *itf;
     sensor_data_t low_thresh;
     sensor_data_t high_thresh;
-    struct sensor_read_ev_ctx *srec;
 
     itf = SENSOR_GET_ITF(sensor);
 
@@ -1369,12 +1368,10 @@ lis2dh12_sensor_set_trigger_thresh(struct sensor *sensor,
 
         os_time_delay((OS_TICKS_PER_SEC * 100)/1000 + 1);
 
-        srec = malloc(sizeof(struct sensor_read_ev_ctx));
-        srec->srec_sensor = sensor;
-        srec->srec_type = type;
+        hal_gpio_irq_release(itf->si_low_pin);
 
-        hal_gpio_irq_init(itf->si_low_pin, lis2dh12_low_int1_irq_handler, srec,
-                          HAL_GPIO_TRIG_FALLING, HAL_GPIO_PULL_NONE);
+        rc = hal_gpio_irq_init(itf->si_low_pin, lis2dh12_low_int1_irq_handler, stt,
+                               HAL_GPIO_TRIG_FALLING, HAL_GPIO_PULL_NONE);
         if (rc) {
             goto err;
         }
@@ -1441,12 +1438,10 @@ lis2dh12_sensor_set_trigger_thresh(struct sensor *sensor,
 
         os_time_delay((OS_TICKS_PER_SEC * 100)/1000 + 1);
 
-        srec = malloc(sizeof(struct sensor_read_ev_ctx));
-        srec->srec_sensor = sensor;
-        srec->srec_type = type;
+        hal_gpio_irq_release(itf->si_high_pin);
 
-        hal_gpio_irq_init(itf->si_high_pin, lis2dh12_high_int2_irq_handler, srec,
-                          HAL_GPIO_TRIG_FALLING, HAL_GPIO_PULL_NONE);
+        rc = hal_gpio_irq_init(itf->si_high_pin, lis2dh12_high_int2_irq_handler, stt,
+                               HAL_GPIO_TRIG_FALLING, HAL_GPIO_PULL_NONE);
         if (rc) {
             goto err;
         }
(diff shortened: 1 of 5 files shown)


hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -1436,12 +1433,10 @@ lis2dh12_sensor_set_trigger_thresh(struct sensor *sensor,
...
-        srec->srec_sensor = sensor;
-        srec->srec_type = type;
+        hal_gpio_irq_release(itf->si_high_pin);
 
-        hal_gpio_irq_init(itf->si_high_pin, lis2dh12_high_int2_irq_handler, srec,
-                          HAL_GPIO_TRIG_FALLING, HAL_GPIO_PULL_NONE);
+        rc = hal_gpio_irq_init(itf->si_high_pin, lis2dh12_high_int2_irq_handler, stt,
+                               HAL_GPIO_TRIG_FALLING, HAL_GPIO_PULL_NONE);
         if (rc) {
             goto err;
```
rymanluk: Please check error label. Looks like we are missing there IRQ disable and release.

vrahane:
@rymanluk I have addressed the comments and I do agree with your comments, I was a bit confused about the unset API since there are different ways of taking care of things but what I have done in the above commits should be good enough considering we have to enable/disable specific interrupts in the sensors registers if all the interrupts are not disabled. Hence, hal_gpio_irq_release() can only be called if all the thresholds are not valid. Hope I am making sense. Please go through the PR once again. Thank you.

hw/sensor/src/sensor.c
```diff
@@ -1922,6 +1922,59 @@ sensor_set_thresh(char *devname, struct sensor_type_traits *stt)
...
+    int rc;
+
+    if (!stt) {
+        rc = SYS_EINVAL;
+        goto err;
+    }
+
+    sensor = sensor_get_type_traits_byname(devname, &stt_tmp,
+                                           stt->stt_sensor_type);
+    if (!sensor) {
```
rymanluk: maybe if (!sensor || !stt_tmp) {}

btw why we care about stt_tmp since this is not used?

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -1463,7 +1485,75 @@ lis2dh12_sensor_set_trigger_thresh(struct sensor *sensor,
...
+        fs = 186;
+    } else {
+        rc = SYS_EINVAL;
+        goto err;
+    }
+
+    /* Set low threshold and enable interrupt */
+    rc = lis2dh12_set_low_thresh(itf, low_thresh, fs, stt);
+    if (rc) {
+        hal_gpio_irq_release(itf->si_low_pin);
```
rymanluk: can we have hal_gpio_irq_release(itf->si_low_pin); in case of error handler inside lis2dh12_set_low_thresh ?
It can easly be put in "err" label there. And since mentioned function does operation on this irq (init/release) it makes sense to handle release in case of error.

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -1463,7 +1485,75 @@ lis2dh12_sensor_set_trigger_thresh(struct sensor *sensor,
...
+        goto err;
+    }
+
+    /* Set high threshold and enable interrupt */
+    rc = lis2dh12_set_high_thresh(itf, high_thresh, fs, stt);
+    if (rc) {
+        hal_gpio_irq_release(itf->si_high_pin);
```
rymanluk: Similar comment as above.  hal_gpio_irq_release(itf->si_high_pin); could be done in "err" label in of lis2dh12_set_high_thresh()

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -1308,44 +1298,62 @@ lis2dh12_sensor_set_trigger_thresh(struct sensor *sensor,
...
-    rc = lis2dh12_get_full_scale(itf, &tmp);
-    if (rc) {
-        goto err;
+    if (!low_thresh.sad->sad_x_is_valid &&
+        !low_thresh.sad->sad_y_is_valid &&
+        !low_thresh.sad->sad_z_is_valid) {
+        /* if all thresholds are not valid, release gpio irq */
+        hal_gpio_irq_release(itf->si_low_pin);
+    } else {
+        /* if one or more thresholds are valid, set trigger */
```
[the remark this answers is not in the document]
vrahane: For lis2dh12, there are multiple interrupts that you can setup. Each axis has its associated interrupt. Unsetting one of the interrupts does not mean the IRQ is disabled. It means one of the interrupts from the sensor is disabled. By doing this, user can unset specific interrupt or all interrupts. I know unsetting an interrupt is actually just clearing registers of the particular interrupt but it also means we need to keep other interrupts active. Almost all the steps in the set thresh function would have to be repeated here which seems more repetitive code. Hence, calling the set thresh function here.

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
vrahane: Agreed. Will move it in.

hw/drivers/sensors/lis2dh12/src/lis2dh12.c
vrahane: Agreed. Will move it in.

PR #719, opened 2018-01-09
 nimble: Add meta-package for HCI transports

andrzej-kaczmarek:
This patch makes net/nimble/transport a separate meta-package to allow select HCI transport via syscfg instead of explicitly adding dependency to particular transport package.

The advantage of this approach is that the same application can be used on multiple targets, regardless of controller being used on target BSP. It is also fully backwards compatible since it is still possible to do
it "the old way", i.e. add explicit dependency to particular transport package instead of meta-package.

By default, combination of RAM transport and NimBLE controller is used which is default configuration for most existing apps.

Each availble transport package has its own syscfg value:
- BLE_HCI_TRANSPORT_EMSPI for net/nimble/transport/emspi
- BLE_HCI_TRANSPORT_RAM for net/nimble/transport/ram
- BLE_HCI_TRANSPORT_SOCKET for net/nimble/transport/socket
- BLE_HCI_TRANSPORT_UART for net/nimble/transport/uart

In addition there's dedicated value to use built-in NimBLE controller, which is most common usage and thus default setting:
- BLE_HCI_TRANSPORT_NIMBLE_BUILTIN

This is virtually the same as enabling RAM transport and adding package with NimBLE controller ( net/nimble/controller) manually, but since package is pulled by syscfg value it can be easily overriden in target settings.


apps/bleprph/src/main.c
```diff
@@ -292,9 +292,6 @@ main(void)
 {
     int rc;
 
-    /* Set initial BLE device address. */
-    memcpy(g_dev_addr, (uint8_t[6]){0x0a, 0x0a, 0x0a, 0x0a, 0x0a, 0x0a}, 6);
```
sjanc: just use static random so that it doesn't use all zero address in combined builds on nRF5x

net/nimble/transport/pkg.yml
```diff
@@ -0,0 +1,42 @@
...
+
+pkg.name: net/nimble/transport
+pkg.description: Meta-package for NimBLE HCI transport
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+    - ble
+    - bluetooth
+
+pkg.deps.BLE_HCI_TRANSPORT_BUILTIN:
```
rymanluk: Maybe BLE_HCI_TRANSPORT_NIMBLE_CTRL_BUILTIN
or BLE_HCI_TRANSPORT_NIMBLE_BUILTIN ?

apps/bleprph/src/main.c
```diff
@@ -125,11 +134,12 @@ bleprph_advertise(void)
         return;
     }
 
+
```
rymanluk: minor nitpic: double empty line

PR #725, opened 2018-01-12
match os_sched() doxygen with mynewt-site

gj262:
Update os_sched() inline doxygen docs to match what is displayed at:

https://mynewt.apache.org/latest/os/core_os/context_switch/os_sched/

@sterlinghughes let me know it you want to go down this route. It will require a lot of updates to source to get their comments to match the current docs.

also @aditihilbert

### DIFF
--- a/docs/os/core_os/API.rst
+++ b/docs/os/core_os/API.rst
@@ -1,0 +1,11 @@
+Core OS API
+-----------
+
+.. contents::
+  :local:
+  :depth: 2
+
+Scheduler
+~~~~~~~~~
+
+.. doxygenfile:: os_sched.h
--- a/docs/os/core_os/context_switch/context_switch.rst
+++ b/docs/os/core_os/context_switch/context_switch.rst
@@ -1,11 +1,9 @@
-#########
 Scheduler
-#########
+=========
 
 Scheduler's job is to maintain the list of tasks and decide which one
 should be running next.
 
-***********
 Description
 ***********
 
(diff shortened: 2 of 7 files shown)


gj262:
screen shot 2018-01-11 at 6 40 46 pm

sterlinghughes:
Hi Gavin, 
I think we should go down this route, although I don't think you should be updating the Mynewt doxygen.  We had a meeting before the new year to discuss, and agreed:

(a) the Mynewt developers will update the doxygen in the tree to make it more complete and usable
(b) We will move to doxygen comments in the header files instead of in the source files, and have that track function API declaration as opposed to implementation.  

I don't think we're going for exact parity here, but we do want to have this be the source of truth going forward.  By maintaining the prototypes externally, we risk them being out of date, especially as we maintain versions of the documentation for each firmware release with the correct API signature.

sterlinghughes:
I should say "all the Mynewt doxygen," obviously have a few example sections that work & rende well is a huge help as a reference for everyone who isn't really familiar with the new structure & format.

gj262:
Makes sense. We can use os_sched() as an example of how one should document a function (after I move it to the header file.). And add this to the "best practices for documenting the code" once I get around to writing that. My plan is to provide the skeleton framework so that devs can easily add inline docs without getting too bogged down with sphinx/breathe/doxygen specifics.

aditihilbert:
looks good. Thanks!

PR #730, opened 2018-01-15
Add Cortex-M3 support

utzig:
This allows MCUBoot built for STM32L152C-Discovery to boot to image, so it's probably complete.

### DIFF
--- a/compiler/arm-none-eabi-m3/syscfg.yml
+++ b/compiler/arm-none-eabi-m3/syscfg.yml
@@ -1,0 +1,20 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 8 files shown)


compiler/arm-none-eabi-m3/syscfg.yml
```diff
@@ -0,0 +1,25 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# Package: compiler/arm-none-eabi-m3
+
+# syscfg.defs:
+#   HARDFLOAT:
```
mkiiskila: I don't think there's hardware FPU on Cortex-M3? Given that, this setting should not be here.

utzig: Hmmm, I thought I had removed all references to `HARDFLOAT`... my bad!

utzig: Yes, I left it commented out because I thought it could have some side effect... forgot to remove later... thanks!

PR #732, opened 2018-01-15
Stm32l152discovery

utzig:
Add initial support for STM32L1xx and STM32L152C-Discovery board.


mkiiskila:
Is this still WIP?

utzig:
I will remove the comment once is not

utzig:
I2C/SPI are not complete but I plan to merge them with the other STM32 HALoverHALs.

kernel/os/src/arch/cortex_m3/os_arch_arm.c
```diff
@@ -210,7 +210,8 @@ os_arch_os_init(void)
             NVIC->IP[i] = -1;
         }
 
-        NVIC_SetVector(SVCall_IRQn, (uint32_t)SVC_Handler);
+        /* SVC_IRQn or SVCall_IRQn depending on specific model */
+        NVIC_SetVector(-5, (uint32_t)SVC_Handler);
```
mkiiskila: If this indeed varies from MCU to MCU, we could declare a name for this, and set that using a #define inside mcu/mcu.h.

PR #741, opened 2018-01-17
nimble/host: Allow low interval on non-connectable advertising

sjanc:
This limiation was removed in BT5 spec so there is no need to keep this
when build with nimBLE controller. Since nimBLE host doesn't verify
controller version for now we allow this on combined build only.

### DIFF
--- a/net/nimble/host/src/ble_hs_hci_cmd.c
+++ b/net/nimble/host/src/ble_hs_hci_cmd.c
@@ -173,6 +173,12 @@ ble_hs_hci_cmd_body_le_set_adv_params(const struct hci_adv_params *adv,
         return -1;
     }
 
+/* When build with nimBLE controller we know it is BT5 compliant so no need
+ * to limit non-connectable advertising interval
+ */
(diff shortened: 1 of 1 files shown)


net/nimble/host/src/ble_hs_hci_cmd.c
```diff
@@ -173,13 +173,20 @@ ble_hs_hci_cmd_body_le_set_adv_params(const struct hci_adv_params *adv,
         return -1;
     }
 
+/* When build with nimBLE controller we know it is BT5 compliant so no need
+ * to limit non-connectable advertising interval
+ */
+#if MYNEWT_VAL(BLE_DEVICE)
```
rymanluk: Maybe we should test also for BSP since it will not apply for nrf51?
For sure  it will work for MYNEWT_VAL(BSP_NRF52840)

rymanluk: hmm for build in nimble it is ok because controller does same checks. So maybe we could remove more checks from host then?

PR #746, opened 2018-01-19
mesh: Reset Heartbeat Publication timer before sending message

michal-narajowski:
The timer should be reset before sending to ensure correct publication period.

This patch allows to pass MESH/NODE/CFG/HBP/BV-02-C.

### DIFF
--- a/net/nimble/host/mesh/src/cfg_srv.c
+++ b/net/nimble/host/mesh/src/cfg_srv.c
@@ -3081,19 +3081,19 @@ static void hb_publish(struct os_event *work)
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


michal-narajowski:
@jhedberg FYI

jhedberg:
Looks ok to me as well. Is anyone planning to port & submit this to Zephyr as well?

MariuszSkamra:
@jhedberg I'll submit the patch.

michal-narajowski:
Updated after comments from @jhedberg. https://github.com/zephyrproject-rtos/zephyr/pull/5747

PR #751, opened 2018-01-21
Driver for debouncing an input pin.

mlampert:
Added hw/drivers/debounce for debouncing an input pin, relies on hal_gpio(_irq) and hal_timer.

See debounce.h for the API description and example use.

### DIFF
--- a/hw/drivers/debounce/include/debounce/debounce.h
+++ b/hw/drivers/debounce/include/debounce/debounce.h
@@ -1,0 +1,229 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef _DEBOUNCE_DEBOUNCE_H_
+#define _DEBOUNCE_DEBOUNCE_H_
+
+/*
+ * debounce
+ *
+ * Driver to debounce a pin and optionally call a callback
+ * when the state changes.
+ *
+ * The implementation registers an IRQ callback for the pin, once triggered
+ * a periodic timer is used to check the state of the pin until it becomes
+ * stable for a certain number of times.
+ *
+ * This way debouncing doesn't consume any processing resources unless its
+ * pin actually changes and it is not susceptible to the thundering herd
+ * problem of IRQs that can be triggered by noisy inputs. See
+ * \c debounce_set_params for a detailed description of how the parameters
+ * influence the minimum time a signal is required to remain stable for it
+ * to be detected.
+ *
+ * The API relies on a structure of type \c debounce_t to remain valid for the
+ * lifetime of debouncing a pin. The structure can be dynamically allocated or
+ * statically defined.
+ *
+ * // ---------------------- Example begin --------------------------
+ *
+ * void buttonPressed(debounce_t *d) {
+ *     if (debounce_state(d)) {
+ *         ... code to process button press event
+ *     } else {
+ *         ... code to process button depress event
+ *     }
+ * }
+ *
+ * debounce_t button;
+ *
+ * int main(int argc, char *argv[]) {
+ *     ...
+ *     debounce_init(&button, BUTTON_PIN, HAL_GPIO_PULL_UP, 0);
+ *     debounce_start(&button, DEBOUNCE_CALLBACK_EVENT_ANY, buttonPressed, NULL);
+ *     ...
+ * }
+ *
+ * // ---------------------- Example end --------------------------
+ *
+ * The driver relies on
+ *   hal_timer ... at least one HW timer needs to be configured
+ *                 and running
+ *   hal_gpio  ... specifically the irq interface is used in order
+ *                 to detect the initial trigger for debouncing
+ */
+
+#include "hal/hal_gpio.h"
+#include "hal/hal_timer.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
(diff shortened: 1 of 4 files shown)


hw/drivers/debounce/include/debounce/debounce.h
```diff
@@ -0,0 +1,235 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Internal structure for debouncing a pin. Application code should only
+ * use the API to access it's data members and not directly manipulate or
+ * retrieve any of them.
+ */
+typedef struct debounce_s {
```
mkiiskila: I would drop the '_s' from the end of debounce_s. 'struct debounce' seems better.

hw/drivers/debounce/include/debounce/debounce.h
```diff
@@ -0,0 +1,235 @@
...
+    struct hal_timer timer;
+} debounce_t;
+
+/*
+ * A callback can be invoked when the debounced pin's state changes. It
+ * can either be invoked when the pin rises, falls or on any state change.
+ * If either \c DEBOUNCE_CALLBACK_NEVER is specified or the callbak itself
+ * is set to \c NULL no callback is invoked.
+ */
+typedef enum debounce_callback_event_e {
```
mkiiskila: Same comment here, 'enum debounce_callback_event' would be better.

hw/drivers/debounce/include/debounce/debounce.h
```diff
@@ -0,0 +1,235 @@
...
+ *
+ *
+ * @param d         Structure to manage debouncing
+ * @param ticks     The # ticks used as the time interval for debouncing
+ * @param count     The # times the pin state has to be stable for
+ *                  debouncing to complete
+ *
+ * @return int  0: no error; -1 otherwise.
+ */
+int debounce_set_params(debounce_t *d, uint16_t ticks, uint8_t count);
```
mkiiskila: Maybe debounce_config()? To use the same term as other drivers.

hw/drivers/debounce/src/debounce.c
```diff
@@ -0,0 +1,128 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "debounce/debounce.h"
+#include "syscfg/syscfg.h"
+#include <string.h>
```
mkiiskila: As a rule, it is better to place syscfg.h before including any header files from within mynewt; syscfg.h contents influence conditional parts in header files.

mlampert:
I am not sure I understand your feature request. Is it that you want the notification at the beginning of a debouncing instead of the end?
That would assume that every 0->1 transition is indeed the beginning of a button press event, which is not necessarily the case. You might also look at a glitch which gets suppressed by the current algorithm.

hw/drivers/debounce/include/debounce/debounce.h
mlampert: The reason I didn't use `_config` is because in all drivers and APIs I looked at the `_config` call is mandatory, whereas setting these parameters in code is only necessary if you have multiple pins to debounce in one application and they need different debouncing parameters.

hw/drivers/debounce/include/debounce/debounce.h
mlampert: I am generally not in favour of data structures that carry the same name as a module, I'll see if I find a different name.

hw/drivers/debounce/src/debounce.c
mlampert: Shouldn't those header files then include syscfg themselves ?
Includes should be alphabetically ordered, makes it a lot easier to figure out what is included and what isn't.

ccollins476ad: Most uses of syscfg involve the `MYNEWT_VAL()` macro.  If a header tries to use `MYNEWT_VAL()` before `syscfg.h` has been included, you'll get a build error.  So, in these cases, it is not terribly important that the `syscfg.h` include go at the top.

However, in rare cases, code needs to check if a syscfg macro has been defined at all (e.g., `log/full/include/log/log.h` uses `#ifndef MYNEWT_VAL_LOG_LEVEL`).  This is needed when you must distinguish an undefined setting from a setting that is defined as `0`.  In these cases, forgetting to included `syscfg.h` won't trigger a build error and can lead to lots of confusion.  My opinion is that preventing this issue is more important than the ability to quickly see which headers are being included, so I agree with @mkiiskila that `syscfg.h` should always be included before any other Mynewt headers.

mlampert: my point was, that header files need to be self contained - they need to include whatever they need and not rely on the fact that another file that included that header includes another header file before that.

So unless this header file needs the contents of syscfg.h in order to determine if it should include another header file or not - it doesn't matter where in the include list it is.

As to promoting including a specific header always first or last (or 3rd), I don't consider that a very good coding style. It encourages and promotes badly formed header and source files. It would be more prudent to fix all headers and sources which rely on syscfg.h but don't include it.

ccollins476ad: I agree that headers should be self-contained.  Including `syscfg.h` early is just protection in case you forget.

Re: good coding style - join the debate :).  I think you'll find everyone has his own opinion on this particular issue (e.g., https://stackoverflow.com/questions/2762568/c-c-include-file-order-best-practices).  This is certainly not a huge issue, but my feeling is that including `syscfg.h` first is a practical solution to an actual problem.  The ability to quickly see which headers are included seems less useful.

mkiiskila:
> On Jan 25, 2018, at 7:35 PM, mlampert <notifications@github.com> wrote:
> 
> I am not sure I understand your feature request. Is it that you want the notification at the beginning of a debouncing instead of the end?
> That would assume that every 0->1 transition is indeed the beginning of a button press event, which is not necessarily the case. You might also look at a glitch which gets suppressed by the current algorithm.
> 
> 

What I want is an ability to see rapid transitions on possibly
noisy lines.

I would not have limited the transition from 0->1, user should
be able to define which direction they want the transition to.

Another option to one-side monitoring would be to have a
‘cooldown' period. If there’s further transitions within cooldown
period, then apply debouncing and notify at the end.
If the transition happens after cooldown period (aka stable signal),
then report the change right away.

I think that would work great for the most common use of
debouncer, a button.

I’m not too worried about solving a problem of avoiding individual
glitch.

mlampert:
I'm still not sure what you want and I get the feeling it is not debouncing of a button and/or noisy line, which is what this driver is meant to do. I'm not opposed to improve and build up the functionality but I would like to get the basic driver in first.

PR #754, opened 2018-01-22
PIC32: Removed unused core dump functions

IMGJulian:


IMGJulian:
I figured that not having it was better than having a it implemented incorrectly. There is a precedent for not having one as set by the rvimac port but I would rather see this as a step towards having one implemented correctly I just don't have the time to sort that out right now.

PR #755, opened 2018-01-22
nimble: mbuf optimization in controller

andrzej-kaczmarek:
This makes some optimizations in the way controller uses memory/mbufs:

* use mbufs for storing adv data so we don't need to allocate a lots of memory for adv data if we're not going to use it
* make ble_phy_tx create PDUs on-the-fly so there is no need to create temporary mbufs and copy data over and over again

It will be especially useful for advertising data chaining, but it speeds things up even for legacy code.

### DIFF
--- a/hw/drivers/nimble/native/src/ble_phy.c
+++ b/hw/drivers/nimble/native/src/ble_phy.c
@@ -67,6 +67,8 @@ struct ble_phy_statistics
 
 struct ble_phy_statistics g_ble_phy_stats;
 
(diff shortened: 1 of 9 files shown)


andrzej-kaczmarek:
Thanks for review, @wes3. Indeed, I missed both cases so will need to address this.

In case of os_mbuf_append failure (or oom in general) we should remove all adv_data as per spec.

As for TX I assumed here that since we just change order of operations and reduce time required to create buffer in some cases it should not affect this, but I'll move check for radio state after filling header bytes and see results in tx_late stat. So far I did not observe any regressions with advertising/connection so hopefully my assumption was correct.

PR #762, opened 2018-01-24
upgrade STM32 Cube for the F4, and re-org to make future upgrades eas…

sterlinghughes:
- Upgrade the STM32F4 Cube library.  Previously we were using a patched version of this library -- instead I've copied the two modified files and included them in our HAL implementation to make upstreaming new STM32 cubes easier.

- I've configured the build of STM32Cube to ignore the "default" files i2c and spi, so instead we build the ones in our HAL.  This should be changed so our HAL has a non-conflicting implementation and is still evil, but will help things in the short term.

- I've added support for the STLINK debug tools, as the Nucleo F413 does not have mainlined openocd support, whereas STLINK is brew installable and reliable.

- I've added a BSP for the Nucleo F413.


utzig:
> I've configured the build of STM32Cube to ignore the "default" files i2c and spi, so instead we build the ones in our HAL. This should be changed so our HAL has a non-conflicting implementation and is still evil, but will help things in the short term.

This is not done correctly I think. This rule:

```
pkg.ign_files:
    - "stm32f4xx_hal_i2c.c"
    - "stm32f4xx_hal_spi.c"
```

ignores both `hw/mcu/stm/stm32f4xx/src/stm32f4xx_hal_(i2c|spi).c` and the same named files inside `ext/Drivers/STM32F4xx_HAL_Driver/Src`, so in practice my projects that have `i2c` or `spi` enabled are not building.

As a temporary change I don't mind this very much (it should be fixed soon!).

I tested this updated HAL on a stm32f4-discovery and everything works fine (have not tested `i2c/spi`).

sterlinghughes:
Hi @utzig  -- I updated the commits to change the files so they build.  

@utzig @mkiiskila -- take a look, paying specific attention to my changes removing the HAL_I2C_MODE_MASTER_SEL.  I did not want to change the vanilla I2C modes, and thought it was fine to check if the bus was busy every time, even if a STOP had not been generated -- although I have not tested this much.  

I instead set the mode to HAL_I2C_MODE_MASTER, so that functions like abort will still work on the bus if a STOP has not been sent.

PR #763, opened 2018-01-25
Integrate nrfx

mlaz:
This pull request integrates the [nrfx standalone driver package](https://github.com/NordicSemiconductor/nrfx) in mynewt-core.
This allows us to bring in all the drivers from [mynewt_nordic](https://github.com/runtimeco/mynewt_nordic) such as PWM and ADC and to stop depending on it.

To-Do:
- [x] Port and test the pwm_nrf52 driver;
- [x] Port and test the adc_nrf52 driver;
- [x] Port the adc_nrf51 driver;
- [x] Test the adc_nrf51 driver (I am unable to test it);
- [x] Test HAL GPIO driver;
- [x] Test HAL SPI driver;
- [x] Test HAL I2C;
- [x] Update the remaining nrf52 boards;


jacobrosenthal:
I can probably take the adc_nrf51 driver

apps/soft_pwm_test/pkg.yml
```diff
@@ -26,9 +26,6 @@ pkg.keywords:
 
 pkg.deps:
     - kernel/os
-   # - hw/hal
-   # - hw/cmsis-core
     - sys/console/full
-   # - hw/bsp/nrf52dk
```
mkiiskila: Instead of commenting lines out as obsolete, it's better to remove them. If you decide later that you want to bring those changes back, they'll be visible via the file history.

mlaz: These comments are the kind of thing I usually address when cleaning up.
I am actually considering to delete this app. We may simply have pwm_test using any pwm driver, by just enabling whatever driver the user wants with newt setting a given syscfg variable.

mkiiskila: That sounds good to me.

jacobrosenthal:
Trying to build this on an nrf52 thingy I get a compilation error

```
Error: repos/apache-mynewt-core/hw/mcu/nordic/nrf52xxx/src/hal_gpio.c: In function 'hal_gpio_read':
repos/apache-mynewt-core/hw/mcu/nordic/nrf52xxx/src/hal_gpio.c:175:42: error: expected expression before ')' token
         (port->OUT >> HAL_GPIO_INDEX(pin)) & 1UL :
                                          ^
repos/apache-mynewt-core/hw/mcu/nordic/nrf52xxx/src/hal_gpio.c:177:1: error: control reaches end of non-void function [-Werror=return-type]
 }
```

Looks like HAL_GPIO_INDEX is blank
```
#ifdef NRF52
#define HAL_GPIO_INDEX(pin)
#define HAL_GPIO_PORT(pin)      (NRF_P0)
#define HAL_GPIO_MASK(pin)      (1 << pin)
#define HAL_GPIOTE_PIN_MASK     GPIOTE_CONFIG_PSEL_Msk
#endif
```

mlaz:
> Looks like HAL_GPIO_INDEX is blank
@jacobrosenthal fixed, in this case(nRF52832) the macro should do nothing, my bad.

jacobrosenthal:
Are we trying to bring this in at the hal level or the driver level?

jacobrosenthal:
Im not sure what you need tested for Test HAL I2C but I ran my drv2605 driver successfully on an nrf52 thingy

sterlinghughes:
@mkiiskila @jacobrosenthal I'm not sure our current HAL uses the nrfx header files or files like that?  I thought nrfx was a replacement for the existing, non-BSD licensed driver pack.  the hal relies on NRF bitfields and a few other definitions, but shouldn't overlap with nrfx?

mlaz:
> I thought nrfx was a replacement for the existing, non-BSD licensed driver pack.

@sterlinghughes It is.

> the hal relies on NRF bitfields and a few other definitions, but shouldn't overlap with nrfx?

IMHO nrfx also should replace both mynewt_nordic and all the files used under hw/mcu/nordic/src/ext/nRF5_SDK_11.0.0_89a8197 namely files containing the NRF bitfields definitions.

jacobrosenthal:
Of note, the old nordic adc and pwm implementations are using the driver(drv) not hal implementations which can get pretty tied into nordics use of timers and other hardware we need to be careful of that and probably document it better.

For instance if Im reading this correctly, simply including nrfx_config.h which both pwm and adc do, will enable everything in that file ( pwm and adc, plus nrfx_systick whatever that is) IE turning on pwm turns on adc. I think well have to replace the hardcoded enables in nrfx_config.h in with checks on MYNEWT_VAL()

Also this PR creates a new sysflag ADC that needs to be enabled in addition, plus ADC_0 for the 0 channel. Its more newty to simply check of the ADC_0 etc, for instance like the STM adc driver currently does
```
#if MYNEWT_VAL(ADC_1)||MYNEWT_VAL(ADC_2)||MYNEWT_VAL(ADC_3)
```

I think I have a lot of comments on the way those drivers implemented. Mainly ADC especially is hugely compile time configured for some reason, and I think they need to be more runtime configuration based. But Im not sure that should hold up the SDK refresh. 

Perhaps the sdk refresh PR needs to be separated from the driver merge and refactor, or at least it would help to rebase this pr into fewer scoped commits to better understand what were changing from the upstream nrfx package, and

jacobrosenthal:
@mlaz Seems like you deleted the adc_test you had in here? What did you run on nrf52 that I can test on nrf51?

mlaz:
> Also this PR creates a new sysflag ADC that needs to be enabled in addition, plus ADC_0 for the 0 channel. Its more newty to simply check of the ADC_0 etc, for instance like the STM adc driver currently does

Agreed, I'll make it work like STM  ADC driver. Probably the PWM macros should work this way as well for pwm_nrf52.

> @mlaz Seems like you deleted the adc_test you had in here? What did you run on nrf52 that I can test on nrf51?

I did because this app will probably only work with adc_nrf52, since the driver makes you use nordic structures namely nrf_saadc_channel_config_t. These test apps, imho, should be generic, i.e. the code should be able to run with whatever driver according to whatever syscfg macros the user sets. Just like pwm_test works with pwm_nrf52 and soft_pwm.

jacobrosenthal:
@mlaz and I worked tonight on the port. Were pulling out the problematic nrfx_config, as its pretty clear the way this was handled in the past was within the bsp pkg.cflags

Those are all out of date now so were fixing them, and also, a bunch of them are enabled even though newt doesnt use any nordic driver code. That must be from a very old version of mynewt where nordic stuff was included in the repo and has been copy and pasted over and over since.

On the flip side, the adc and pwm which there are drivers for we are enabling, but behind the pkg.cflags.ADC_0 conditional which is more recent development I believe

However, I would argue that this shouldnt be done in the bsp but rather in the driver package so that theyre only enabled if you bring the driver in and dont have to be edited out of the bsp if someone wants to write a new newt driver not based on any of the nordic 'driver' code. However when I do that theyre not included in the correct order to keep the nrfx headers from bitching theyre not defined. Not sure if thats an easy fix under the newt hood?

jacobrosenthal:
ADC is now in with minimal changes (for better or worse)

Were getting close to squashing this.

But mlaz and I are going back and forth on why you might use ign.files like
https://github.com/runtimeco/mynewt_nordic/blob/master/hw/mcu/nordic_sdk/pkg.yml#L35

There are one or two files in the sdk that dont compile and were not using them so those obviously go on the list, but should literally all the files were not using be on that list, does that speed up newt compiles at all?

hw/bsp/ble400/pkg.yml
```diff
@@ -29,52 +29,49 @@ pkg.keywords:
...
+    - '-DNRFX_TIMER2_ENABLED=0'
+    - '-DNRFX_TWI_ENABLED=0'
+    - '-DNRFX_TWI0_ENABLED=0'
+    - '-DNRFX_TWI1_ENABLED=0'
+    - '-DNRFX_UART_ENABLED=0'
+    - '-DNRFX_UART0_ENABLED=0'
+    - '-DNRFX_WDT_ENABLED=0'
+
+#todo this needs to be in the adc driver file instead, but that doesnt seem to work currently
+pkg.cflags.ADC_0:
```
jacobrosenthal: Still looking for newt feedback on these. They should be in the adc driver IMPL as theyre very related to our (wrapped abstraction) style and someone who writes a hal style doesnt want these enabled or they might be stomped on by the nordic files.

hw/bsp/bbc_microbit/pkg.yml
```diff
@@ -28,54 +28,49 @@ pkg.keywords:
...
-    - '-DTIMER4_ENABLED=0'
-    - '-DTWI0_CONFIG_SCL=0'
-    - '-DTWI0_CONFIG_SDA=1'
-    - '-DTWI0_ENABLED=1'
-    - '-DTWI1_ENABLED=0'
-    - '-DTWIS0_ENABLED=0'
-    - '-DTWIS1_ENABLED=0'
-    - '-DUART0_ENABLED=1'
-    - '-DWDT_ENABLED=1'
+    - '-DNRFX_CLOCK_ENABLED=0'
```
jacobrosenthal: These are stand in for the nordic driver template (in this case nrx_config.h) Note that were turning all of these 'off' as Im frankly not sure why any of them were enabled previously. They should only be enabled if were using nordic driver files, which ideally we shouldnt be except in cases of the ADC where we are wrapping their driver with ours.

hw/drivers/adc/adc_nrf51/src/adc_nrf51.c
```diff
@@ -0,0 +1,409 @@
...
+    if (rc != NRFX_SUCCESS) {
+        goto err;
+    }
+
+    return (0);
+err:
+    return (rc);
+}
+
+//WHY DOES THIS EXIST! why doesnt it work without it!
```
jacobrosenthal: Heh. I dont understand why calling nrfx_adc_buffer_convert would 'release' the buffer, but when I remove this it breaks....

hw/mcu/nordic/pkg.yml
```diff
@@ -26,14 +26,101 @@ pkg.keywords:
 
 pkg.type: sdk
 
+pkg.ign_files.BSP_NRF52:
+    - "nrfx_adc.c"
```
jacobrosenthal: These are the ign_files were asking about in the comments section. Should we be ignoring everything for speed or something? or just leaving them so if someone were to 'enable' a driver later they wouldnt have to do anything here.

mkiiskila:
Looks pretty good. Here's one target that seems to be failing:

[marko@IsMyLaptop:~/src/incubator-mynewt-blinky]$ newt target show bleprph_nrf52 
targets/bleprph_nrf52
    app=@apache-mynewt-core/apps/bleprph
    bsp=@apache-mynewt-core/hw/bsp/nrf52840pdk
    build_profile=debug
    syscfg=BLE_EXT_ADV=1:BLE_LL_CFG_FEAT_LE_2M_PHY=1:BLE_LL_CFG_FEAT_LE_CODED_PHY=1
[marko@IsMyLaptop:~/src/incubator-mynewt-blinky]$ newt build -j 1 bleprph_nrf52
Building target targets/bleprph_nrf52
Compiling repos/apache-mynewt-core/hw/drivers/nimble/nrf52/src/ble_phy.c
Error: repos/apache-mynewt-core/hw/drivers/nimble/nrf52/src/ble_phy.c: In function 'ble_phy_get_cur_rx_phy_mode':
repos/apache-mynewt-core/hw/drivers/nimble/nrf52/src/ble_phy.c:796:35: error: subscripted value is neither array nor pointer nor vector
         phy = NRF_RADIO->RESERVED7[0] & 0x06 ?
                                   ^

exit status 1

mkiiskila:
LGTM.

hw/drivers/nimble/nrf52/src/ble_phy.c
```diff
@@ -773,14 +773,16 @@ ble_phy_get_cur_rx_phy_mode(void)
...
          * XXX CI field value is bits [2:1] in NRF_RADIO->PDUSTAT which is
-         *     available in SDK v14.0 - it's NRF_RADIO->RESERVED7[0] in
-         *     SDK v11.0 so let's use it for now
+         *     available in nrfx for nrf52840 only - it's NRF_RADIO->RESERVED7[0]
+         *     for other nrf52 vairants.
          */
-        phy = NRF_RADIO->RESERVED7[0] & 0x06 ?
-                                        BLE_PHY_MODE_CODED_500KBPS :
-                                        BLE_PHY_MODE_CODED_125KBPS;
+#if MYNEWT_VAL(BSP_NRF52840)
```
mkiiskila: This does not seem right. You can have other BSPs with this particular nordic MCU in them, and have this code be built. Are the original #ifdef's not sufficient?

PR #766, opened 2018-01-25
nimble/ll: Fix HCI LE Read Remote Features command

andrzej-kaczmarek:
Commit 04bf9592 removed automatic feature exchange procedure after
connection when in slave role. This breaks HCI LE Read Remote Features
command since it expected that this procedure is always started so it
did not even try to start it, just sets bit in state machine that reply
is pending.

This patch fixes this regression by starting features exchange procedure
if not already done.


MariuszSkamra:
LGTM

PR #767, opened 2018-01-25
nimble/ll_dtm: Couple of fixes for DTM code

rymanluk:
With this patch DTM works properly when XTAL is not ON all the time.

### DIFF
--- a/net/nimble/controller/src/ble_ll_dtm.c
+++ b/net/nimble/controller/src/ble_ll_dtm.c
@@ -314,6 +314,12 @@ ble_ll_dtm_rx_start(void)
 
     ble_ll_state_set(BLE_LL_STATE_DTM);
 
+#ifdef BLE_XCVR_RFCLK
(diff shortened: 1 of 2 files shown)


net/nimble/controller/src/ble_ll_dtm.c
```diff
@@ -315,6 +315,10 @@ ble_ll_dtm_rx_start(void)
 
     ble_ll_state_set(BLE_LL_STATE_DTM);
 
+#ifdef BLE_XCVR_RFCLK
+    ble_ll_xcvr_rfclk_timer_start(os_cputime_get32());
```
andrzej-kaczmarek: this will try to setup timer for starting clock which is pointless since you want to start it now.

if (ble_ll_xcvr_rfclk_state() == BLE_RFCLK_STATE_OFF) {
    ble_ll_xcvr_rfclk_start_now(os_cputime_get32());
}

rymanluk: agree.

PR #769, opened 2018-01-26
nimble/phy: Fix broken encryption on nRF52/nRF52840

andrzej-kaczmarek:
This fixes regression from PR #755 + minor cleanup


andrzej-kaczmarek:
I tested this on both nRF52 and nRF52840 and seems to work fine even with data length extension.

Another options here would be to start key-stream generation immediately, as previously, but remove ENDKSGEN->CRYPT shortcut and use PPI to trigger CRYPT on TXREADY. This however requires extra PPI channel (there's no preconfigured one) so let's keep it with short enabled for now.

PR #780, opened 2018-01-29
Run sim on FreeBSD.

mkiiskila:


mkiiskila:
Probably should be cc. It's been a while since I used FreeBSD as a system (last extended exposure was FreeBSD6), and things have changed. I wanted this in just to get past the initial hump. Good that newtmgr is now out of newt, so at least BLE library dependency does not stop builds. 

Outstanding issue is that 'newt test all' fails with net/oic test;  probably something missing for ipv6 socket usage for multicast, or might be just be my VM setup. Did not investigate (yet). This is not a full complete port yet, wanted to see if folks are interested in this :)

PR #786, opened 2018-01-31
nimble/ll: Fix for disconnection with LL response timeout with one vendor

rymanluk:


net/nimble/controller/src/ble_ll_conn_hci.c
```diff
@@ -1512,6 +1512,16 @@ ble_ll_conn_hci_le_set_phy(uint8_t *cmdbuf)
         return BLE_ERR_CMD_DISALLOWED;
     }
 
+    /*
+     * If we are a slave and the master has initiated the channel map
+     * update procedure we should deny the slave request for now.
```
andrzej-kaczmarek: I don't think we should reject HCI command here. Procedure collision occurs at LL level and at this point in fact there is no collision: we did not start PHY Update procedure and we do not have to start it now, we can wait for current one to finish which is the proper way to handle this IMO. Additionally, spec Vol 6 Part B section 5.3 (Procedure Collisions) states that error code used here should be used in either rejection PDU or disconnection event (i.e. when connection occurred in LL) which is also not the case here.

BTW, the same applies for other procedures with instants so we would need more generic fix for procedures collisions in LL.

net/nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -1473,6 +1473,9 @@ ble_ll_ctrl_rx_reject_ind(struct ble_ll_conn_sm *connsm, uint8_t *dptr,
         ble_ll_ctrl_proc_stop(connsm, BLE_LL_CTRL_PROC_PHY_UPDATE);
         break;
 #endif
+    case BLE_LL_CTRL_PROC_DATA_LEN_UPD:
+        ble_ll_ctrl_proc_stop(connsm, BLE_LL_CTRL_PROC_DATA_LEN_UPD);
```
andrzej-kaczmarek: We should have a comment here that in fact this is workaround for broken peer device - as per spec (Vol 6 Part B section 5.1.9) collisions in Data Length Update procedure are harmless and both devices should respond as normal.

net/nimble/controller/src/ble_ll_conn_hci.c
rymanluk: well I agree with you. I will create Jira for this as there is at least one other place where we solve collision in the same way.  I will remove this patch from this PR then

net/nimble/controller/src/ble_ll_ctrl.c
rymanluk: agree

PR #788, opened 2018-02-01
nimble/dtm: Disable whitening for testing

rymanluk:


andrzej-kaczmarek:
I'm not sure why you swapped function names? I guess your intention was to have functions to enable/disable whitening, but right now you enable DTM by calling 'ble_phy_disable_dtm' which is counter-intuitive.

Anyway, I'd keep original names and have PHY API to enable/disable DTM rather than whitening only.

PR #793, opened 2018-02-06
lora: make the lora driver not dependent on lora node

wes3:
The lora node radio driver should not be dependent on the lora node code. The radio
driver should exist as an independent entity. The changes made for this commit
are to make the driver stand-alone.

Note that part of this change removed some of the statistics. These were initial
stats to get the code up and running and were useful for the initial lora apps. Soon
there will be lora mfrg fw so not sure if the original apps are still useful. If they
are, the statistics should be added to the sx1276 code and should be added to the driver
as opposed to lora node.


wes3:
Chris:

Certainly not in the form it is in. I am glad you pointed this out as it is a piece that I was going to come back to reconcile but never did. Will address this issue (not sure how yet but will add to PR when done). Thanks!

wes3:
I am going to merge this now and address the hal_timer_config issue in a different PR as I want to merge this prior to committing sx1272 driver support

PR #794, opened 2018-02-06
nimble: Fixes for extended scanner

andrzej-kaczmarek:


net/nimble/transport/ram/syscfg.yml
```diff
@@ -42,4 +42,4 @@ syscfg.defs:
         value: 255
 
 syscfg.vals.BLE_EXT_ADV:
-    BLE_HCI_EVT_BUF_SIZE: 274
+    BLE_HCI_EVT_BUF_SIZE: 257
```
rymanluk: Since now we send event in chunks to host I would remove this special case at all. Just keep old BLE_HCI_EVT_BUF_SIZE

rymanluk:
this solves MYNEWT-853

PR #803, opened 2018-02-09
compiler: Define __MYNEWT__=1 for all builds

andrzej-kaczmarek:
This patch adds -D__MYNEWT__=1 for all compiler packages so we have
common symbol defined for all Mynewt builds. This will e.g. allow to
easily recognize Mynewt build when building external repositories.


andrzej-kaczmarek:
newt tool adds similar define already...

PR #814, opened 2018-02-15
charge-control: Adding sensor-like interface for battery charge controller ICs

brianpax:
Upstreaming to allow Vipul to review/comment/modify


brianpax:
@vrahane

PR #819, opened 2018-02-18
Feature/blecsc app: Cycling Speed and Cadence app

mjurczak:
Added a new example BLE app implementing Cycling Speed and Cadence service.

The app was build from blehr app and may have inherited some issues. I consider implementation of Cycling Power sample app as well after this.

Tested with Polar V650.

### DIFF
--- a/apps/blecsc/src/main.c
+++ b/apps/blecsc/src/main.c
@@ -43,18 +43,20 @@
 #define CSC_SIM_SPEED_KPH_MAX                     35
 
 /* Log data */
-struct log blehr_log;
+struct log blecsc_log;
 
+/* Noticication status */
 static bool notify_state = false;
 
+/* Connection handle */
+static uint16_t conn_handle;
+
+/* Advertised device name  */
 static const char *device_name = "blecsc_sensor";
+static uint8_t blecsc_addr_type;
 
-static int blehr_gap_event(struct ble_gap_event *event, void *arg);
-
-static uint8_t blehr_addr_type;
-
-/* Sending notify data timer */
-static struct os_callout blecsc_tx_timer;
+/* Measurement and notification timer */
+static struct os_callout blecsc_measure_timer;
 
 /* Variable holds current CSC measurement state */
 static struct ble_csc_measurement_state csc_measurement_state;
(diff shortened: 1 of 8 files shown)


mjurczak:
One comment. I did not figure out how to implement one SC control point requirement:

_"(...)Collector writes an Op Code to the SC Control Point while the SC Control Point
characteristic is not configured for indication (i.e. via the Client Characteristic
Configuration descriptor), it will receive an ATT Error Response with the Application
error code set to **Client Characteristic Configuration Descriptor Improperly Configured**"_

It is fairly easy to check if indication is configured. But I could not find appropriate error code and I'm not sure what is the right way to pass ATT error response in such case.

net/nimble/host/services/gap/include/services/gap/ble_svc_gap.h
```diff
@@ -37,6 +37,8 @@ struct ble_hs_cfg;
 
 const char *ble_svc_gap_device_name(void);
 int ble_svc_gap_device_name_set(const char *name);
+int ble_svc_gap_device_appearance(void);
+int ble_svc_gap_device_appearance_set(const uint16_t appearance);
```
andrzej-kaczmarek: I'd actually make this a syscfg value (BLE_SVC_GAP_APPEARANCE)
the reason for this is that appearance is read-only so it cannot (and should not) be changed in runtime.
getter is ok though.

apps/blecsc/src/main.c
```diff
@@ -0,0 +1,323 @@
...
+    case BLE_GAP_EVENT_DISCONNECT:
+        BLECSC_LOG(INFO, "disconnect; reason=%d\n", event->disconnect.reason);
+        conn_handle = 0;
+        /* Connection terminated; resume advertising */
+        blecsc_advertise();
+        break;
+
+    case BLE_GAP_EVENT_ADV_COMPLETE:
+        BLECSC_LOG(INFO, "adv complete\n");
+        blecsc_advertise();
```
andrzej-kaczmarek: CSCP does not allow concurrent connections to single sensor so we should not restart advertising here - only after disconnection and failed connection

mjurczak: Fair enough. 
The spec is ab it ambiguous: "2.3 (...) There are no concurrency limitations or restrictions for the Collector and CSC Sensor roles imposed by this profile."

andrzej-kaczmarek: This means a device can implement different roles concurrently so it ca be both a sensor and a collector, which probably does not make sense, but iirc only a handful of profiles impose any concurrency restrictions only if necessary (like HoGP).

PR #824, opened 2018-02-20
Add stm32 common hal for shared drivers

utzig:
This is an effort to remove duplication in STM32 based models, moving common hal infrastructure to a shared pkg and parametrize stuff for the specific models.

NOTE: I tested this on F4, L1, F1 and F7. Extra drivers will go in new PRs. F3 was not updated (will wait for f3-discovery merge!).


mkiiskila:
hal_debugger_connected() if-statement should be behind MCU specific macro too. E.g. I'm not sure it's possible to do that check on STM32L1.

utzig:
@mkiiskila Nice catch! It should work on STM32L1 because it is a Cortex-M3. But it will probably fail on STM32L0 which is Cortex-M0+ so makes sense to make it a macro too.

mkiiskila:
Indeed, STM32L0 with it’s M0 was the one I was thinking of.

PR #829, opened 2018-02-21
add driver for ADXL345 sensor

matthewwarnes:


hw/drivers/sensors/adxl345/src/adxl345.c
```diff
@@ -0,0 +1,1262 @@
...
+ * under the License.
+ */
+
+#include <string.h>
+#include <errno.h>
+#include <assert.h>
+
+#include "defs/error.h"
+#include "os/os.h"
+#include "sysinit/sysinit.h"
```
mkiiskila: Move this line a bit higher. There are syscfg settings which impact how includes from os/os.h are read in.

matthewwarnes: I will sort the requested changes. I should also be able to add in IRQ bits, I mostly haven't done that yet as I wasn't sure how to implement it yet.

vrahane:
Sounds good. Support for thresholds and tap detection are quite useful in the Sensor API.

PR #830, opened 2018-02-21
nimble/ll: Fix BLE_LL_CONN_INIT_MIN_WIN_OFFSET

andrzej-kaczmarek:
The code which calculates earliest connection event start does not take
BLE_LL_CONN_INIT_MIN_WIN_OFFSET into account, but we still update tx
windows offset in state machine with its value. As a result when this
setting is set to non-zero value we schedule first connection event
earlier than what slave calculates from CONNECT_IND PDU so connection
cannot be established.


andrzej-kaczmarek:
@jacobrosenthal

jacobrosenthal:
Confirm this fixes problem I was seeing.

PR #822, opened 2018-02-19
Adding int/float easing library

mlaz:
This PR adds an easing library. 
This library provides the most [common easing functions usually available](http://easings.net/).

- [x] Integer version;
- [x] Float version;

PR #836, opened 2018-02-22
PWM add cycle interrupt handling to pwm_nrf52 driver

mlaz:
To-Do:

- [x] have the user set interrupt priority  
- [x] have a default and a n cycle mode, where pwm plays for n cycles.  
- [x] update pwm_test in order to have this working there (use easing lib? https://github.com/apache/mynewt-core/pull/822)
- [x] cleanup/refactor

### DIFF
--- a/hw/drivers/pwm/include/pwm/pwm.h
+++ b/hw/drivers/pwm/include/pwm/pwm.h
@@ -99,6 +99,8 @@ typedef int (*pwm_get_resolution_bits_func_t) (struct pwm_dev *);
  */
 typedef int (*pwm_disable_func_t) (struct pwm_dev *, uint8_t);
 
+typedef void (*user_handler_t) (void*);
+
 struct pwm_driver_funcs {
     pwm_configure_channel_func_t pwm_configure_channel;
     pwm_enable_duty_cycle_func_t pwm_enable_duty_cycle;
@@ -121,11 +123,16 @@ struct pwm_dev {
  *
  * pin - The pin to be assigned to this pwm channel.
  * inverted - Whether this channel's output polarity is inverted or not.
- * data - A pointer do a driver specific parameter.
+ * user_handler - A pointer to a user_handler_t function for a cycle interrupt.
+ * cycle_int_prio - The cycle interrupt priority.
+ * data - A pointer do a driver specific parameter, if cycle_handler is set it
+ *  will be passed to it as a parameter.
  */
 struct pwm_chan_cfg {
     uint8_t pin;
     bool inverted;
+    user_handler_t cycle_handler;
+    uint32_t cycle_int_prio;
     void* data;
 };
 
--- a/hw/drivers/pwm/include/pwm/pwm.h
+++ b/hw/drivers/pwm/include/pwm/pwm.h
@@ -135,7 +135,7 @@ struct pwm_dev {
  * pin - The pin to be assigned to this pwm channel.
  * inverted - Whether this channel's output polarity is inverted or not.
  * n_cycles - The number of cycles for the channel to play. 0 for loop mode.
- * interrupts - Whether interruptions are to be used or not.
+ * interrupts - Whether interruptions are to be configured or not.
  * data - Driver specific data.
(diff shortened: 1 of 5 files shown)


hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -177,9 +249,10 @@ nrf52_pwm_close(struct os_dev *odev)
 
     dev = (struct pwm_dev *) odev;
     inst_id = dev->pwm_instance_id;
+    assert(instances[inst_id].in_use);
```
jacobrosenthal: I think probably all these asserts needs to be just return codes. My understanding is that asserts are for system startup where theres no one to return to or no reason to keep going on failure, not runtime errors especially when the function already allows a return code

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -177,9 +249,10 @@ nrf52_pwm_close(struct os_dev *odev)
...
 
     nrfx_pwm_uninit(&instances[inst_id].drv_instance);
-    instances[inst_id].playing = false;
+    cleanup_instance(inst_id);
```
jacobrosenthal: I take it the assert is whats currently protecting us from a regressions from this recent commit?
https://github.com/apache/mynewt-core/pull/821/files

mlaz: The problem which the commit you mentioned fixes is the misuse of in_use flag. When you reported the problem, what was "breaking" execution was in fact an assert on pwm_open. Only yesterday I noticed that fix is not coherent with how the flag should be used.

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
ccollins476ad: I think an assert is appropriate for certain failure conditions:
* software bugs (i.e., logical contradiction)
* unlikely hardware failure

In the case of a software bug, my view is that it is best to terminate as soon as possible.  This makes it easier to identify and fix the bug, and generally there is no reasonable recovery mechanism.  For "safety critical" systems, presumably there are hardware recovery mechanisms or redundancies that protect against a software crash.

I am not familiar with this PWM driver, but it looks like this `assert` will only fail if there is a firmware bug.  There are similar asserts that check this same condition elsewhere in this file.

Anyway, I'm sure opinions vary, so that is just my take.

mlaz: > I think probably all these asserts needs to be just return codes.

Agreed, on this case it makes perfect sense to have error codes. 

IMHO, I'd say use error codes on recoverable situations, use asserts on unrecoverable situations (i.e. checking NULL pointers, array lengths, etc). I've seen here code produced under this practice, I think if everyone agrees, it would be

sterlinghughes:
is this ready for merge?

hw/drivers/pwm/include/pwm/pwm.h
```diff
@@ -121,18 +134,43 @@ struct pwm_dev {
  *
  * pin - The pin to be assigned to this pwm channel.
  * inverted - Whether this channel's output polarity is inverted or not.
- * data - A pointer do a driver specific parameter.
+ * n_cycles - The number of cycles for the channel to play. 0 for loop mode.
+ * interrupts - Whether interruptions are to be used or not.
+ * data - Driver specific data.
```
mlampert: Would it make sense to add a comment here that if `interrupts == true` this has to be a point to a `struct pwm_dev_interrupt_cfg` ?

mlampert: never mind - I misread the interface.

mlaz: Eventhough the comment on 'interrupts' covers both possible states, it is misleading, sorry.

PR #837, opened 2018-02-22
nimble/ll: Fix sending HCI Number of Completed Packets event

andrzej-kaczmarek:
The calculations for sending this HCI event periodically seem strange,
probably something was mixed up in the past here.

After fix I believe this does what it is supposed to do: every X ticks
we'll just go through all connections and send HCI event no matter if
there is something to send or not.

Also added some with reference to spec section.

### DIFF
--- a/net/nimble/controller/src/ble_ll_conn_hci.c
+++ b/net/nimble/controller/src/ble_ll_conn_hci.c
@@ -279,8 +279,8 @@ ble_ll_conn_num_comp_pkts_event_send(struct ble_ll_conn_sm *connsm)
      * have completed but there are data packets in the controller buffers
      * (i.e. enqueued in a connection state machine).
      */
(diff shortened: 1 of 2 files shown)


andrzej-kaczmarek:
Ok, so I had to write this step-by-step on a paper in order to find out that it seems to work, but I find it really hard to understand - especially the part that 'rate' is added to 'last time' at the end of function and the fact that BLE_NUM_COMP_PKT_RATE is defined in ticks but in code it is multiplied by OS_TICKS_PER_SEC only add to the mystery :-)

Also I found an issue with my change as well so in new patch version I updated only two things I find most difficult to follow.

wes3:
I certainly agree the original code is very confusing and could be written differently to make it clearer. I would have used a timer for it but wanted to save memory. So all good...

PR #838, opened 2018-02-22
bsp: puckjs

jacobrosenthal:
We dont have an nrf52 with XTAL_32768_SYNTH so figured Id do it up right and PR it.
Fun clean little button/led/ir/nfc xmit design.

Check out https://www.puck-js.com and https://www.adafruit.com/product/3372


jacobrosenthal:
Ah. that snuck in here, Ill pull it out and send that separately.

PR #844, opened 2018-02-24
Bug fix for SCL being held low after NACK on NRF52

alvarop:
For some reason SCL is held low by the nrf52832 after a NACK.
For an example of the bug (and fix) using the NRF52, see:
https://github.com/alvarop/i2ctest


andrzej-kaczmarek:
Did you try to just trigger TASK_STOP in case of error instead of full reset? Perhaps TWI is just waiting for more data, regardless of an error. Just wondering, I don't really know.

alvarop:
@andrzej-kaczmarek You were absolutely correct! That worked as well.
I've updated the CL (and also added another missing STOP in the read timeout handling)

Issue #856, opened 2018-02-28
nrf52840 hal spi is broken


wes3:
The current hal spi for the nordic nrf52840 is broken as it assumes SPI pins will use port 0 (pins 0 - 31). If pins greater than 31 are used incorrect ports will be accessed.

PR #857, opened 2018-02-28
hw/mcu/nordic/nrf52xxx: HAL spi broken for 52840.

wes3:
The incorrect port was being accessed if any SPI pin was using
a pin number greater than 31 (on port 1). This issues should be
fixed with this commit.


wes3:
Fixes #856

PR #861, opened 2018-03-01
bsp: allow override of i2c, spis and spim pins

jacobrosenthal:
Uart pins are allowed to be overridden but not i2c and spis/spim. If this is wanted Ill do the other boards as well

### DIFF
--- a/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
+++ b/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
@@ -79,18 +79,18 @@ static const struct uart_bitbang_conf os_bsp_uart1_cfg = {
  * and is handled outside the SPI routines.
  */
 static const struct nrf52_hal_spi_cfg os_bsp_spi0m_cfg = {
-    .sck_pin      = 12,
-    .mosi_pin     = 13,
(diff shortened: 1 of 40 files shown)


wes3:
Changes look good to me.

jacobrosenthal:
Hopefully I havent broken anything. But as a result Ive left nrf51-blenano and calliope_mini unchanged as they had existing definitions for i2c freq and I didnt want to break them (yet)

Also of note, we have another inconsistency in that many nrf51 boards seem to create a spi_0_master and spi_1_slave but nrf52 boards seem to create an spi_0_slave. We could attempt to clean this up now, but again that would break existing bsps...

wes3:
I guess I should have commented on this here. The nrf51 does not have a spi0 slave; that is the reason.

wes3:
@jacobrosenthal After seeing all the files that have changed (my original "looks good" was only after seeing changes to one bsp) I am thinking that it would probably have been best to put the MYNEWT_VAL definitions for the pins in the MCU as opposed to each BSP. I wonder what others think? Seems to make sense to me as it will mean we will have to duplicate less stuff in each bsp syscfg.

PR #867, opened 2018-03-02
Add option do configure spi master/slave number

utzig:
This adds new syscfg options to pass the number of the slave/master peripheral to use in the tests. This customization is required when testing on boards that have SPI0 pins already used for other purposes or not easily reachable in the conn headers.

This also allows for using a single board doing spi testing between two different SPI peripherals.


utzig:
I will try to test master and slave on the same board before I merge this, thanks for the review @wes3

utzig:
TODO: still needs testing on a single board with SPI_0 <-> SPI_1 communication, to be done in the near future.

PR #870, opened 2018-03-06
nimble/phy: Fix receiver configuration on late RX

andrzej-kaczmarek:
If we are late when scheduling RX, receiver is enabled anyway since we
may just want to scan for "anything". However, current code only enables
receiver but does not configure it so the results seem unpredictable.

This patch fixes this by ensuring that radio is properly configured for
RX in both cases.


andrzej-kaczmarek:
to clarify: I do not see any particular problem due to this potential issue - found this by looking through code and was puzzled how this could actually work.

PR #873, opened 2018-03-06
[WIP] sys/metrics: Add framework for logging events data

andrzej-kaczmarek:
This is proposed framework for logging aggregated data for events.

The idea is simple: there is an event which has multiple metrics we want to log. Each metric is either a simple value (overwritten each time) or data series (appended each time). When event is (re)started each data logged from now on is aggregated internally. At the end of event its contents can be serialized using CBOR and written to mbuf for further processing or writing to logging subsystem (need to add append_mbuf API to logging for this). Now event can be restarted and this starts again.

There is no limit for number of events other than memory capacity.
There is tentative limit for number of metrics for each event (32) due to internal state variabled, but can be increased in future if necessary.

There is sample app which shows intended usage, however there is one thing that needs more explanation. Once metrics are defined, each have name and value type, e.g.:
```
METRICS_SECT_START(my_metrics)
    METRICS_SECT_ENTRY(val1, EVENT_METRIC_TYPE_SINGLE)
    METRICS_SECT_ENTRY(u8, EVENT_METRIC_TYPE_SERIES_U8)
    METRICS_SECT_ENTRY(u16, EVENT_METRIC_TYPE_SERIES_U16)
    METRICS_SECT_ENTRY(u32, EVENT_METRIC_TYPE_SERIES_U32)
METRICS_SECT_END;
```
For data series there are separate types for storing 8, 16 and 32-bit values since it does make difference in terms of memory usage. For single values it does not really matter so internally this is always stored as 32-bit value.
The metric name is, unfortunately, only used for serializing and accessing it via e.g. shell (tbd) as it does not generate any symbol to actually pass it to APIs. Thus, APIs identify each metric by ordinal number - as defined, starting from 0. I'd really like to make this smarter, but without adding to much extra macros or need for additional definitions in user code - have no idea at the moment so any comments are welcome here. As a workaround, app can define a simple enum with symbols corresponding to each metric and since its auto-numbered from 0 it can be easily rearranged as required.

### DIFF
--- a/apps/metrics/pkg.yml
+++ b/apps/metrics/pkg.yml
@@ -1,0 +1,29 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+pkg.name: apps/metrics
+pkg.type: app
+pkg.description: 
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+
+pkg.deps:
+    - kernel/os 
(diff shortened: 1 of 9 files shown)


sterlinghughes:
cc @InternetOfAwesome, @wes3, @vrahane, @ccollins476ad

andrzej-kaczmarek:
latest commits are not cleaned up nicely so I'd recommend to look through all the changes at once - will fix this later

andrzej-kaczmarek:
This now depends on changes from PR #933 (included here but apparently GitHub was not smart enough to figure this out automatically).

InternetofAwesome:
Is there anything blocking getting this in? I am waiting on this PR to get some application level logging merged in in our code.

sterlinghughes:
Can we update & merge this?

sterlinghughes:
hey @andrzej-kaczmarek -- just wanted to see about updating & merging.

Also, one thought occurred to me: it would be good to capture the start time and end time of the metrics events in wall-clock time if available, that way we'd know when things actually happened.

PR #888, opened 2018-03-08
kernel/os: Make sure correct stack aligment 

rymanluk:
This PR has been done after testing of  #882

### DIFF
--- a/kernel/os/include/os/os_task.h
+++ b/kernel/os/include/os/os_task.h
@@ -36,6 +36,10 @@
 extern "C" {
 #endif
 
+#define OS_TASK_STACK_DEFINE(__name, __size) \
+    static os_stack_t __name [OS_STACK_ALIGN(__size)] \
+        __attribute__((aligned(OS_STACK_ALIGNMENT)));
+
 /** Highest priority task */
 #define OS_TASK_PRI_HIGHEST (0)
 /** Lowest priority task */
--- a/kernel/os/src/os.c
+++ b/kernel/os/src/os.c
@@ -40,12 +40,12 @@
  */
 
 struct os_task g_idle_task;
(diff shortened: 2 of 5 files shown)


sterlinghughes:
I agree - I don't know who this DENIFE is.  :-) Why is attribute(aligned) required, shouldn't OS_STACK_ALIGN and the size of os_stack_t give you that?

andrzej-kaczmarek:
@sterlinghughes `OS_STACK_ALIGN` will align size to 8 bytes, but since `os_stack_t` is `uint32_t` array will be aligned to 4 bytes only. We could change `os_stack_t` to `uint64_t` but I don't think this makes sense since word size is still 32-bits, just the alignment should be different.

sterlinghughes:
makes sense.  @rymanluk can you fix typo, and then we can merge?

sterlinghughes:
btw, i can't believe this is the only place we define OS stacks.  we should definitely highlight this for users, as this will affect anybody who declares a task and uses floating point.

ccollins476ad:
Good catch.  It looks like there is another related issue.  `os_malloc()` doesn't always return 8-byte aligned blocks.  Some apps (e.g., slinky) malloc task stacks.  If the allocated stack is only 4-byte aligned, then we'll have this same issue.  This happened to me when I was testing the floating point printf PR with the slinky app.

Just for completeness (n1570, 7.22.3 (Memory management functions)):
> The pointer returned if the allocation succeeds is suitably aligned so that it may be assigned to a pointer to any type of object with a fundamental alignment requirement

I will file a separate PR for this.

rymanluk:
fixed that typo - ehh. Please review.
I planned to update documentation anyway, will do it

PR #889, opened 2018-03-08
implementation of common button events

sdalu:
Common dev board or iot projects usually have a minimal set of buttons (1 or 2),  This package will ease the use of theses buttons by handling common user interaction (click, double click, long click, auto-repeating, 3 button emulation). Code size can be reduced by using various syscfg.

### DIFF
--- a/hw/util/button/include/button/button.h
+++ b/hw/util/button/include/button/button.h
@@ -1,0 +1,374 @@
+#ifndef _BUTTON_H_
+#define _BUTTON_H_
+
+#include <stdbool.h>
+#include <os/os.h>
+
+/**
+ * This library generate events in the default queue to deal 
+ * with user interactions. These events need to be processed
+ * (by the main program or a thread) so that the associated 
+ * callback is triggered.
+ *
+ * According to the selected mode it is possible to deal with actions
+ * such as: click, double click, long click, long double click.
+ * And to make the last action auto-repeat (like a key).
+ *
+ * It can also generate event to reflect the actual button state,
+ * which is a mix of: pressed / doubled / long / repeating.
+ *
+ * The action differ from the state, as the action is generated
+ * when the button is released or after a small timeout.
+ *
+ * When a button is actively participating in an emulation, the state
+ * change event is emited but the action is not.
+ * For example button 3 = button 1 + button 2
+ *    a simultaneous click on button 1 and 2, will emit 
+ *    a click for button 3, but neither for 1 or 2.
+ *
+ *
+ *
+ * Small example, using the debounce package to drive the buttons.
+ *
+ * 
+ *
+ * button_t btns[] = {
+ *    { .id       = 1, // optional
+ *      .children = (button_t *[]){ &btns[2], NULL },
+ *      .mode     = BUTTON_MODE_MOUSE ,
+ *    },
+ *    { .id       = 2, // optional
+ *      .children = (button_t *[]){ &btns[2], NULL },
+ *      .mode     = BUTTON_MODE_TOUCH | BUTTON_FLG_REPEATING, 
+ *    },
+ *    { .id       = 3, // optional
+ *      .emulated = (button_t *[]){ &btns[0], &btns[1], NULL },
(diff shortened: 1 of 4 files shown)


sterlinghughes:
Stephane -- thank you for this pull request, it looks like a great addition.  I have a number of comments regarding coding standards and documentation, but generally I'd love to have it as a part of Mynewt.

As a small nit, are you OK with renaming it "button" versus "btn" -- I think button is not _that_ many more characters and reads nicer.  I'll make more comments inline.

util/btn/include/btn/btn.h
```diff
@@ -0,0 +1,148 @@
...
+	uint8_t mode;
+	uint8_t fsm_state;
+#if MYNEWT_VAL(BTN_USE_EVENT_FILTERING) &&	\
+    ( MYNEWT_VAL(BTN_GENERATE_EVENT_STATE) ||	\
+      MYNEWT_VAL(BTN_GENERATE_EVENT_ACTION) )
+	struct {
+	    bool enabled;
+#if MYNEWT_VAL(BTN_GENERATE_EVENT_STATE)
+	    uint8_t state;
+#endif
```
sterlinghughes: For smaller variables like this, I'd prefer to just have them always declared in the struct itself.  You are only wasting a couple of bytes for a few buttons, and it makes it easier to debug not to have variable elements in structures.  I think the button emulation above makes sense to #ifdef out, but these variables should probably just be there IMO.

util/btn/include/btn/btn.h
```diff
@@ -0,0 +1,148 @@
...
+	} filter;
+#endif
+#if MYNEWT_VAL(BTN_USE_DOUBLE) || \
+    MYNEWT_VAL(BTN_USE_LONG  ) || \
+    MYNEWT_VAL(BTN_USE_REPEAT)
+	struct os_callout callout;
+#endif
+    } notify;
+} btn_t;
```
sterlinghughes: Can you add doxygen documentation to the header file, this will make it easier to autogenerate this as a part of the docs.

sterlinghughes:
also, in general, should this maybe go into drivers?  I know there isn't a ton of HW specific code, but it seems to me to be more of a driver... interested in thoughts.

jacobrosenthal:
without showing any preference, we have another debounce 'driver' right now
https://github.com/apache/mynewt-core/tree/master/hw/drivers/debounce

sterlinghughes:
I think this actually leverages the debounce driver for the button interaction, no?

sdalu:
Improved the doxygen documentation (I didn't documented fields only used internally). Renamed btn to button, removed some #ifdef .

I'm wondering if the dependency with the debounce package should be removed and let the programmer deal with the few lines of code (that's only initialising the debounce driver and have it call button_handler)

sdalu:
I preferred to remove the dependency with the debounce package: this simplified initialisation structure,  and  perhaps is less confusing about it's role (more an utility than a hardware driver)

sdalu:
Let me know if you have other feedback

sterlinghughes:
Hi Stephane,  there were a bunch of coding style comments I originally made.  I'd really like to merge this PR, so I can either merge & then fix coding style and then re-push -- or it would be great if you could make the changes.  Let me know if you'd be willing to do this, we'd love to get you submitting lots of PRs to Mynewt :-) 

The major ones are:

- declaring function brace on a new line 
i.e.

int 
bla(void) 
{

and not

int 
bla(void) {

- Variables should only be assigned constants on declaration at the top of a function:

i.e.

int 
bla(int *c)
{
  int b;

   b = *c;
}

and not:

int 
bla(int *c)
{
    int b = *c;
}

But OK is:

int 
bla(int *c)
{
  int a = 10;

sdalu:
I guess it's the last time to decide, if it should go to util or hw? Has anybody some preferences?

sterlinghughes:
my preference would be in  hw/util/button, but I'd be fine with util as well.

sdalu:
moved to hw/util

PR #893, opened 2018-03-09
libc: Conditionally compile float support in printf

michal-narajowski:
Libs doing float computations take a lot of text space,
so it shouldn't be compiled by default.


mkiiskila:
Test this please

PR #897, opened 2018-03-09
Move cmsis_nvic.h files frm BSP to MCU

utzig:
This moves `cmsis_nvic.h` from BSPs to MCUs. Where customizations where necessary, new `syscfg` options were added to the MCU to be overridden by the BSP, eg, number of peripheral interrupts. This can be seen for `stm32f4` and `stm32f7`...

I made a build script that builds `mcuboot` + `slinky` for all BSPs. A lot of them are not building anymore which I will fix in other PRs (it is not related to this change, also happens on master). I ended up actually testing this PR (as in: install, verify, etc) on stm32f4discovery, stm32f429discovery, stm32f767-nucleo, ble400 (nrf51), nrf52dk, frdm_k64f, stm32l1discovery and nucleo-401re.

PS: I am fairly confident this is OK for all MCUs but it *might* break nrf52840pdk, so I would like someone to test this board (will most probably work!). Also the nrf52dk was the only board where `slinky` shell did not work, but it also happens on master here, so probably not related to this PR.

PS: I have not updated porting documentation. Nice to have reviews before I do that...


utzig:
@mlaz ^

sterlinghughes:
yes.

utzig:
A little food for thought: this PR may be a good start, but it could still be considered *too much* duplication. What could be done instead (or later!) is to have a `mcu/cortex-m` pkg that every Cortex-Mx `mcu` depends on, which itself carries the `bsp/cmsis_nvic.h` file using `syscfg`s for configurations that MCUs+BSPs would just need to override...

utzig:
Btw, the vector amount config is only for `stm32` right now...

wes3:
OK, I suspected that might be the case. Again, just my opinion, but I think having defines in cmsis_nvic.h for the various MCU's. For example:

#ifdef __STM32F407xx_H
#define NUM_VECTORS 81
#endif
 
Excuse my shorthand but you get the idea. For some reason seeing the #include syscfg.h in cmsis_nvic.h bothers me.

ccollins476ad:
Looks great!  It's good to delete so much code.  I agree with Will that it is better if the number of interrupts isn't configurable.

vrahane:
Agreed, this looks great. Making the BSP simpler sounds like a good idea. There might be more things we would want to move to the mcu which should probably all be handled as part of the same release since this kind of breaks backwards compatibility for custom BSPs and having most of the backwards incompatible changes in a single release makes sense so that BSPs do not keep breaking every release. Not sure when or how that should happen but might be this is one of many changes we should consider.

utzig:
Hardcoded stm32f4x and stm32f7x number of vectors with 2f99168

jacobrosenthal:
Speaking of duplication, the empty boards.h in nordic targets was me back in the day
https://github.com/apache/mynewt-core/commit/c37a30581aa7cbbb763baf31e9000d32f75f40ce

I just tested and with new sdk Im not seeing any errors on an nrf51 or nrf52 target

Shall we wipe that out too?

wes3:
Removal of boards.h has been a long time coming...

utzig:
Cmon, I agree, but not here! :P

wes3:
Oh sorry; I did not mean to add it to this PR necessarily. Just that I thought it a good idea.

PR #901, opened 2018-03-10
ble service: Device Information Service

sdalu:
Main characteristics for the Device Information Service

### DIFF
--- a/net/nimble/host/services/dis/src/ble_svc_dis.c
+++ b/net/nimble/host/services/dis/src/ble_svc_dis.c
@@ -1,0 +1,178 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <assert.h>
+#include <string.h>
+#include "sysinit/sysinit.h"
+#include "host/ble_hs.h"
+#include "services/dis/ble_svc_dis.h"
+
+
+#if (MYNEWT_VAL(BLE_SVC_DIS_MODEL_NUMBER     ) == 0) && \
+    (MYNEWT_VAL(BLE_SVC_DIS_SERIAL_NUMBER    ) == 0) && \
+    (MYNEWT_VAL(BLE_SVC_DIS_HARDWARE_REVISION) == 0) && \
+    (MYNEWT_VAL(BLE_SVC_DIS_FIRMWARE_REVISION) == 0) && \
+    (MYNEWT_VAL(BLE_SVC_DIS_SOFTWARE_REVISION) == 0) && \
+    (MYNEWT_VAL(BLE_SVC_DIS_MANUFACTURER_NAME) == 0)
+#error "Remove the ble_svc_dis package or use one of the characteristics"
+#endif
+
+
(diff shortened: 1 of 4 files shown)


andrzej-kaczmarek:
Few comments:
- characteristics permissions should be configurable - for consistency, I'd make it the same way as configurable write permissions in GAP service so there's no need for separate enable and permissions syscfgs
- also for consistency I'd use helpers to set values instead of using extern struct
- there's no need for #error in case none of characteristics were enabled - this is explicitly allowed by spec
- default value for characteristics with no value set should be an empty string
- there are multiple consecutive empty lines in few places - use single empty line as separator

As an addition, it may be useful to have default value for some/each characteristics in syscfg (same as e.g. Device Name in GAP) so there's no need to set all of them in application.

sdalu:
* I think "Model Number", "Manufacturer Name" will generally be know at compile time and won't change, so I don't feel comfortable at preallocating memory as for the Device Name in GAP. Is there a preprocessor trick to do conditional with string value?
* For the Firmware/Software we could use the image versioning, but I'm not sure which one makes more sense.

jacobrosenthal:
@sdalu I took a shot at one of these a while ago and got busy and never PRed so thanks!

BTW I ended up finding newt's 'config' stuff, and filled a bunch of these from there
https://github.com/jacobrosenthal/mynewt-nimble-services/blob/master/services/dis/src/ble_svc_dis.c#L105

None of these are 'right', I just wanted to point out I think this is where a lot of these values are stored 'the newt way' right now for people who use newtmgr

net/nimble/host/services/dis/syscfg.yml
```diff
@@ -0,0 +1,80 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# Package: ble/host/services/dis
+
+syscfg.defs:
+    BLE_SVC_DIS_MODEL_NUMBER_READ_PERM:
+        description: >
+            Defines permissions for writing "Model Number" characteristics. Can
```
jacobrosenthal: Same thing here, these should all be Defines permissions for *reading

net/nimble/host/services/dis/src/ble_svc_dis.c
```diff
@@ -0,0 +1,249 @@
...
+        }, }
+    },
+
+    {
+        0, /* No more services. */
+    },
+};
+
+/**
+ * Simple write access callback for the device information service
```
jacobrosenthal: *read access callback

net/nimble/host/services/dis/syscfg.yml
andrzej-kaczmarek: Also looking at this now I wonder if we should have just one read_perm syscfg for all characteristics and enable for each characteristic separately? I am not sure if there is scenario for setting different permissions for each characteristic here.

Not really sure what is the proper solution here, just thinking out loud.

andrzej-kaczmarek:
My proposal with default values was not to propagate each characteristic with "computed" value but simply have syscfg with simple string for all/some characteristics - just to e.g. define manufacturer or model name from app or settings easily. I think DIS implementation should be just generic one as it is now.

@jacobrosenthal I think reading these values from config and configuring DIS accordingly could be a nice example to include in one of sample apps - blehr or blecsc perhaps?

jacobrosenthal:
@andrzej-kaczmarek Yeah either seems cool to me. We should figure out what some good newt defaults would be, that was my first pass a year ago.

@sdalu  No opinon on some defaults here, Technically theyre all optional in the spec so you could do it empty, so maybe ifdef guard this?

But currently if the user doesnt set any syscfg read permissions you get a compile error so well have to guard or default
```
Error: repos/apache-mynewt-core/net/nimble/host/services/dis/src/ble_svc_dis.c:101:1: error: 'ble_svc_dis_access' defined but not used [-Werror=unused-function]
 ble_svc_dis_access(uint16_t conn_handle, uint16_t attr_handle,
 ^
```

net/nimble/host/services/dis/syscfg.yml
sdalu: Perhaps a `BLE_SVC_DIS_DEFAULT_READ_PERM`?
And propose the `MYNEWT_VAL_BLE_SVC_DIS_DEFAULT_READ_PERM` as a possible alternative for the `BLE_GATT_CHR_F_READ_*` values in `BLE_SVC_DIS_*_READ_PERM`

sdalu:
See: https://github.com/apache/mynewt-nimble/pull/12

PR #902, opened 2018-03-10
Feature: stm32 f3/f7 pll clk off HSI

mlampert:
This change adds support to use the internal PLL as the system clock, driven by the HSI.
The default for this setting is currently off to preserve existing behaviour.

It also consolidates the STM32F3 external file location with the other STM32 ports.


utzig:
Looks good. Would you mind also applying the changes you did for the `nucleo-f303*` to `stm32f3discovery`?

mlampert:
I certainly can do that - but you'll have to review very carefully, I have no way of testing.
Do you also want me to add support to the `stm32f7discovery` (also have none) ?

utzig:
Ok, this now has conflicts because I removed `hal_system.c` from all models to a common location shared between all. Yours added the clock config code so I would suggest you make a new file like `hal_system_clock.c` for example when rebasing on latest master.

mlampert:
way ahead of you ;)

hw/mcu/stm/stm32f3xx/src/hal_gpio.c
```diff
@@ -187,69 +187,69 @@ hal_gpio_clk_enable(uint32_t port_idx)
...
             break;
+#endif
 #if HAL_GPIO_PORT_COUNT > 3
         case 3:
             if (!__HAL_RCC_GPIOD_IS_CLK_ENABLED()) {
                 __HAL_RCC_GPIOD_CLK_ENABLE();
             }
             break;
+#endif
 #if HAL_GPIO_PORT_COUNT > 4 && defined GPIOE_BASE
```
utzig: Cool catching this! I would just have changed this line to:

```
#if HAL_GPIO_PORT_COUNT > 4
  #if defined(GPIOE_BASE)
  ...
  #endif
#if HAL_GPIO_PORT_COUNT > 5
```
I believe it would fix the same issue without moving all those `#endif` lines around.

utzig:
Thanks a lot for this!

PR #903, opened 2018-03-11
service/ans: don't depent on libm for creating bitmask

sdalu:
Removed pow function, and math.h include

### DIFF
--- a/net/nimble/host/services/ans/src/ble_svc_ans.c
+++ b/net/nimble/host/services/ans/src/ble_svc_ans.c
@@ -19,19 +19,12 @@
 
 #include <assert.h>
 #include <string.h>
-#include <math.h>
-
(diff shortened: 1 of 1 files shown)


sdalu:
I'm not particularly good at git, could you give me the command line to do it.

utzig:
Yes, you can "squash and merge" and rewrite the commit message directly in the interface.

andrzej-kaczmarek:
@sdalu In general if you just want to fix something in your last commit then do `git commit --amend` and then push the result to the same branch with `-f` option - this will replace your remote branch with fixed commit and PR will be also updated. This way for simple fixes we'll just have single commit with all changes instead of series of fixes-for-fixes-for-fix. For editing/squashing multiple commits take a look here: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History.

For this simple change I guess we can squash from GitHub ;-)

PR #904, opened 2018-03-11
ble service: Battery Service

sdalu:
Main characteristics for the Battery Service

### DIFF
--- a/net/nimble/host/services/bas/src/ble_svc_bas.c
+++ b/net/nimble/host/services/bas/src/ble_svc_bas.c
@@ -27,7 +27,7 @@
 #include "services/bas/ble_svc_bas.h"
 
 /* Charachteristic value handles */
-#if MYNEWT_VAL(BLE_SVC_BAS_BATTERY_LEVEL_NOTIFY) > 0
+#if MYNEWT_VAL(BLE_SVC_BAS_BATTERY_LEVEL_NOTIFY_ENABLE) > 0
 static uint16_t ble_svc_bas_battery_handle;
 #endif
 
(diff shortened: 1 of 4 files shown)


net/nimble/host/services/bas/syscfg.yml
```diff
@@ -0,0 +1,32 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+syscfg.defs:
+    BLE_SVC_BAS_BATTERY_LEVEL_PERM:
+        description: >
+            Defines permissions for writing "Battery Level" characteristics. Can
+            be zero to allow write without extra permissions or combination of:
```
jacobrosenthal: Defines permissions for *reading "Battery Level"
and
Can be zero to allow *read without extra permissions or combination of

And maybe BLE_SVC_BAS_BATTERY_LEVEL_PERM should be BLE_SVC_BAS_BATTERY_LEVEL_READ_PERM but I dont particularly care either way.

andrzej-kaczmarek: +1 for READ_PERM

net/nimble/host/services/bas/syscfg.yml
```diff
@@ -0,0 +1,32 @@
...
+                BLE_GATT_CHR_F_READ_ENC
+                BLE_GATT_CHR_F_READ_AUTHEN
+                BLE_GATT_CHR_F_READ_AUTHOR
+        value: 0
+    BLE_SVC_BAS_BATTERY_LEVEL_NOTIFY:
+        description: >
+            Set to BLE_GATT_CHR_F_NOTIFY to support notification,
+            or 0 to disable it.
+        value: BLE_GATT_CHR_F_NOTIFY
```
andrzej-kaczmarek: this should be a simple 0/1 value - it's not a permissions mask, just simple bool to either enable or disable notify prop

net/nimble/host/services/bas/src/ble_svc_bas.c
```diff
@@ -0,0 +1,148 @@
...
+ * If greater than 100, it will be silently set to 100.
+ */
+int
+ble_svc_bas_battery_level_set(uint8_t level) {
+    if (level > 100)
+	level = 100;
+    ble_svc_bas_battery_level = level;
+#if MYNEWT_VAL(BLE_SVC_BAS_BATTERY_LEVEL_NOTIFY) > 0
+    return ble_gattc_notify(ble_svc_bas_conn_handle, 
+                            ble_svc_bas_battery_handle);
```
andrzej-kaczmarek: we probably should not notify if value did not change, i.e. if ble_svc_bas_battery_level == level

net/nimble/host/services/bas/syscfg.yml
jacobrosenthal: or I suppose the spec allows to permission these? So you could make it a permissions mask?

andrzej-kaczmarek: Based on some informations scattered in different places of spec, notifications should most likely follow permissions of CCCD. However, NimBLE does not allow to set them and obviously also does not enforce them so there's nothing to set here actually :)

I'd leave this as NOTIFY_ENABLE 0/1 syscfg for now and I'll look into adding appropriate support in Mynewt - we can then add NOTIFY_PERM syscfg and update code accordingly.

sdalu:
See: https://github.com/apache/mynewt-nimble/pull/13

PR #907, opened 2018-03-12
nimble: Remove NimBLE code from core repository

andrzej-kaczmarek:
This PR removes NimBLE code from core repository and sets dependency for each package to new apache-mynewt-nimble repository. Latest `newt` is required for this to work properly, i.e. download dependency and build code using external repository.

Current master @ apache-mynewt-nimble is basically rewritten history of apache-mynewt-core with only commits related to NimBLE included. Each commit has `X-Original-Commit` added with commit id of source commit in apache-mynewt-core repository to make lookups easier.


jacobrosenthal:
no problems here with a basic bleprph test on an nrf51 device.

andrzej-kaczmarek:
PR is now updated to latest master and repository.yml points to proper repository for apache-mynewt-nimble (i.e. master on apache/mynewt-nimble).

PR #910, opened 2018-03-12
Common hal for stm32f3

utzig:
This makes stm32f3 part of common HAL, also fixes small issues with flash and flash areas on stm32f3discovery.


hw/mcu/stm/stm32f3xx/src/hal_flash.c
```diff
@@ -127,7 +127,7 @@ stm32f3_flash_dev()
         stm32f3_flash_dev_.hf_base_addr = FLASH_BASE;
         stm32f3_flash_dev_.hf_size = HAL_FLASH_SIZE;
         stm32f3_flash_dev_.hf_sector_cnt = HAL_FLASH_SIZE / HAL_FLASH_SECTOR_SIZE;
-        stm32f3_flash_dev_.hf_align = FLASH_PAGE_SIZE;
+        stm32f3_flash_dev_.hf_align = 2;
```
mlampert: I'm sure you know what you're doing - I would like to understand as well. Could you explain why the alignment is `2` and doesn't have to be the `FLASH_PAGE_SIZE` please?

mlampert:
I like your refactoring better than mine - I'll rebase the PLL clock on your changes once they're in.

hw/mcu/stm/stm32f3xx/src/hal_flash.c
utzig: The naming might confuse but this is the `write_size` of the flash. Since the F3 is using half-word writes the correct value here is `2`.

mlampert: got it - thanks for the clarification.

PR #912, opened 2018-03-13
nrf targets: remove empty boards.h

jacobrosenthal:
As discussed within https://github.com/apache/mynewt-core/pull/897 boards.h doesnt seem necessary with current nrfx sdx and can be removed


jacobrosenthal:
Nope. Well not in newt, I cant recall but the old sdk apparently wanted a boards.h file somewhere in it.

PR #914, opened 2018-03-13
boot_serial - Allow UART string for serial mode.

ccollins476ad:
*Note: This feature was requested by a user.  If we are unsure if this feature belongs in the boot loader in its current state, I am happy to just point him to this branch in the meantime.  Also, maybe this should be filed against MCUboot instead?*
 
If enabled, the serial boot loader listens on the UART for a string for a configurable amount of time.  If the string is detected, the boot loader enters serial mode.  If the string is not detected when the timeout expires, the boot loader proceeds in normal (non-serial) mode.

This feature is intended for setups that lack a suitable pin to use for enabling serial mode (either the pins are inaccessible or inconvenient to use).

Enabling this feature added 248 bytes to the boot loader binary when built on the nRF52dk with the default string setting ("nmgr").

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -491,6 +491,70 @@ boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
     return 0;
 }
 
+#if MYNEWT_VAL(BOOT_SERIAL_DETECT_TIMEOUT) != 0
+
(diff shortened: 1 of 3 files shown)


shankuuuu:
Tested OK this feature. I believe it is a good feature to be present in the boot loader where only UART port is exposed to outside world.

sterlinghughes:
chris, can you merge this & also submit similar to mcuboot?

PR #915, opened 2018-03-14
Expose timeval <-> clocktime conversion functions from time/datetime library

ncasaril:
The conversion functions used internally in the time/datetime library are quite useful. I see no reason why they should not be exposed. Hence this small PR :)


ncasaril:
The original datetime used -1 as error return value. I changed it to OS_EINVAL in hopes to make it conform.

PR #919, opened 2018-03-15
nimble/mesh: Fix simulator segmentation fault on mesh init

rymanluk:


sterlinghughes:
when r we fixing conflicts & merging?

PR #921, opened 2018-03-15
datetime: export clocktime

sdalu:
Exporting clocktime structure and conversion from/to timeval
These methods can be usefull outside the parsing context of `datetime_parse`. For example to implement the bluetooth current time service


sdalu:
Seems I'm one day late....

PR #925, opened 2018-03-15
Fix stm32f1xx/olimex-p103 UART

utzig:
This fixes the UART configuration on Olimex STM32-P103 and also fixes the uart driver for STM32F1 which uses a slightly different mechanism for configuring alternate function for GPIOs. Hopefully this will live a short life and be replaced by a unified stm32-common uart driver!


utzig:
Any feedback here?

PR #926, opened 2018-03-15
Add Driver for LIS2DW12

matthewwarnes:


jacobrosenthal:
Im about to do an LIS2DS12 myself. Was this a fork of the existing lis2dh? How different did you find them?

matthewwarnes:
I did start from the LIS2DH. There were some similarities but a lot of the register layout was different. Also the available interrupts were quite different.

jacobrosenthal:
As it forks from @vrahane work on the lis2dh pinging him

vrahane:
Thank you @jacobrosenthal ... will take a look at it in detail. Looks good for the most part.

vrahane:
@matthewwarnes , couples of things: 
As our understanding grows for sensors, we try to add more functionality to old sensors. One such example is BMA253. With the recent changes to the bma253 driver what happens is that sensor_read() either does a poll read or a stream read (interrupt based) based on a config setting. I would like to see something like that happen in this driver as well. I am going through it more to understand if there is something that would need more attention.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -0,0 +1,1664 @@
...
+    if (rc) {
+        goto err;
+    }
+
+    rc = lis2dw12_write8(itf, LIS2DW12_REG_CTRL_REG7, 0x30);
+    if (rc) {
+        goto err;
+    }
+
+    rc = lis2dw12_set_int1_pin_cfg(itf, 0);
```
vrahane: I think the configuration for the interrupts should come from the lis2dw12_cfg which is not the case. I agree that none of it should be enabled at init time but everytime a change to the config is made, you want to be able to change interrupt config settings, which should not be 0 here.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -0,0 +1,1664 @@
...
+    if (rc) {
+        goto err;
+    }
+
+    rc = lis2dw12_set_int2_pin_cfg(itf, 0);
```
vrahane: I think the configuration for the interrupts should come from the lis2dw12_cfg which is not the case. I agree that none of it should be enabled at init time but everytime a change to the config is made, you want to be able to change interrupt config settings, which should not be 0 here.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1136,45 +1310,78 @@ init_intpin(struct lis2dw12 * lis2dw12, hal_gpio_irq_handler_t handler,
...
     itf = SENSOR_GET_ITF(sensor);
+    pdd = &lis2dw12->pdd;
+
+    pdd->int_enable &= ~int_to_disable;
     
     /* disable int pin */
-    hal_gpio_irq_disable(lis2dw12->sensor.s_itf.si_ints[lis2dw12->pdd.int_num]
-                         .host_pin);
+    if (pdd->int_enable == 0) {
+        hal_gpio_irq_disable(lis2dw12->sensor.s_itf.si_ints[lis2dw12->pdd.int_num]
```
sterlinghughes: can't this be taken directly off of itf, assigned above?   that would also solve the wrap around structure access?

PR #928, opened 2018-03-15
datetime: use timezone for clocktime_to_timeval

sdalu:
Take in account time zone for converting clocktime <-> timeval. Was use in timeval_to_clocktime but not clocktime_to_timeval

### DIFF
--- a/time/datetime/include/datetime/datetime.h
+++ b/time/datetime/include/datetime/datetime.h
@@ -68,7 +68,7 @@ int datetime_parse(const char *input, struct os_timeval *utctime,
  *
  * Returns 0 on success and OS_EINVAL on invalid input data
  */
-int clocktime_to_timeval(const struct clocktime *ct, struct os_timeval *tv);
+int clocktime_to_timeval(const struct clocktime *ct, const struct os_timezone *tz, struct os_timeval *tv);
 
 /*
  * Converts from struct os_timeval to struct clocktime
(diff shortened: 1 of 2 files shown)


time/datetime/include/datetime/datetime.h
```diff
@@ -68,7 +68,7 @@ int datetime_parse(const char *input, struct os_timeval *utctime,
  *
  * Returns 0 on success and OS_EINVAL on invalid input data
  */
-int clocktime_to_timeval(const struct clocktime *ct, struct os_timeval *tv);
+int clocktime_to_timeval(const struct clocktime *ct, const struct os_timezone *tz, struct os_timeval *tv);
```
utzig: I'm not a big fan of changing APIs, but since this was only recently exported I think this is not a problem.

time/datetime/src/datetime.c
```diff
@@ -140,6 +140,10 @@ clocktime_to_timeval(const struct clocktime *ct, struct os_timeval *tv)
         ct->sec;
     tv->tv_usec = ct->usec;
 
+    /* Convert localtime to utctime */
+    tv->tv_sec += tz->tz_minuteswest * 60;
```
utzig: Should allow for passing no `tz` (aka `NULL`).

time/datetime/include/datetime/datetime.h
sdalu: better fixing it now before it's part of an official release

utzig:
Looks good, please squash your commits!

sdalu:
Seems I manage to squash it

PR #931, opened 2018-03-18
Adding support for SPI_2 for nrf52

ncasaril:
NRF52832 has support for a third SPIM or SPIS peripheral. Just making that available. 
Credit to Marcus Engholm for helping with this.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/syscfg.yml
+++ b/hw/mcu/nordic/nrf52xxx/syscfg.yml
@@ -71,6 +71,13 @@ syscfg.defs:
             - "!SPI_1_MASTER"
             - "!I2C_1"
 
+    SPI_2_MASTER:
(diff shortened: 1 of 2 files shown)


hw/mcu/nordic/nrf52xxx/syscfg.yml
```diff
@@ -71,6 +71,13 @@ syscfg.defs:
             - "!SPI_1_MASTER"
             - "!I2C_1"
 
+    SPI_2_MASTER:
```
simonratner: Should this have a `!SPI_2_SLAVE` restriction, and respectively for the slave?

ncasaril: Good point. Adding this. Thanks :)

PR #932, opened 2018-03-19
ADC: Added missing function for blocking read of channel.

ncasaril:
The adc api was missing the adc_read_channel function.


sterlinghughes:
adding jacob & mlaz for review, looks good to me.

jacobrosenthal:
Seems like youd need one of those.

mlaz:
Looks good to me.

PR #933, opened 2018-03-19
sys/log: Better handling for long/non-text entries

andrzej-kaczmarek:
Few improvements to logging subsystem to allow putting more data to log entries:
- allow to log directly from mbuf
- add data type to log entry (to distinguish between text/CBOR/whatever entries)
- special log handler to put data on unused slot1

Things to consider:
- Log version is bumped from 2 to 3, but there is no backwards compatibility retained - this means data from fcb written with v2 will be unreadable. Does it matter? Previous updates to log header did not care about it.
- Should slot1 be unlockable i.e. slot1 is locked, data is collected to cbmem, then unlocked and data is copied back to fcb. I removed this piece of code from PR since it was not working well for me yet, but I wonder if we'd actually need it. I imagine when image is uploaded to slot1 we have restart after which image is swapped by bootloader and we use slot1.

### DIFF
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -198,15 +198,16 @@ log_register(char *name, struct log *log, const struct log_handler *lh,
     return (0);
 }
 
-int
-log_append(struct log *log, uint16_t module, uint16_t level, void *data,
-        uint16_t len)
+static int
+log_append_prepare(struct log *log, uint16_t module, uint16_t level,
+                   struct log_entry_hdr *ue)
 {
-    struct log_entry_hdr *ue;
     int rc;
     int sr;
     struct os_timeval tv;
     uint32_t idx;
+
+    rc = 0;
 
     if (log->l_name == NULL || log->l_log == NULL) {
         rc = -1;
@@ -227,8 +228,6 @@ log_append(struct log *log, uint16_t module, uint16_t level, void *data,
         goto err;
     }
 
-    ue = (struct log_entry_hdr *) data;
-
     OS_ENTER_CRITICAL(sr);
     idx = g_log_info.li_next_index++;
     OS_EXIT_CRITICAL(sr);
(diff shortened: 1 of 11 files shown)


andrzej-kaczmarek:
Thanks for review, Chris!

I made few modificatoins, here's short summary wrt your comments:
1. I added `LOG_VERSION` which defaults to v2 so existing app won't break. If v3 is enabled, log entry type field is available and app can log typed messages other than string.
I am reluctant to add version to log entry as this means we need to maintain code which can read different versions of entries. Log version compatibility applies only to FCB handler (or some custom ones if someone implemented in own app) and in FCB we already have version field which can be utilized to detect mismatched versions and act accordingly. For now I just added check to FCB code so it will fail when trying to init area with different version. But I am open for discussion here :-)
2. I removed fallback to cbmem so this is now just a (very) thin wrapper over standard FCB - it uses the same log structure and will just return OS_EBUSY if slot 1 was locked. An open question here is when this slot should be locked - probably somewhere in imgmgr code or sth, I'm not familiar with that code at all.
3. Added.
4. Agree, but since I removed fallback to cbmem and this is just a wrapper over standard FCB this race does not occur anymore.
5. Updated.

I will upload modified version soon.

sterlinghughes:
How are we looking on this PR, what is blocking it?

sys/log/full/src/log.c
```diff
@@ -62,7 +62,7 @@ log_init(void)
     log_inited = 1;
     log_written = 0;
 
-    g_log_info.li_version = LOG_VERSION_V2;
+    g_log_info.li_version = LOG_VERSION_V3;
```
ccollins476ad: Should this be:
```
g_log_info.li_version = MYNEWT_VAL(LOG_VERSION);
```
?

Actually, I am not sure I understand why we even need `g_log_info.li_version`.  Can we just use `MYNEWT_VAL(LOG_VERSION)` instead?  I might have missed something in the code, so feel free to correct me :).

andrzej-kaczmarek: correct, should be assigned from `MYNEWT_VAL`. will fix.

`g_log_info.li_version` is used in log_fcb and log_reboot to initialize fcb version but obviously this can also be just `MYNEWT_VAL(LOG_VERSION)`. since `g_log_info` is a global struct so I kept it since someone could use it - but tbh I'm not sure what for, so I doubt anyone would really miss it ;-)

PR #934, opened 2018-03-20
util/parse - Additional parsing functions.

ccollins476ad:
* Allow base to be specified for byte streams.
* Parsing function for IPv6 network string.

### DIFF
--- a/net/ip/mn_socket/include/mn_socket/mn_socket.h
+++ b/net/ip/mn_socket/include/mn_socket/mn_socket.h
@@ -187,6 +187,10 @@ int mn_close(struct mn_socket *);
 /*
  * Address conversion
  */
+
+#define MN_INET_ADDRSTRLEN  16
+#define MN_INET6_ADDRSTRLEN 46
+
 int mn_inet_pton(int af, const char *src, void *dst);
 const char *mn_inet_ntop(int af, const void *src, void *dst, int len);
 
--- a/util/parse/src/parse.c
+++ b/util/parse/src/parse.c
@@ -47,20 +47,6 @@ parse_num_base(const char *sval)
     }
 }
 
-/**
- * Parses a long long within an imposed range.
- *
- * @param sval                  The string to parse.
- * @param min                   Values less than this are rejected.
- * @param max                   Values greater than this are rejected.
- * @param out_status            Written on completion;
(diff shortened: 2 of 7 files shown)


sterlinghughes:
@mkiiskila can you look at these, and if good merge?

ccollins476ad:
FYI- there was a bug in parsing address strings with leading double colons (e.g., "::1:2:3:4").  I just force pushed a fix.

sterlinghughes:
@mkiiskila can you review?

net/ip/mn_socket/include/mn_socket/mn_socket.h
```diff
@@ -187,6 +187,9 @@ int mn_close(struct mn_socket *);
 /*
  * Address conversion
  */
+
+#define MN_INET6_MAX_LEN    (39)    /* Does not include terminating \0. */
```
mkiiskila: To match the naming of the other items should use the following instead. Also note the size. And if adding this one, probably should add the MN_INET_ADDRSTRLEN there as well :)

#define MN_INET6_ADDRSTRLEN        46

util/parse/src/parse.c
```diff
@@ -261,3 +190,67 @@ parse_byte_stream_exact_length(const char *sval, uint8_t *dst, int len)
...
+int
+parse_byte_stream_exact_length(const char *sval, uint8_t *dst, int len)
+{
+    return parse_byte_stream_exact_length_base(sval, 0, dst, len);
+}
+
+bool
+parse_bool(const char *sval, int *out_status)
+{
+    if (strcmp(sval, "false") == 0) {
```
mkiiskila: strcasecmp() would be more user-friendly.

ccollins476ad:
Thanks, Marko.  I agree with all your comments.  I have made the following changes:

* Rename `MN_INET6_MAX_LEN` to `MN_INET6_ADDRSTRLEN`.
* Increase max IPv6 address string length from 39 to 46 (now includes null-terminator).
* Add `#define MN_INET_ADDRSTRLEN  16` to `mn_socket.h`.
* Use `strcasecmp` instead of `strcmp` in `parse_bool`.
* Add `#include "mn_socket/mn_socket.h"` to `parse.c`.

PR #936, opened 2018-03-20
hw/bsp: Clean up NRFX config on all bsps

michal-narajowski:
This patch cleans up the NRFX configuration. There are several
reasons why we should fix this:

- it makes the whole configuration cleaner and easier to port, removes
redundancy (see PWM configuration)
- we are using nrfx_config which was designed for this purpose
- this approach results in shorter compilation commands

If everyone is okay with this I will fix other bsps too.


mlaz:
This approach seems legit, although I do have two questions here:

1. Aren't we diverging from doing thing in a mynewty way? What do you think, @jacobrosenthal ?
2. If we're taking this approach let's then do it for every chip (and BSP consequently).

michal-narajowski:
@mlaz, I see only BSPs using Nordic SDK are doing it this way. This PR suggests using the SDK the way it was designed to be used.

mlaz:
@michal-narajowski, as I said, It seems OK to me(i.e. I _agree_ with you). Anyway if we are doing this we need to keep this consistent through all BSPs which use mcu/nordic (i.e. doing it for all bsps, probably having nrfx_config.h under mcu/nordic/include).

jacobrosenthal:
Good with me. As Miguel mentioned, we need all the bsps altered or to move it up to hw/mcu.

jacobrosenthal:
Still needs all the other bsps changed as well

michal-narajowski:
@mlaz @jacobrosenthal Updated all bsps

PR #939, opened 2018-03-20
Upgrade lwIP to 2.0.3

ccollins476ad:
This upgrades to the latest stable version of lwIP: 2.0.3.  Currently, we are using 2.0.2.


utzig:
You need to remove `include/netif/ppp/pppoe.h` and `src/netif/ppp/pppoe.c` because they are category X license (4-clause BSD).

ccollins476ad:
Thanks for the reminder up, @utzig!  I remember the nightmare those files caused in a previous release.  I have removed the two files.

utzig:
I will definitely not test this, but PR looks ok!

PR #940, opened 2018-03-21
nrf52xxx_i2c: Fix to allow hal_i2c_clear_bus to read from input-buffer 

ncasaril:
I'm working on a board on which I read a lot of data from the i2c bus. Resets in the middle of one of these transfers can cause a slave on the bus to hold the sda pin low - i.e. hang the i2c bus. Even though the hal_i2c_init runs the hal_i2c_clear_bus() function at boot, which should be able to clear the i2c bus at startup, the system does not recover from a hanged i2c bus without a power cycle. 

hal_i2c_clear_bus clears the bus by checking if it's held low and outputs clock cycles on scl until the slave lets go. However, because of the weird mode the sda pin was in - output but still with the input_buffer enabled input buffer, the actual value from outside was never read. This because hal_gpio_read(sda_pin) only reads the input buffer if the port-DIR was set to input. 

Not sure this is the cleanest solution, but it corresponds better with what's in the NRF SDK's nrf_drv_twi.c and it works well on my board.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
@@ -163,12 +163,12 @@ __ASM volatile (
     : "+r" (delay));
 }
 
-/*
- * Special version of hal_gpio_read_in which reads the in buffer
- * regardless of what direction the pin is set to.
+/**
+ * Reads the input buffer of the specified pin regardless
+ * of if it is set as output or input
  */
 static int
-hal_gpio_read_in(int pin)
+read_gpio_inbuffer(int pin)
 {
     NRF_GPIO_Type *port;
     port = HAL_GPIO_PORT(pin);
@@ -202,7 +202,7 @@ hal_i2c_clear_bus(const struct nrf52_hal_i2c_cfg *cfg)
     hal_i2c_delay_us(4);
 
     for (i = 0; i < 9; i++) {
-        if (hal_gpio_read_in(cfg->sda_pin)) {
+        if (read_gpio_inbuffer(cfg->sda_pin)) {
             if (i == 0) {
                 /*
                  * Nothing to do here.
--- a/hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
@@ -39,6 +39,14 @@
       (GPIO_PIN_CNF_DIR_Input     << GPIO_PIN_CNF_DIR_Pos))
 #define NRF52_SDA_PIN_CONF NRF52_SCL_PIN_CONF
 
+#define NRF52_SCL_PIN_CONF_CLR                                  \
+     ((GPIO_PIN_CNF_SENSE_Disabled << GPIO_PIN_CNF_SENSE_Pos) | \
+      (GPIO_PIN_CNF_DRIVE_S0D1     << GPIO_PIN_CNF_DRIVE_Pos) | \
+      (GPIO_PIN_CNF_PULL_Pullup    << GPIO_PIN_CNF_PULL_Pos)  | \
+      (GPIO_PIN_CNF_INPUT_Connect  << GPIO_PIN_CNF_INPUT_Pos) | \
+      (GPIO_PIN_CNF_DIR_Output     << GPIO_PIN_CNF_DIR_Pos))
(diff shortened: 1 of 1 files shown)


hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
```diff
@@ -171,13 +192,13 @@ hal_i2c_clear_bus(const struct nrf52_hal_i2c_cfg *cfg)
     hal_gpio_write(cfg->scl_pin, 1);
     hal_gpio_write(cfg->sda_pin, 1);
 
-    /* Still with input buffer connected, the direction is output */
-    NRF_GPIO->DIRSET = ((1<<cfg->scl_pin)|(1<<cfg->sda_pin));
+    NRF_GPIO->PIN_CNF[cfg->scl_pin] = NRF52_SCL_PIN_CONF_CLR;
+    NRF_GPIO->PIN_CNF[cfg->sda_pin] = NRF52_SDA_PIN_CONF_CLR;
```
vrahane: You are changing the GPIO config, so you might want to do the following again after finishing clocking bits:
```
    NRF_GPIO->PIN_CNF[cfg->scl_pin] = NRF52_SCL_PIN_CONF;
    NRF_GPIO->PIN_CNF[cfg->sda_pin] = NRF52_SDA_PIN_CONF;
```

ncasaril: Thanks, when the clear_bus command is run from outside of the hal_i2c_init function changing the io back is important. <- fix added. 

I'm not sure I'm following you on the second issue. The hal_gpio_read_in function I added does resolve the port and pins: 

```
static int
hal_gpio_read_in(int pin)
{
    NRF_GPIO_Type *port;
    port = HAL_GPIO_PORT(pin);

    return (port->IN >> HAL_GPIO_INDEX(pin)) & 1UL;
}
```

If the restoring of the pins' config is the problem for the nrf52840, then we have the same problem in hal_i2c_init() after the hal_i2c_clear_bus(); line, right?  

```
...    
    hal_i2c_clear_bus(cfg);

    NRF_GPIO->PIN_CNF[cfg->scl_pin] = NRF52_SCL_PIN_CONF;
    NRF_GPIO->PIN_CNF[cfg->sda_pin] = NRF52_SDA_PIN_CONF;

    regs->PSELSCL = cfg->scl_pin;
...
```

vrahane: Sorry, I should have been more clear `hal_gpio_read_in()` is correct as the implementation takes into account ports and specifics pins belonging to the ports. What seems to be problematic is assigning PIN_CNF to a gpio directly without selecting a particular port.

vrahane:
@ncasaril hope you got what I mean in the above comment ?

ncasaril:
@vrahane I think I understand that it would need to be resolved to a port, but the same thing would then apply to the hal_i2c_init function? And also to the NRF_>PIN_CNF[] section above my change in hal_i2c_clear_bus, right? 

Or, are we in fact looking at two different changes that needs to happen to the hal_i2c.c. One to allow the clearing of the bus and one to allow the hal_i2c* to work with the nrf52840?

ncasaril:
Ok, I'll change this PR to address both the clearing of the bus as well as providing support for i2c for nrf52840. Should be able to push the remaining changes and possibly rebase to clean it up a bit later today.

vrahane:
Sounds good to me.

ncasaril:
I've changed so the ports are taken into account. Let me know if you think I should move the hal_gpio_read_in to hal_gpio.c as well. I left it in hal_i2c because I didn't know if it would be useful anywhere else.

vrahane:
Yes, I think you are right, all gpio related hal functionality should go in hal_gpio.c. So, hal_gpio_read_in() should probably be moved to to hal_gpio.c. It would not be static at that point since hal_i2c needs to use it.

ncasaril:
The only problem with moving the hal_gpio_read_in() to hal_gpio is that it will not be defined in the hal_gpio.h unless I change that as well. If this is only specific to the Nordic, and only really to when clearing the i2c bus, perhaps it's better to not change the hal_gpio, but rather just rename the function and keep it in hal_i2c. 

I've renamed the function and kept it in hal_i2c. See what you think about this @vrahane .

ncasaril:
ping @vrahane ?

sterlinghughes:
@vrahane can you look at this?  I would like to get it merged.

PR #941, opened 2018-03-21
Stm32 common SPI

utzig:
This adds an initial common shared SPI driver for all STM32 supported MCUs.

For now I disabled support for slave on F3/F7, which have internal FIFOs. The TX FIFO is not configurable on size always generates a `TXE` interrupt when it is half full, and it does not seem to have a flush operation, so when working as slave, after SS is asserted there is no way to flush whatever was already presented in the FIFO and it will be sent later after a new communication, breaking functionality of some apps like spitest. I will open an issue to look into this after all common driver framework is in place.


utzig:
PS: there are some `#if 0`, `FIXME` and `TODO` in the code. I would prefer to leave them there if it is in one of the two driver files and `hw/mcu/stm/stm32_common`!

utzig:
@ccollins476ad LOL, of course, it works on F1/L1/F4 as master/slave and F3/F7 as master only (for now)!

PR #942, opened 2018-03-21
Fix printf fault before UART was initialized

utzig:
This fixes an issue were `os_default_irq` is trying to print dump information before UART initialization was completed, causing a fault and losing all stack/backtrace data, making debug impossible.


sterlinghughes:
@michal-narajowski ping on review?

PR #947, opened 2018-03-22
Refactor/easing library

mlampert:
I ran into some performance issues on an 8MHz STM32F3 (Cortex-M4) when playing with the easing library. I refactored the library a bit and think the performance improvements are worth the somewhat lesser readability.

### DIFF
--- a/util/easing/src/easing.c
+++ b/util/easing/src/easing.c
@@ -69,7 +69,7 @@ static inline float quadratic_in(float step, float max_steps, float max_val)
 {
 	float ratio = step / max_steps;
 
-	return max_val * pow(ratio, 2);
+	return max_val * (ratio * ratio);
 }
 
 static inline float quadratic_out(float step, float max_steps, float max_val)
@@ -84,7 +84,7 @@ static inline float quadratic_io(float step, float max_steps, float max_val)
 	float ratio = step / (max_steps / 2.0f);
 
 	if (ratio < 1)
-		return max_val / 2.0f * pow(ratio, 2);
+		return max_val / 2.0f * (ratio * ratio);
 
     ratio = (step - (max_steps/2.0f)) / (max_steps/2.0f);
     return (max_val / 2.0f) - (max_val / 2.0f) * ratio * (ratio - 2.0f);
@@ -95,14 +95,14 @@ static inline float cubic_in(float step, float max_steps, float max_val)
 {
 	float ratio = step / max_steps;
 
-	return max_val * pow(ratio, 3);
+	return max_val * (ratio * ratio * ratio);
 }
 
 static inline float cubic_out(float step, float max_steps, float max_val)
 {
 	float ratio = step / (max_steps - 1.0f);
 
-	return max_val * (pow(ratio, 3) + 1);
+	return max_val * ((ratio * ratio * ratio) + 1);
 }
 
 static inline float cubic_io(float step, float max_steps, float max_val)
@@ -110,9 +110,10 @@ static inline float cubic_io(float step, float max_steps, float max_val)
 	float ratio = step / (max_steps / 2.0f);
 
 	if (ratio < 1)
-		return max_val / 2 * pow(ratio, 3);
+		return max_val / 2 * (ratio * ratio * ratio);
 
-	return max_val / 2 * (pow(ratio - 2, 3) + 2);
+    ratio -= 2;
+	return max_val / 2 * ((ratio * ratio * ratio) + 2);
 }
 
 /* Quartic */
@@ -120,14 +121,14 @@ static inline float quartic_in(float step, float max_steps, float max_val)
 {
 	float ratio = step / max_steps;
 
-	return max_val * pow(ratio, 4);
+	return max_val * (ratio * ratio * ratio * ratio);
 }
 
 static inline float quartic_out(float step, float max_steps, float max_val)
 {
 	float ratio = (step / max_steps) - 1.0f;
 
-	return max_val + max_val * pow(ratio, 5);
+	return max_val + max_val * (ratio * ratio * ratio * ratio * ratio);
 }
 
 static inline float quartic_io(float step, float max_steps, float max_val)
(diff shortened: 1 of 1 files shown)


mlaz:
Looks good to me. Have you tested these for mathematical correctness?

mlampert:
The math is identical to the original implementation. Is there anything specific you're looking for?

util/easing/src/easing.c
```diff
@@ -43,12 +43,12 @@ static inline float exponential_out(float step, float max_steps, float max_val)
 {
 	return (step == max_steps) ?
         max_val :
-        max_val - pow(max_val, 1.0 - (float)step/max_steps);
+        max_val - pow(max_val, 1.0f - (float)step/max_steps);
```
utzig: Since you are moving everything to `float` you could as well get rid of those `(float)` casts. This one would be the same as `max_val - pow(max_val, 1.0f - step/max_steps)`. This happens across much of the refactoring.

util/easing/src/easing.c
```diff
@@ -69,125 +69,128 @@ static inline float quadratic_in(float step, float max_steps, float max_val)
...
 {
-	float ratio = step / (max_steps / 2.0);
+	float ratio = step / (max_steps / 2.0f);
 
 	if (ratio < 1)
-		return max_val / 2 * pow(ratio, 4);
+		return max_val / 2 * (ratio * ratio * ratio * ratio);
 
-	return max_val + max_val / 2 * pow(ratio -2, 5);
+    ratio -= 2;
```
utzig: Weird identation

util/easing/src/easing.c
```diff
@@ -69,125 +69,128 @@ static inline float quadratic_in(float step, float max_steps, float max_val)
...
 {
-	float ratio = step / (max_steps / 2.0);
+	float ratio = step / (max_steps / 2.0f);
 
 	if (ratio < 1)
-		return max_val / 2 * pow(ratio, 3);
+		return max_val / 2 * (ratio * ratio * ratio);
 
-	return max_val / 2 * (pow(ratio - 2, 3) + 2);
+    ratio -= 2;
```
utzig: Weird identation

util/easing/src/easing.c
```diff
@@ -69,125 +69,128 @@ static inline float quadratic_in(float step, float max_steps, float max_val)
...
+    return (max_val / 2.0f) - (max_val / 2.0f) * ratio * (ratio - 2.0f);
 }
 
 /* Cubic */
 static inline float cubic_in(float step, float max_steps, float max_val)
 {
 	float ratio = step / max_steps;
 
-	return max_val * pow(ratio, 3);
+	return max_val * (ratio * ratio * ratio);
```
utzig: Whenever you have a `pow` or `sqrt` on the code, your variables are automatically promoted to `double` and the operation is done at a higher precision, then what is done now but simply multiplying `float`s. I am not sure how much this would impact on precision but I assume this is why @mlaz might have asked the previous question about the accuracy.

util/easing/src/easing.c
mlampert: the weird identation comes from the original source code having tab characters in some places and github uses tabstop=8. I didn't clean up the tabs because it introduces a lot of noise into the code review but can certainly do that.

util/easing/src/easing.c
mlampert: I'm not sure why the original implementation had that cast there, it was promoted to a `double` anyway. I presume that early on `step` and `max_step` were integer values and it was a remnant of refactoring. I can certainly clean those up.

util/easing/src/easing.c
mlampert: Understood - but I cannot answer that question without a reference. The interface does not suggest that there is `double` precision involved, it's all based on `float` so as a user that's as much as I can expect. In order to answer the question if this refactoring falsifies the result we would need a definition as to what level of deviation is acceptable.

In addition none of the computations uses accumulation which is where `float`'s really start bleeding, all values are monotonically increasing or decreasing and the result is converted to a float. So if the initial value and the result fits into a `float` then every intermediary value will also fit.

mlampert:
A general comment on the "precision" question.

I love the library, use it all the time ever since the PR was posted the first time. Yet, as is it is IMHO not very practical because it's resource requirements are too high. The reason I started looking into this was because the output of `console` started to have pauses (in the middle of printing a word) when I used the library for a single LED.

Having this library is insanely cool but I cannot afford 0.5-1ms execution times in a timer interrupt callback.

Maybe I'm missing a use case here - is there one that requires `double` precision?

util/easing/src/easing.c
mlaz: The indentations where inherited from a previous implementation I had while validating this mathematically.

util/easing/src/easing.c
mlaz: I believe that at some point, not using the the functions will in fact change the results, we just need to make sure wont break the functions. The method I used to validate these mathematically is to look at the function here and here, then implementing it on wolfram (or some other equivalent tool) run the lib's code in order to produce some values an compare them to what I get on wolfram.

mlaz:
@mlampert as you can see this lib is far from being optimized.
My approach to this is to have an implementation which is accurate at first.
Then let's talk about performance, take your LED case for instance: right now you may implement an easing_i_func_t which instead of calling one of the implemented functions may just use a lookup table.
This is one approach, the other way we may work this out is to implement an alternate, architecture level optimized math lib (which is not something one is able to do overnight, see an exaple of this in 1.) and use it on our functions. 

Some references here:
1. https://web.archive.org/web/20161219143458/http://http.developer.nvidia.com:80/Cg/index_stdlib.html
2. http://people.math.sfu.ca/~cbm/aands/abramowitz_and_stegun.pdf

sdalu:
A bit of topic, but I'm wondering if it would be possible to combine a lookup table (generated by the easing function) with access to the easydma feature of the nrf52840? (perhaps ST also has the same kind of feature?)

jacobrosenthal:
FWIW Id rather have performance over accuracy. Easing is generally used for animation and is not mission critical. But if we want both options, separate functions or a separate dependency seems viable.

mlaz:
@jacob i think we may agree here that in this particular case accuracy is having growth ratios near to what we have now. I never opposed to having these changes merged. We just need to check if by any means the behaviour is consistent.

PR #954, opened 2018-03-23
Add CLI to LIS2DW12

matthewwarnes:


hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -2047,8 +2042,6 @@ lis2dw12_config(struct lis2dw12 *lis2dw12, struct lis2dw12_cfg *cfg)
         goto err;
     }
     lis2dw12->cfg.tap_cfg = cfg->tap_cfg;
-
-    lis2dw12_write8(itf, 0x34, 0x80);
```
jacobrosenthal: Is this necessary inside of tap configuration, or just not necessary at all?

matthewwarnes: I don't believe it is necessary at all, single and double tap detection work fine without it.

jacobrosenthal:
Its a bit sparse, could probably hook up some of the existing public function calls, especially around troubleshooting and setup like run_self_test

BTW does your run_self_test work? mine fails.. but I just copied yours :)

Also youll want to add your _shell_init to the apps/sensors_test/src/main.c

I added cli to mine as well. 
https://github.com/jacobrosenthal/mynewt-core/commit/ada23aa3f529772494a967e5fa541b7a0dae78f7

I also added a poke and peek I used on drv2605, and for better or worse I generalized your dump for developer friendliness because I was too lazy to hand code that. You may want these

matthewwarnes:
The run_self_test is definitely working for me. 

I'll take a look at your CLI.

PR #961, opened 2018-03-27
hw/util: blinker/beeper implementation

sdalu:
Provide an implementation of a blinker/beeper, only need to provide an on/off method (gpio, pwm, ...).
Make it easy to provide a simple user feedback


sdalu:
Going XKCD 1597
See new PR: https://github.com/apache/mynewt-core/pull/1055

PR #963, opened 2018-03-27
Fix imgmgr to follow flash write alignment

utzig:
This fixes an issue found on stm32f3 where newtmgr after writing an odd sized packet, would hardfault on the next write due to trying to write an odd address which is not valid on stm32f3 flash.


utzig:
Hmmm, but then it would return with an `imgr_state.upload.off` that is the last image position minus 1, and `newtmgr` should send the last packet with 1 byte, no?

utzig:
Well, your suggestion makes sense, I will update this.

utzig:
@ccollins476ad Updated it to not break the last write, per your suggestion.

PR #964, opened 2018-03-27
Update to 2k sector size on stm32f3

utzig:
This updates `hal_flash` on the stm32f3 to use the native sector size for this mcu (2k).


utzig:
@mlampert When you submitted the initial flash for stm32f3 you decided to use 16k for flash sectors instead of 2k. I had some issues doing swap updates on the bootloader with it (although might be errors of mine!). Did you have a strong a reason to do this? (I would presume it could be due to maximum amount of sectors allowed for a swap?). Would you mind moving to 2k?

mlampert:
No problem at all, the sector size was arbitrary.

For posterity: I saw the sector sizes in the other STM32 ports and wanted to keep it simpler than that. I think the smallest sector I found in there was 16k for the bootloader so I just went with that.

PR #953, opened 2018-03-23
Stm32 common I2C

utzig:
This is completely functional on F3x, F4x, F7x, L1x. Still having issues on F1x. Will fix on another PR.

PR #966, opened 2018-03-27
Misc BSP fixes

ccollins476ad:
I encountered a few build errors when trying to build blinky for all BSPs.  These are some small changes that fix the errors.

### DIFF
--- a/hw/mcu/stm/stm32f1xx/include/mcu/stm32f1_bsp.h
+++ b/hw/mcu/stm/stm32f1xx/include/mcu/stm32f1_bsp.h
@@ -42,6 +42,12 @@ struct stm32f1_uart_cfg {
     IRQn_Type suc_irqn;                 /* NVIC IRQn */
 };
 
+/*
+ * Internal API for stm32f1xx mcu specific code.
+ */
+int hal_gpio_init_af(int pin, uint8_t af_type, enum hal_gpio_pull pull,
+                     uint8_t od);
+
 struct hal_flash;
 extern struct hal_flash stm32f1_flash_dev;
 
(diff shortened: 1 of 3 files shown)


hw/mcu/stm/stm32f1xx/include/mcu/stm32f1_bsp.h
```diff
@@ -42,6 +42,12 @@ struct stm32f1_uart_cfg {
     IRQn_Type suc_irqn;                 /* NVIC IRQn */
 };
 
+/*
+ * Internal API for stm32f1xx mcu specific code.
+ */
+int hal_gpio_init_af(int pin, uint8_t af_type, enum hal_gpio_pull pull,
```
utzig: This one is strange... I removed it intentionally because AFs work differently on stm32f1, and on `master` only `i2c` still uses it, so I can only assume you have `I2C_0: 1`. Btw, PR #953 already removed this dependency, but I'm OK with merging temporarily, and I'll remove it again with that PR.

utzig:
I'm not sure with the dos<->unix thing... that file will always be based on sources from stm32cube so will always have to do the extra step of saving as unix...

hw/mcu/stm/stm32f1xx/include/mcu/stm32f1_bsp.h
ccollins476ad: Hmm, that is strange.  I don't have `I2C_0` enabled.  My target is defined as follows:
```
targets/blinky-olimex-p103
    app=apps/blinky
    bsp=@apache-mynewt-core/hw/bsp/olimex-p103
    build_profile=debug
```

PR #968, opened 2018-03-28
Fix build errors - Add required bootutil dependency to a few apps

ccollins476ad:
This PR modifies two apps:
* bleuart
* bsnprph

These two apps depend on `boot/split`, which requires the `bootloader` API.  This PR satisfies this requirement by adding the `boot/bootutil` package to both apps' dependency lists.

### DIFF
--- a/apps/bleuart/pkg.yml
+++ b/apps/bleuart/pkg.yml
@@ -23,6 +23,7 @@ pkg.homepage: "http://mynewt.apache.org/"
 pkg.keywords:
 
 pkg.deps:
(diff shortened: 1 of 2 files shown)


utzig:
Ah, I see what you mean... `boot/split` already has the requirement... well, I add `@mcuboot/boot/mynewt" to my `boot` targets...

ccollins476ad:
Thanks, @utzig.  Hopefully we'll get the Mynewt boot loader and MCUboot consolidated soon.

PR #973, opened 2018-03-29
Add BSD compatibility queue.h

### LINKED ISSUE #972: os: moving os/queue.h to sys/queue.h
Moving os/queue.h to it's expected place: sys/queue.h (as it is a pretty standard include)
So that's possible to write portable code across OS (otherwise there will be a clash when trying to import our own copy of sys/queue.h)

Possibilities are (in my order of preference):
* effectively moving file to `sys/queue.h` and creating an `os/queue.h` (for compatibilty with previous mynewt code)
* creating a `sys/queue.h` wrapper: 
~~~
#ifnded _SYS_QUEUE_H_
#define _SYS_QUEUE_H_
#include "os/queue.h"
#endif
~~~
* replacing define with `_SYS_QUEUE_H_` instead of  `_QUEUE_H_`. This will avoid inclusion of developper own copy of `sys/queue.h`, but a `sys/queue.h` will need to be provided

utzig:
This creates a new `sys` package and exports `<sys/queue.h>` to ease portability of external apps. `<os/queue.h>` was updated to use `<sys/queue.h>`.

This fixes #972


utzig:
@sdalu This little experiment is one possible solution for #972

sdalu:
Using a package is a good idea.

vrahane:
So, what about components breaking off of mynewt like NFFS or NFC or nimble, which header would they include ? I would assume `sys/queue.h`.

utzig:
`sys/queue.h` is definitely better for portability.

PR #976, opened 2018-03-29
Fix console echo in minimal console

utzig:
console minimal was using `console_echo()` function to set echo, but this function is empty on the minimal version. It is only implemented on `console/full`.


utzig:
Thanks, updated as suggested!

PR #978, opened 2018-03-29
Fix boot_serial to follow flash write alignment

utzig:
This fixes an issue found on stm32f3 where newtmgr after writing an odd sized packet, would hardfault on the next write due to trying to write an odd address which is not valid on stm32f3 flash.


utzig:
@ccollins476ad @mkiiskila This basically applies the alignment fix in image upload, this time on `serial_boot`. I have not tested this one, but I just opened a PR on `mcuboot` with exactly the same change, and I have tested that one!

PR #980, opened 2018-03-30
util/button: Use the button id in callback instead of button pointer

sdalu:
It doesn't look useful to have the whole button structure exposed in the user callback

### DIFF
--- a/hw/util/button/include/button/button.h
+++ b/hw/util/button/include/button/button.h
@@ -33,15 +33,15 @@
  * 
  *
  * button_t btns[] = {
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
Separate from this. But thoughts on returning os callbacks to make these
more newty. Could put which queue you want it returned in. I find I end up
wrapping all this when I use it anyway.

jacobrosenthal:
Yep that :)

sdalu:
I would prefer to have this small api change merge now.

PR #984, opened 2018-03-31
hw/util/button: allow to specify eventq

sdalu:
Allow to specify event queue for internal processing. And eventq (per button if requested) for the callback processing

### DIFF
--- a/hw/util/button/syscfg.yml
+++ b/hw/util/button/syscfg.yml
@@ -74,4 +74,4 @@ syscfg.defs:
         description: >
             Allow to specify per button event queue for callback processing.
             By default os_eventq_dflt_get()
-        value: 1
+        value: 0


ccollins476ad:
First, the code looks great.  I want to say that first, because this PR opens quite the can of worms, and I don't want my feedback to sound completely negative :).

I get a little uncomfortable with specifying C code in syscfg settings (e.g., `os_eventq_dflt_get()`).  My worry is that this has the potential for some confusing debugging sessions in the case of a typo or syntax error.  There are also some thorny semantics issues, e.g., what if the code references the setting in an incorrect context (in an ISR, before the code is ready to run, etc.)?  I should acknowledge that there are already other syscfg settings that use C identifiers, so there is some precedent for this.

This particular issue is something that has been lingering in my mind for a while. I think it would be good if the group decided on some form of project-wide policy, so I will kick off a discussion on the dev list.

In the specific case of event queues, I believe all the other packages require the application to do the configuration at runtime.  For example, in the shell package: https://github.com/apache/mynewt-core/blob/2f35d09b6fa29e2b715d713d82015a1988bc87d6/sys/shell/src/shell.c#L48.  I am not sure that is necessarily "right", but it is how it is done elsewhere.

sdalu:
Moved to function instead of syscfg. Either will need to squash or remove last commit.

jacobrosenthal:
This (actually not this, but the original PR) doesnt appear to build with the arm-gcc on macosx
```
Error: repos/apache-mynewt-core/hw/util/button/src/button.c:49:15: error: missing braces around initializer [-Werror=missing-braces]
 static struct button_event button_event[MYNEWT_VAL(BUTTON_EVENT_MAX)] = { 0 };
               ^
repos/apache-mynewt-core/hw/util/button/src/button.c:49:15: error: (near initialization for 'button_event[0]') [-Werror=missing-braces]
repos/apache-mynewt-core/hw/util/button/src/button.c: In function 'button_alloc_event':
repos/apache-mynewt-core/hw/util/button/src/button.c:55:5: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
     for (int i = 0 ; i < MYNEWT_VAL(BUTTON_EVENT_MAX) ; i++) 
     ^
repos/apache-mynewt-core/hw/util/button/src/button.c:55:5: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
repos/apache-mynewt-core/hw/util/button/src/button.c: In function 'button_init':
repos/apache-mynewt-core/hw/util/button/src/button.c:509:5: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
     for (unsigned int i = 0 ; i < count ; i++) {
     ^
repos/apache-mynewt-core/hw/util/button/src/button.c:517:23: error: redefinition of 'i'
     for (unsigned int i = 0 ; i < count ; i++) {
                       ^
repos/apache-mynewt-core/hw/util/button/src/button.c:509:23: note: previous definition of 'i' was here
     for (unsigned int i = 0 ; i < count ; i++) {
                       ^
repos/apache-mynewt-core/hw/util/button/src/button.c:517:5: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
     for (unsigned int i = 0 ; i < count ; i++) {
     ^

```

sdalu:
Is there a rule/reason not to use c99 or c11 standard? Otherwise I can add it to the pkg.cflags

sdalu:
Going XKCD 1597: https://xkcd.com/1597/
New PR: https://github.com/apache/mynewt-core/pull/1054

PR #993, opened 2018-04-04
lps33hw: Added sensor set config function to LPS33HW driver

brianpax:


brianpax:
set_config function added in a different PR

vrahane:
thanks @brianpax

PR #995, opened 2018-04-05
sys/log: Updates to handle CBOR entries via nmgr

andrzej-kaczmarek:
With v3 we can now have CBOR (and binary data, but let's skip this for now...) stream as log entry and this should be properly transmitted via nmgr. This makes necessary changes to do this by sending `msg` field as valid CBOR stream from log entry, though `newtmgr` probably needs some updates to read it properly.

### DIFF
--- a/sys/log/full/src/log_nmgr.c
+++ b/sys/log/full/src/log_nmgr.c
@@ -52,6 +52,11 @@ static struct mgmt_handler log_nmgr_group_handlers[] = {
     [LOGS_NMGR_OP_LOGS_LIST] = {log_nmgr_logs_list, NULL}
 };
 
+struct log_encode_data {
+    uint32_t counter;
+    CborEncoder *enc;
+};
+
 /**
  * Log encode entry
  * @param log structure, log_offset, dataptr, len
@@ -67,7 +72,7 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
     int rc;
     int rsp_len;
     CborError g_err = CborNoError;
-    CborEncoder *penc = (CborEncoder*)log_offset->lo_arg;
+    struct log_encode_data *ed = log_offset->lo_arg;
     CborEncoder rsp;
     struct CborCntWriter cnt_writer;
     CborEncoder cnt_encoder;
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
I also analyzed if it would be feasible to keep 'msg' as string and transfer CBOR/binary data encided in base64. The problem is that these entries can be (very) long and tinycbor does not have API that would allow to encode string provided in chunks. This means it would be necessary to encode data to some intermediate buffer and we may simply run our of memory here.

sterlinghughes:
plus it will increase size over the air: i think binary makes more sense here.

ccollins476ad:
Just a question: why is there a log type for CBOR?  Would the "binary" type work just as well for CBOR data?

andrzej-kaczmarek:
@ccollins476ad it would work, but I see two advantages with separate type:
- for newtmgr transfer binary data has to be base64 (or sth) encoded which increases size over the air; CBOR can be just included as-is
- any reader can immediately display these data without extra information what's exactly inside binary log entry (can be e.g. used for pretty-printing).

ccollins476ad:
Thanks for responding, Andrzej.

I don't believe the first point is correct; we can send binary data in a newtmgr response.  Perhaps you're thinking of the serial transport, which requires responses to be base64 encoded?

I see what you mean in the second point - it could be useful for newtmgr clients to automatically know how to display a response.  On the other hand, I would expect the consumer of logs to already know how to parse them by convention.

andrzej-kaczmarek:
Yes, the data over newtmgr can be binary but actually I had log entries that are sent in response over newtmgr in mind - sorry for confusion. In this case, each log entry included in response has 'msg' parameter which is the log message and it is always encoded to CBOR response as string by log_nmgr.c (which was a fair assumption given that only string entries were logged). Now, I'd like this 'msg' to be binary data which I don't think is possible without extra encoding in general, but since the response itself is CBOR data, the log entry of type CBOR can be just appended there. In JSON notation (kind of) instead of
`{ 'msg' : 'text log entry }`
it would be:
`{ 'msg' : { 'foo' : 1, 'bar' : 'test' } }`
where `{ 'foo' : 1, 'bar' : 'test' }` is contents of log entry. I believe `newtmgr` does not understand such response since it seems to be stuck trying to receive it so it would need to check extra 'type' parameter and act accordingly. Nothing changes for string logs - they are handled exactly as previously.

HTH :-)

sys/log/full/src/log_nmgr.c
```diff
@@ -99,6 +105,35 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
...
+        dlen = len - sizeof(ueh);
+        break;
+    case LOG_ETYPE_BINARY:
+        /* XXX just some placeholder for now, perhaps should be base64 encoded? */
+        strcpy(data, "<binary>");
+        dlen = strlen(data);
+        break;
+    default:
+        assert(0);
+        strcpy(data, "<??\?>");
```
ccollins476ad: Accidental backslash? :)

andrzej-kaczmarek:
@ccollins476ad Fixed. Sorry it took so long, I totally forgot about this PR...

sys/log/full/src/log_nmgr.c
andrzej-kaczmarek: It's there to prevent `??>` being interpreted as trigraph. I could have used some other string though...

ccollins476ad: Oh, right.  I totally missed that.

PR #999, opened 2018-04-07
os_time: Prevent 0 ticks returned when OS_TICKS_PER_SEC < 1000

ncasaril:
The os_time_ms_to_ticks assumes the multiplication always proceeds the division in the following piece of code: 

```
 ticks = (uint64_t)ms * OS_TICKS_PER_SEC / 1000;
```

With an OS_TICKS_PER_SEC < 1000 the division could result in 0 and give the wrong number of ticks. This PR adds () to force multiplication before division. Thus:

```
 ticks = ((uint64_t)ms * OS_TICKS_PER_SEC) / 1000;
```


ccollins476ad:
Thanks, Niklas.  Do you get different results after adding the parentheses?  Multiplication and division should always be evaluated left-to-right with respect to each other, per the C standard.

That said, I agree the parentheses clarify the expression.  I was just curious if they made a difference.

PR #1000, opened 2018-04-07
sys/config: Checking results on rc2 instead of rc

ncasaril:
Surely the if statements should be checking if rc2 is non-zero rather than the (unchanging in the loops) rc.


ncasaril:
Hmmm, I'm probably wrong on this. I'll have another look tomorrow and probably close this pr. Sorry :)

ncasaril:
Yep, I was wrong. Please ignore this.

PR #1003, opened 2018-04-08
STM32: common hal_gpio

mlampert:
Note that I cannot test actual functional correctness for stm32f1 and stm32f4. If somebody could do a real world test on those platforms - that would be great.


hw/mcu/stm/stm32_common/src/hal_gpio.c
```diff
@@ -479,8 +524,16 @@ int hal_gpio_init_out(int pin, int val)
     cfg.Pin = GPIO_MASK(pin);
     cfg.Mode = GPIO_MODE_OUTPUT_PP;
     cfg.Pull = GPIO_NOPULL;
+#if (defined GPIO_SPEED_FREQ_VERY_HIGH)
```
utzig: while this should be OK, it would be better to write it the standard way: `#if defined(x)`

hw/mcu/stm/stm32_common/src/hal_gpio.c
```diff
@@ -479,8 +524,16 @@ int hal_gpio_init_out(int pin, int val)
...
+    cfg.Speed = GPIO_SPEED_FREQ_VERY_HIGH;
+#elif (defined GPIO_SPEED_HIGH)
```
utzig: same as above

hw/mcu/stm/stm32_common/src/hal_gpio.c
```diff
@@ -503,9 +556,17 @@ hal_gpio_init_af(int pin, uint8_t af_type, enum hal_gpio_pull pull, uint8_t od)
     } else {
         gpio.Mode = GPIO_MODE_AF_OD;
     }
-    gpio.Speed = GPIO_SPEED_HIGH;
     gpio.Pull = pull;
+#if (defined GPIO_SPEED_FREQ_VERY_HIGH)
```
utzig: same as above on this "block"

PR #1007, opened 2018-04-10
LIS2DW12: Interrupt Line Remapping

brolan-juul:
Added mapping of int2 to int1.
User configurable.


vrahane:
Changes requested.

brolan-juul:
Made the changes with get / set functions.

PR #1010, opened 2018-04-11
os: Add os_time_ticks_to_ms

andrzej-kaczmarek:
This is complementary to os_time_ms_to_ticks and allows to have more
OS abstraction in code by avoiding use of OS_TICKS_PER_SEC explicitly.


andrzej-kaczmarek:
fyi, I want to use this in NimBLE code so porting layer does not need to define OS_TICKS_PER_SEC macro

sterlinghughes:
can we merge?

PR #1012, opened 2018-04-12
LIS2DW12 add bitfields

matthewwarnes:
Change config structure to use bitfields to reduce structure size. 

Fix bit shift value and if else style as well.

### DIFF
--- a/hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
+++ b/hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
@@ -49,15 +49,10 @@ extern "C" {
 #define LIS2DW12_ST_MODE_MODE1                  0x40
 #define LIS2DW12_ST_MODE_MODE2                  0x80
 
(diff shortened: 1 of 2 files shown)


hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
```diff
@@ -182,19 +182,23 @@ struct lis2dw12_cfg {
...
+    uint8_t inactivity_sleep_enable : 1;
+    uint8_t low_noise_enable : 1;
+    uint8_t stationary_detection_enable : 1;
+    uint8_t double_tap_event_enable : 1;
 
-    uint8_t int_pp_od;
-    uint8_t int_latched;
-    uint8_t int_active;
     uint8_t slp_mode;
     uint8_t self_test_mode;
```
vrahane: Since you are doing bit fields for config, making sure these get converted to bitfields as well LIS2DW12_PM_LP_MODEX could be 5 bits and LIS2DW12_HPF_M_XXX could be 3 bits.

PR #1015, opened 2018-04-13
nrf52840 support for qspi flash

kasjer:
This adds support for QSPI to NRF52840 MCU and NRF52840PDK


sterlinghughes:
👍

hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
```diff
@@ -0,0 +1,251 @@
...
+#include "os/mynewt.h"
+#if MYNEWT_VAL(QSPI_ENABLE)
+#include <mcu/cmsis_nvic.h>
+#include <hal/hal_flash_int.h>
+#include "mcu/nrf52_hal.h"
+#include "nrf.h"
+#include <nrfx/hal/nrf_qspi.h>
+
+static int
+nrf52k_qspi_read(const struct hal_flash *dev, uint32_t address,
```
mlaz: Why nrf52k ?

kasjer: same naming convention was used in hw/mcu/nordic/nrf52xxx/src/hal_flash.h
nrf52k_flash_read

kasjer:
@wes3 I though about putting all definition in MCU. I moved some of them to the BSP because there are not sensible default values for pins or flash size. If you think that it is more mynewt way to put them in MCU and then override them in BSP I have no objections.

sterlinghughes:
@kasjer I agree with @wes3, i think they should be defined in the MCU package.  I see your point regarding lack of sensible values -- the BSP will have to override them, but they are specific to configuring the Nordic MCU, and so that's where they should be defined.

wes3:
@kasjer I understand about the lack of sensible defaults. One way I have seen this done (and I am not saying this is good or bad) is that "nonsensical" values are assigned to these values by default. These nonsensical values are then checked in the code and if they are still nonsensical, meaning not overridden somewhere, the code fails to compile. I guess this could be done run time but generally it is done by the preprocessor. For example:

#if MYNEWT_VAL(QPSI_FLASH_SECTOR_SIZE) == 0
#error "Must override qspi flash sector size"
#endif

Or some such thing; I am sure you get the idea.

kasjer:
@wes3 I agree, I will move definitions to MCU and I will add preprocessor checks to detect unchanged values as you suggested, thanks.

utzig:
LGTM

hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
```diff
@@ -0,0 +1,279 @@
...
+#include "os/mynewt.h"
+#if MYNEWT_VAL(QSPI_ENABLE)
+#include <mcu/cmsis_nvic.h>
+#include <hal/hal_flash_int.h>
+#include "mcu/nrf52_hal.h"
+#include "nrf.h"
+#include <nrfx/hal/nrf_qspi.h>
+
+#if MYNEWT_VAL(QSPI_FLASH_SECTOR_COUNT) < 1
+#error QSPI_FLASH_SECTOR_COUNT must be set to the correct value in bsp.yml
```
mkiiskila: Not in bsp.yml but BSP's syscfg.yml. The same is true for the subsequent #error messages.

hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
```diff
@@ -0,0 +1,279 @@
...
+    const char src_not_in_ram = (((uint32_t) src) & 0xE0000000) != 0x20000000;
+
+    while ((NRF_QSPI->STATUS & QSPI_STATUS_READY_Msk) == 0)
+        ;
+
+    while (num_bytes != 0) {
+        const uint32_t page_start = address
+                & ~(MYNEWT_VAL(QSPI_FLASH_PAGE_SIZE) - 1);
+        const uint32_t page_limit = page_start
+                + MYNEWT_VAL(QSPI_FLASH_PAGE_SIZE);
```
mkiiskila: Coding style: move the page_start, page_limit declaration in the beginning of the function.
When you have a statement which expands to multiple lines, put the operator at the end of the line.
See coding style doc in the root of the repository.

hw/mcu/nordic/nrf52xxx/syscfg.yml
```diff
@@ -114,6 +114,76 @@ syscfg.defs:
         description: 'SOFT PWM'
         value: 0
 
+    QSPI_ENABLE:
+        descritption: 'NRF52 QSPI'
+        value: 0
+
+    QSPI_READOC:
+        descritption: >
```
mkiiskila: Typo: descritption should be description. Alos see QSPI_WRITEOC

hw/mcu/nordic/nrf52xxx/syscfg.yml
```diff
@@ -114,6 +114,76 @@ syscfg.defs:
...
+    QSPI_SCK_FREQ:
+        description: '32MHz clock divider (0-31). Clock = 32MHz / (1+divider)'
+        value: 0
+    QSPI_SPI_MODE:
+        description: 'SPI 0=Mode0 or 1=Mode3'
+        value: 0
+
+    QSPI_FLASH_SECTOR_SIZE:
+        description: 'QSPI sector size'
+        value: 4096
```
mkiiskila: Set default value to 0.

hw/mcu/nordic/nrf52xxx/syscfg.yml
```diff
@@ -114,6 +114,76 @@ syscfg.defs:
...
+    QSPI_FLASH_PAGE_SIZE:
+        description: 'QSPI page size. Writes can only be perfrmed to one page at a time.'
+        value: 256
```
mkiiskila: Set default value to 0.

PR #1024, opened 2018-04-14
lis2dw12: Support sleep change event for notifications

### LINKED ISSUE #1020: lis2dw12: int2 not routable for sensor notifications
int2 not routable for sensor notifications

### LINKED ISSUE #1022: lis2dw12: Support sleep change event for notifications
Support sleep change event for notifications which is part of int2

vrahane:
- lis2dw12: Support sleep change event for notifications
- Also cleaning up a bit, removing extraneous spaces, adding comments
- Also fixes #1020, lis2dw12: int2 not routable for sensor notifications
- This fixes #1022

### DIFF
--- a/hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
+++ b/hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
@@ -33,7 +33,7 @@ extern "C" {
 #define LIS2DW12_PM_LP_MODE4                    0x03
 #define LIS2DW12_PM_HIGH_PERF                   0x04
 #define LIS2DW12_PM_ON_DEMAND                   0x08
-    
+
 #define LIS2DW12_DATA_RATE_0HZ                  0x00
 #define LIS2DW12_DATA_RATE_1_6HZ                0x10
 #define LIS2DW12_DATA_RATE_12_5HZ               0x20
@@ -107,7 +107,6 @@ extern "C" {
 #define LIS2DW12_SIXD_SRC_XH                    0x02
 #define LIS2DW12_SIXD_SRC_XL                    0x01
 
-    
 enum lis2dw12_ths_6d {
     LIS2DW12_6D_THS_80_DEG = 0,
     LIS2DW12_6D_THS_70_DEG = 1,
@@ -131,19 +130,17 @@ enum lis2dw12_fifo_mode {
     LIS2DW12_FIFO_M_BYPASS_TO_CONTINUOUS = 4,
     LIS2DW12_FIFO_M_CONTINUOUS           = 6
 };
-    
+
 enum lis2dw12_read_mode {
     LIS2DW12_READ_M_POLL = 0,
     LIS2DW12_READ_M_STREAM = 1,
 };
-    
+
 struct lis2dw12_tap_settings {
-    uint8_t en_x;
-    uint8_t en_y; 
-    uint8_t en_z;
-    
-    uint8_t en_4d;
-    enum lis2dw12_ths_6d ths_6d;
+    uint8_t en_x  : 1;
+    uint8_t en_y  : 1;
+    uint8_t en_z  : 1;
+    uint8_t en_4d : 1;
     enum lis2dw12_tap_priority tap_priority;
 
     /* ths data is 5 bits, fs = +-2g */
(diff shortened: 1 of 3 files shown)


hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1946,89 +1943,106 @@ init_intpin(struct lis2dw12 * lis2dw12, hal_gpio_irq_handler_t handler,
...
-    rc = lis2dw12_set_int1_pin_cfg(itf, pdd->int_enable);
+    if (int_num == 0) {
+        rc = lis2dw12_set_int1_pin_cfg(itf, int_to_enable);
+    } else {
+        rc = lis2dw12_set_int2_pin_cfg(itf, int_to_enable);
+    }
+
     if (rc) {
-        return rc;
+        disable_interrupt(sensor, int_to_enable, int_num);
```
brolan-juul: Can you create a prototype for this function, or move it above? The function is statically defined and is defined below. I think depending on the C standard use, this may cause an error.

Error: repos/apache-mynewt-core/hw/drivers/sensors/lis2dw12/src/lis2dw12.c: In function 'enable_interrupt':
repos/apache-mynewt-core/hw/drivers/sensors/lis2dw12/src/lis2dw12.c:1994:9: error: implicit declaration of function 'disable_interrupt' [-Werror=implicit-function-declaration]
         disable_interrupt(sensor, int_to_enable, int_num);

hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
```diff
@@ -131,85 +130,99 @@ enum lis2dw12_fifo_mode {
...
        0 = 2*1/ODR, LSB = 4*1/ODR */
     uint8_t shock;
-    
+
+};
+
+/* Read mode configuration */
+struct lis2dw12_read_mode_cfg {
+    enum lis2dw12_read_mode mode;
+    uint8_t int_num:1;
```
brolan-juul: I am a bit unclear what this is used for given that lis2dw12_cfg struct contains int1_pin_cfg and int2_pin_cfg.

brolan-juul: Also, sensor_creator.c in master has this assigned to an enum type only.

https://github.com/apache/mynewt-core/blob/e0d46d6f58bcc81a74536591e03560458f349b14/hw/sensor/creator/src/sensor_creator.c#L801

Looks like this was changed to a struct after. sensor_creator needs to be updated if someone decides to use LIS2DW12_OFB.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1946,89 +1943,106 @@ init_intpin(struct lis2dw12 * lis2dw12, hal_gpio_irq_handler_t handler,
...
 
-    /*update which interrupts are enabled */
-    pdd->int_enable |= int_to_enable;
-    
+    pdd->int_enable++;
+
     /* enable interrupt in device */
-    rc = lis2dw12_set_int1_pin_cfg(itf, pdd->int_enable);
+    if (int_num == 0) {
+        rc = lis2dw12_set_int1_pin_cfg(itf, int_to_enable);
```
brolan-juul: Doesn't lis2dw12_set_int1_pin_cfg( ) need to read the register contents and "OR" it with the next interrupt? It doesn't right now and currently wipes the register clean it is invoked.

lis2dw12_set_int1_pin_cfg( ) is called in two different places:

- lis2dw12_config ( ) 
- lis2dw12_sensor_set_notification ( ) which invokes enable_interrupt( ) which then invokes lis2dw12_set_int1_pin_cfg ()

Thoughts?

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -2790,8 +2836,9 @@ lis2dw12_config(struct lis2dw12 *lis2dw12, struct lis2dw12_cfg *cfg)
         goto err;
     }
 
-    lis2dw12->cfg.stream_read_interrupt = cfg->stream_read_interrupt;
-    lis2dw12->cfg.read_mode = cfg->read_mode;    
+    lis2dw12->cfg.read_mode.int_cfg = cfg->read_mode.int_cfg;
+    lis2dw12->cfg.read_mode.int_num = cfg->read_mode.int_num;
+    lis2dw12->cfg.read_mode.mode = cfg->read_mode.mode;
     lis2dw12->cfg.mask = cfg->mask;
```
brolan-juul: Is the mask intended to encompass sensor_type_t & sensor_event_type_t?

vrahane: Nope, that is not the case.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
vrahane: I see what you mean here, `lis2dw12_set_int1_pin_cfg()` probably should just OR it then if we were to support multiple APIs from enabling interrupts and we can add a interrupt clear API to clear the int cfg. The initial thought was, notification/config would just override each other but that sounds like might be too limiting.

matthewwarnes: Previously it was ORed in the enable_interrupt function. If it should be done in the set_int1_pin_cfg function that is also fine. I believe it is quite important because without the ORing if a stream read is performed when even a single notification is registered the stream read will reset the interrupt for the notification.

PR #1029, opened 2018-04-16
sys/log: Improve log handler for slot1

andrzej-kaczmarek:
This PR adds multiple improvements to handling of log_fcb_slot1 handler. Most importantly:
- an optional cbmem buffer can be specified which is used as intermediate log when slot1 is locked
- slot1 can unlocked in runtime and data from intermediate buffer will be copied back to fcb
- convenient callback for app to reinitialize fcb when slot is unlocked
- automatic slot1 lock management by image manager
- log API to read entry to mbuf
- allow to register per-user module with a given name (this will be e.g. returned by newtmgr)
- add imgmgr command to erase image header (so slot1 can be unlocked if we decide not to perform update)


mgmt/imgmgr/src/imgmgr.c
```diff
@@ -28,12 +28,14 @@
 #include "bootutil/image.h"
 #include "bootutil/bootutil.h"
 #include "mgmt/mgmt.h"
+#include "log/log_fcb_slot1.h"
```
mkiiskila: I recommend making these #include lines to be behind #ifdef MYNEWT(LOG_FCB_SLOT1), as sys/log/<xxx> might not be visible to mgmt/imgmgr package.

Also, I think you're missing a change to pkg.yml, which conditionally includes sys/log/full iff LOG_FCB_SLOT1 is set. This should be full name of the package, not the 'log' api, as you have the API definition header file in sys/log/full only.

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -340,6 +356,45 @@ imgr_erase(struct mgmt_cbuf *cb)
...
+    CborError g_err = CborNoError;
+
+    area_id = imgmgr_find_best_area_id();
+    if (area_id >= 0) {
+        rc = flash_area_open(area_id, &fa);
+        if (rc) {
+            return MGMT_ERR_EINVAL;
+        }
+
+        rc = flash_area_erase(fa, 0, sizeof(struct image_header));
```
mkiiskila: Ignoring this rc? Looks like there's a similar issue in imgr_erase()...

sys/log/full/src/log_fcb_slot1.c
```diff
@@ -18,82 +18,282 @@
  */
 
 #include <string.h>
+#include "os/os.h"
 #include "log/log.h"
 #include "log/log_fcb_slot1.h"
 
 #if MYNEWT_VAL(LOG_FCB_SLOT1)
 
+static struct os_mutex g_mutex;
```
mkiiskila: g_mutex is a bad name. Use something more descriptive,  e.g. g_log_slot1_mtx.

util/cbmem/src/cbmem.c
```diff
@@ -277,6 +277,46 @@ cbmem_read(struct cbmem *cbmem, struct cbmem_entry_hdr *hdr, void *buf,
...
+        goto err;
+    }
+
+    rc = os_mbuf_append(om, (uint8_t *) hdr + sizeof(*hdr) + off, len);
+    if (rc != 0) {
+        cbmem_lock_release(cbmem);
+        goto err;
+    }
+
+    rc = cbmem_lock_release(cbmem);
```
mkiiskila: I would ignore return code from cbmem_lock_release() here. Call to os_mutex_release() is not going to fail.

mgmt/imgmgr/src/imgmgr.c
andrzej-kaczmarek: I added slot1 APIs to `sys/log/stub` and added `pkg.req_apis` here
this also allows to build app which uses slot1 with `sys/log/stub' if necessary

mgmt/imgmgr/src/imgmgr.c
andrzej-kaczmarek: added also commit that fixes this in `imgr_erase()`

andrzej-kaczmarek:
fixed comments and rebased

PR #1030, opened 2018-04-16
mempool: Add os_mempool_clear function

sjanc:
This is used to clear mempool (ie all block taken are invalidated)
without doing full initialisation of it. Unlike init this function
can be called multiple time on same mempool.

### DIFF
--- a/kernel/os/include/os/os_mempool.h
+++ b/kernel/os/include/os/os_mempool.h
@@ -189,6 +189,15 @@ os_error_t os_mempool_ext_init(struct os_mempool_ext *mpe, uint16_t blocks,
                                uint32_t block_size, void *membuf, char *name);
 
 /**
+ * Clears a memory pool.
+ *
+ * @param mp            The mempool to clear.
+ *
+ * @return os_error_t
+ */
+os_error_t os_mempool_clear(struct os_mempool *mp);
+
+/**
  * Performs an integrity check of the specified mempool.  This function
  * attempts to detect memory corruption in the specified memory pool.
  *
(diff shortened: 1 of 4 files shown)


sjanc:
Main motivation for this is https://github.com/apache/mynewt-nimble/issues/46 where we need to clear mempool on HCI reset.

kernel/os/src/os_mempool.c
```diff
@@ -135,6 +135,45 @@ os_mempool_ext_init(struct os_mempool_ext *mpe, uint16_t blocks,
     return 0;
 }
 
+os_error_t
+os_mempool_clear(struct os_mempool *mp)
```
rymanluk: Can this patch make use of this function already in os_mempool_init ?

kernel/os/include/os/os_mempool.h
```diff
@@ -188,6 +188,15 @@ os_error_t os_mempool_init(struct os_mempool *mp, uint16_t blocks,
 os_error_t os_mempool_ext_init(struct os_mempool_ext *mpe, uint16_t blocks,
                                uint32_t block_size, void *membuf, char *name);
 
+/**
+ * Clears a memory pool.
+ *
+ * @param mp            Pointer to a pointer to a mempool
```
ccollins476ad: Too many levels of indirection in this comment :). How about: "Mempool to clear"?

sjanc: copy and paste :) will fix that

kernel/os/src/os_mempool.c
sjanc: I was thinking on that too but in the end decided not to due to fact that clear takes initialized mempool

PR #1033, opened 2018-04-17
lis2dw12: Add support for set/clear int cfg bits

vrahane:
- Set/Clear individual interrupt cfg bits
- Change config in sensor creator
- Change int_active to int_active_low to be more explicit
- Change lis2dw12_set_int_active() to lis2dw12_set_int_active_low()

### DIFF
--- a/hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
+++ b/hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
@@ -261,7 +261,7 @@ struct lis2dw12 {
 /**
  * Reset lis2dw12
  *
- * @param The sensor interface
+ * @param itf The sensor interface
  *
  * @return 0 on success, non-zero on failure
  */
@@ -270,16 +270,16 @@ int lis2dw12_reset(struct sensor_itf *itf);
 /**
  * Get chip ID
  *
- * @param sensor interface
- * @param ptr to chip id to be filled up
+ * @param itf The sensor interface
+ * @param chip_id Ptr to chip id to be filled up
  */
 int lis2dw12_get_chip_id(struct sensor_itf *itf, uint8_t *chip_id);
 
 /**
  * Sets the full scale selection
  *
- * @param The sensor interface
- * @param fs to set
+ * @param itf The sensor interface
+ * @param fs The full scale value to set
  *
  * @return 0 on success, non-zero on failure
  */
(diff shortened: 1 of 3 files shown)


hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1949,92 +2034,93 @@ init_intpin(struct lis2dw12 *lis2dw12, hal_gpio_irq_handler_t handler,
...
     pdd = &lis2dw12->pdd;
 
-    rc = lis2dw12_clear_int(itf, &reg);
-    if (rc) {
-        goto err;
-    }
-
-    /* if no interrupts are currently in use enable int pin */
+    /* disable int pin */
     if (!pdd->int_enable) {
```
ccollins476ad: Should this be:
```
if (pdd->int_enable == 1) {
```
?

hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
```diff
@@ -579,6 +579,26 @@ lis2dw12_set_int1_pin_cfg(struct sensor_itf *itf, uint8_t cfg);
  */
 int lis2dw12_set_int2_pin_cfg(struct sensor_itf *itf, uint8_t cfg);
 
+/**
+ * Clear interrupt pin configuration for interrupt 1
+ *
+ * @param the sensor interface
```
ccollins476ad: `itf` is missing from this `@param` comment (same for `lis2dw12_clear_int1_pin_cfg`).

ccollins476ad:
I know I already mentioned this, but I get a little nervous about tracking the count of enabled interrupts without knowing which ones are actually enabled.  I don't know this sensor code very well, so I may well be worrying over something that doesn't actually matter.  I would just be sure that it isn't possible to disable the same interrupt twice, and/or disable interrupts too many times and cause the interrupt count to roll over.  There, I said my piece :).

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
vrahane: Taking care of this.

hw/drivers/sensors/lis2dw12/include/lis2dw12/lis2dw12.h
vrahane: Do you mean the `itf` text in the comments ? None of my comments have that. Is it needed ?

vrahane:
@ccollins476ad changed all the comments to match doxygen format. Please check it out.

PR #1037, opened 2018-04-18
hw/drivers/trng: Add driver for True Random Number Generator (TRNG)

andrzej-kaczmarek:
This is proposed hal_trng which abstracts TRNG (True Random Number Generator) available on many SoCs. It exposed just few simple APIs which should be enough for typical operation:
- read single u32 value
- fill buffer with random data - blocking and non-blocking

### DIFF
--- a/apps/trng_test/pkg.yml
+++ b/apps/trng_test/pkg.yml
@@ -1,0 +1,28 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+pkg.name: apps/trng_test
+pkg.type: app
+pkg.description: 
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+
+pkg.deps:
+    - kernel/os
+    - sys/console/full
+    - hw/drivers/trng
--- a/apps/trng_test/src/main.c
+++ b/apps/trng_test/src/main.c
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <stddef.h>
+#include <stdint.h>
+#include "sysinit/sysinit.h"
+#include "os/os.h"
+#include "console/console.h"
+#include "trng/trng.h"
+
+static uint8_t buf[128];
+
+static void
+print_buffer(void *ptr, size_t size)
+{
+    while (size) {
+        console_printf("%02x", *((uint8_t *) ptr));
+
+        size--;
+        ptr++;
+    }
+
+    console_printf("\n");
+}
+
+int
+main(void)
+{
+    struct trng_dev *trng;
+
+    sysinit();
+
+    trng = (struct trng_dev *) os_dev_open(MYNEWT_VAL(APP_TRNG_DEV),
+                                           OS_TIMEOUT_NEVER, NULL);
+    assert(trng);
+
+    os_time_delay(OS_TICKS_PER_SEC);
+
+    trng_read(trng, buf, sizeof(buf));
+    console_printf("%d bytes from os_dev:\n", sizeof(buf));
+    print_buffer(buf, sizeof(buf));
+
+    while (1) {
+        console_printf("os_dev -> %08x\n", (unsigned) trng_get_u32(trng));
+
+        os_time_delay(OS_TICKS_PER_SEC / 4);
+    }
+
+    return 0;
+}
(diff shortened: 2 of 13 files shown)


utzig:
LGTM

mkiiskila:
Good, we need something like this. However, I think this should be a hw/driver interface instead.
I'd like to be able to use random number generator from other sources than MCU only; like semtech SX chip which samples radio noise.

mkiiskila:
Exactly; I think that kind of flexibility would come handy for this.

andrzej-kaczmarek:
@mkiiskila I pushed additional patches which implements the same functionality but as a driver, we can choose now which one looks better ;-)

@wes3 yep, I implemented this more or less the same as in NimBLE controller in case of hal :-) as for driver implementation the question is whether this should be generic way of how it works or rather how it is implemented in particular driver - I think the latter is just easier to manage since `hw/drivers/trng` can be just a shim layer without extra logic inside.

apps/trng_test/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+
+//    hal_trng_read(buf, sizeof(buf));
+    trng_read(trng, buf, sizeof(buf));
+    for (int i = 0; i < sizeof(buf); i++) {
+        console_printf("%02x ", buf[i]);
+    }
+    console_printf("\n");
+
+    while (1) {
+//        console_printf("%08x\n", (unsigned) hal_trng_get_u32());
```
utzig: will this stay here?

apps/trng_test/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+main(void)
+{
+    struct trng_dev *trng;
+
+    sysinit();
+
+    trng = (struct trng_dev *) os_dev_open("nrf_rng", OS_TIMEOUT_NEVER, NULL);
+    assert(trng);
+
+//    hal_trng_read(buf, sizeof(buf));
```
utzig: will this stay here?

hw/bsp/nrf52840pdk/src/hal_bsp.c
```diff
@@ -200,6 +204,13 @@ hal_bsp_init(void)
     /* Make sure system clocks have started */
     hal_system_clock_start();
 
+    hal_trng_init();
+
+    rc = os_dev_create(&os_bsp_trng.dev, "nrf_rng",
```
utzig: wouldn't it be better to call it just `trng`?

apps/trng_test/pkg.yml
```diff
@@ -0,0 +1,28 @@
...
+pkg.type: app
+pkg.description: 
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+
+pkg.deps:
+    - kernel/os
+    - sys/console/full
+    - hw/drivers/trng/trng_nrf52
```
utzig: This breaks abstraction on anything other than `nrf52`, the deps should be `hw/drivers/trng`

hw/mcu/nordic/nrf52xxx/src/hal_trng.c
```diff
@@ -0,0 +1,165 @@
...
+        NRF_RNG->EVENTS_VALRDY = 0;
+        rng_cache[rng_cache_tail] = NRF_RNG->VALUE;
+        rng_cache_tail++;
+        if (rng_cache_tail >= sizeof(rng_cache)) {
+            rng_cache_tail = 0;
+        }
+    }
+
+    NRF_RNG->INTENCLR = 1;
+    NRF_RNG->EVENTS_VALRDY = 0;
```
utzig: This bit was already cleared above, inside the `if`, should it really be cleared again here?

andrzej-kaczmarek: no need to :)

hw/bsp/nrf52840pdk/src/hal_bsp.c
andrzej-kaczmarek: agree, but how about this: we may have both `hal_trng` and `hw/drivers/trng`. on platforms with TRNG in SoC it can be available as HAL and also as os_dev called `hal_trng`. obviously driver will use HAL implementaion instead of duplicating code as it is now.

apps/trng_test/pkg.yml
andrzej-kaczmarek: I did not intend to merge this PR with app (thus `[tmp]` prefix in commit message) and just pushed it for reference for discussion, but I can fix all the above comments and include it as well

hw/bsp/nrf52840pdk/src/hal_bsp.c
andrzej-kaczmarek: ...or it can be even called `trng` as suggested - like a main TRNG on BSP ;-)

utzig:
Why is not all functionality under `hw/drivers/trng/...`? Any reason for having a `hal_` as well?

andrzej-kaczmarek:
well, since there no conclusion here which would be better I did both :)
you can choose to either use `hal_` directly (i.e. you know you want TRNG from MCU) or you can use `hw/driver/trng` and select currently available HAL or other driver at any time (i.e. you want to test things). I know it's a bit redundant, but does not cost us anything.

mkiiskila:
I would've only implemented the nonblocking version of the read; presumably the user would prefer to have some of sleep to let entropy buildup in between calls to rng. But I'm ok leaving it there as is.

utzig:
Ok, I'd rather have this merged now, but I still think it is a bad idea to leave the duplication. I would prefer to have only the `hw/drivers/trng` (or just `hal_trng` with good reason). I plan to do single interface style on stm32 (no hal_).

andrzej-kaczmarek:
@utzig any reason why you do not want to implement `hal_trng`? it's basically the same code as for driver, just in different place (and `trng_hal` driver will provide `os_dev`-like access). I can just remove `hal_trng` from this PR since it does not make sense to merge something that will not be used.

@mkiiskila non-blocking only makes sense to me since blocking one is just thin wrapper on this which app can easily do if needed - will remove blocking version

utzig:
@andrzej-kaczmarek I don't see how could be a good idea that every `trng` has to implement both a "hal driver" and a `hw/driver/trng`. Is that what you are suggesting? and even leaving it loose for the user to choose, that breaks abstraction if some MCUs implement a hal while others implement a `hw/drivers/trng`. Seems better to just settle on a common/single interface.

andrzej-kaczmarek:
with current implementation:
- MCU implements `hal_trng` which can be accessed either directly or via generic driver `trng_hal` (which in fact is just shim layer on top of `hal_trng`)
- non-MCU TRNGs need dedicated driver

so there is no need to implement both `hal_trng` and TRNG driver for MCU - just `hal_trng` is enough to have both interfaces.

see e.g. https://github.com/apache/mynewt-core/pull/1037/commits/54f2e34328d1320a19d2f7563f3e0b391314279f

andrzej-kaczmarek:
Ok, so since most opinions are in favour of driver for TRNG, I pushed new code which contains only TRNG driver with implementation for nRF52xxx. `nrf52840pdk` has this driver include as `trng` device with `TRNG` syscfg flag.

Also there is only non-blocking version of read API since blocking one is just a wrapper on this one so no need for special driver API here. The `get32` API is still there since it's just convenient way to get simple integer and this one is obviously blocking.

PR #1038, opened 2018-04-18
sys/console: Add console history functionality

michal-narajowski:
Use up and down arrows to browse history.
Duplicate lines are dropped.

### DIFF
--- a/sys/console/full/src/console.c
+++ b/sys/console/full/src/console.c
@@ -172,6 +172,14 @@ cursor_backward(unsigned int count)
 }
 
 static inline void
+cursor_clear_line(void)
+{
+    console_out(ESC);
+    console_out('[');
+    console_out('K');
+}
+
+static inline void
 cursor_save(void)
 {
     console_out(ESC);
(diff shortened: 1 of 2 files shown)


utzig:
There are some weird C++ style comments and no spaces after some `if` and `while` clauses, but apart from that looks good.

michal-narajowski:
Thanks, @utzig. Fixed.

sys/console/full/syscfg.yml
```diff
@@ -43,6 +43,12 @@ syscfg.defs:
     CONSOLE_MAX_INPUT_LEN:
         description: 'Maximum input line length'
         value: 256
+    CONSOLE_HIST:
```
andrzej-kaczmarek: I'd call this `CONSOLE_HISTORY` (but I'm ok with `console_hist` prefix in code)

sys/console/full/src/console.c
```diff
@@ -248,6 +248,186 @@ del_char(char *pos, uint16_t end)
...
+    cur = 0;
+    while (cur < end) {
+        console_out(' ');
+        ++cur;
+    }
+    if (cur) {
+        cursor_backward(cur);
+    }
+    end = 0;
+    cur = 0;
```
andrzej-kaczmarek: I think this can be substituted by simple escape sequence conbination of `ESC [G` (move to 1st column) + `ESC [K` (clear from cursor to eol)

sys/console/full/src/console.c
```diff
@@ -248,6 +248,186 @@ del_char(char *pos, uint16_t end)
...
+    /* Trim trailing space */
+    end = str + strlen(str) - 1;
+    while (end > str && isspace((unsigned char)*end)) {
+        end--;
+    }
+
+    end++;
+
+    /* Set output size to minimum of trimmed string length and buffer size minus 1 */
+    out_size = (end - str) < len - 1 ? (end - str) : len - 1;
```
andrzej-kaczmarek: we have `min()` macro ;-)

michal-narajowski:
@andrzej-kaczmarek Thanks, fixed. I couldn't use `ESC [G` because that would erase also the prompt.

PR #840, opened 2018-02-23
Feature/stm32f3

mlampert:
Adds support for the STM32F3 family of processors and two of their development boards:
 - nucleo-f303k8
 - nucleo-f303re

PR #1039, opened 2018-04-18
Add stm32 common hal_timer (f4/f7/f3/l1)

utzig:
Add stm32 common hal_timer (f4/f7/f3/l1/f1)


utzig:
1) @mkiiskila On the F1 I had an issue where two consecutive calls to `hal_timer_read` could return a smaller value in the second call. I fixed it by applying this patch: https://gist.github.com/utzig/59f2be8e4c94d76245f140003a24d0b2. I have not tracked the exact reason for the race but the patch fixes it and doesn't seem to break any of  the other models.

2) @mlampert While merging the `hal_timer` from all models I found this commit you did on the STM32F3: https://github.com/apache/mynewt-core/pull/840/commits/440eb1b6fca50e6c3df3ac44a8a8224368480e44. I have removed it, but I could suggest you open a new PR after this one is merged with just that patch.

mlampert:
I added that mostly for myself ;) There aren't any guards of this type anywhere in the system, not sure if it should be brought back.

Thanks for tackling this - desperately needed for the PWM driver!

hw/bsp/stm32f429discovery/src/hal_bsp.c
```diff
@@ -42,17 +42,18 @@
...
     [0] = {
-        .suc_uart = USART2,
-        .suc_rcc_reg = &RCC->APB1ENR,
-        .suc_rcc_dev = RCC_APB1ENR_USART2EN,
-        .suc_pin_tx = MCU_GPIO_PORTD(5),	/* PD5 */
-        .suc_pin_rx = MCU_GPIO_PORTD(6),	/* PD6 */
+        .suc_uart = USART1,
+        .suc_rcc_reg = &RCC->APB2ENR,
+        .suc_rcc_dev = RCC_APB2ENR_USART1EN,
+        .suc_pin_tx = MCU_GPIO_PORTA(9),
```
mkiiskila: You might want to send a note out when changing console UART. If there's someone using this BSP, they'll be surprised after pulling in this change.

hw/mcu/stm/stm32_common/src/hal_timer.c
```diff
@@ -308,26 +409,80 @@ stm32f7_hw_setdown(TIM_TypeDef *regs)
...
-    stm32f7_hw_setup(num, tmr->sht_regs);
+    stm32_hw_setup(num, regs);
 
     /*
-     * Stop the timers at debugger. XXX Which TIM?
+     * Stop the timers at debugger.
      */
-    DBGMCU->APB1FZ |= 0x1ff; /* TIM2 - TIM7, TIM12-TIM14 */
-    DBGMCU->APB2FZ |= 0x70003; /* TIM1, TIM8-TIM11 */
+#ifdef TIM1
```
mkiiskila: Opinion: this is a lot of code space being used for very little gain.

hw/mcu/stm/stm32f1xx/include/mcu/stm32_hal.h
```diff
@@ -62,6 +62,13 @@ struct stm32_hal_spi_cfg {
 #include "stm32f1xx_hal_uart.h"
 #include "mcu/stm32f1_bsp.h"
 
+/* hal_timer */
+#include "stm32f1xx_hal_tim.h"
+
+#define STM32_HAL_TIMER_MAX     (3)
+#define STM32_OFLOW_VALUE       (0x10000UL)
+#define STM32_NSEC_PER_SEC      (1000000000UL)
```
mkiiskila: I would keep STM32_OFLOW_VALUE and STM32_NSEC_PER_SEC inside c-file. They are of no use as an exported API.

hw/bsp/stm32f429discovery/src/hal_bsp.c
mkiiskila: That is a good question. Having console go via that seems like a better option for ease of use.

hw/mcu/stm/stm32_common/src/hal_timer.c
utzig: What bothers you here is the amount of code (size of .c file) or the amount of generated code (size of .elf/.bin file)?

utzig: In any case I don't see any problem in just freezing them all, but I'll just use the "macro" names instead.

PR #1045, opened 2018-04-20
LIS2DW12: Reset Update

brolan-juul:
LIS2DW12: Updated reset per application note procedures.


brolan-juul:
Ok. Closing as there were no previous side effects from the original code.

PR #1049, opened 2018-04-24
kernel/os: Add helpers for converting ms<->ticks

andrzej-kaczmarek:
These helpers do not check for overflow so should be used with care,
but in general they allow for more compact code.

Note that would be nice to have them as static inline but since we need
OS_TICKS_PER_SEC in header file, there are some weird dependencies in
os/os_*.h files which needs to be resolved first so code can build
properly.

### DIFF
--- a/kernel/os/include/os/os_time.h
+++ b/kernel/os/include/os/os_time.h
@@ -213,6 +213,31 @@ int os_time_ms_to_ticks(uint32_t ms, uint32_t *out_ticks);
  */
 int os_time_ticks_to_ms(uint32_t ticks, uint32_t *out_ms);
 
+
+/**
+ * Converts milliseconds to OS ticks.
+ *
+ * This function does not check if conversion overflows and should be only used
+ * in cases where input is known to be small enough not to overflow.
+ *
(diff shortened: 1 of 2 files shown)


ccollins476ad:
We already have `os_time_ms_to_ticks()` (which does test for overflow).  Do we need both?

andrzej-kaczmarek:
it's a convenience function instead of using expressions with `OS_TICKS_PER_SEC` explicitly, e.g.
https://github.com/apache/mynewt-nimble/pull/61/commits/41e06a279a8b74d348546fead661d7dfbbfc76ab#diff-468a6a208a7bf9df02d7a92471431fefL357

in case of nimble this is included in porting layer so we don't need define `OS_TICKS_PER_SEC` - probably it would be just enough for any system, but I don't really know

ccollins476ad:
So the new functions are useful because they return the result rather than a status code?  Makes sense.  Perhaps  `os_time_ticks_to_ms32()` should be implemented by calling the existing function.  I haven't tried it, but I imagine gcc would optimize away the range check if the return code is unused.

PR #1050, opened 2018-04-24
 kernel/os: Make os_callout_remaining_ticks thread-safe

andrzej-kaczmarek:
Not sure why these 2 APIs expected to be called within critical section
only, but let's make them callable from anywhere just like other APIs.

### DIFF
--- a/kernel/os/src/os_callout.c
+++ b/kernel/os/src/os_callout.c
@@ -170,15 +170,18 @@ os_callout_wakeup_ticks(os_time_t now)
 os_time_t
 os_callout_remaining_ticks(struct os_callout *c, os_time_t now)
 {
+    os_sr_t sr;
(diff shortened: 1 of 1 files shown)


mkiiskila:
These are only to be called from clock ticker interrupt handler, or idle task for figuring out how long to put MCU to sleep for. These are not part of API exported to regular packages, just to MCU/BSP.

So the OS_ASSERT_CRITICAL() is intentional, I don't think it should be changed.

andrzej-kaczmarek:
ok, so I see `os_callout_wakeup_ticks` is indeed hidden but `os_callout_remaining_ticks` seems like a public API (used by mesh) so should be changed for consistency

mkiiskila:
Sure, that should be fine.

PR #1054, opened 2018-04-25
hw/util/button: allow to specify eventq

sdalu:
Allow to specify event queue for internal processing. And eventq (per button if requested) for the callback processing


sdalu:
@ccollins476ad, can it be merged?

PR #1058, opened 2018-04-26
SensorAPI: lis2dw12: Breaking up sensor notifications

vrahane:
- Earlier the notifications were stacked to give one single callback
  for a combination of event notifications, as part of this commit the
  notifications events are handled as separate events so that events do
  not get missed while the notifications are being serviced
- Adding an event pool for the notifications with a default size of 5,
  this is configurable using SENSOR_NOTIF_EVENTS_MAX mynewt syscfg
  variable
- Adding stats for notifications to track which event notification got
  triggered
- Fixing doxygen style comments in sensor.h.
- Changing other drivers to use the event_put API per type of event,
  specifying an event
- Adding Sleep and Wake Up event support

### DIFF
--- a/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
+++ b/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
@@ -33,6 +33,15 @@
 #include "log/log.h"
 #include "stats/stats.h"
 
+const struct lis2dw12_notif_cfg dflt_notif_cfg[] = {
+    { SENSOR_EVENT_TYPE_SINGLE_TAP,   0, LIS2DW12_INT1_CFG_SINGLE_TAP  },
+    { SENSOR_EVENT_TYPE_DOUBLE_TAP,   0, LIS2DW12_INT1_CFG_DOUBLE_TAP  },
+    { SENSOR_EVENT_TYPE_SLEEP,        1, LIS2DW12_INT2_CFG_SLEEP_STATE },
+    { SENSOR_EVENT_TYPE_FREE_FALL,    0, LIS2DW12_INT1_CFG_FF          },
+    { SENSOR_EVENT_TYPE_WAKEUP,       0, LIS2DW12_INT1_CFG_WU          },
+    { SENSOR_EVENT_TYPE_SLEEP_CHANGE, 1, LIS2DW12_INT2_CFG_SLEEP_CHG   }
+};
+
 static struct hal_spi_settings spi_lis2dw12_settings = {
     .data_order = HAL_SPI_MSB_FIRST,
     .data_mode  = HAL_SPI_MODE3,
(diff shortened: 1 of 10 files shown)


hw/sensor/src/sensor.c
```diff
@@ -76,15 +76,15 @@ static struct os_event sensor_interrupt_event = {
...
-
-  /** OS event - for doing a sensor read */
+/** OS event - for doing a sensor read */
 static struct os_event sensor_read_event = {
     .ev_cb = sensor_read_ev_cb,
 };
 
+static struct os_mempool sensor_notify_evt_pool;
+static uint8_t sensor_notify_evt_area[OS_MEMPOOL_BYTES(MYNEWT_VAL(SENSOR_NOTIF_EVENTS_MAX),
+      sizeof(struct os_event))];
```
ccollins476ad: Shouldn't this be `sizeof (struct sensor_notify_os_ev)`?

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -2549,25 +2567,76 @@ lis2dw12_sensor_set_config(struct sensor *sensor, void *cfg)
...
+    (void)snoe;
+
     lis2dw12 = (struct lis2dw12 *)SENSOR_GET_DEVICE(sensor);
     itf = SENSOR_GET_ITF(sensor);
 
+    if (lis2dw12->pdd.notify_ctx.snec_evtype & SENSOR_EVENT_TYPE_SLEEP) {
+        /*
+         * We need to read this register only if Sleep event if we are
+         * interested in the sleep event
+         */
```
ccollins476ad: Nitpick: This comment is a bit odd :)

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -2549,25 +2567,76 @@ lis2dw12_sensor_set_config(struct sensor *sensor, void *cfg)
...
 lis2dw12_sensor_handle_interrupt(struct sensor *sensor)
 {
+    struct sensor_notify_os_ev *snoe;
     struct lis2dw12 *lis2dw12;
     struct sensor_itf *itf;
     uint8_t int_src;
+    uint8_t int_status;
     int rc;
 
+    (void)snoe;
```
ccollins476ad: `snoe` never gets used.  Can it be removed from this function?

hw/sensor/src/sensor.c
```diff
@@ -1162,18 +1184,26 @@ sensor_interrupt_ev_cb(struct os_event *ev)
...
-    const struct sensor_notify_ev_ctx *ctx;
+    struct sensor_notify_os_ev *snoe;
     const struct sensor_notifier *notifier;
 
-    ctx = ev->ev_arg;
+    snoe = ev->ev_arg;
+
+    if (!snoe) {
+        return;
+    }
```
ccollins476ad: It doesn't seem like `snoe` can ever be null, but if it does happen, then the event gets leaked here.  The event callback should never return without freeing the event.

hw/sensor/src/sensor.c
```diff
@@ -1162,18 +1184,26 @@ sensor_interrupt_ev_cb(struct os_event *ev)
...
+            notifier->sn_func(snoe->snoe_sensor,
                               notifier->sn_arg,
-                              ctx->snec_evtype);
+                              snoe->snoe_evtype);
+            break;
         }
     }
+
+    /* Remove notify os event from pool */
+    os_memblock_put(&sensor_notify_evt_pool, snoe);
```
ccollins476ad: Nitpick: I think this comment isn't quite right.  This line puts the event back into the pool rather than removing it.

vrahane: cool, will fix it.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
vrahane: yup, correct. sorry about that.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
vrahane: I'll fix that.

hw/sensor/src/sensor.c
vrahane: yup, good catch. Thanks.

vrahane:
I think your comment about the `sensor_interrupt_event` is correct, we haven't hit that use case yet but it does make sense to make it per sensor.  I will fix that as well. Thanks a lot for going through it.

hw/sensor/src/sensor.c
vrahane: Actually, snoe is the event, I agree chances of it becoming NULL are slim but it definitely cannot be put back into the pool since, it is NULL and the block address would be invalid then.

vrahane:
@ccollins476ad I made changes as per review comments. I have also made one more change to     - decouple events wrt bits in int_src register from interrupts to allow checking of different bits for the same interrupt for a different event. Adding a config for this so that developer can change the notification config if required. It uses the default config generally if not over written.

PR #1067, opened 2018-05-02
Battery manager

kasjer:
Battery management provides abstraction layer so an application
can get information about battery without hardware specific (fuel gauge)
code.

The battery is associated with set of properties (provided by hardware
driver or battery manager itself). Properties represent physical
values measured by fuel gauges or ADC converter that application may
be interested in.
The application specify how frequently battery information should be read
and can subscribe to some subset of data that is interesting for the application.
Typically application will subscribe to voltage, temperature or current readings.

On the lower layer battery API specifies simple interface that fuel gauge need to
provide.

The battery information can be provided by several drivers (fuel gauge and ADC).

This pull request also extends bq27z561 driver to conform to battery API
It adds battery ADC driver that adds additional property to battery.
Battery shell is also added

### DIFF
--- a/hw/battery/include/battery/battery.h
+++ b/hw/battery/include/battery/battery.h
@@ -1,0 +1,308 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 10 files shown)


kasjer:
@wes3 Will I'm not able to add reviewer so far, but I heard that when mentioned person do get notification, so fill free to take a look.

PR #1070, opened 2018-05-04
btshell: Fix scan command in legacy mode

sjanc:
When starting legacy scan code was also trying to start extended scan
with garbage parameters.


sjanc:
Hi Sterling,

This is something we were discussing with guys in the office, the thing is that our BLE sample apps are Mynewt applications and it feels like those should be in mynewt-core repo. Those just depends on apache-nimble for functionality.  

That said, we could have apps/mynewt apps/freertos apps/linux etc in apache-nimble and keep those there per OS supported. I'm fine with that too.  I'll create issue for this.

PR #1072, opened 2018-05-05
drv2605 erm support

davidzuhn:
This patch adds support for ERM motors to the DRV2605 module, which heretofore only supported LRA devices.   

There are also a handful of documentation corrections for the drv2605 driver included as well.


jacobrosenthal:
Cant build at the moment, and dont have an ERM on me, but looks pretty straightforward. LGTM

PR #1075, opened 2018-05-09
sys/config: Automatically restore on startup

ccollins476ad:
Before PR: the application needed to manually restore the persisted configuration via a call to `conf_load()`, typically from `main()`.

After PR: `conf_load()` gets called automatically at startup.  It doesn't get called directly from `sysinit()`, as the underlying storage hasn't been configured at that point.  Instead, values are loaded by an event which config init function enqueues to the default event queue.  When `main()` has completed its runtime configuration, including that of the sys/config storage, it processes the default eventq, at which point `conf_load()` gets called.

There is one danger here: some packages may try to use a configuration setting before `conf_load()` has been called.  However, this is not a new issue; high priority tasks, created by `sysinit()`, would already preempt `main()` before `conf_load()` could be called.  To address this problem, I believe we need a separate fix.

### DIFF
--- a/apps/bleprph/src/main.c
+++ b/apps/bleprph/src/main.c
@@ -331,8 +331,6 @@ main(void)
     phy_init();
 #endif
 
-    conf_load();
-
     /* If this app is acting as the loader in a split image setup, jump into
      * the second stage application instead of starting the OS.
      */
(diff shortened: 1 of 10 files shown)


mkiiskila:
I think this PR warrants an email to dev-list, as this will have effect on all apps using sys/config.

PR #1076, opened 2018-05-09
Battery charger ADP5061 driver

kasjer:
Battery charger driver to be used with charge-control API.


hw/drivers/chg_ctrl/adp5061/src/adp5061.c
```diff
@@ -0,0 +1,583 @@
...
+ * under the License.
+ */
+
+#include <string.h>
+#include <errno.h>
+#include <assert.h>
+
+#include <defs/error.h>
+#include <os/os.h>
+#include <sysinit/sysinit.h>
```
mkiiskila: It would be better to replace these 3 include lines with one #include <os/mynewt.h>. We'll want to include syscfg.h before anything else from mynewt tree, and that convenience header file does exactly that.

hw/drivers/chg_ctrl/adp5061/include/adp5061/adp5061.h
```diff
@@ -0,0 +1,965 @@
+#ifndef _ADP5061_H
```
mkiiskila: Need to add apache license as the first thing.

hw/drivers/chg_ctrl/adp5061/syscfg.yml
```diff
@@ -0,0 +1,10 @@
+syscfg.defs:
```
mkiiskila: Need to add apache license.

hw/drivers/chg_ctrl/adp5061/src/adp5061.c
```diff
@@ -0,0 +1,583 @@
...
+    return rc;
+}
+
+int
+adp5061_set_reg(struct adp5061_dev *dev, uint8_t addr, uint8_t value)
+{
+    int rc = 0;
+    uint8_t payload[2] = {};
+    payload[0] = addr;
+    payload[1] = value;
```
mkiiskila: We try to keep variable declarations in the top, and not declare vars in the middle. To make that true here, you could do e.g.:
uint8_t payload[2] = { addr, value };

kasjer:
@mkiiskila all suggestions addressed in forced update, thanks for the review

PR #1078, opened 2018-05-10
blemesh: Remove bt_mesh_register_gatt from app code

rymanluk:
Since it is possible to have mesh along with regular BLE, lets allow to
make it use in btshell.

### DIFF
--- a/apps/blemesh/src/main.c
+++ b/apps/blemesh/src/main.c
@@ -427,7 +427,6 @@ main(int argc, char **argv)
 
     hal_gpio_init_out(LED_2, 0);
 
-    bt_mesh_register_gatt();
     health_pub_init();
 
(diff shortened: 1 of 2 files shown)


apps/btshell/src/main.c
```diff
@@ -2082,6 +2086,10 @@ main(int argc, char **argv)
     rc = gatt_svr_init();
     assert(rc == 0);
 
+#if MYNEWT_VAL(BLE_MESH) == 1
+    bt_mesh_register_gatt();
```
andrzej-kaczmarek: why not just do this as mesh pkg init? this is what other packages do.

rymanluk:
This should go along with https://github.com/apache/mynewt-nimble/pull/86

PR #1090, opened 2018-05-15
app/btshell: Increase block count if mesh enabled

kolodgrz:
Memory error may occure while operating with enabled mesh and additional
instances of advertising. This update was tested agains initialized mesh
and one separate instance of advertising.


kolodgrz:
@andrzej-kaczmarek @rymanluk @michal-narajowski

PR #1092, opened 2018-05-15
Fix nucleo767 BSP naming convention

utzig:


mlaz:
LGTM

mkiiskila:
Changing name of BSP. Maybe send an email to devlist reporting that this happened? In case of users.

PR #1097, opened 2018-05-16
crypto/tinycrypt: Allow to set uECC RNG at package init

andrzej-kaczmarek:
Right now uECC RNG is not set by TinyCrypt so package which uses it
should configure it on its own. This may create conflicts when multiple
packages set own uECC RNG function so it's not quite clear which one
will be set last (and effectively will be used).

This patch adds option to set uECC RNG at package init so there is no
need for other packages to override this setting. The only available
option for now is TRNG driver.

Note that other packages may still overwrite uECC RNG function. This
patch is not to force any particular uECC RNG, just to have settings
to set this by default.


utzig:
LGTM

PR #1100, opened 2018-05-16
RFC - upgrade CMSIS to 5.3.0

utzig:
This is a RFC for upgrading to CMSIS 5.3.0.

CMSIS 5.x provides per arch `NVIC_SetVector` and `NVIC_GetVector` which is being used here for CM3/4/7. For CM0 I added a custom header which uses the `__vector_tbl_reloc__` symbol from the linker files. This might not be the most clean way to do it, because I ended up adding defines to the m0 compiler flags.

I tested this on at least one target of each currently supported ARM cpu.

### DIFF
--- a/hw/cmsis-core/src/ext/mynewt_cm0_vectab.h
+++ b/hw/cmsis-core/src/ext/mynewt_cm0_vectab.h
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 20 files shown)


hw/cmsis-core/src/ext/mynewt_cm0_vectab.h
```diff
@@ -0,0 +1,28 @@
+#ifndef MYNEWT_CM0_VECTAB_H
```
mkiiskila: Apache copyright missing.

utzig:
Thanks @mkiiskila!

PR #1101, opened 2018-05-16
HiFive1 add missing peripheral driver

kasjer:
Hal was missing SPI driver.
Now two SPI controllers can be configured.

Those two controller can only work as masters.
There is no 16 bits support nor slave (hardware limitation).
MISO/MOSI/SCK pins are hardwired and can't be configured.
Hardware suport slave select pin selection, but SPI HAL does cover chip select handling.

Optimization flag -O1 change to -Og for batter debug experience.

### DIFF
--- a/hw/mcu/sifive/fe310/src/hal_spi.c
+++ b/hw/mcu/sifive/fe310/src/hal_spi.c
@@ -1,0 +1,410 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
+#include <os/mynewt.h>
+#include <assert.h>
+
+#include <hal/hal_spi.h>
+#include <hal/hal_gpio.h>
+
+#include <defs/error.h>
+#include <syscfg/syscfg.h>
+
+#include <mcu/fe310_hal.h>
+#include <mcu/sys_clock.h>
+#include <mcu/plic.h>
+#include <sifive/devices/spi.h>
+#include <env/freedom-e300-hifive1/platform.h>
+
+struct fe310_hal_spi {
+    struct hal_spi_settings spi_cfg;
+
+    uint32_t spi_base;
+
+    int len;
+    int txleft;
+    int rxleft;
+
+    /* Pointers to tx/rx buffers */
+    uint8_t *txbuf;
+    uint8_t *rxbuf;
+
+    /* Callback and arguments */
+    hal_spi_txrx_cb txrx_cb_func;
+    void            *txrx_cb_arg;
+};
+
+#if MYNEWT_VAL(SPI_1)
+struct fe310_hal_spi fe310_hal_spi1 = {
+    .spi_base = SPI1_CTRL_ADDR
+};
+#endif
+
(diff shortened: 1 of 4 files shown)


hw/mcu/sifive/fe310/src/hal_spi.c
```diff
@@ -0,0 +1,384 @@
...
+#include <syscfg/syscfg.h>
+
+#include <mcu/fe310_hal.h>
+#include <mcu/sys_clock.h>
+#include <mcu/plic.h>
+#include <sifive/devices/spi.h>
+#include <env/freedom-e300-hifive1/platform.h>
+
+struct fe310_hal_spi
+{
```
mkiiskila: Coding style nit: move this '{' on the line above.

hw/mcu/sifive/fe310/src/hal_spi.c
```diff
@@ -0,0 +1,384 @@
...
+#endif
+
+#if MYNEWT_VAL(SPI_2)
+struct fe310_hal_spi fe310_hal_spi2 = {
+    .spi_base = SPI2_CTRL_ADDR
+};
+#endif
+
+static struct fe310_hal_spi *fe310_hal_spis[3] = {
+    NULL, /* SPI 0 used to access flash */
```
mkiiskila: Is this always the case for this MCU?

hw/mcu/sifive/fe310/src/hal_spi.c
```diff
@@ -0,0 +1,384 @@
...
+        rc = SYS_EINVAL;                                    \
+        goto err;                                           \
+    }
+
+static void
+spi_interrupt_handler(int int_num)
+{
+    struct fe310_hal_spi *spi = fe310_hal_spis[int_num - INT_SPI0_BASE];
+    uint32_t val;
+    for (;;) {
```
mkiiskila: Is it possible to cap this to something? It would be best if we would have some limit on time spent within interrupt handler. I'm not familiar with this MCUs interrupt delivery mechanism, nor priorities. So I'm not sure if earlier return would gain anything, or if the system would get stuck without being able to any other work anyway.

hw/mcu/sifive/fe310/src/hal_spi.c
```diff
@@ -0,0 +1,384 @@
...
+        goto err;
+    }
+
+    plic_set_handler(INT_SPI0_BASE + spi_num, spi_interrupt_handler, 3);
+    if (spi_num == 1 && MYNEWT_VAL(SPI_1)) {
+        _REG32(GPIO_CTRL_ADDR, GPIO_IOF_SEL) &= ~(7 << 3);
+        _REG32(GPIO_CTRL_ADDR, GPIO_IOF_EN) |= (7 << 3);
+    } else if (spi_num == 2 && MYNEWT_VAL(SPI_2)) {
+        _REG32(GPIO_CTRL_ADDR, GPIO_IOF_SEL) &= ~(7 << 27);
+        _REG32(GPIO_CTRL_ADDR, GPIO_IOF_EN) |= (7 << 27);
```
mkiiskila: Is there any, more descriptive, defines to use instead of '3', '7' and '27'?

hw/mcu/sifive/fe310/src/hal_spi.c
kasjer: done

hw/mcu/sifive/fe310/src/hal_spi.c
kasjer: Mathilda: Is life always this hard, or is it just when you're a kid?
Léon: Always like this.

SPI0 is always for QSPI flash. Pins can't be configured as GPIO.
SPI2 is not routed to external Pins (in current hardwarefor now)
SPI1 is the only SPI that can be used.

hw/mcu/sifive/fe310/src/hal_spi.c
kasjer: Added sanity cap

hw/mcu/sifive/fe310/src/hal_spi.c
kasjer: Proper defines found at last.

kasjer:
@mkiiskila Thanks for review, I addressed all you points. I also allowed to have rxbuf set to NULL. Before I checked STM32 code I saw that it will not work for rxbuf == NULL and I thought this is how HAL was designed. Then I noticed that @wes3 used rxbuf = NULL for led driver. So now risc-v mcu allows rxbuf = NULL as well.

PR #1103, opened 2018-05-17
sensor/lis2dw12: Fix lis2dw12_set_tap_cfg

kasjer:
Z tap threshold was wrongly masked with register number LIS2DW12_REG_TAP_THS_Z
instead of mask LIS2DW12_TAP_THS_Z_THS.


vrahane:
Ok, sounds fine. We are not using 7 as a threshold value for double taps, hence probably missing this issue. Thank you for fixing it.

PR #1105, opened 2018-05-17
hw/drivers: Add TLC5971 driver

wes3:
This is the first cut at a TLC5971 driver. This allows the user to control the global brightness and individual RGB leds in each of the four channels of the chip.

### DIFF
--- a/hw/drivers/tlc5971/src/tlc5971.c
+++ b/hw/drivers/tlc5971/src/tlc5971.c
@@ -19,7 +19,6 @@
 #include <string.h>
 #include "os/mynewt.h"
 #include "tlc5971/tlc5971.h"
-#include "hal/hal_gpio.h"
 #include "hal/hal_spi.h"
 
 /**
@@ -31,7 +30,7 @@
  * @param wait The amount of time to wait to open (if needed).
  * @param arg device open arg (not used)
  *
- * @return int
+ * @return int 0:success; error code otherwise
  */
 static int
 tlc5971_open(struct os_dev *odev, uint32_t wait, void *arg)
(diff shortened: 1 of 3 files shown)


wes3:
Thanks for the review Jerzy! I addressed all of your comments except the global buffer as opposed to stack. Since it is fairly small I decided to leave it global as it might be nice to look at in the debugger if something is going wrong.

hw/drivers/tlc5971/src/tlc5971.c
```diff
@@ -0,0 +1,270 @@
...
+        return -1;
+    }
+
+    /*
+     * XXX: for now, disable interrupts around write as it is possible that
+     * too long a gap will cause mis-program of device.
+     */
+    tlc5971_construct_packet(dev);
+
+    OS_ENTER_CRITICAL(sr);
```
vrahane: This is not needed. The user/library/app should use a mutex to access the device anyways. The packet is really big (217 bits). Let me know what you think ?

wes3: The reason the OS_ENTER_CRITICAL() is in the code is due to the chip specification. I might be reading this wrong but if the chip sees a gap of 8 clocks (well, no clock activity for 8 clocks) it will think the transfer is done and latch the bits in the shift register. Someone familiar with the chip also mentioned that a gap in the spi transfer would cause undesirable behavior.

Thus, even if we have exclusive access to the chip an interrupt could occur that would delay the spi transfer and create a gap in it. I think the only way around this is to use the non-blocking SPI interface (assuming the mcu being used has one).

The current spi frequency is 8MHz, or 1MHz per byte. This means that the entire transfer will take 28 usecs (give or take). Seems ok to have interrupts disabled for that long.

vrahane: Yes, at that high frequency it definitely sounds fine.

PR #1108, opened 2018-05-18
Fix/nrf52 i2c hang

benmccrea:
The nRF52 TWI peripheral (I2C master) gets in a weird state when writing a reset byte out to the lp5523 (LED driver chip). Once in this state, the TWI will fail to output anything when I2C operations are attempted. A workaround is to reset the TWI by disabling and re-enabling it, after which it will function normally again. 
The issue may be related to an errata documented in nRF52832 Rev 2 Errata v1.1, "TWIS: Stopped event occurs twice if the STOP event is triggered during a transaction".
The attached image shows a Saleae capture of an offending frame that causes the issue. The first sign of trouble is a glitch on SDA at the end of the 2nd byte. Also note the double STOP condition at the end of the frame. 
glitch during write to lp5523 reset register

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
@@ -290,12 +290,14 @@ hal_i2c_master_write(uint8_t i2c_num, struct hal_i2c_master_data *pdata,
     int rc = -1;
     int i;
     uint32_t start;
+    int tries = 0;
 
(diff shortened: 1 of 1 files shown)


vrahane:
@benmccrea I think you did not mean to commit the first two changes. Please remove those and update the PR.

benmccrea:
@vrahane Fixed.

vrahane:
I tested it out on the with LP5523 chip as I was having issues with it on a custom board. It works well. Looks good. Thank you.

PR #1110, opened 2018-05-18
hal; add convenience routine to return reason for reset as string.

mkiiskila:


hw/hal/src/hal_common.c
```diff
@@ -57,3 +57,27 @@ hal_spi_data_mode_breakout(uint8_t data_mode, int *out_cpol, int *out_cpha)
...
     }
 }
+
+const char *
+hal_reset_cause_str(void)
+{
+    enum hal_reset_reason cause;
+
+    cause = hal_reset_cause();
+    switch (cause) {
```
utzig: could not this just be `switch (hal_reset_cause()) {` eliminating the var declaration?

mkiiskila: I often use local variables to make the code easier to read. It's up to compiler to optimize them out :)

PR #1113, opened 2018-05-18
apps/btshell: Add cmd_set_scan_opts command

MariuszSkamra:
This patch extends btshell with cmd_set_scan_opts command, that
limit the data printed out to the serial console during
scanning. It has been noticed that the large amount of data
sent on UART might cause the LE Extended Adverising packets
to be truncated.

The command params are:
* ignore_legacy - to not print legacy advertisements
* decode_limit - to limit the data to be printed out to serial
console.

### DIFF
--- a/apps/btshell/src/btshell.h
+++ b/apps/btshell/src/btshell.h
@@ -81,6 +81,11 @@ struct btshell_conn {
     struct btshell_l2cap_coc_list coc_list;
 };
 
+struct btshell_scan_opts {
+    uint16_t limit;
+    uint8_t ignore_legacy:1;
+};
+
 extern struct btshell_conn btshell_conns[MYNEWT_VAL(BLE_MAX_CONNECTIONS)];
 extern int btshell_num_conns;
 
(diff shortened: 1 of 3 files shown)


MariuszSkamra:
@sjanc @andrzej-kaczmarek

MariuszSkamra:
@rymanluk

apps/btshell/src/btshell.h
```diff
@@ -81,6 +81,11 @@ struct btshell_conn {
     struct btshell_l2cap_coc_list coc_list;
 };
 
+struct btshell_scan_cb_args {
```
rymanluk: I would call it btshell_scan_opts

apps/btshell/src/cmd.c
```diff
@@ -1084,6 +1084,9 @@ static struct kv_pair cmd_scan_ext_types[] = {
     { NULL }
 };
 
+
+static struct btshell_scan_cb_args scan_cb_args;
```
rymanluk: I would call it g_scan_opts;

apps/btshell/src/btshell.h
MariuszSkamra: done

apps/btshell/src/cmd.c
MariuszSkamra: done

MariuszSkamra:
@rymanluk As discussed offline, I pushed another version putting those parameters in separate command, so that those could be preserved between subsequent scan calls

rymanluk:
Thanks. Merged now!

PR #1121, opened 2018-05-23
Put extra jtag commands at end of command

InternetofAwesome:
This allows the previous arguments specified in the script to be overridden by the user. 
This has been an issue for us, since some of our devices require lower speeds to program/debug properly.

### DIFF
--- a/hw/scripts/jlink.sh
+++ b/hw/scripts/jlink.sh
@@ -53,7 +53,7 @@ jlink_load () {
     # downloading somewhere in the flash. So need to figure out how to tell it
     # not to do that, or report failure if gdb fails to write this file
     #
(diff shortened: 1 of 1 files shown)


mkiiskila:
Can't move $EXTRA_JTAG_CMD, it is currently being used for selecting particular jlink to debug with. And this has to be in the beginning.
Instead you'll have to add another environment variable. I suggest naming it $CFG_POST_INIT, like openocd.sh has.

InternetofAwesome:
@mkiiskila it looks like that is what `-device $JLINK_DEV` is doing.  This works fine for me, though I haven't tried it with more than one jlink attached.

PR #1127, opened 2018-05-23
util/cmd: Package for parsing <key>=<value> pairs.

ccollins476ad:
Utility package for parsing `<key>=<value>` pairs among shell command arguments.

### DIFF
--- a/util/cmdarg/include/cmdarg/cmdarg.h
+++ b/util/cmdarg/include/cmdarg/cmdarg.h
@@ -1,0 +1,696 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef H_UTIL_CMDARG_
+#define H_UTIL_CMDARG_
+
+#include <inttypes.h>
+#include <stdbool.h>
+struct mn_in6_addr;
+
+/**
+ * @brief Finds the argument with the specified key.
+ *
+ * Given a key, finds the first argument with the following contents:
+ *
+ *     <key>=[val]
+ *
+ * @param argc                  The argument count.
+ * @param argv                  The argument list.
+ * @param key                   The key to search for.
+ * @param out_val               On success, points to the start of the located
+ *                                  value string.
+ *
+ * @return                      Pointer to argument on success; NULL on
+ *                                  failure.
+ */
+char *cmdarg_find(int argc, char **argv, const char *key, char **out_val);
+
+/**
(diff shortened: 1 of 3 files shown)


util/cmd/include/cmd/cmd.h
```diff
@@ -0,0 +1,694 @@
...
+ * This function modifies the argument list such that this argument won't be
+ * found on a subsequent search.
+ *
+ * @param argc                  The argument count.
+ * @param argv                  The argument list.
+ * @param key                   The key to search for.
+ * @param out_val               On success, points to the start of the located
+ *                                  value string.
+ *
+ * @return                      Pointer to argument on success; -1 on failure.
```
kasjer: NULL on failure

util/cmd/include/cmd/cmd.h
```diff
@@ -0,0 +1,694 @@
...
+ *
+ *     <key>=[val]
+ *
+ * @param argc                  The argument count.
+ * @param argv                  The argument list.
+ * @param key                   The key to search for.
+ * @param out_val               On success, points to the start of the located
+ *                                  value string.
+ *
+ * @return                      Pointer to argument on success; -1 on failure.
```
kasjer: NULL on failure

util/cmd/include/cmd/cmd.h
```diff
@@ -0,0 +1,694 @@
...
+ *
+ * @param argc                  The argument count.
+ * @param argv                  The argument list.
+ * @param key                   The key to search for.
+ * @param out_s                 On success, points to the start of the located
+ *                                  value string.
+ *
+ * @return                      0 on success; system error code on failure.
+ */
+int cmd_arg_find_str(int argc, char **argv, const char *key, char **out_s);
```
kasjer: argv and out_s could be const char ** since strings are not changed just array, this applies to all the functions, but this could be a valid design decision to not have const

ccollins476ad: Thanks, Jerzy.  This is going to be really long-winded.  You raise a good point, so I feel I should give a proper response.

I think `const` gets weird when double indirection is involved (e.g., "pointer-to-pointer").  I agree that something like const would be helpful here, but I think the cure is worse than the disease in this case.

With a single layer of indirection, there is no problem.  When a function accepts a `const char *` parameter, the caller can pass in either `pointer-to-char` or `pointer-to-const-char`, and it all works, because the non-const argument is implicitly converted to const.

When there are two layers of indirection, it puts an additional burden on the caller.  `char **` and `const char **` are not compatible types, so if the function takes a `const char **` argument, the user must declare his data const.

In other words, if the function were changed to:
```
int cmd_arg_find_str(int argc, const char **argv, const char *key, const char **out_s);
```

Then the following would be a constraint violation:
```
char *argv[] = { "a=b", "c=d", 0 };
const char *val;
cmd_arg_find_str(2, myargs, "a", &val);
```

since `argv` is an array of pointer-to-char, not an array of pointer-to-const-char.  The shell package produces array-of-char (and rightfully so, since the caller may need to modify the arguments later), so it would be quite a nuisance for this package to expect array of pointer-to-const-char.

If `argv` must be non-const, then it seems reasonable to make the extracted value (`out_s`) also non-const.

Of course, if the user wants to pass in const data, then they face the opposite problem.  They'll need to use a cast and assume a sane C implementation.

I think this exposes two fundamental problems with C:
1. Inconsistent semantics of `const` (actual read-only data vs. can't-change through this particular pointer).
2. Limit of one return value per function.

Anyway, my feeling is that the current API is not ideal, but it is "less bad" than the alternative.

ccollins476ad:
Good catch re: "NULL on failure."  I have fixed the comments.

andrzej-kaczmarek:
Nice! I would just change package name as I feel `util/cmd` is too generic. You already have `cmd_arg` prefix for all APIs so what about `util/cmdarg`?

ccollins476ad:
Thanks, all.  I have renamed the package to `util/cmdarg`.

PR #1128, opened 2018-05-23
lwIP static routing for IPv6

ccollins476ad:
Add the ipv6_static_routing lwIP addon.

### DIFF
--- a/net/ip/lwip_contrib/addons/ipv6_static_routing/include/ipv6_static_routing/ip6_route_table.h
+++ b/net/ip/lwip_contrib/addons/ipv6_static_routing/include/ipv6_static_routing/ip6_route_table.h
@@ -1,0 +1,94 @@
+/**
+ * @file
+ *
+ * IPv6 static route table.
+ */
+
+/*
+ * Copyright (c) 2015 Nest Labs, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 3. The name of the author may not be used to endorse or promote products
+ *    derived from this software without specific prior written permission.
+ *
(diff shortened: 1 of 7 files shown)


net/ip/lwip_contrib/addons/ipv6_static_routing/src/ip6_route_table.c
```diff
@@ -0,0 +1,248 @@
...
+
+  /* Check if the table is full */
+  if (static_route_table[LWIP_IPV6_NUM_ROUTE_ENTRIES - 1].netif != NULL) {
+    retval = ERR_MEM;
+    goto exit;
+  }
+
+  /* Shift all entries down the table until slot is found */
+  for (i = LWIP_IPV6_NUM_ROUTE_ENTRIES - 1;
+       i > 0 && (ip6_prefix->prefix_len > static_route_table[i - 1].prefix.prefix_len); i--) {
```
mkiiskila: Could look at static_route_table[i - 1].netif being NULL before memcpy().

net/ip/lwip_contrib/addons/ipv6_static_routing/include/ipv6_static_routing/ip6_route_table.h
```diff
@@ -0,0 +1,94 @@
...
+  
+/**
+ * LWIP_IPV6_NUM_ROUTES: Number of IPV6 routes that can be kept in the static route table.
+ */
+#ifndef LWIP_IPV6_NUM_ROUTE_ENTRIES
+#define LWIP_IPV6_NUM_ROUTE_ENTRIES         (8)
+#endif
+
+#define IP6_MAX_PREFIX_LEN                  (128)
+#define IP6_PREFIX_ALLOWED_GRANULARITY      (8)
```
mkiiskila: This is somewhat questionable. Network prefixes don't follow byte boundaries.
See https://tools.ietf.org/html/rfc4632#section-5.1 and https://tools.ietf.org/html/rfc7608

ccollins476ad:
Thanks, Marko.  I agree with both comments.  Just to clarify- this static route implementation is part of the lwIP "contrib" repo; I didn't implement it.  I thought it would be better to use the existing library, despite its shortcomings, than use our own.

Now that I think about it, it would not be too much work to write our own implementation with the  use of lwIP's routing "hook".  When I was first looking at this, I was thinking we either had to use the existing implementation or fork lwIP.

Let's close this PR for now while I think about the options.

PR #1129, opened 2018-05-23
lwip_base: Make IPv6 neighbor discovery cfgable

ccollins476ad:
IPv6 neighbor discovery is enabled by default.  Define `LWIP_ND6` equal to 0 to disable it.

This is useful in case you don't want lwIP sending packets automatically.  If accepted, we will want to try to get this change upstreamed into lwIP.

### DIFF
--- a/net/ip/lwip_base/src/core/ipv6/nd6.c
+++ b/net/ip/lwip_base/src/core/ipv6/nd6.c
@@ -43,7 +43,7 @@
 
 #include "lwip/opt.h"
 
-#if LWIP_IPV6  /* don't build if not configured for use in lwipopts.h */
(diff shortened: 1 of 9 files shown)


net/ip/lwip_base/src/core/ipv6/nd6.c
```diff
@@ -703,6 +703,7 @@ nd6_input(struct pbuf *p, struct netif *inp)
 }
 
 
+#if LWIP_ND6_TMR
```
mkiiskila: I don't see this define set/declared anywhere. Is this check supposed to be here?

ccollins476ad:
You're right about the stray `#if`, good catch.  I have force pushed a fix.

PR #1135, opened 2018-05-24
log: Add better support for long log entries in nmgr

andrzej-kaczmarek:
This changes `msg` encoding to byte string so it can carry strings, CBOR and any binary data. Since indefinite-length strings are used each entry can carry virtually unlimited number of data (tinycbor was extended to support this).

Note that this also seems to work perfectly fine with existing `newtmgr` implementation for traditional (string) log entries of any length - only update to handle non-string data based on `type` field has to be added.


andrzej-kaczmarek:
@bgiori

sys/log/full/src/log_nmgr.c
```diff
@@ -159,21 +129,41 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
...
+            break;
+        }
+        g_err |= cbor_encode_byte_string(&str_encoder, data, rc);
+        off += rc;
+    }
+    g_err |= cbor_encoder_close_container(&rsp, &str_encoder);
 #else
     g_err |= cbor_encode_text_stringz(&rsp, "msg");
-    g_err |= cbor_encode_text_stringz(&rsp, data);
+    g_err |= cbor_encode_byte_string(&rsp, data, strlen((char *)data));
```
ccollins476ad: We shouldn't change the format of V2 log entries, should we?  I.e., shouldn't `msg` still be a text string here?

andrzej-kaczmarek: I just checked that `newtmgr` handles byte strings in v3 just fine so I just changed v2 to byte strings without much thinking about it. But I agree, let's keep v2 untouched - I'll change it back to text string.

vrahane: I agree with Chris, log v2 should be untouched and the byte string change along with the `type` field should be only in v3.

andrzej-kaczmarek:
@ccollins476ad @vrahane fixed

PR #1150, opened 2018-05-30
log/shell: Add support for logging cbor logs in shell

kolodgrz:
This patch adds possibility to pretty print cbor logs in log shell.

### DIFF
--- a/sys/log/full/src/log_shell.c
+++ b/sys/log/full/src/log_shell.c
@@ -35,6 +35,9 @@
 #endif
 #include "shell/shell.h"
 #include "console/console.h"
+#if MYNEWT_VAL(LOG_VERSION) > 2
+#include "tinycbor/cbor_mbuf_reader.h"
+#endif
 
 static int
 shell_log_dump_entry(struct log *log, struct log_offset *log_offset,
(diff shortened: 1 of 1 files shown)


kolodgrz:
@andrzej-kaczmarek @rymanluk @michal-narajowski

sys/log/full/src/log_shell.c
```diff
@@ -44,6 +47,12 @@ shell_log_dump_entry(struct log *log, struct log_offset *log_offset,
     char data[128];
```
vrahane: This kind of restricts the data length for the cbor entry. You might want to be able to read a longer indefinite length entry.

sterlinghughes:
can we update, so we can merge?

sys/log/full/src/log_shell.c
kolodgrz: @vrahane this is not a change in this PR. I can add separate patch to this PR with adoption of this log data size - @vrahane have any idea how big it should be ?

andrzej-kaczmarek:
fixed in https://github.com/apache/mynewt-core/pull/1548

PR #1151, opened 2018-05-30
lis2dw12: Fix self testing function error

kasjer:
Per datasheet configuration that was supposed to switch on
positive self test mode, did it, but also turned off register
auto increment functionality.
This effectively disabled possibility to read
normal values from accelerometer since those reading are
done in blocks and not single registers.
After the self test, original configuration was meant to be restored,
alas it used multi register write that was just turned off,
so the configuration was not restored.

Now IF_ADD_INC is set in per-datasheet configuration.

### DIFF
--- a/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
+++ b/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
@@ -1929,7 +1929,7 @@ int lis2dw12_run_self_test(struct sensor_itf *itf, int *result)
     int i;
     uint8_t prev_config[6];
     /* set config per datasheet, with positive self test mode enabled. */
-    uint8_t st_config[] = {0x44, 0x08, 0x40, 0x00, 0x00, 0x10};
+    uint8_t st_config[] = {0x44, 0x04, 0x40, 0x00, 0x00, 0x10};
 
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Shouldnt the fact that readlen and writelen exist mean that this should be enabled int the _config, and then just preserved in the self test function as it already preserves registers?

kasjer:
@jacobrosenthal Flag that allows to read and write few registers at once is the default one. It should be set before self test and after. Someone decided for self test to prepare all control registers and sat them at once. I'm not sure what really was meant to be set for testing. I guess that maybe performance settings and maybe something else that's why I did not want to read current settings add self test flags only.
But settings were wrong at least for auto increment flag, and just as new self-test-settings were written auto increment flag was canceled preventing reads of several registers and also preventing write the original settings with just one write.

jacobrosenthal:
I agree I dont like the batch register setting not using any defines. I
have lots of questions about that PR but so far Ive been ignored, shrug.

Oh I see: it is default 1
lis2dw12 8.5 page 44/63
IF_ADD_INC Register address automatically incremented during multiple byte
access with a serial interface (I2C or SPI). *Default value: 1 *(0:
disabled; 1: enabled)

PR #1152, opened 2018-05-31
native hal_flash; use mktemp() instead of tmpnam() to generate name for temp flash file.

mkiiskila:
New MacOS headers are complaining about tmpnam() being deprecated.

### DIFF
--- a/hw/mcu/native/src/hal_flash.c
+++ b/hw/mcu/native/src/hal_flash.c
@@ -93,11 +93,11 @@ static void
 flash_native_file_open(char *name)
 {
     int created = 0;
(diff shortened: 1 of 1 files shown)


sjanc:
From man 3 mktemp :-)

BUGS
       Never  use  mktemp().  Some implementations follow 4.3BSD and replace XXXXXX by the current process ID and a single letter, so that at most 26 different names can be returned.  Since on the one hand the names
       are easy to guess, and on the other hand there is a race between testing whether the name exists and opening the file, every use of mktemp() is a  security  risk.   The  race  is  avoided  by  mkstemp(3)  and
       mkdtemp(3).

PR #948, opened 2018-03-22
WIP - sensors: lis2ds12

jacobrosenthal:
Altered lis2dw12 for lis2ds12

notable changes:
* rate and power modes are intertwined, not exposing high freq and high res for now
* no filtering/noise/offsets
* removed freefall wasnt hooked up to interrupts anyway
* removed interrupt 2 wasnt really hooked up anyway
* removed intx_pin_cfg - dont think I like overriding interrupts without an explicit function call
* self test doesnt work? but I dont know that it did on other devices I guess, just ported it for this
* Havent tested fifo yet

PR #1154, opened 2018-06-02
Lis2dw12 fixes from lis2ds12 port

jacobrosenthal:
Needs testing, I dont have this device this is just my set of notes from porting. Happy to rebase or remove any commits.

Id especially like to see if self test is still passing for people as it had a few problems still. I dont personally feel like its properly implemented even yet.

AN5038 Section 8 says

> The procedure consists of:
> 1. enabling the accelerometer
> 2. averaging five samples before enabling the self-test
> 3. averaging five samples after enabling the self-test
> 4. computing the difference in absolute value for each axis and verifying that it falls within
> a given range. The min and max values are provided in the datasheet.

And Im not sure I consider the current implementation following that quote
This https://github.com/apache/mynewt-core/pull/948/commits/787e2a2a915b9f3e4da2fce6ea60989b5e6cc1ba is passing for me and my current recommendation, but I dont have your board.

### DIFF
--- a/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
+++ b/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
@@ -1295,7 +1295,7 @@ int lis2dw12_get_tap_cfg(struct sensor_itf *itf, struct lis2dw12_tap_settings *c
  * Set freefall detection configuration
  *
  * @param the sensor interface
- * @param freefall duration (5 bits LSB = 1/ODR)
+ * @param freefall duration (6 bits LSB = 1/ODR)
  * @param freefall threshold (3 bits)
  * @return 0 on success, non-zero on failure
  */
@@ -1359,7 +1359,7 @@ int lis2dw12_get_freefall(struct sensor_itf *itf, uint8_t *dur, uint8_t *ths)
  *
  * @param the sensor interface
  * @param FIFO mode to setup
- * @patam Threshold to set for FIFO
+ * @param Threshold to set for FIFO
  * @return 0 on success, non-zero on failure
  */
 int lis2dw12_set_fifo_cfg(struct sensor_itf *itf, enum lis2dw12_fifo_mode mode, uint8_t fifo_ths)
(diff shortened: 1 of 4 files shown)


vrahane:
I am a bit concerned about the self test, rest of the changes look good and thank you for adding the comments @jacobrosenthal. I haven't got to test your changes on real hardware yet. I will try to do that but is there a way you can try it yourself ?

jacobrosenthal:
Well existing is wrong. It's using the wrong full scale to calculate. I'd
prefer the self test in my repo which works on my boards and follows the
app notes. No I don't have this chip. Lots of other silent people cc'ed
here not speaking?

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1920,52 +1924,115 @@ int lis2dw12_get_int1_on_int2_map(struct sensor_itf *itf, uint8_t *val)
...
+        return rc;
+    }
+
+    rc = lis2dw12_write8(itf, LIS2DW12_REG_CTRL_REG1, config[0]);
     if (rc) {
         return rc;
     }
 
+    /* wait 200ms */
+    os_time_delay(OS_TICKS_PER_SEC / 200);
```
mkiiskila: Just peeking here: OS_TICKS_PER_SEC / 200 means 5 milliseconds, comment says 200msec?

jacobrosenthal: Good catch. Thanks fixed in eaad01c

vrahane:
@jacobrosenthal I tried out your changes on actual hardware, I am getting a return value of `-6`. Something is broken.

jacobrosenthal:
On which commit?

vrahane:
Whatever you have on your branch currently.

jacobrosenthal:
I'll but your board I guess let's rebase out self test and do that in a
second pr

jacobrosenthal:
All self test changes were squashed, removed and added to https://github.com/apache/mynewt-core/issues/1171

jacobrosenthal:
My LIS2DW12 board arrived. My freefall setter was incorrect and has been rebased out in c7f0dfa and also pr'ed to my branch.

One thing I noticed is that the taps are VERY sensitive. I would align with the double tap example as I have done in mine. Shrug...

the only other destructive edit left to argue over in here is the getter

The upstream code
```
cfg.freefall_dur = 6;
cfg.freefall_ths = 3; /* ~312mg */
```
and in register 0x36 (dec 54)
```
lis2dw12 peek 54
008298 reg 0x36(54) = 0x33
```
IE dur 6, ths 3, So it has a correct setter. but, lis2dw12_get_freefall reads
```
001055 dur:0x06, ths:0x00
```
incorrect getter.

And you wont be surprised that in this pr, lis2dw12_get_freefall
```
002681 dur:0x06, ths:0x03
```

PR #1158, opened 2018-06-04
sys/shell; CLI command to reset device.

mkiiskila:
Optionally set nvreg to enter boot_serial in bootloader.


ccollins476ad:
Looks good.  It seems I screwed up the operator precedence in newt (parens shouldn't be necessary in the restrictions string, but are).

PR #1159, opened 2018-06-05
syscfg to configure NFC pins to be GPIO

InternetofAwesome:
Break out a syscfg value to configure the NFC pins as GPIOs


vrahane:
@wes3 spoke to Sam, it does make sense. I do agree as well. Let's merge it ?

wes3:
Yep; looks good!

PR #1163, opened 2018-06-06
kernel/os - Add os_mbuf_widen() function

ccollins476ad:
Increases the length of an mbuf chain by inserting a gap at the specified
offset.  The contents of the gap are indeterminate.  If the mbuf chain
contains a packet header, its total length is increased accordingly.

This function never frees the provided mbuf chain.

@param om                    The mbuf chain to widen.
@param off                   The offset at which to insert the gap.
@param len                   The size of the gap to insert.

@return                      0 on success; SYS_[...] error code on failure.

### DIFF
--- a/kernel/os/src/os_mbuf.c
+++ b/kernel/os/src/os_mbuf.c
@@ -1059,3 +1059,89 @@ os_mbuf_trim_front(struct os_mbuf *om)
     return om;
 }
 
+int
+os_mbuf_widen(struct os_mbuf *om, uint16_t off, uint16_t len)
+{
+    struct os_mbuf *first_new;
+    struct os_mbuf *edge_om;
+    struct os_mbuf *prev;
+    struct os_mbuf *cur;
+    uint16_t rem_len;
+    uint16_t sub_off;
+    int rc;
+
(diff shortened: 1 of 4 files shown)


kernel/os/src/os_mbuf.c
```diff
@@ -1059,3 +1059,83 @@ os_mbuf_trim_front(struct os_mbuf *om)
...
+        if (prev != NULL) {
+            SLIST_NEXT(prev, om_next) = cur;
+        }
+        prev = cur;
+    }
+
+    /* Move the misplaced data from the edge mbuf over to the right side of the
+     * gap.
+     */
+    os_mbuf_append(prev, edge_om->om_data + sub_off, edge_om->om_len - sub_off);
```
mkiiskila: I think there's a chance of failure here; os_mbuf_append() might fail if it turns out that number of bytes copied requires a new mbuf to be allocated. Of course, this requires a bit unusual config, i.e. one where edge_om mbuf came from pool of bigger mbufs than where prev came from. So not sure if this is something worth worrying.

ccollins476ad: Good catch.  The return code should be checked.

PR #1165, opened 2018-06-06
tinycbor: Fix cmp return value from readers

andrzej-kaczmarek:
cmp functions are used as IteratorFunc and expected behaviour is to
return non-zero value on success so the opposite of what memcmp-like
function do.


andrzej-kaczmarek:
for reference, original implementation in tinycbor: https://github.com/intel/tinycbor/blob/master/src/cborparser.c#L1066

PR #1166, opened 2018-06-06
syscfg to enable nRF52 pin reset

lance-proxy:
A syscfg value to enable pin reset on the `nrf52xxx` instead of using it as a normal GPIO.


mlaz:
LGTM

PR #1169, opened 2018-06-06
nrf52_pwm: catch NRFX_PWM_EVT_STOPPED events

mlaz:
This PR makes pwm_stop not crash when called from seq_end_handler.

### DIFF
--- a/hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
+++ b/hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
@@ -123,9 +123,10 @@ static void handler_0(nrfx_pwm_evt_type_t event_type)
         break;
 
     case NRFX_PWM_EVT_FINISHED :
-        instances[0].playing = false;
-        nrfx_pwm_uninit(&instances[0].drv_instance);
         instances[0].seq_end_handler(instances[0].seq_end_data);
+        break;
+
+    case NRFX_PWM_EVT_STOPPED :
         break;
 
     default:
@@ -145,9 +146,10 @@ static void handler_1(nrfx_pwm_evt_type_t event_type)
         break;
 
     case NRFX_PWM_EVT_FINISHED :
-        instances[1].playing = false;
-        nrfx_pwm_uninit(&instances[1].drv_instance);
         instances[1].seq_end_handler(instances[1].seq_end_data);
+        break;
+
+    case NRFX_PWM_EVT_STOPPED :
         break;
 
     default:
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
This does indeed solve my crash. 

Should nrf52_pwm_close just call nrf52_pwm_disable instead of calling 
```
    if (!instances[inst_id].playing) {
        nrfx_pwm_uninit(&instances[inst_id].drv_instance);
    }
```

Will nrfx_pwm_stop ever be called if you nrf52_pwm_close without nrf52_pwm_disable first?

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -336,7 +342,7 @@ nrf52_pwm_close(struct os_dev *odev)
     }
 
     if (!instances[inst_id].playing) {
-        nrfx_pwm_uninit(&instances[inst_id].drv_instance);
+        nrf52_pwm_disable(dev);
```
jacobrosenthal: Actually, you're duplicating all these checks now, they mostly exist inside nrf52_pwm_disable. Take another look at this

jacobrosenthal: why does nrf52_pwm_disable not check for playing before uninit,but nrf52_pwm_close does?

jacobrosenthal: Again, whether it should call nrf52_pwm_disable from nrf52_pwm_close I dont have an opinion, it just seems like were sharing code, but doing things differently in both... which seems odd

mlaz: > Actually, you're duplicating all these checks now, they mostly exist inside nrf52_pwm_disable. Take another look at this

There is only a single repeated check, which is `instances[inst_id].in_use`, we can have a static fuction (i.e. disable_instance) for stopping this being called from both `nrf52_pwm_disable` and `nrf52_pwm_close`. 

>why does nrf52_pwm_disable not check for playing before uninit,but nrf52_pwm_close does?
 
On close we don't need to stop and uninit the instance if it is not playing which is what we may expect any caller for disable to do.  However nrfx_pwm_uninit may not need the instanced to be initialized nrfx_pwm_stop does. Adding this check to nrf52_pwm_disable is a good idea, but we'd still need it on close.  

>Again, whether it should call nrf52_pwm_disable from nrf52_pwm_close I dont have an opinion, it just seems like were sharing code, but doing things differently in both... which seems odd

The device should be uninitialized on close. Do we need to stop it before? No. Do I think we should try to do both uninitializations the same way?(ultimately for the sake consistency/readability) Yes.

jacobrosenthal:
Ok so make them consistent and don't have close call disable then?

mlaz:
Using nrfx_stop is pointless, since we're uninitializing whenever we want to stop and nrfx_pwm_uninit also stops the device. This way both close and stop call uninit, which will never crash. Both functions are simpler while consistent and doing what's expected.

jacobrosenthal:
You would know best having worked with this api all this time, just
pointing out what looks odd to me.

PR #1174, opened 2018-06-07
log/modlog - Module-mapped logging

ccollins476ad:
The modlog facility allows log entries to be written to numeric modules identifiers.  In typical usage, startup code maps 8-bit module IDs to one or more log objects, and other parts of the application log events by writing entries to module IDs.  This usage differs from the underlying `sys/log` facility, which requires clients to provide a log object to write to rather than just a module identifier.

Benefits provided by the modlog package are:
* Improved modularity - configuration and usage are distinct.
* Ability to write to several logs with a single function call.  If one module ID is mapped to several logs, a write to that ID causes all mapped logs to be written.
* Default mappings.  Writes to unmapped module IDs get written to an optional set of default logs.
* Minimum log level per mapping.  Writes specifying a log level less than the module's minimum level are discarded.
* Decreased RAM usage; multiple unrelated packages can share the same `struct log` object rather than defining their own (`sizeof (struct log)` is 20).

Costs of using modlog rather than the bare `sys/log` facility are:
* Increased RAM usage (`MODLOG_MAX_MAPPINGS` * 12).
* Increased CPU usage - each log write requires a lookup in the set of configured modlog mappings.

Example usage:

```
    /* Map module 100 to the console log. */
    rc = modlog_register(100, log_console_get(), LOG_LEVEL_DEBUG, NULL);
    assert(rc == 0);

    /* Log a message to module 100. */
    MODLOG_INFO(100, "hello from module %d\n", 100);
```

### DIFF
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -83,6 +83,10 @@ log_init(void)
     rc = log_nmgr_register_group();
     SYSINIT_PANIC_ASSERT(rc == 0);
 #endif
+
+#if MYNEWT_VAL(LOG_CONSOLE)
+    log_console_init();
+#endif
 }
 
 struct log *
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -34,7 +34,6 @@ struct log_info g_log_info;
 
 static STAILQ_HEAD(, log) g_log_list = STAILQ_HEAD_INITIALIZER(g_log_list);
 static const char *g_log_module_list[ MYNEWT_VAL(LOG_MAX_USER_MODULES) ];
-static uint8_t log_inited;
 static uint8_t log_written;
 
 #if MYNEWT_VAL(LOG_CLI)
@@ -63,12 +62,9 @@ log_init(void)
 
     (void)rc;
 
-    if (log_inited) {
-        return;
-    }
-    log_inited = 1;
     log_written = 0;
 
+    STAILQ_INIT(&g_log_list);
     g_log_info.li_version = MYNEWT_VAL(LOG_VERSION);
     g_log_info.li_next_index = 0;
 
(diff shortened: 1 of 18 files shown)


ccollins476ad:
Note: I thought the registration API was unwieldy, so I simplified it.  The current state of this PR reflects the latest changes.

sys/log/full/src/log.c
```diff
@@ -360,11 +360,15 @@ log_append_typed(struct log *log, uint8_t module, uint8_t level, uint8_t etype,
...
 int
-log_append_mbuf_typed(struct log *log, uint8_t module, uint8_t level,
-                      uint8_t etype, struct os_mbuf *om)
+log_append_mbuf_typed_no_free(struct log *log, uint8_t module, uint8_t level,
+                              uint8_t etype, struct os_mbuf **om_ptr)
 {
+    struct os_mbuf *om;
     int rc;
 
+    /* Remove a loyer of indirection for convenience. */
```
vrahane: not really that important but `loyer` should be `layer` :-)

sys/log/full/src/log.c
```diff
@@ -83,6 +79,10 @@ log_init(void)
     rc = log_nmgr_register_group();
     SYSINIT_PANIC_ASSERT(rc == 0);
 #endif
+
+#if MYNEWT_VAL(LOG_CONSOLE)
+    log_console_init();
```
vrahane: The other types of logs needed to be defined in the app before being initialized. Only `LOG_CONSOLE` would become different now after the changes introduced by this PR. I guess if MODLOG wants to use the console to log stuff, it should be the one initializing a log and anyone using the MODLOG actually gets the console log as an effect of using the MODLOG. Just a thought though, not important enough functionally.

PR #1175, opened 2018-06-08
pwm_nrf52: fix crash on close not checking playing status

jacobrosenthal:
If you disable pwm and then close the device, it attempts to nrfx_pwm_uninit the same device twice and assert crashes.

Now we check if a device is 'playing' before we nrfx_pwm_uninit

### DIFF
--- a/hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
+++ b/hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
@@ -339,7 +339,10 @@ nrf52_pwm_close(struct os_dev *odev)
         return (EINVAL);
     }
 
-    nrfx_pwm_uninit(&instances[inst_id].drv_instance);
+    if(instances[inst_id].playing) {
+        nrfx_pwm_uninit(&instances[inst_id].drv_instance);
+    }
+
     cleanup_instance(inst_id);
 
     if (os_started()) {
@@ -545,8 +548,13 @@ nrf52_pwm_disable(struct pwm_dev *dev)
         return (-EINVAL);
     }
 
+    if (!instances[inst_id].playing) {
+        return (-EINVAL);
+    }
+
     nrfx_pwm_uninit(&instances[inst_id].drv_instance);
     instances[inst_id].playing = false;
+
     return (0);
 }
 
--- a/hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
+++ b/hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
@@ -280,7 +280,7 @@ cleanup_instance(int inst_id)
  *             it can be a nrf_drv_pwm_config_t, to override the default
  *             configuration.
  *
- * @return 0 on success, non-zero on failure.
+ * @return 0 on success, non-zero error code on failure.
  */
 static int
 nrf52_pwm_open(struct os_dev *odev, uint32_t wait, void *arg)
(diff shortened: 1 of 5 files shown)


hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -339,7 +339,10 @@ nrf52_pwm_close(struct os_dev *odev)
         return (EINVAL);
     }
 
-    nrfx_pwm_uninit(&instances[inst_id].drv_instance);
+    if(instances[inst_id].playing) {
```
kasjer: missing space after if

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -545,8 +548,13 @@ nrf52_pwm_disable(struct pwm_dev *dev)
         return (-EINVAL);
     }
 
+    if (!instances[inst_id].playing) {
```
kasjer: missing space after if

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -545,8 +548,13 @@ nrf52_pwm_disable(struct pwm_dev *dev)
...
+        return (-EINVAL);
```
kasjer: EINVAL usage is inconsistent in this file some functions return -EINVAL some EINVAL. I don't know which should be used but having both version looks kind of strange.

jacobrosenthal:
Im with you on these, but this is all in the original file. We need a linter set up so we dont have to go through this nit pick stuff every time.

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
mlaz: It is stated on the comments that this function returns a negative. However we can go with the non zero(i.e. use non-negative error codes) 
in this case becausee we got no other data being returned.

jacobrosenthal:
@mlaz is this a workable fix? Do you want to take a look and see if any more checks are missing?

mlaz:
@jacobrosenthal on it

jacobrosenthal:
ping @mlaz lets land this before 1.4.1

mlampert:
I defer to @utzig

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -437,7 +440,7 @@ nrf52_pwm_configure_channel(struct pwm_dev *dev,
     struct nrf52_pwm_dev_global *instance = &instances[inst_id];
     nrfx_pwm_config_t *config = &instance->config;
 
-    if (!instance->in_use) {
+    if (!instance->) {
```
utzig: funny syntax!

jacobrosenthal: heh

jacobrosenthal: @mlaz were you trying to alter the check here, or did the in_use just get lost.

mlaz: It did get lost, apparently before I commit, and after I compiled.

mlaz: https://github.com/jacobrosenthal/mynewt-core/pull/2 @jacobrosenthal

jacobrosenthal:
Rebased out that error. Going to go ahead and put this out of its misery finally.

PR #1183, opened 2018-06-11
lis2dw12: fix self test to match an5038 rev2

### LINKED ISSUE #1171: lis2dw12 self test
It utilizes fs 2g even though the device is in fs 4g
https://github.com/apache/mynewt-core/blob/26607457d6eb8db4918412834576784131f56794/hw/drivers/sensors/lis2dw12/src/lis2dw12.c#L1952

I also dont think it properly implements the app note. AN5038 Section 8 says the procedure consists of:
- enabling the accelerometer
- averaging five samples before enabling the self-test
- averaging five samples after enabling the self-test
- computing the difference in absolute value for each axis and verifying that it falls within
- a given range. The min and max values are provided in the datasheet.

The current one seems to add a positive reading to a negative one, and not do a ground truth reading at all.

jacobrosenthal:
This works with no changes from the previous patch on my freshly shipped hardware. Resubmitting. Closes #1171

### DIFF
--- a/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
+++ b/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
@@ -38,6 +38,8 @@
  */
 #define LIS2DW12_MAX_INT_WAIT (4 * OS_TICKS_PER_SEC)
 
+#define LIS2DW12_ST_NUM_READINGS 5
+
 const struct lis2dw12_notif_cfg dflt_notif_cfg[] = {
     { SENSOR_EVENT_TYPE_SINGLE_TAP,   0, LIS2DW12_INT1_CFG_SINGLE_TAP  },
     { SENSOR_EVENT_TYPE_DOUBLE_TAP,   0, LIS2DW12_INT1_CFG_DOUBLE_TAP  },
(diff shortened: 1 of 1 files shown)


hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1920,52 +1922,115 @@ int lis2dw12_get_int1_on_int2_map(struct sensor_itf *itf, uint8_t *val)
...
 
+    //discard
+    //TODO poll DRDY in STATUS (27h) instead?
+    rc = lis2dw12_get_data(itf, fs, &(data[0]), &(data[1]), &(data[2]));
+    if (rc) {
+        return rc;
+    }
+
+    /* take no st offset reading */
+    for(i=0; i<LIS2DW12_ST_NUM_READINGS; i++) {
```
mlaz: Why not `i = 0` instead of `i=0` ? You use both ways throughout the code.

mlaz: Although there is no specific rule for this one.

jacobrosenthal: Eh. I dont have a preference. Though the original file has i=0 in the lis2dw12_run_self_test function.

Thanks for the look, though Im mostly looking for a review from @vrahane as he previously had issues with this

sterlinghughes:
Can we merge this?

jacobrosenthal:
I would hate to merge on top of  @matthewwarnes  and @InternetofAwesome without some buy in from them

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1920,52 +1922,115 @@ int lis2dw12_get_int1_on_int2_map(struct sensor_itf *itf, uint8_t *val)
...
+
+    /* wait 200ms */
+    os_time_delay(OS_TICKS_PER_SEC / 5 + 1);
+
+    rc = lis2dw12_get_fs(itf, &fs);
     if (rc) {
         return rc;
     }
 
+    //discard
```
kasjer: It is not clearly stated in application note, but test procedure will only work if FIFO is not enabled during self test described there.
To make this function work correctly I would suggest disabling FIFO during test, or make sure that FIFO does not have anything before reading self test data. Second method would take longer so disabling FIFO is batter IMHO.

hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -1920,52 +1922,115 @@ int lis2dw12_get_int1_on_int2_map(struct sensor_itf *itf, uint8_t *val)
...
+        rc = lis2dw12_get_data(itf, fs, &(data[0]), &(data[1]), &(data[2]));
+        if (rc) {
+            return rc;
+        }
+        scratch[0] += data[0];
+        scratch[1] += data[1];
+        scratch[2] += data[2];
+    }
+
+    //average
```
kasjer: C++ style comment introduced here, C-style was used in this file so why not stick to it.

kasjer:
@jacobrosenthal I will provide separate PR to cover style and FIFO changes

PR #1186, opened 2018-06-12
kernel/os: Fix APIs to use os_time_t for ticks arguments

andrzej-kaczmarek:
This patch changes callout, mutex, sem and time APIs to use os_time_t
consistently in all APIs that take or return ticks as arguments/return
value. This way we define explicitly which parameters expect values in
ticks rather than other time units.

This should not affect any existing code since most args were uint32_t
anyway which is the same as os_time_t.

### DIFF
--- a/kernel/os/include/os/os_callout.h
+++ b/kernel/os/include/os/os_callout.h
@@ -44,7 +44,7 @@ struct os_callout {
     /** Pointer to the event queue to post the event to */
     struct os_eventq *c_evq;
     /** Number of ticks in the future to expire the callout */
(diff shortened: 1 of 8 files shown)


andrzej-kaczmarek:
Note that argument in `os_time_delay` was `int32_t` and defined as _<=0_ for on timeout_. I changed this to `os_time_t` anyway so it's now _=0 for no timeout_. I am not sure how many application this may affect - we can put this in release notes for 1.5 or use `os_stime_t` instead, but I think this would leave things inconsistent so `os_time_t` seems like a better solution here.

PR #1189, opened 2018-06-12
time/timesched: Add package to create timers using clock time

andrzej-kaczmarek:
This patch adds new package which implements timer operating on clock time. These timers do not replace e.g. `hal_timer` since they do not guarantee high accuracy, but are just a convenience wrapper over `os_callout` APIs to fire at specified clock time (`os_timeval`) instead of at ticks interval.

### DIFF
--- a/kernel/os/include/os/os_time.h
+++ b/kernel/os/include/os/os_time.h
@@ -110,6 +110,19 @@ void os_time_delay(os_time_t osticks);
 #define OS_TIME_TICK_GT(__t1, __t2) ((os_stime_t) ((__t1) - (__t2)) > 0)
 #define OS_TIME_TICK_GEQ(__t1, __t2) ((os_stime_t) ((__t1) - (__t2)) >= 0)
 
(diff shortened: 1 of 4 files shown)


andrzej-kaczmarek:
@ccollins476ad @wes3 @mkiiskila @sjanc @rymanluk

ccollins476ad:
I like the idea, and it looks good to me.  However, I do think this should go in a separate package.  I think we will want to attach several "hooks" to a time change event, so rescheduling the list of timers would just be another one.  Of course, that isn't implemented yet, and this is :).

sterlinghughes:
I agree this should be a separate package, and I like the idea.

andrzej-kaczmarek:
moved to separate package
since there are no hook in OS (yet) to notify about time change I simply reduced max callout time to 1min so basically every 1min schedule is recalculated. I am not really sure if 1min is the proper value here, perhaps we can make it even longer at e.g. 10mins since I can imagine time on device does not change too often and not by too much (more like adjustments) - like by syncing to some external time source.

PR #1196, opened 2018-06-14
sys/log/full: Add "body" functions to API

### LINKED ISSUE #1173: Awkward log API requires padded buffer
The `log_append()` function writes the contents of a flat buffer to the specified log.  However, the provided buffer must contain `LOG_ENTRY_HDR_SIZE` bytes of padding at the beginning, since the function populates this space with the log entry header.  So, to log the string "abc", you must pass the following arguments to this function:
* `data`: `<padding>abc`   (total of `LOG_ENTRY_HDR_SIZE`+3 bytes.)
* `len`: 3

Similarly, `log_append_mbuf()` requires that the mbuf contain `LOG_ENTRY_HDR_SIZE` of padding at the start.  The provided mbuf must have these characteristics:
* `om_data`: `<padding>abc`
* `om_len`: `LOG_ENTRY_HDR_SIZE`+3

I understand why the API is like this - some underlying storage media, namely FCB, requires writes to be done in a single chunk.  However, I think this API is too awkward and error-prone to endure.  Perhaps we need to modify the fcb code to allow split writes.

ccollins476ad:
This fixes #1173.

This PR adds five functions to the log API:
* `log_append_body`
* `log_append_mbuf_body`
* `log_read_hdr`
* `log_read_body`
* `log_read_mbuf_body`
* `log_walk_body`

The "append" functions are similar to `log_append_typed` and `log_append_mbuf_typed`.  The difference is that the provided flat buf or mbuf does not need to contain padding at the beginning for holding a log entry header.

`log_read_hdr` just reads the specified entry header from the log.

The "read body" functions are very similar to the existing "read" function.  The only difference is that the specified offset starts from the beginning of the message body, not the from the beginning of the entry.  That is, the user doesn't need to add `LOG_ENTRY_HDR_SIZE` to the offset that gets passed in.

`log_walk_body` is similar to `log_walk`, except it passes the message header and body separately to the callback.  This allows the caller to read the log message without knowing how the header is encoded.

This PR does not affect backwards compatibility.  In opinion, the new functions are less error-prone than the old ones, so it might be a good idea to deprecate some of the old functions.

There is one related change that I think should be made:
* Make the `append_mbuf` functions compatible with flash-alignments greater than 1.


kernel/os/src/os_mbuf.c
```diff
@@ -370,6 +370,20 @@ _os_mbuf_copypkthdr(struct os_mbuf *new_buf, struct os_mbuf *old_buf)
     new_buf->om_data = new_buf->om_databuf + old_buf->om_pkthdr_len;
 }
 
+uint16_t
+os_mbuf_len(const struct os_mbuf *om)
+{
+    uint16_t len;
+
+    len = 0;
+    while (om != NULL) {
```
mkiiskila: Did you consider SLIST_FOREACH()? This is fine too.

sys/log/full/src/log.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one
+* Licensed to the Apache Software Foundation (ASF) under one
```
mkiiskila: Accidental whitespace deletion? Might be worth adding back, not sure how rat tool checks these...

sys/log/full/src/log.c
```diff
@@ -63,6 +63,7 @@ log_init(void)
     (void)rc;
 
     log_written = 0;
+    STAILQ_INIT(&g_log_list);
```
mkiiskila: Already done two lines below.

sys/log/full/src/log_fcb.c
```diff
@@ -97,46 +100,185 @@ log_fcb_append(struct log *log, void *buf, int len)
 }
 
 static int
-log_fcb_append_mbuf(struct log *log, struct os_mbuf *om)
+log_fcb_first_chunk_sz(uint8_t align)
 {
+    uint8_t mod;
+
+    mod = sizeof (struct log_entry_hdr) % align;
```
mkiiskila: % operator is very costly. I think we can assume that the alignment restriction is a power of 2, and then do the math assuming that.

sys/log/full/src/log_fcb.c
```diff
@@ -97,46 +100,185 @@ log_fcb_append(struct log *log, void *buf, int len)
...
+    int len;
+    int rc;
+
+    fcb_log = (struct fcb_log *)log->l_arg;
+    fcb = &fcb_log->fl_fcb;
+
+    /* This function expects to be able to write each mbuf without any
+     * buffering.
+     */
+    if (fcb->f_align != 1) {
```
mkiiskila: Maybe a /* XXXX */ comment here to call attention in case someone ends up missing this, and wondering where to start add support for it :)

sys/log/full/src/log.c
ccollins476ad: Huh... not sure how that happened.  Definitely not intentional :).

kernel/os/src/os_mbuf.c
ccollins476ad: I don't think `SLIST_FOREACH` is applicable here since that macro requires an slist head.

ccollins476ad:
Thanks for the review, Marko.  Here are the changes I made:

* Restore missing whitespace in Apache license.
* Remove redundant invocation of `SLIST_INIT()`
* Remove use of `%` operator.
* Add `/* XXX: Fix this; buffer mbuf writes to satisfy flash alignment. */` to draw attention to incomplete implementation of `log_fcb_append_mbuf()`.

Regarding the removal of `%` - I made the following changes:
*`fcb_init()`:*
```
    /* Require alignment to be a power of two.  Some code depends on this
     * assumption.
     */
    assert((fcb->f_align & (fcb->f_align - 1)) == 0);
```

*`log_fcb.c`:*
```
    /* Assume power-of-two alignment for faster modulo calculation. */
    assert((align & (align - 1)) == 0);

    mod = sizeof (struct log_entry_hdr) & (align - 1);
    if (mod == 0) {
        return 0;
    }
```

PR #1197, opened 2018-06-14
sys/console: Fix RTT input polling

andrzej-kaczmarek:
Timeout counter was missing "static" keyword thus we always used 50ms
as polling interval for every single character in buffer. This commit
fixes the problem and also refactors polling code a bit to make RTT
input more responsive.

We will now read all available data on RTT in single run and then start
with polling interval of 10ms, increasing it by 10ms on every idle run
until it reaches max interval set by syscfg value (default 250ms). This
means there may be tiny delay when reading 1st character after longer
idle period, but then complete input can be processed in no time.

### DIFF
--- a/sys/console/full/src/rtt_console.c
+++ b/sys/console/full/src/rtt_console.c
@@ -50,27 +50,30 @@ console_out(int character)
 }
 
 #if MYNEWT_VAL(CONSOLE_INPUT)
+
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
No code comments, but OH MAN. thats much better. Great job

PR #1211, opened 2018-06-21
lis2dw12: Orient notification change notification

vrahane:
- Add support for X_, Y_ and Z_ orientation change notification
- Add some more notification stats
- Add LIS2DW12_NOTIF_STATS for making notification stats conditional
- Add lis2dw12_find_notif_cfg_by_event() and use event to find notif_cfg

### DIFF
--- a/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
+++ b/hw/drivers/sensors/lis2dw12/src/lis2dw12.c
@@ -39,12 +39,66 @@
 #define LIS2DW12_MAX_INT_WAIT (4 * OS_TICKS_PER_SEC)
 
 const struct lis2dw12_notif_cfg dflt_notif_cfg[] = {
-    { SENSOR_EVENT_TYPE_SINGLE_TAP,   0, LIS2DW12_INT1_CFG_SINGLE_TAP  },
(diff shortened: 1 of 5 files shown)


hw/drivers/sensors/lis2dw12/src/lis2dw12.c
```diff
@@ -2695,18 +2798,107 @@ lis2dw12_sensor_handle_interrupt(struct sensor *sensor)
...
+         * YL, XH and XL
+         */
+
+        sensor_mgr_put_notify_evt(&lis2dw12->pdd.notify_ctx,
+                                  SENSOR_EVENT_TYPE_ORIENT_X_CHANGE);
+
+        lis2dw12_inc_notif_stats(SENSOR_EVENT_TYPE_ORIENT_X_CHANGE);
+    }
+
+    notif_cfg = lis2dw12_find_notif_cfg_by_event(SENSOR_EVENT_TYPE_ORIENT_Y_CHANGE,
```
ccollins476ad: I don't know if this is something worth worrying about or not, but I wonder if there is an efficiency concern here.  Each call to `lis2dw12_find_notif_cfg_by_event` results in a linear traversal of the notification array.  This function (`lis2dw12_find_notif_cfg_by_event`) calls the find function quite a few times (nine by my count).  Like I said, I don't know if this matters in practice.  If it does, it seems it would be more efficient to do a single walk of the notification list, processing each element as they are encountered.  I think it depends on how frequently this function gets called.

PR #1212, opened 2018-06-22
mgmt/imgr: Add support to resume image upload

andrzej-kaczmarek:
1st chunk of data can include "datahash" field which is simply SHA256
computer over all data to be uploaded. This hash is stored in upload
state so in case transfer was interrupted we can detect this on next
attempt and simply resume transfer at previous location by replying
with latest offset from upload state.

"datahash" field is not mandatory so if it is missing we'll just start
new upload as usual.

### DIFF
--- a/mgmt/imgmgr/src/imgmgr.c
+++ b/mgmt/imgmgr/src/imgmgr.c
@@ -407,11 +407,13 @@ static int
 imgr_upload(struct mgmt_cbuf *cb)
 {
     uint8_t img_data[MYNEWT_VAL(IMGMGR_MAX_CHUNK_SIZE)];
+    uint8_t data_hash[IMGMGR_DATA_HASH_LEN];
     long long unsigned int off = UINT_MAX;
     long long unsigned int size = UINT_MAX;
     size_t data_len = 0;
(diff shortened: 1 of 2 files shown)


andrzej-kaczmarek:
corresponding change in `newtmgr`: https://github.com/apache/mynewt-newtmgr/pull/89

ccollins476ad:
Looks good to me.

I wonder if we should have done image management differently.  Rather than automatically erase the image slot, the user would have to manually erase it.  Then, any attempt to upload to a partially-filled slot would could be assumed to be a resumption of an incomplete upload.  Then the  client wouldn't need to decide whether to include a hash or not; it would just try to upload the image and react to the response.

Anyway, I think it is too late to change any of this.  Just musing :).

andrzej-kaczmarek:
Actually it would be possible to do this without `datahash` because upload state keeps track of both offset and size, so we could use this information to resume. However, the reason why I decided to add extra field is to make sure that client tries to upload exactly the same image so it's safe to resume transfer - adding hash seems like an easy and clean solution for this which is also fully backwards compatible.

PR #1213, opened 2018-06-22
NRF52/ADC: Allow concurrent usage

kasjer:
Even though NRF52 have multiple chanels that could be used for
reading current implementation allowd only one task to open
device.
This patch allows to use single channel readings from channels
by different task by using mutex that was previously used
for locking device in open.
    
Mutex is no longer acquired in open.
Multiple opens are allowed.
Mutex is used to synchronize concurrent readings.

### DIFF
--- a/hw/drivers/adc/adc_nrf52/src/adc_nrf52.c
+++ b/hw/drivers/adc/adc_nrf52/src/adc_nrf52.c
@@ -91,7 +91,8 @@ static int
 nrf52_adc_open(struct os_dev *odev, uint32_t wait, void *arg)
 {
     struct adc_dev *dev;
-    int rc;
+    int rc = 0;
+    int unlock = 0;
 
     dev = (struct adc_dev *) odev;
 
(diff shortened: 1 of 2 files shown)


hw/drivers/adc/adc_nrf52/src/adc_nrf52.c
```diff
@@ -135,19 +135,31 @@ static int
...
 
-    nrfx_saadc_uninit();
+    if (os_started()) {
+        rc = os_mutex_pend(&dev->ad_lock, OS_TIMEOUT_NEVER);
+        if (rc != OS_OK) {
+            goto err;
+        }
+        unlock = 1;
+    }
+    if (dev->ad_dev.od_open_ref == 1) {
```
wes3: Where does od_open_ref get decremented?

rymanluk: It is in os_dev.c . Actually I'm not sure why os_dev.c always calls od_open() and od_close() no matter on reference counter.

Maybe we could change a bit logic here and  make  os_dev.c to call od_open() only once when reference count is 0 and calls od_close() only once when reference count drops back to 0. In this way, driver does not have to  check od_open_ref.

wes3: Yeah, I think I had made a similar comment on the dev list. That changes seems reasonable to me.

kasjer:
@wes3 @rymanluk new proposition that should not be prone to race condition. ref count is kept in adc_dev and checked only when lock is taken. Otherwise same logic as in previous patch.

wes3:
Sounds good to me.

PR #1218, opened 2018-06-25
mgmt/imgmgr: Update data hash handling (for image resume)

andrzej-kaczmarek:
This changes SHA256 to SHA1 to be used for data check before allowing
image upload to be resumed.

### DIFF
--- a/mgmt/imgmgr/src/imgmgr.c
+++ b/mgmt/imgmgr/src/imgmgr.c
@@ -125,6 +125,7 @@ static struct {
     uint32_t size;
 
     /** Hash of image data; used for resumption of a partial upload. */
+    uint8_t data_sha_len;
     uint8_t data_sha[IMGMGR_DATA_SHA_LEN];
 } imgr_state;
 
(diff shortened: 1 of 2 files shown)


andrzej-kaczmarek:
@bgiori fyi

utzig:
Could as well use `crc16` and use only 2 bytes if security is not a concern! :-P

andrzej-kaczmarek:
Sure, except I think hash function fits better here :-)

I mean, I do not think we need anything fancy here since this is to check that client tries to resume upload of the same image. I just have no idea what is reasonable to use here so went with the most obvious (I think) SHA hash. But I am happy to change it to anything most people will agree on.

simonratner:
You can also truncate SHA256, it is perfectly valid to truncate it down to as few as 128 bits which still gets you better collision protection in fewer bits than SHA1 (and smaller code size, if you already use SHA256 elsewhere).

sjanc:
Agree with Simon,  we should not promote known broken hash functions :-)

andrzej-kaczmarek:
yeah, Simon's proposal absolutely makes sense to me. the only question is if we should just have this fixed on 128 bits or allow both full or truncated sha - just to avoid some problems in future in case we need to scale it again ;-)

andrzej-kaczmarek:
I updated this PR to:
- have `sha` field name for hash instead of `datahash` (this will be used by `newtmgr` as previous name caused issues with mcuboot)
- accept full SHA256 or value trimmed to 1/2 or 1/4 of original size

I hope 2nd change makes sense :-)

andrzej-kaczmarek:
@ccollins476ad I can change it to accept anything that has at least let's say 4 or 8 characters - it works for me perfecly. I am just not sure if it makes sense in terms of how SHA works... but actually, do we really care? this is just informational anyway :-)

utzig:
> I am just not sure if it makes sense in terms of how SHA works

If using only for integrity checking, should be perfectly OK to use 2, 3, 4, whatever number of bytes: https://en.wikipedia.org/wiki/Avalanche_effect

simonratner:
Just remember about collision frequency, that's still a function of the number of bits. With two bytes, there's a 1 in 256 chance of a collision completely at random. Obviously there is still the signature check, but this may just be frequent enough to cause annoying errors in the wild.

ccollins476ad:
@simonratner True (but shouldn't that be 1 in 65536?).  But I'm just wondering if it is the firmware's place to enforce a minimum hash length.  The hash field is optional as it is; why not let the client decide how much to include?

andrzej-kaczmarek:
Good point, @ccollins476ad. Also if needed this length will be only adjusted in single client (`newtmgr` or whatever other app is used) without worrying about compatibility.

PR #1219, opened 2018-06-25
Battery Shell

kasjer:
This adds shell command line that allows to read battery properties provided by battery drivers.

This PR also adds driver for reading battery voltage using ADC.

### DIFF
--- a/hw/drivers/bq27z561/src/bq27z561.c
+++ b/hw/drivers/bq27z561/src/bq27z561.c
@@ -663,6 +663,11 @@ bq27z561_battery_property_get(struct battery_driver *driver,
         rc = -1;
         assert(0);
     }
+    if (rc == 0) {
+        property->bp_valid = 1;
+    } else {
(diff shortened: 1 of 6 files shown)


hw/drivers/bq27z561/src/bq27z561.c
```diff
@@ -663,6 +663,7 @@ bq27z561_battery_property_get(struct battery_driver *driver,
         rc = -1;
         assert(0);
     }
+    property->bp_valid = rc == 0;
```
wes3: I personally do not like "cryptic" lines of code like this. Makes people think too hard. There are other ways to write that line of code that would not force anyone to think about it :-)

sterlinghughes: i agree.

kasjer: I will change accordingly it and merge if there no more doubts.

PR #1223, opened 2018-06-27
util/rwlock: Readers–writer lock

ccollins476ad:
I needed this to implement a different feature, but I am submitting it as a separate PR.

This is a lock for multiple readers, single writer.

This lock is write-preferring.  That is:
* If there is no active writer and no pending writers, read-acquisitions do not block.
* If there is an active writer or a pending writer, read-acquisitions block.
* When the last active reader or the active writer releases the lock, it is acquired by a pending writer if there is one.  If there are no pending writers, the lock is acquired by all pending readers.

### DIFF
--- a/util/rwlock/include/rwlock/rwlock.h
+++ b/util/rwlock/include/rwlock/rwlock.h
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef H_RWLOCK_
+#define H_RWLOCK_
+
+#include "os/mynewt.h"
+
+/**
+ * @brief Readers–writer lock - lock for multiple readers, single writer.
+ *
+ * This lock is write-preferring.  That is:
+ *     o If there is no active writer and no pending writers, read-acquisitions
+ *       do not block.
+ *     o If there is an active writer or a pending writer, read-acquisitions
+ *       block.
+ *     o When the last active reader or the active writer releases the lock, it
+ *       is acquired by a pending writer if there is one.  If there are no
+ *       pending writers, the lock is acquired by all pending readers.
+ *
+ * All struct fields should be considered private.
+ */
+struct rwlock {
+    /** Protects access to rwlock's internal state. */
+    struct os_mutex mtx;
+
+    /** Blocks and wakes up pending readers. */
+    struct os_sem rsem;
+
+    /** Blocks and wakes up pending writers. */
+    struct os_sem wsem;
+
+    /** The number of active readers. */
+    uint8_t num_readers;
+
+    /** Whether there is an active writer. */
+    bool active_writer;
+
+    /** The number of blocked readers. */
+    uint8_t pending_readers;
+
+    /** The number of blocked writers. */
+    uint8_t pending_writers;
+
+    /**
+     * The number of ownership transfers currently in progress.  No new
+     * acquisitions are allowed until all handoffs are complete.
+     */
+    uint8_t handoffs;
+};
+
+/**
+ * @brief Acquires the lock for use by a reader.
+ *
+ * @param lock                  The lock to acquire.
+ */
+void rwlock_acquire_read(struct rwlock *lock);
+
+/**
+ * Releases the lock from a reader.
+ *
+ * @param lock                  The lock to release.
+ */
+void rwlock_release_read(struct rwlock *lock);
+
+/**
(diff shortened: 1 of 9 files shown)


util/rwlock/include/rwlock/rwlock.h
```diff
@@ -0,0 +1,90 @@
+#ifndef H_RWLOCK_
```
mkiiskila: Missing license blurbs.

util/rwlock/src/rwlock.c
```diff
@@ -0,0 +1,163 @@
...
+
+    os_mutex_pend(&lock->mtx, OS_TIMEOUT_NEVER);
+    if (rc != 0 && rc != OS_NOT_STARTED) {
+        /* Semaphore timed out; undo increment and return error. */
+        lock->pending_readers--;
+        rc = SYS_ETIMEOUT;
+    } else {
+        /* Success; record reader ownership. */
+        lock->num_readers++;
+        rc = 0;
```
mkiiskila: This should go back to beginning to see if lock could be acquired. Writer might've grabbed the lock before reader task locked the mutex.

util/rwlock/src/rwlock.c
```diff
@@ -0,0 +1,163 @@
...
+
+    os_mutex_pend(&lock->mtx, OS_TIMEOUT_NEVER);
+    if (rc != 0 && rc != OS_NOT_STARTED) {
+        /* Semaphore timed out; undo increment and return error. */
+        lock->pending_writers--;
+        rc = SYS_ETIMEOUT;
+    } else {
+        /* Success; record writer ownership. */
+        lock->active_writer = true;
+        rc = 0;
```
mkiiskila: This also should re-evaluate lock state once sem_pend() has returned, after lock->mtx was acquired.

ccollins476ad:
Thanks, Marko.  You are definitely right about the the bug in `rwlock_acquire_read()`.  As you say, this needs to handle the case where a writer swoops in before the function is able to lock the mutex and increment `lock->num_readers`.  So that will probably save about a week of painful debugging effort :).

I am not sure about the other comment, though (`rwlock_acquire_write()`).  Since this function has already incremented `lock->pending_writers()`, I believe that prevents anyone else (reader or writer) from stealing the lock while the mutex is unlocked.  Does this sound right to you, or do you still think there is a bug?

ccollins476ad:
I have (hopefully) fixed the race condition bugs as follows: Add a new member to `struct rwlock` called `handoffs`.  This value represents the number of ownership transfers currently in progress. 
No new acquisitions are allowed until all handoffs are complete.

Each time a token is added to a semaphore, the `handoffs` variable is incremented (while the mutex is still locked).  Each time a user is unblocked, it decrements `handoffs` after locking the mutex.

(also, I added the apache license to each of the new files.)

ccollins476ad:
I also removed the `timeout` parameter from the "acquire" functions.  Now, these functions never time out.  I couldn't think of a safe way to implement a timeout; it seems there will always be a race condition here.  If a task tries to add a token to a semaphore, it is impossible to determine if the other task is still blocking on the semaphore or if it has just timed out.

util/rwlock/test/src/testcases/rwlock_test_case_basic.c
```diff
@@ -0,0 +1,147 @@
+#include "rwlock/rwlock.h"
```
mkiiskila: Missing license.

util/rwlock/test/src/rwlock_test.h
```diff
@@ -0,0 +1,10 @@
+#ifndef H_RWLOCK_TEST_H
```
mkiiskila: Missing license.

util/rwlock/src/rwlock.c
```diff
@@ -0,0 +1,213 @@
...
+
+    os_mutex_release(&lock->mtx);
+}
+
+int
+rwlock_init(struct rwlock *lock)
+{
+    int rc;
+
+    *lock = (struct rwlock) { 0 };
```
mkiiskila: LLVM does not like this kind of initializations.

ccollins476ad: Then LLVM is broken :).  It works in my clang (`Apple LLVM version 7.3.0 (clang-703.0.31)`).  Out of curiosity, which version of clang are you using?

mkiiskila: Just upgraded, and ended up with Apple LLVM version 9.1.0 (clang-902.0.39.2). Upgrade broke my gcc-5 installation, so I tried running with clang for a bit. A lot of issues when building unit tests. Some of them due to this kind of initialization; so it is fresh in my mind :)
In the end, I ended up falling back to gcc :)

ccollins476ad: Wow, that would be quite a regression for Apple to introduce.  I have an equally recent version of clang on my linux machine (5.0.2) which doesn't have this bug.

I wonder if it is correct to change code to accommodate a broken version of a compiler...

ccollins476ad:
Thanks, Marko. Just a question:
> Only thing I'm a bit uneasy about is that the task waiting on this lock should not end up in run queue unless it is woken by a call to rwlock_unblock().
> Also there's the problem of priority inversion, but I'm not too worried about that, given the nature of the lock.

Aren't these two the same issue? Or is there another reason, aside from priority inversion, why it is bad that a blocked task is not on the run queue.

I do see what you mean about priority inversion.  I had not considered that.  It would be possible to implement priority inheritance here, but I am not sure I trust myself to do it :).

mkiiskila:
We can fix priority inversion later, if it becomes an issue. In practice I don't feel priority inversion is a hindrance ;) But to add it, I would probably replace current read/write semaphores with linked list similar to what we have with other locks, maybe ordered by priority? Then checking whether priority boost is needed should be cheap.

I believe nowhere we have calls to os_sched_wakeup() unless task is truly sleeping, based on my latest read of the code where it gets called. So I shouldn't worry about that either :)

PR #1226, opened 2018-06-27
adc_stm32f4: Fix build when not all ADC are present

kasjer:
Some STM32F4xx MCUs do not have 3 ADCs.
This change allows to build ADC code for those MCUs that
do not have some ADCs.


kasjer:
@utzig, @andrzej-kaczmarek can you take a look. I'm not able to add reviewers yet

PR #1230, opened 2018-06-28
adp5061: Decode register fields in dump

kasjer:
dump command now has optional parameter 'decode' which allows
to decode register fields and makes it easier for humans.

help command is explicitly added


hw/drivers/chg_ctrl/adp5061/src/adp5061_shell.c
```diff
@@ -27,16 +27,546 @@
...
 
 static struct adp5061_dev *adp5061_dev;
 
+/*
+ * Struct for field name description.
+ * Used for argument validation and filed decoding
+ */
+struct reg_field
+{
+    /* Name of filed */
```
wes3: Was this supposed to be field instead of filed?

PR #1234, opened 2018-06-30
mgmt/imgmgr: Allow app to reject image upload reqs

ccollins476ad:
Allow a callback to configured in the imgmgr module.  If one is configured, the callback is called immediately before the request is processed.  The callback has the following properties:

```
The callback's return code determines whether the upload request is
accepted or rejected.  If the callback returns 0, processing of the
upload request proceeds.  If the callback returns nonzero, the request
is rejected with a response containing an `rc` value equal to the return
code.

@param offset                The offset specified by the incoming
                                request.
@param size                  The total size of the image being uploaded.
@param arg                   Optional argument specified when the
                                callback was configured.

@return                      0 if the upload request should be accepted;
                             nonzero to reject the request with the
                                specified status.

typedef int imgr_upload_fn(uint32_t offset, uint32_t size, void *arg);
```

This might be a somewhat painful review; the "clean up" commit introduces a lot of changes.  I explain my rationale for these changes in the commit message.  I also thought some of image upload code could use a bit of cleanup, so I probably changed more than was strictly necessary.  Ignoring the added callback feature, the code should be the same as before functionally, with the following exceptions:

1. Response status codes changed.  There were a few conditions for which the code responded with a `MGMT_ERR_EINVAL` status code, but not because of an invalid parameter.  I changed these all to `MGMT_ERR_EUNKNOWN`.

2. The flash area does not stay open for the duration of the upgrade.  Now the flash area gets opened each time an image request is received.  I think this is the correct behavior.  In the old code, there was a possibility that the flash area would remain open indefinitely if an upgrade never completes (though I don't know if there are any consequences to this).

### DIFF
--- a/mgmt/imgmgr/include/imgmgr/imgmgr.h
+++ b/mgmt/imgmgr/include/imgmgr/imgmgr.h
@@ -44,6 +44,26 @@ extern "C" {
 #define IMGMGR_STATE_F_ACTIVE           0x04
 #define IMGMGR_STATE_F_PERMANENT        0x08
 
+/** @typedef imgr_upload_fn
+ * @brief Application callback that is executed when an image upload request is
+ * received.
+ *
+ * The callback's return code determines whether the upload request is accepted
+ * or rejected.  If the callback returns 0, processing of the upload request
+ * proceeds.  If the callback returns nonzero, the request is rejected with a
+ * response containing an `rc` value equal to the return code.
+ *
+ * @param offset                The offset specified by the incoming request.
+ * @param size                  The total size of the image being uploaded.
+ * @param arg                   Optional argument specified when the callback
+ *                                  was configured.
+ *
+ * @return                      0 if the upload request should be accepted;
+ *                              nonzero to reject the request with the
+ *                                  specified status.
+ */
+typedef int imgr_upload_fn(uint32_t offset, uint32_t size, void *arg);
+
 extern int boot_current_slot;
 
 void imgmgr_module_init(void);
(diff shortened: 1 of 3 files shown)


ccollins476ad:
Thanks, Andrzej.  Regarding `ULLONG_MAX` vs `UINT32_MAX`- I think it is correct, but please let me know if I got it wrong.  The only two values that are compared with `ULLONG_MAX` are: 1) `req->off` and `req->size`.  Both of these members are of type `unsigned long long`, and are both initialized to `ULLONG_MAX` in `imgr_upload()` (line 587).

When I look at it now, I think this is really confusing.  `struct imgr_upload_action`'s `size` member has type `uint32_t`, *not* `unsigned long long`.  This inconsistency seems weird to me, so now I think I should change this member's type to `unsigned long long`.  The reason `imgr_upload_req` uses `unsigned long long` is just that that is what the CBOR library uses for decoding, and I didn't want to waste stack space by allocating separate temporaries to decode into.

Regarding the callback on each received request- I was thinking the application may want to reject an image upgrade due to low battery.  If the battery level became critically low in the middle of the upgrade, the application would want to abort the procedure at that point.

ccollins476ad:
As I mentioned, I have changed `struct imgr_upload_action`'s `size` member from `uint32_t` to `unsigned long long`.

I noticed this line in `imgr_upload_inspect()`:
```
        action->size = req->size;
```

Which may indeed have caused problems when the two `size` members had different types.  Is this the issue you were commenting on, Andrzej?  My latest change should fix this.

PR #1237, opened 2018-07-02
Initial TSL2591 driver

microbuilder:
This is an initial attempt at a driver for the [TSL2591 light sensor](https://www.adafruit.com/product/1980) since the currently supported TSL2561 is EOL.

This driver includes basic shell config support, and read operations, but does not currently include interrupt support for threshold triggering. This can be added in a future PR once the concerns below are addressed.

Before this can be merged, it would be beneficial to discuss the timing constraints around this sensor (and others like the TSL2561). It has a variable integration time from 100..600ms, and any reads before this initial delay or less than this between conversions will results in faulty or stale data being returned.

An experimental flag called `TSL2591_ITIME_DELAY` has been added to the package syscfg.yml file, but this is really just an attempt to get a discussion started on how to handle situations where the sensor may be polled at a rate inconsistent with the HW limitations, and how to best handle these timing disparities.

This timing delay can be dynamic in certain situations, such as implementing auto-ranging support in the driver to accommodate a wide range of lighting conditions. As the light level lowers, you would increase the integration time and thus the overall sensitivity, and vice versa in high light situations. These changes may be transparent to the user at the application level, meaning they need some way to determine the minimum delay between **valid** sensor reads, which isn't part of the current sensor API design but should perhaps be considered as an addition via a new field required by all sensor drivers (`.min_delay`, etc.)?

Any thoughts (and testing) is welcome!

### DIFF
--- a/hw/drivers/sensors/tsl2591/include/tsl2591/tsl2591.h
+++ b/hw/drivers/sensors/tsl2591/include/tsl2591/tsl2591.h
@@ -1,0 +1,193 @@
+/**************************************************************************/
+/*!
+    @file     tsl2591.h
+    @author   Kevin Townsend
+
+    @section LICENSE
+
+    Software License Agreement (BSD License)
+
+    Copyright (c) 2018, Kevin Townsend
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+    1. Redistributions of source code must retain the above copyright
+    notice, this list of conditions and the following disclaimer.
+    2. Redistributions in binary form must reproduce the above copyright
+    notice, this list of conditions and the following disclaimer in the
+    documentation and/or other materials provided with the distribution.
+    3. Neither the name of the copyright holders nor the
+    names of its contributors may be used to endorse or promote products
+    derived from this software without specific prior written permission.
+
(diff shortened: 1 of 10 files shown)


microbuilder:
@vrahane Is this something you would be able to review? If you don't have access to a TSL2591 let me know and I'll figure out how to get you one. The TSL2561 is EOL and won't be easy to source in the near future, so it makes sense to get an updated driver into the codebase.

microbuilder:
I was really hoping to get some feedback on this, particularly opinions on the delay mechanism [here](https://github.com/apache/mynewt-core/pull/1237/files#diff-27697b7935dcde394e64d942f4716f0fR357), but with the great northern hemisphere weather I get that everyone would probably rather be in the sun with a cold drink. 

If HW is an issue to testing this, let me know and I'll figure out a solution to get something to the appropriate person(s).  I just wanted to get this merged so that I can put some effort into a simulator for this sensor, just to have a proof of concept of how that might work. (See also #1251).

microbuilder:
Closing this due to API changes. A new PR will be generate shortly.

jacobrosenthal:
@microbuilder Noticed something on this, it was missing the Interface locking that was introduced recently, https://github.com/apache/mynewt-core/pull/1250
Something to add when you reopen

PR #1244, opened 2018-07-05
Battery property write

kasjer:
Battery drivers can have writable properties (i.e. threshold values).

This extends bq27z561 driver with access to threshold values and also exposes those values to battery API.

Write command for battery properties was also added so property writes can be tested.

### DIFF
--- a/hw/drivers/bq27z561/include/bq27z561/bq27z561.h
+++ b/hw/drivers/bq27z561/include/bq27z561/bq27z561.h
@@ -69,6 +69,14 @@
 #define BQ27Z561_REG_DCAP       (0x3C)
 #define BQ27Z561_REG_MFRG_ACC   (0x3E)
 #define BQ27Z561_REG_CHKSUM     (0x60)
+#define BQ27Z561_REG_VOLT_HI_SET_TH (0x62)
+#define BQ27Z561_REG_VOLT_HI_CLR_TH (0x64)
+#define BQ27Z561_REG_VOLT_LO_SET_TH (0x66)
+#define BQ27Z561_REG_VOLT_LO_CLR_TH (0x68)
+#define BQ27Z561_REG_TEMP_HI_SET_TH (0x6A)
+#define BQ27Z561_REG_TEMP_HI_CLR_TH (0x6B)
+#define BQ27Z561_REG_TEMP_LO_SET_TH (0x6C)
+#define BQ27Z561_REG_TEMP_LO_CLR_TH (0x6D)
 
 
 /* Alt Manufacturer Command List */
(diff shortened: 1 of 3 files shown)


hw/drivers/bq27z561/include/bq27z561/bq27z561.h
```diff
@@ -232,6 +294,60 @@ int bq27z561_get_temp(struct bq27z561 *dev, float *temp_c);
...
+ *
+ * When interrupt is active and voltage goes above this threshold interrupt
+ * is cleared.
+ *
+ * @param dev pointer to device
+ * @param voltage voltage, in mV
+ *
+ * @return int 0: success, -1 error
+ */
+int bq27z561_set_voltage_lo_set_threshold(struct bq27z561 *dev,
```
wes3: This appears to be the same prototype as the one above. I think it is misnamed.

hw/drivers/bq27z561/include/bq27z561/bq27z561.h
```diff
@@ -232,6 +294,60 @@ int bq27z561_get_temp(struct bq27z561 *dev, float *temp_c);
...
+ * When interrupt is active and voltage goes below this threshold interrupt
+ * is cleared.
+ *
+ * @param dev pointer to device
+ * @param voltage voltage, in mV
+ *
+ * @return int 0: success, -1 error
+ */
+int bq27z561_set_voltage_hi_set_threshold(struct bq27z561 *dev,
+        uint16_t voltage);
```
wes3: Same comment as above. This is the same prototype as the one above. Is it mis-named?

hw/drivers/bq27z561/src/bq27z561.c
```diff
@@ -440,12 +490,181 @@ bq27z561_get_temp(struct bq27z561 *dev, float *temp_c)
...
+}
+
+int
+bq27z561_set_voltage_hi_clr_threshold(struct bq27z561 *dev, uint16_t voltage)
+{
+    int rc;
+
+    rc = bq27z561_wr_std_reg_word(dev, BQ27Z561_REG_VOLT_HI_CLR_TH, voltage);
+
     return rc;
```
wes3: I am not saying that this is right or wrong, but I wonder what the compiler is going to do with the above functions. I do not trust compilers so part of the reason that these "helper wrappers" were not added in the original driver is that they add a bit of code just to hide the underlying function calls. All of these could be made macros as well. The use of static inline (and even always inline) does not appear to be a guarantee these things are inlined. Not a big deal but still.

hw/drivers/bq27z561/include/bq27z561/bq27z561.h
kasjer: fiexed

hw/drivers/bq27z561/include/bq27z561/bq27z561.h
kasjer: fixed

PR #1245, opened 2018-07-05
sys/config; add a custom filtering option when compressing FCB config.

mkiiskila:
This allows user to pick and choose what persisted data they want
to keep.


mkiiskila:
void *arg is not a bad idea here, I'll add that one.

PR #1253, opened 2018-07-09
hw/mcu/nordic: Disconnect input buffer when gpio made an output.

wes3:
This change disconnects the input buffer on a gpio when the gpio
is configured to be an output. This should save current.


wes3:
Not sure about other MCU's as I have not looked at how their gpio are handled but it is a good question.

PR #1255, opened 2018-07-10
hw/sensor: Report read errors

ccollins476ad:
Allow "sensor error listeners" to be registered with a sensor.  When a read from a sensor fails, all error listeners that match the read attempt are notified of the error.

### DIFF
--- a/hw/sensor/include/sensor/sensor.h
+++ b/hw/sensor/include/sensor/sensor.h
@@ -246,6 +246,20 @@ typedef int
 (*sensor_notifier_func_t)(struct sensor *, void *, sensor_event_type_t);
 
 /**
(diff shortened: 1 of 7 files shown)


microbuilder:
@ccollins476ad This is a good idea, thanks for adding it. Any change of perhaps adding a test case, just as built in documentation of how it works? The test cases are usually the first place I look personally when working with new packages or features, and since it's part of the build process they generally are kept up to date with API changes. (Obviously just a convenience request, feel free to ignore it 😉)

ccollins476ad:
@microbuilder, that is a good idea.  I have added a very basic unit test for reporting of sensor read errors.

PR #1259, opened 2018-07-11
TSL2591 driver w/sim stubs

microbuilder:
This adds a driver for the [TSL2591](https://cdn-shop.adafruit.com/datasheets/TSL25911_Datasheet_EN_v1.pdf) light sensor from Taos (now AMS), which should be used on new designs in place of the older (**EOL**) TSL2561.

The driver includes support for the sensor API, shell support, statistics and logging, as well as basic simulation stubs for the new native I2C simulation support added in #1251, and has been tested using the [Adafruit TSL2591 breakout](https://www.adafruit.com/product/1980).

- The **HW driver** includes support for basic reads, writes and device configuration, but doesn't yet include interrupt threshold support.
- The **sim driver** is currently a work in progress, and includes the basic plumbing needed to test the new native hal_i2c code, but needs further additions to process the I2C packets, and simulate the various registers available on the device.


microbuilder:
I'm not sure at this point what the best approach is to model the I2C slave/peripheral behaviour in the simulator. I'm thinking through it now, but presumably some sort of finite state machine would be useful to handle the various read/write combinations where you first write the address of the register(s) you want, then read an appropriate number of bytes/registers back, and handle error conditions appropriately.

Should this be implemented entirely in `tsl2591_sim.c`, or should a helper package be put together to make this easier to implement in other I2C simulators in the future (I have plans to write a couple more myself)???

Any suggestions on a sustainable, elegant solution to this are welcome, since it makes sense for this first simulated driver to be well written since it will serve as an example for future development.

microbuilder:
I updated this to reflect the recent modlog and lock changes in mynewt-core.

microbuilder:
I think this can be merged in at this point. Full sensor simulation isn't implemented yet, but I think this can be the object of another PR since there is the need for further discussion around how I2C bus simulation should be implemented, and if helper package(s) are required, etc.

microbuilder:
I think #1273 is causing the new merge conflict here, but it's also become hard to read the PR changes via 'Files Changed' due to keeping the PR up to date with recent changes. Should I submit this as a new PR based on the latest code, removing all of the commit duplicates  shown here from update attempts? Otherwise, the review will need to happen by looking at the individual commits from me, which is kind of an annoying way of doing things.

If anyone has a TSL2591 to test this against, any feedback would be appreciate. I'm also happy to send some TSL2591s to Runtime as a freebie order from Adafruit, and just need the order number to clear it with shipping, but send me a private message to follow up on that.

jacobrosenthal:
You just need a rebase on master. The file thats going to whine is this one 
hw/sensor/creator/pkg.yml

hw/drivers/sensors/tsl2591/src/tsl2591.c
```diff
@@ -0,0 +1,706 @@
...
+        goto err;
+    }
+
+    rc = tsl2591_write8(itf, TSL2591_COMMAND_BIT | TSL2591_REGISTER_CONTROL,
+                        int_time | gain);
+    if (rc) {
+        goto err;
+    }
+
+    #if MYNEWT_VAL(TSL2591_ITIME_DELAY)
```
mkiiskila: We don't indent preprocessor #if lines.

hw/drivers/sensors/tsl2591/src/tsl2591.c
```diff
@@ -0,0 +1,706 @@
...
+
+int
+tsl2591_get_data(struct sensor_itf *itf, uint16_t *broadband, uint16_t *ir)
+{
+    int rc;
+    uint8_t itime;
+    uint16_t igain;
+    uint16_t maxval;
+
+    #if !(MYNEWT_VAL(TSL2591_AUTO_GAIN))
```
mkiiskila: Same here; #if should not be indented, and it should not cause code inside the preprocessor conditional to be indented further.

hw/drivers/sensors/tsl2591/src/tsl2591_sim.c
```diff
@@ -0,0 +1,134 @@
...
+#include <string.h>
+#include <errno.h>
+#include "os/mynewt.h"
+#include "tsl2591/tsl2591.h"
+#include "tsl2591_priv.h"
+#include "mcu/mcu_sim_i2c.h"
+
+/* Default register values for this sensor */
+static uint8_t g_tsl2591_sim_regs[TSL2591_REGISTER_CHAN1_HIGH] = {
+  /* 0x00 = TSL2591_REGISTER_ENABLE */
```
mkiiskila: 4 bytes of indentation rather than 2.

hw/drivers/sensors/tsl2591/src/tsl2591.c
```diff
@@ -0,0 +1,706 @@
...
+
+    /* Determine the ideal gain setting */
+    igain = maxval / (*broadband > *ir ? *broadband : *ir);
+
+    /* Find the closest gain <= igain */
+    if (igain < 25) {
+        /* Gain 1x, return the current sample */
+        return 0;
+    }
+    else if (igain < 428) {
```
mkiiskila: See https://github.com/apache/mynewt-core/blob/master/CODING_STANDARDS.md regarding whitespaces and braces.

hw/drivers/sensors/tsl2591/src/tsl2591.c
```diff
@@ -0,0 +1,706 @@
...
+    else {
+        /* Set gain to ~9876x */
+        rc = tsl2591_set_gain(itf, TSL2591_LIGHT_GAIN_MAX);
+        if (rc) {
+            goto err;
+        }
+    }
+
+    /* Get a new data sample */
+    /* Note: double-read required to empty cached values with prev gain */
```
mkiiskila: Good to have this comment explaining the double read.

microbuilder:
@mkiiskila Thanks for the review, stylistic changes pushed out.

PR #1262, opened 2018-07-12
Add lock to charge control and adp5061 and bq27z561 fix

vrahane:
- Fix bq27z561 and add lock for adp5061


hw/drivers/bq27z561/include/bq27z561/bq27z561.h
```diff
@@ -429,6 +429,8 @@ bq27z561_err_t bq27x561_rd_flash(struct bq27z561 *dev, uint16_t addr,
 bq27z561_err_t bq27x561_rd_alt_mfg_cmd(struct bq27z561 *dev, uint16_t cmd,
                                        uint8_t *val, int val_len);
 
+bq27z561_err_t bq27x561_wr_alt_mfg_cmd(struct bq27z561 *dev, uint16_t cmd,
+                                       uint8_t *buf, int len);
```
wes3: This is an interesting addition in this commit. Was it intended?

vrahane: Yup, it was intended because I think it was left out by mistake.

PR #1266, opened 2018-07-12
battery/adc: Add activation pin support

kasjer:
Battery voltage reading often uses resistor divider that is
actiaved with GPIO pin for energy saving (divider is not powered
when not needed).
This change allows to specify GPIO pin that must be activated
before ADC can be used.
GPIO level also can be specified allowing divider activation with 0 and 1.

### DIFF
--- a/hw/battery/src/battery_adc.c
+++ b/hw/battery/src/battery_adc.c
@@ -23,6 +23,7 @@
 #include <battery/battery_drv.h>
 #include <battery/battery_adc.h>
 #include <adc/adc.h>
+#include <hal/hal_gpio.h>
 
 /* Battery manager interface functions */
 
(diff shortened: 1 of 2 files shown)


hw/battery/src/battery_adc.c
```diff
@@ -104,6 +122,18 @@ battery_adc_open(struct os_dev *dev, uint32_t timeout, void *arg)
         /* Setup channel configuration to use for battery voltage */
         adc_chan_config(bat_adc->adc_dev, bat_adc->cfg.channel,
                 bat_adc->cfg.adc_channel_cfg);
+
+        /* Additional GPIO needed before measurement ? */
+        if (bat_adc->cfg.activation_pin != -1) {
+            /* Should it be activate once or before each measurement */
+            if (bat_adc->cfg.activation_pin_on_open) {
```
wes3: I have to say I do not understand why you need the activation_pin_on_open flag. Why have the extra logic?

PR #1268, opened 2018-07-12
util/debounce: Debouncing / hysteresis utility pkg

ccollins476ad:
I needed this utility for something, and I thought it was generally useful enough to be added to core.  Perhaps this would be better placed in a different repo.

I struggled a lot deciding on a width for the integers in `struct debouncer`.  `uint32_t` requires a bit more memory than I was comfortable with.  I thought about using macros to define several instances of the API (`debouncer8`, `debouncer16`, `debouncer32`), but that was making the code too complicated to be readable.  I also considered a single syscfg setting which determined the integer width for all debouncers, but I didn't think that was very useful, and it still made the interface difficult to read due  to the need for macros.  In the end, I just settled on what I think works in most cases: `uint16_t`.


util/debounce/include/debounce/debounce.h
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * @brief Indicates which of the two states the provided debouncer is in.
+ *
+ * @param debouncer             The debouncer to query.
+ *
+ * @return                      true if the debouncer is in the high state;
+ *                              false if the debouncer is in the low state.
+ */
+static inline bool
+debouncer_high(const debouncer_t *debouncer)
```
andrzej-kaczmarek: I would make it return `int` instead with `0` for low, `1` for high - just as logic state.
and change name to `debouncer_get/state_get/read` or something like this.

ccollins476ad: Thanks, Andrzej.  I have changed `debouncer_high` to `debouncer_state`, and made it return 0 or 1 rather than a bool.

PR #1275, opened 2018-07-16
tinycbor/oic; byte string encoding with data as iovec.

mkiiskila:
This avoid memory copies when encoding byte strings where data is not contiguous in memory.


mkiiskila:
Re: having doxygen stuff in .h vs in .c. I did it like this here as the rest of the package is doing it this way.

PR #1277, opened 2018-07-16
log/modlog: Allow LOG_[...] macros to use modlog

ccollins476ad:
Create a syscfg setting: `MODLOG_LOG_MACROS`.  If enabled, this setting retires the old `LOG_[...]` macros and redefines them to use modlog.


sterlinghughes:
I think this is a useful setting, and after the next release should be on by default -- thoughts?

PR #1281, opened 2018-07-18
Add SVC IRQ name to mcu.h to avoid naming issues

utzig:
STM32L0 changed the naming as had previously happened on Cortex-M3 which was fixed on #851.


utzig:
@mkiiskila Done!

PR #1285, opened 2018-07-19
move sensor_itf to driver_itf WIP please evaluate

jacobrosenthal:
The sensor interface model is really nice way to properly remove state from driver implementations and a missing piece in the driver model.

Further Ive been pushing that sensors should rather be a wrapper around slimmer driver implementations.

I can and will break this out, but wanted it all in one place for discussion and testing. This pr:
* brings what I think is the bare minimum from sensor over to a driver pkg 
* converts the lis2dxxx sensors.
* removes all sensor stuff from the drv2605 which was using sensor just to get the itf.

Note the major change was making the _config take a driver_itf instead of a os_dev object. This is the major change and Im not sure how to fix it differently other than adding a third arg like init.

But It does brings up an argument Ive made before, no drivers/sensors (excepting pwm, uart, adc) utilize os_dev's open/close handlers and we now have sensor (or driver in this case) interface locking.. are we getting our memorys worth with os_devs in sensors/drivers?? 

Thoughts?


sterlinghughes:
Why not just add this functionality into the os_dev framework?  It seems like a fairly low code & memory footprint, and useful/generalizable?

vrahane:
I agree. The only issue I have with a generic interface is that some hardware requires more context for physical interfaces than the other. For example, sensors might have interrupt pins, led chips generally don't and so on.

jacobrosenthal:
My main thrust is getting to writing thinner drivers instead of driver/sensor combos and the absolute quickest path to that was pulling sensor interface out. Im happy to see it in os_dev or wherever but I just dont have a vision for what os_dev should look like. More specific brainstorming would be appreciated. 

As a refresher, we previously had this discussion here, which seem to be concluded PRs welcome.
https://lists.apache.org/thread.html/%3C113E05F9-CD70-4DAD-9E56-73E4F248CDEC@runtime.io%3E

sterlinghughes:
PRs are welcome (at least -- I'm glad you are looking at this!) - i think generally this looks like a good start, but it should go into os_dev.  os_dev was meant to be a skeleton to build on, and I agree with you that this is a common pattern, seen in sensor drivers, but also something that most people who build a driver end up implementing themselves as well.  I'll review a bit closer as well.

jacobrosenthal:
2 new totally non functional commits just to take a look at what driver and sensor separation feels like. I pulled the low level driver code out of the lis2dw12 sensor. The driver builds, but wont be functional yet. And all i did was remove code on whats left of the lis2dw12 sensor so that wont even build. It does seem like a clean and useful separation to me.

Im presuming the sensor would call the drivers init and config? or is that too hardcoded? Though its interesting to look at how all a single lis2dxxx sensor would look utilizing either one of the three lis2dxxx drivers...

jacobrosenthal:
Interesting fact. It lookes to me like the init function and init_arg are already attached to os_dev and therefore interface has been available there all this time.

ccollins476ad:
This looks like it has stalled. Can this be closed?

PR #1294, opened 2018-07-25
Encrypted flash

mkiiskila:
This is done with a pseudo flash device, which provides hal_flash interface.
Device is split into common part, and architecture specific part. Only nrf51/nrf52 implementation is within this PR. So the current split might need adjusting when/if other platforms are added. Split is done so that you can take advantage of crypto acceleration when available.

Internally it uses AES-128 CTR mode where counter values comes from block offset.

At the moment the use cases are a bit limited; we'd need to augment hal_flash interface a bit to allow use of this device for things like FCB and image storage. Specifically, they'd need a routine with which to query the flash device whether underlying flash is empty or not.

As a example I modified one BSP to show how to set this up.

### DIFF
--- a/fs/fcb/src/fcb.c
+++ b/fs/fcb/src/fcb.c
@@ -150,9 +150,6 @@ fcb_get_len(uint8_t *buf, uint16_t *len)
     int rc;
 
     if (buf[0] & 0x80) {
-        if (buf[0] == 0xff && buf[1] == 0xff) {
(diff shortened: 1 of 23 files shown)


utzig:
My useful comment of the day: it should have been called `crypt_flash`!

mkiiskila:
Thanks for taking a look, good points in all comments.

PR #1297, opened 2018-07-26
Fix some issues with the L1/L0 flash driver

utzig:
- Fix alignment;
- Disable write protection just while erasing/writing instead of always leaving the flash unlocked. Since the watchdog was left on in Mynewt when asking for an upgrade, and not initialized on MCUBoot, the tickle in this driver was causing an invalid write to the flash, that was unprotected, as well as a restart failing verification of the swapped image. A related watchdog issue is being submitted to MCUBoot;
- Other small refactorings applied to ease merging L4/L1/L0 into one single configurable flash driver.

### DIFF
--- a/hw/mcu/stm/stm32l1xx/src/hal_flash.c
+++ b/hw/mcu/stm/stm32l1xx/src/hal_flash.c
@@ -18,6 +18,7 @@
  */
 
 #include <string.h>
+#include <syscfg/syscfg.h>
 #include "stm32l1xx_hal_def.h"
 #include "stm32l1xx_hal_flash.h"
 #include "stm32l1xx_hal_flash_ex.h"
@@ -50,7 +51,7 @@ const struct hal_flash stm32l1_flash_dev = {
     .hf_base_addr = 0x08000000,
     .hf_size = _FLASH_SIZE,
     .hf_sector_cnt = _FLASH_SIZE / _FLASH_SECTOR_SIZE,
-    .hf_align = 8,
+    .hf_align = MYNEWT_VAL(MCU_FLASH_MIN_WRITE_SIZE),
 };
 
 static int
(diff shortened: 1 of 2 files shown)


utzig:
Btw, this probably also affects L4 as well, but maybe because its flash is erased to `0xff` it hadn't been detected before so I am not fixing that one (although I hope to merge both drivers soon).

hw/mcu/stm/stm32l1xx/src/hal_flash.c
```diff
@@ -65,31 +65,44 @@ static int
...
     int rc;
+    uint8_t align;
     uint32_t num_words;
 
     if (!num_bytes) {
         return -1;
     }
 
+    align = dev->hf_align;
+    num_words = ((num_bytes - 1) / dev->hf_align) + 1;
```
mkiiskila: Divide is expensive, bitwise operations are cheap. I would use those rather, given that alignment is a power of 2.

utzig: Yup, agreed but `udiv` is not thaaaat bad. I guess it will be worse on L0 which is Cortex-M0 and will need emulation. Btw, I took a look at the generated listing and there are only 3 divisions, this one + two in the clock initialization code inside the HAL!

utzig: Btw, I don't know of any bit-fiddling that can do `log2` so I would use a small lookup table (alignments are 1 to 8 only) or simply a `switch`, but I wonder how better that would be in number of cycles than just doing a `udiv` here. It's probably much better on the L0 but may be worse on the L1/L4!

mkiiskila: Luckily I have this code around. And actually, I'd be quite interested in seeing what different it does make :)
https://github.com/mkiiskila/incubator-mynewt-core/tree/flash_speed_test

Include test/flash_test package with slinky, and then from console:
'flash_speed 0 0x08000000 0 0 1'
I admit, the arguments to this might not be the greatest.

If you have multiple possible alignments, and this code will be shared by all, I'd do:
#if FLASH_ALIGNMENT == 8
#define FLASH_ALIGNMENT_SHIFT 3
#elif FLASH_ALIGNMENT == 4
#define FLASH_ALIGNMENT_SHIFT 2
...

num_words = (((num_bytes - 1) >> FLASH_ALIGNMENT_SHIFT) + 1;

Anyway, running the speed test would show if udiv is worth avoiding or not.

utzig:
@mkiiskila Ok, call me an ass but I didn't even consider to use `#ifdefs` with `MYNEWT_VAL(MCU_FLASH_MIN_WRITE_SIZE)`. It no doubt beats speed of using division (but generates same code size). Now I'm wondering why only the native flash driver makes `.hf_align = MYNEWT_VAL(MCU_FLASH_MIN_WRITE_SIZE)` since this seems like a duplication for every BSP/MCU!

mkiiskila:
Native uses syscfg setting so you can validate/test package user behavior with different flash alignment restrictions.

utzig:
@mkiiskila Sure, but the question was: why don't everyone else?

PR #1299, opened 2018-07-27
[RFC] [NRF52] Remove uint64_t div in watchdog init

utzig:
Because of precision requirements the watchdog was using a uint64_t division cause it allowed any possible timeout to be used. This updates the code to rely only on uint32_t division with the limitation of a maximum timeout of around 17 min. It shrinks the code due to not linking in the div emulation code.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_watchdog.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_watchdog.c
@@ -44,13 +44,14 @@ nrf52_wdt_irq_handler(void)
 int
 hal_watchdog_init(uint32_t expire_msecs)
 {
-    uint64_t expiration;
-
     NRF_WDT->CONFIG = WDT_CONFIG_SLEEP_Msk;
 
-    /* Convert msec timeout to counts of a 32768 crystal */
-    expiration = ((uint64_t)expire_msecs * 32768) / 1000;
-    NRF_WDT->CRV = (uint32_t)expiration;
+    if (expire_msecs >= 44739243) {
+        /* maximum allowed time is near 12.5 hours! */
+        assert(0);
+    } else {
+        NRF_WDT->CRV = (expire_msecs * 32) + ((expire_msecs * 96) / 125);
+    }
 
     NVIC_SetVector(WDT_IRQn, (uint32_t) nrf52_wdt_irq_handler);
     NVIC_SetPriority(WDT_IRQn, (1 << __NVIC_PRIO_BITS) - 1);


utzig:
@mkiiskila @wes3 This is mostly a Proof of Concept, based on Marko's comments here: https://github.com/runtimeco/mcuboot/pull/322

Some numbers, based on mcuboot loader:

Without watchdog linked the size is `9660` bytes. With the current watchdog initialization it is `10568`. This amount is due to this code generated because of the `uint64_t` division:

```
/* Convert msec timeout to counts of a 32768 crystal */                            
expiration = ((uint64_t)expire_msecs * 32768) / 1000;                              
 85c:»··f44f 727a »·mov.w»··r2, #1000»··; 0x3e8                                    
 860:»··2300      »·movs»···r3, #0                                                 
 862:»··03c0      »·lsls»···r0, r0, #15                                            
 864:»··0c49      »·lsrs»···r1, r1, #17                                            
 866:»··f001 fdd9 »·bl»·241c <__aeabi_uldivmod>                                    
NRF_WDT->CRV = (uint32_t)expiration;
```

With the updated code the size is `9860` bytes. I suppose the change would be less significant in a final firmware where `__aeabi_uldivmod` might be linked in anyway, so leaving this here for RFC only now!

hw/mcu/nordic/nrf52xxx/src/hal_watchdog.c
```diff
@@ -44,13 +44,20 @@ nrf52_wdt_irq_handler(void)
...
     NRF_WDT->CONFIG = WDT_CONFIG_SLEEP_Msk;
 
-    /* Convert msec timeout to counts of a 32768 crystal */
-    expiration = ((uint64_t)expire_msecs * 32768) / 1000;
-    NRF_WDT->CRV = (uint32_t)expiration;
+    if (expire_msecs >= 1048576) {
+        /* watchdog with more than 17 minutes? */
+        assert(0);
+    } else {
+        uint32_t e = expire_msecs, rtc = 32768, d = 1000;
```
wes3: This is definitely not mynewt coding style: declaring multiple local variables in one line and initializing them all on the same line, too.

hw/mcu/nordic/nrf52xxx/src/hal_watchdog.c
```diff
@@ -44,13 +44,20 @@ nrf52_wdt_irq_handler(void)
...
+        while (e >= 131072) {
+            e /= 2;
+            rtc /= 2;
+            d /= 2;
```
wes3: I think you could do this in the following manner and get much more range. I probably should have done this from the  start but did not worry about the 65-bit divide. We are trying to do the following: (x * 32768) / 1000. This is the same as doing this: x * 32.768. Since we do not want to use floating point you can do this: (x * 32) + ((x * 768) / 1000) which is also this: (x * 32) + ((x * 96) / 125). You can then get much more range.

utzig: Cool!

```
#include <stdint.h>
#include <stdio.h>

int main(void)
{
    uint64_t r;
    uint32_t s;

    for (uint32_t u = 0; u < 0xffffffff; u++) {
        r = ((uint64_t)u * 32768) / 1000;
        s = (u * 32) + ((u * 96) / 125);

        if (r != s) {
            printf("u=%u, r=%lu, s=%u\n", u, r, s);
            break;
        }
    }

    return 0;
}
```

```
$ ./test
u=44739243, r=1466015514, s=1431655776
```

I guess you just made it work upto 12 hours, you're a genius @wes3!

wes3: Thanks Fabio but not really warranted.

utzig:
And what if someone needs the watchdog to tickle every 24h? ;-)

PR #1301, opened 2018-07-28
kernel/os: Add debug mode syscfg setting

ccollins476ad:
This commit adds a new syscfg setting:

    OS_DEBUG_MODE

This setting doesn't do anything by itself, but enabling it causes several other settings to be enabled.  Currently, this setting enables the following:

    OS_CRASH_STACKTRACE
    OS_CTX_SW_STACK_CHECK
    OS_MEMPOOL_CHECK
    OS_MEMPOOL_POISON
    HAL_FLASH_VERIFY_WRITES
    HAL_FLASH_VERIFY_ERASES

### DIFF
--- a/hw/hal/src/hal_flash.c
+++ b/hw/hal/src/hal_flash.c
@@ -104,15 +104,13 @@ static int
 hal_flash_cmp_erased(const struct hal_flash *hf, uint32_t address,
   uint32_t num_bytes)
 {
-    uint8_t empty[MYNEWT_VAL(HAL_FLASH_VERIFY_BUF_SZ)];
     uint8_t buf[MYNEWT_VAL(HAL_FLASH_VERIFY_BUF_SZ)];
 
     uint32_t off;
     uint32_t rem;
     int chunk_sz;
     int rc;
-
-    memset(empty, 0xff, sizeof empty);
+    int i;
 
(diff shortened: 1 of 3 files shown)


hw/hal/src/hal_flash.c
```diff
@@ -91,11 +93,98 @@ hal_flash_read(uint8_t id, uint32_t address, void *dst, uint32_t num_bytes)
...
+        } else {
+            chunk_sz = rem;
+        }
+
+        rc = hf->hf_itf->hff_read(hf, address + off, buf, chunk_sz);
+        if (rc != 0) {
+            return rc;
+        }
+
+        if (memcmp(buf, empty, chunk_sz) != 0) {
```
andrzej-kaczmarek: I would just do a simple loop here and remove `empty` buffer from stack - it seems redundant to me

ccollins476ad: Good point.  I'm not sure why I didn't think of that.

ccollins476ad:
Thanks, Andrzej.  I have fixed that ridiculous extra buffer allocation that I should never have done in the first place :).  The fix is in a separate commit to make reviewing it easier.  I will combine the two commits before merging.

PR #1306, opened 2018-08-01
sys/console; rename global variables to make them more identifiable

mkiiskila:
as console variables. Current naming was confusing while running code
in debugger.

### DIFF
--- a/sys/console/full/include/console/console.h
+++ b/sys/console/full/include/console/console.h
@@ -43,7 +43,7 @@ struct console_input {
 
 typedef void (*console_rx_cb)(void);
 typedef int (*console_append_char_cb)(char *line, uint8_t byte);
-typedef void (*completion_cb)(char *str, console_append_char_cb cb);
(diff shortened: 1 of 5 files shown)


sys/console/full/include/console/console.h
```diff
@@ -43,7 +43,7 @@ struct console_input {
 
 typedef void (*console_rx_cb)(void);
 typedef int (*console_append_char_cb)(char *line, uint8_t byte);
-typedef void (*completion_cb)(char *str, console_append_char_cb cb);
+typedef void (*console_completion_cb)(char *str, console_append_char_cb cb);
```
utzig: since you seem to have added `cons_` prefix to everything else I think this should also be `cons_` instead of `console_`

mkiiskila: The thought I followed here was that all names visible through public API would start with
'console_', and internal symbols with 'cons_' for brevity.
Therefore the console_completion_cb.

I can re-rename everything with 'console_', I'd be ok with that as well.

PR #1307, opened 2018-08-01
Boot loader: Add assert for sector count

kasjer:
If sector count declared in bsp is greater than boot loader
specified 120 sectors, random memory can be overwritten.
This type of error can be difficult to detect.
It is much easier to add assert to detect error condition earlier.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -328,6 +328,11 @@ boot_read_sectors(void)
     if (rc != 0) {
         return BOOT_EFLASH;
     }
+    /*
+     * Make sure that BSP specific number of sectors will not result in
+     * random memory clobber
+     */
+    assert(num_sectors_slot0 <= BOOT_MAX_IMG_SECTORS);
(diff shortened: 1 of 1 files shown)


ccollins476ad:
I think elsewhere in the code, this condition is handled by calling `flash_area_to_sectors()` twice.  In the first call, `NULL` is passed for the last argument to indicate that the caller just wants to know the sector count.  After verifying that the count is not too large, the caller calls `flash_area_to_sectors()` again, but passes the actual buffer for the last argument.  An example where this is done is `config_init_fcb()`.  I think this is preferable because it catches the problem before corruption occurs, ensuring the sector count is accurately read.

It is odd that `boot_read_sectors()` assigns `num_sectors_slot0` and `num_sectors_slot1`, since they get overwritten by `flash_area_to_sectors()` anyway.  I know you didn't write this part, but these two lines should probably be deleted.

mkiiskila:
The logic Chris mentioned used to exist in the bootloader app. I'm guessing that (presumably accidental) omission of functionality happened when boot_read_sectors() was introduced. I would not mind having that check back.

PR #1308, opened 2018-08-01
Nucleo l476 fixes

kasjer:
With sector size 2048 bytes and 120 max sector count in bootloader
it is impossible to have more than 240kB per slot.
This fix does not changes slot positions and is considered to be
intermediate fix that allows to use nucleo board.
Proper fix should also change layout or sector size and placement.

File board/st_nucleo_l4.cfg is not present in openocd releases.
Debug and download script now have correct path:
board/st_nucleo_l476rg.cf

### DIFF
--- a/hw/bsp/nucleo-l476rg/bsp.yml
+++ b/hw/bsp/nucleo-l476rg/bsp.yml
@@ -40,11 +40,11 @@ bsp.flash_map:
         FLASH_AREA_IMAGE_0:
             device: 0
             offset: 0x08010000
-            size:  464kB
+            size:  240kB
         FLASH_AREA_IMAGE_1:
             device: 0
             offset: 0x08084000
-            size: 464kB
+            size: 240kB
         FLASH_AREA_IMAGE_SCRATCH:
             device: 0
             offset: 0x080f8000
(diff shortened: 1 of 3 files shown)


utzig:
Did you have any issues with the previous openocd board config? Also I used `MCUBoot` with `BOOTUTIL_MAX_IMG_SECTORS` == 256 to enable swap upgrade on this board. Not sure about reducing the size of slots to adhere to limitation in the bundled bootloader, also not sure there is any way to ask for MCUBoot in a BSP, so I am undecided for now!

kasjer:
Since mcuboot handles slots with higher sector count slot size we can drop second commit.

kasjer:
file st_nucleo_l4.cfg is not in openocd, maybe it was named like that before releases but right now such file does not exists so one that is there is proposed.

mkiiskila:
My installation of openocd has both st_nucleo_l4.cfg, as well as st_nucleo_l476rg.cfg.

kasjer:
hm, I also found some binary distributions that have st_nucleo_l4.cfg file instead of l476 one.
Right now I'm not sure how we should handle this.

kasjer:
Fabio maybe it would be better for you to clear this as you know what you are talking about. I just tried to use nucleo board that I have at home and failed on first try. I would gladly abandon this PR if you would fix what need to be fixed in correct way.

utzig:
@kasjer Ok, added with https://github.com/apache/mynewt-core/pull/1311. Please drop https://github.com/apache/mynewt-core/pull/1308/commits/4df3ae5068d74b0e5cffad33c98bad98e9fe00a8

PR #1309, opened 2018-08-01
STM32L0

utzig:
Adds initial STM32L0 support

Everything apart from SPI slave is working, some FIXMEs to get rid off. Also depends on support for flash erased with zero to be opened in another PR.

Will add reviewers when complete!


hw/mcu/stm/stm32_common/src/hal_timer.c
```diff
@@ -28,6 +28,11 @@
 #include "mcu/stm32_hal.h"
 #include "stm32_common/stm32_hal.h"
 
+#if MYNEWT_VAL(MCU_STM32L0)
+void TIM_Base_SetConfig(TIM_TypeDef *TIMx, TIM_Base_InitTypeDef *Structure);
+void TIM_CCxChannelCmd(TIM_TypeDef* TIMx, uint32_t Channel, uint32_t ChannelState);
```
kasjer: `*` used inconsequently in those two lines

utzig: I fixed the inconsistency on latest rebase (this was just copy/paste from stm32cube because those functions were made `static`). Thanks!

PR #1310, opened 2018-08-02
apps: Remove BLE sample applications

sjanc:
Those were moved to apache-nimble repository.


utzig:
Looks good. Are all the tutorials/docs updated to refer to the apps in `mynewt-nimble`?

utzig:
Btw, the reason I asked is because I was looking here: https://mynewt.apache.org/latest/network/docs/mesh/sample.html and it still uses `@apache-mynewt-core/apps/blemesh` (but I haven't checked if the docs sources are up to date!), so just to ensure tutorials, etc are updated.

sjanc:
good point, we should update docs too

PR #1316, opened 2018-08-06
adc/adc_nrf52: Handle ADC calibration done event.

rymanluk:
This PR  fixes assert() when ADC offset calibration is done with nrfx_saadc_calibrate_offset()

This PR does not provide HAL API to trigger ADC calibration (maybe it should be simple adc_offset_calibrate() ?),
however last patch adds support to send ADC_EVENT_CALIBRATED to the user when calibration is completed. 
https://github.com/apache/mynewt-core/commit/ec294361aa11bb3ea425f13f45b9404b7a7cf784

For now I just need first two patches, however open to hear comments on API for this.

### DIFF
--- a/hw/drivers/adc/adc_nrf52/src/adc_nrf52.c
+++ b/hw/drivers/adc/adc_nrf52/src/adc_nrf52.c
@@ -65,7 +65,9 @@ nrf52_saadc_event_handler(const nrfx_saadc_evt_t *event)
                                        done_ev->size * sizeof(nrf_saadc_value_t));
             break;
         case NRFX_SAADC_EVT_CALIBRATEDONE:
-            /*TODO Expose to the user? */
+            rc = global_adc_dev->ad_event_handler_func(global_adc_dev,
+                                       global_adc_dev->ad_event_handler_arg,
+                                       ADC_EVENT_CALIBRATED, NULL, 0);
             break;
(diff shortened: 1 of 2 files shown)


hw/drivers/adc/adc_nrf52/src/adc_nrf52.c
```diff
@@ -58,14 +58,25 @@ nrf52_saadc_event_handler(const nrfx_saadc_evt_t *event)
     /* Right now only data reads supported, assert for unknown event
      * type.
      */
```
andrzej-kaczmarek: this comment should be removed

hw/drivers/adc/adc_nrf52/src/adc_nrf52.c
```diff
@@ -58,14 +58,25 @@ nrf52_saadc_event_handler(const nrfx_saadc_evt_t *event)
...
-    assert(event->type == NRFX_SAADC_EVT_DONE);
-
-    done_ev = (nrfx_saadc_done_evt_t * const) &event->data.done;
+    switch(event->type) {
```
andrzej-kaczmarek: missing space before (

rymanluk:
@andrzej-kaczmarek thx - fixed

PR #1321, opened 2018-08-08
Add simple PM for UART device driver

andrzej-kaczmarek:
This is not complete, just an attempt to be able to disable UART to save some power without hacks like closing and reopening it. The whole `os_dev_suspend`/`os_dev_resume` framework needs more work since it was likely never used before so I just quickly fixed it to make this work.

### DIFF
--- a/hw/drivers/uart/include/uart/uart.h
+++ b/hw/drivers/uart/include/uart/uart.h
@@ -72,12 +72,6 @@ struct uart_driver_funcs {
     void (*uf_blocking_tx)(struct uart_dev *, uint8_t);
 };
 
-struct uart_dev {
-    struct os_dev ud_dev;
-    struct uart_driver_funcs ud_funcs;
-    void *ud_priv;
-};
-
 /*
  * ***Note that these enum values must match what hw/hal/hal_uart.h***
  */
@@ -92,6 +86,14 @@ enum uart_flow_ctl {
     UART_FLOW_CTL_RTS_CTS = 1   /* RTS/CTS */
 };
 
+struct uart_conf_port {
+    uint32_t uc_speed;          /* baudrate in bps */
+    uint8_t uc_databits;        /* number of data bits */
+    uint8_t uc_stopbits;        /* number of stop bits */
+    enum uart_parity uc_parity;
+    enum uart_flow_ctl uc_flow_ctl;
+};
+
 struct uart_conf {
     uint32_t uc_speed;          /* baudrate in bps */
     uint8_t uc_databits;        /* number of data bits */
@@ -102,6 +104,13 @@ struct uart_conf {
     uart_rx_char uc_rx_char;
     uart_tx_done uc_tx_done;
     void *uc_cb_arg;
+};
+
+struct uart_dev {
+    struct os_dev ud_dev;
+    struct uart_driver_funcs ud_funcs;
+    struct uart_conf_port ud_conf_port;
+    void *ud_priv;
 };
 
 /*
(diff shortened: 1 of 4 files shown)


hw/drivers/uart/include/uart/uart.h
```diff
@@ -104,6 +98,14 @@ struct uart_conf {
     void *uc_cb_arg;
 };
 
+struct uart_dev {
+    struct os_dev ud_dev;
+    struct uart_driver_funcs ud_funcs;
+    struct uart_conf ud_conf;
+    void *ud_priv;
+
```
mkiiskila: Unnecessary empty line.

hw/drivers/uart/include/uart/uart.h
```diff
@@ -104,6 +98,14 @@ struct uart_conf {
     void *uc_cb_arg;
 };
 
+struct uart_dev {
+    struct os_dev ud_dev;
+    struct uart_driver_funcs ud_funcs;
+    struct uart_conf ud_conf;
```
mkiiskila: I would recommend remembering the fields from speed -> flow_ctl settings, instead of retaining the function pointers (which the hal driver will remember). This'll save some RAM. I think the call to hal_uart_init_cbs() from uart_hal_resume() is unnecessary.

hw/drivers/uart/uart_hal/src/uart_hal.c
```diff
@@ -82,31 +104,70 @@ uart_hal_open(struct os_dev *odev, uint32_t wait, void *arg)
...
+     * then.
+     */
+
+    assert(dev);
+
+    /*
+     * It seems as for now we have no way of making decision whether we can
+     * suspend or not, so let's just allow force suspend at "now".
+     */
+    if ((suspend_at > os_time_get()) || !force) {
```
mkiiskila: Time comparison does not take into account rollover. Use OS_TIME_TICK_XX() macros.

hw/drivers/uart/uart_hal/src/uart_hal.c
```diff
@@ -82,31 +104,70 @@ uart_hal_open(struct os_dev *odev, uint32_t wait, void *arg)
...
+     * and enable UART to have some basic power saving. The whole suspend/resume
+     * framework probably should be revised and proper implementation added here
+     * then.
+     */
+
+    assert(dev);
+
+    /*
+     * It seems as for now we have no way of making decision whether we can
+     * suspend or not, so let's just allow force suspend at "now".
```
mkiiskila: Indeed, you'd need to add a new function to hal_uart interface to figure that one out. That can be done in some other PR.

hw/drivers/uart/uart_hal/src/uart_hal.c
```diff
@@ -82,31 +104,70 @@ uart_hal_open(struct os_dev *odev, uint32_t wait, void *arg)
...
+
+    /*
+     * XXX this should not be used as an example on how to add suspend/resume
+     * to os_dev since it's just provisional implementation to allow disable
+     * and enable UART to have some basic power saving. The whole suspend/resume
+     * framework probably should be revised and proper implementation added here
+     * then.
+     */
+
+    assert(dev);
```
mkiiskila: assert() is unnecessary here. The caller would've already come via os_dev_suspend()

andrzej-kaczmarek:
fixed all comments and rebased (except of extending UART HAL which needs some more work in another PR later).

PR #1323, opened 2018-08-09
Allow flash devices to specify erase value

utzig:
This updates support for devices that are erased to zero instead of 0xff.

@mkiiskila I slightly changed the low-level interfaces you have initially written but should be 100% compatible from the user perspective.

### DIFF
--- a/hw/bsp/nucleo-f401re/src/hal_bsp.c
+++ b/hw/bsp/nucleo-f401re/src/hal_bsp.c
@@ -60,7 +60,8 @@ const struct hal_flash stm32f4_flash_dev = {
     .hf_base_addr = 0x08000000,
     .hf_size = 512 * 1024,
     .hf_sector_cnt = NAREAS - 1,
-    .hf_align = 1
+    .hf_align = 1,
+    .hf_erased_val = 0xff,
 };
 
 #if MYNEWT_VAL(UART_0)
--- a/hw/bsp/nucleo-f413zh/src/hal_bsp.c
+++ b/hw/bsp/nucleo-f413zh/src/hal_bsp.c
@@ -69,7 +69,8 @@ const struct hal_flash stm32f4_flash_dev = {
     .hf_base_addr = 0x08000000,
     .hf_size = 1536 * 1024,
     .hf_sector_cnt = NAREAS - 1,
-    .hf_align = 1
+    .hf_align = 1,
+    .hf_erased_val = 0xff,
 };
 
 #if MYNEWT_VAL(UART_0)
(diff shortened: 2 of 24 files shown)


utzig:
Associated `mcuboot` PR: https://github.com/runtimeco/mcuboot/pull/327

utzig:
I tested this on the L1/L4 (0 and 0xff erased flash respectively). I am pretty confident that this is working OK now, and unit tests are passing. Still need to update design doc in `mcuboot` and some other small fixes there but here everything looks about right...

mkiiskila:
This will not work for encrypted flash.
Here's dump of data from erased sector of flash through encrypting driver. 

000734 compat> flash read 0x7c000 32 1                                          
001319 Read 0x7c000 + 20                                                        
001319   0x7c000: 0xa4 0xac 0x2e 0x6f 0xb7 0x11 0xbb 0x98                       
001321   0x7c008: 0xd3 0xac 0x3d 0x28 0x74 0x5d 0x3e 0x71                       
001322   0x7c010: 0x85 0x40 0x7c 0x18 0xfe 0x83 0x65 0x83                       
001324   0x7c018: 0x9e 0x67 0xe5 0x83 0xa4 0x09 0xbc 0xa8

utzig:
@mkiiskila Yup, missed that one, on it!

mkiiskila:
The changes do not address encrypted flash scenario still.

Check with the following target:
```
[marko@IsMyLaptop:~/src2/incubator-mynewt-blinky]$ newt target show slinky_nrf52 
targets/slinky_nrf52
    app=@apache-mynewt-core/apps/slinky
    bsp=@apache-mynewt-core/hw/bsp/nrf52dk
    build_profile=debug
    syscfg=BOOTUTIL_USE_MBED_TLS=1:BOOT_SERIAL_NVREG_INDEX=0:CONSOLE_HISTORY_SIZE=16:ENC_FLASH_DEV=1:IMGMGR_CLI=0:IMGMGR_COREDUMP=1:OS_COREDUMP=1:SHELL_OS_SERIAL_BOOT_NVREG=1
```

Now the encrypted flash overlay is over internal flash. You can test from within gdb

```
(gdb) call hal_flash_isempty(0, 0x7c000, 32)
$1 = 1
(gdb) call hal_flash_isempty(1, 0x7c000, 32)
$2 = 0
```
Return code from both calls should be 1.

utzig:
@mkiiskila I have not updated it before! I now rolled back the logic of using `hff_isempty` just to make it easier to approve this one before doing more thorough changes. Also had missed some recent "verify flash writes" stuff. Btw, both users of enc_flash (native/nrf52dk) don't have calls to `enc_flash_setkey`. How are keys being managed?

mkiiskila:
Yeah, the sample inclusion within BSP does not set the key. It does mangle the data without setting it already, so you can see try out how that flash driver behaves :)

Actual use case would be to use encrypted flash over external SPI flash, and keep the key within internal flash (could also be  STM32 OTP area or NRF52 user area). And then set it within hal_bsp_init(). That way encrypted flash could even be used from within bootloader.

PR #1328, opened 2018-08-13
libc: Add optimized memcmp() for Cortex-M3/4/7

andrzej-kaczmarek:
This takes advantage of unaligned access on Cortex-M3/4/7 and compares
memory by-word instead of by-byte, if possible. Any remaining data at
the end are compared by-byte.

### DIFF
--- a/libc/baselibc/src/memcmp.c
+++ b/libc/baselibc/src/memcmp.c
@@ -6,14 +6,50 @@
 
 int memcmp(const void *s1, const void *s2, size_t n)
 {
+    int d = 0;
+
+#if defined(ARCH_cortex_m3) || defined(ARCH_cortex_m4) || defined(ARCH_cortex_m7)
+    asm (".syntax unified                   \n"
+         "       push  {r4, r5, r6}         \n"
+         "       mov   r5, #0               \n"
+         "       and   r6, r2, #0xfffffffc  \n"
+         "       b     test1                \n"
+         "loop1: ldr   r3, [r0, r5]         \n"
+         "       ldr   r4, [r1, r5]         \n"
+         "       cmp   r3, r4               \n"
+         "       bne   res1                 \n"
+         "       add   r5, #4               \n"
+         "test1: cmp   r5, r6               \n"
+         "       bne   loop1                \n"
+         "       b     test2                \n"
+         "res1:  rev   r3, r3               \n"
+         "       rev   r4, r4               \n"
+         "       subs  r3, r4               \n"
+         "       ite   hi                   \n"
+         "       movhi r3, #1               \n"
+         "       movls r3, #-1              \n"
+         "       b     done                 \n"
+         "loop2: ldrb  r3, [r0, r5]         \n"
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
for comparison, below are execution times (in us) for `memcmp` on nRF52840 on equal buffers: `A` = buffer size, `B` = standard `memcmp`, `C` = optimized `memcmp`
```
   A        B        C
   1      5.6      5.7
   2      6.9      5.7
   4      6.3      5.8
 256     81.3     20.8
 512    157.3     35.8
1024    309.3     65.8
2048    613.3    125.8
```

libc/baselibc/src/memcmp.c
```diff
@@ -6,14 +6,47 @@
...
+         "       b    test1                 \n"
+         "loop1: ldr  r3, [r0, r5]          \n"
+         "       ldr  r4, [r1, r5]          \n"
+         "       cmp  r3, r4                \n"
+         "       bne  loop2                 \n"
+         "       add  r5, #4                \n"
+         "test1: cmp  r5, r6                \n"
+         "       bne  loop1                 \n"
+         "       b    test2                 \n"
+         "res1:  rev  r3, r3                \n"
```
wes3: I am not the most conversant with inline assembly but how does the code ever get to the res1 label?

andrzej-kaczmarek: Good catch - it should jump here from line 20. This however still works since it jumps to `loop2` instead which compares word byte by byte. I added `res1` label later and forgot to update branch :-)

mkiiskila:
It believe return codes are not right. I get a incorrect return code in the 3rd loop in the included code snippet when i = 1 and j = 81.

See http://pubs.opengroup.org/onlinepubs/9699919799/functions/memcmp.html.

```
uint8_t data1[64];
uint8_t data2[64];

static void
memcmp_check(void)
{
    int i;
    int j;
    int rc;

    for (i = 0; i < 4; i++) {
        for (j = 1; j < 32; j++) {
            memset(data1, 0xa, sizeof(data1));
            memset(data2, 0xa, sizeof(data2));

            rc = memcmp(data1 + i, data2 + i, j);
            assert(rc == 0);
        }
    }
    for (i = 0; i < 4; i++) {
        for (j = 1; j < 32; j++) {
            memset(data1, 0, sizeof(data1));
            memset(data1 + i, 0xaf, j);
            memset(data2, 0xaf, sizeof(data2));

            rc = memcmp(data1 + i, data2 + i, j);
            assert(rc == 0);
        }
    }

    for (i = 0; i < 4; i++) {
        for (j = 0; j < 255; j++) {
            memset(data1, 0xaf, sizeof(data1));
            memset(data2, 0xaf, sizeof(data2));

            data2[5] += j;
            rc = memcmp(data1 + i, data2 + i, 16);
            if (data1[5] < data2[5]) {
                assert(rc < 0);
            } else if (data1[5] == data2[5]) {
                assert(rc == 0);
            } else {
                assert(rc > 0);
            }
        }
    }
}
```

PR #1331, opened 2018-08-13
test/runtest: Move testbench code into runtest

ccollins476ad:
Prior to this PR, the testbench app predefined a set of four tasks for use by unit tests.  Unit tests that needed to use generic tasks would assume the presence of tasks called "task1", "task2", etc.  In addition, the testbench was responsible for logging test results.

This approach had some problems:
* Introduction of unit tests requires major changes to the app:
    * Define new tasks and stacks and add code to initialize and suspend
    * Create a cbmem and register a test log.
    * Write dedicated "driver" code for each test suite.
* Different test code for SELFTEST vs. real-hardware (the testbench app is not present during a self-test).
* Number of generic tasks limited to four.

This PR adds a generic task-allocation mechanism to the `test/runtest` library.  When a unit test needs a new task, it requests one from runtest.  When a test completes, runtest suspends all the allocated tasks.  In addition, runtest registers a cbmem-backed test log if configured to do so (`RUNTEST_LOG`).


test/runtest/syscfg.yml
```diff
@@ -19,11 +19,42 @@
 # Package: test/runtest
 
 syscfg.defs:
+    RUNTEST_LOG:
+        description: >
+            Causes test results to be logged to a cbmem-backed logged.
+        value: 0
```
mkiiskila: Maybe default this to 1? This might make more sense to be defaulted on. I'm fine either way :)

PR #1333, opened 2018-08-14
[STM32L4] Remove error clearing on flash driver

utzig:
This removes custom error clearing code. Calls to `erase` and `program` already do pretty good error checking and clearing on both beginning/end. This code was causing flash corruption when swap
upgrades were performed with interruptions.


utzig:
btw, I am not entirely sure why just resetting error bits was causing corruption and it's a bit complicated to debug since it only happened when swap upgrading with interruption (aka resets), but I am tested it thoroughly enough to be sure it works 100% after this change.

PR #1324, opened 2018-08-10
Flash speed test

mkiiskila:

PR #1336, opened 2018-08-16
Add flash-id parameter to flash cli command

utzig:
Adds `<flash-id>` option to allow flash command to work on a device that is not 0 (optional on dump command).

> `flash <flash-id>` -- dumps sector map
> `flash <flash-id> read <offset> <size>` -- reads bytes from flash
> `flash <flash-id> write <offset>  <size>` -- writes incrementing data pattern 0-8 to flash
> `flash <flash-id> erase <offset> <size>` -- erases flash

This probably conflicts with https://github.com/apache/mynewt-core/pull/1324 so might wait for that one to be merged.

### DIFF
--- a/test/flash_test/src/flash_test/flash_test.c
+++ b/test/flash_test/src/flash_test/flash_test.c
@@ -41,69 +41,99 @@ flash_cli_cmd(int argc, char **argv)
     uint32_t sz = 1;
     int sec_cnt;
     int i;
+    int devid;
     int soff;
     char *eptr;
     char tmp_buf[8];
     char pr_str[80];
 
-    hf = hal_bsp_flash_dev(0);
-    if (!hf) {
-        console_printf("No flash device present\n");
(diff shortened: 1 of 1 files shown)


test/flash_test/src/flash_test/flash_test.c
```diff
@@ -41,69 +41,99 @@ flash_cli_cmd(int argc, char **argv)
...
     char *eptr;
     char tmp_buf[8];
     char pr_str[80];
 
-    hf = hal_bsp_flash_dev(0);
-    if (!hf) {
-        console_printf("No flash device present\n");
+    if (argc > 1 && (!strcmp(argv[1], "?") || !strcmp(argv[1], "help"))) {
+        console_printf("Commands Available\n");
+        console_printf("flash <flash-id> -- dumps sector map \n");
```
ccollins476ad: Should this be `[flash-id]` (i.e., this argument is optional)?

utzig: Yup, that one is optional, updated, thanks!

PR #1338, opened 2018-08-20
Drv2605 odds and ends

jacobrosenthal:
I noticed some driver/sensor changes that werent catching drv2605 probably because its not in the sensors folder. I hope I caught them all.


hw/drivers/drv2605/src/drv2605.c
```diff
@@ -78,6 +79,11 @@ drv2605_write8(struct sensor_itf *itf, uint8_t reg, uint8_t value)
         .buffer = payload
     };
 
+    rc = sensor_itf_lock(itf, MYNEWT_VAL(DRV2605_ITF_LOCK_TMO));
+    if (rc) {
+        goto err;
```
ccollins476ad: Be careful not to unlock an already-unlocked mutex.  This should be `return rc` rather than `goto err`.  This same comment applies to the other calls to `sensor_itf_lock()`.

PR #1339, opened 2018-08-20
sys/config: Add API to use FCB as generic key-value storage

andrzej-kaczmarek:
This adds simple API to use config_fcb engine as generic key-value
storage without need to use config module. The main difference is that
config values need to be somehow handled by registered module thus it is
useful for pre-defined settings.

With config_fcb_kv API it is possible to read and write any key at any
time and they are immediately persisted and loaded directly from FCB so
there is no need to store them in RAM. This can be used as kind of a
scratchpad and is definitely not intended as replacement for sys/config
module.

### DIFF
--- a/sys/config/src/config_fcb.c
+++ b/sys/config/src/config_fcb.c
@@ -26,6 +26,7 @@
 
 #include "config/config.h"
 #include "config/config_fcb.h"
+#include "config/config_generic_kv.h"
 #include "config_priv.h"
 
 #define CONF_FCB_VERS		1
@@ -33,6 +34,12 @@
 struct conf_fcb_load_cb_arg {
     load_cb cb;
     void *cb_arg;
+};
+
+struct conf_kv_load_cb_arg {
+    const char *name;
+    char *value;
+    size_t len;
 };
 
 static int conf_fcb_load(struct conf_store *, load_cb cb, void *cb_arg);
@@ -150,11 +157,11 @@ conf_fcb_var_read(struct fcb_entry *loc, char *buf, char **name, char **val)
     return rc;
 }
 
-void
-conf_fcb_compress(struct conf_fcb *cf,
-                  int (*copy_or_not)(const char *name, const char *val,
-                                     void *cn_arg),
-                  void *cn_arg)
+static void
+conf_fcb_compress_internal(struct fcb *fcb,
+                           int (*copy_or_not)(const char *name, const char *val,
+                                              void *cn_arg),
+                           void *cn_arg)
 {
     int rc;
     char buf1[CONF_MAX_NAME_LEN + CONF_MAX_VAL_LEN + 32];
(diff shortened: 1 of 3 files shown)


sys/config/src/config_fcb.c
```diff
@@ -255,14 +262,84 @@ conf_fcb_append(struct conf_fcb *cf, char *buf, int len)
...
+    return conf_fcb_kv_save(&cf->cf_fcb, name, value);
+}
+
+void
+conf_fcb_compress(struct conf_fcb *cf,
+                  int (*copy_or_not)(const char *name, const char *val,
+                                     void *cn_arg),
+                  void *cn_arg)
+{
+    return conf_fcb_compress_internal(&cf->cf_fcb, copy_or_not, cn_arg);
```
ccollins476ad: This function returns `void`, so there is no need for `return`.

sys/config/src/config_fcb.c
```diff
@@ -255,14 +262,84 @@ conf_fcb_append(struct conf_fcb *cf, char *buf, int len)
...
+}
+
+static int
+conf_kv_load_cb(struct fcb_entry *loc, void *arg)
+{
+    struct conf_kv_load_cb_arg *cb_arg = arg;
+    char buf[CONF_MAX_NAME_LEN + CONF_MAX_VAL_LEN + 32];
```
ccollins476ad: I see this `32` magic number quite a bit in this file.  Do you know what it represents?  I couldn't figure out why this extra buffer space is required (I know this was already here before your change).

ccollins476ad:
Looks good.  My only request is that minor `return` thing.

sys/config/src/config_fcb.c
mkiiskila: 32 is probably carryover from older, happier times. I think '2' is now the correct magic number to add.

sys/config/src/config_fcb.c
```diff
@@ -255,14 +262,84 @@ conf_fcb_append(struct conf_fcb *cf, char *buf, int len)
...
+    }
+
+    strncpy(cb_arg->value, val_str, cb_arg->len);
+    cb_arg->value[cb_arg->len - 1] = '\0';
+
+    return 0;
+}
+
+int
+conf_fcb_kv_load(struct fcb *fcb, const char *name, char *value, size_t len)
```
mkiiskila: So what is the functionality expected here? The last key/value pair restored on load only?

andrzej-kaczmarek: I added doxygen comment for both functions :-)
This one will return last stored value (i.e. internally it obviously walks FCB and fills output buffer with each value so the last one stored will be returned).

sys/config/src/config_fcb.c
andrzej-kaczmarek: Yep, seems like `2` is length of element data written to FCB here:
https://github.com/apache/mynewt-core/blob/master/fs/fcb/src/fcb_append.c#L106

However, since we read data by `data_off` and `data_len` I think there's no need to add any magic value here and just keep buffer large enough for element data only. I dug in code a bit and it seems this will work, but would prefer that someone more familiar with FCB confirm...

andrzej-kaczmarek:
rebased, fixed return and added doxygen in new header file - no other changes

sys/config/src/config_fcb.c
mkiiskila: FCB does not use extra space. config uses key=value'\0'. So therefore 2 extra. I don't think we check CONF_MAX_NAME_LEN nor CONF_MAX_VAL_LEN separately either; only together. Those numbers are there just to guide the user. sys/config tries to accomodate what it can.

PR #1341, opened 2018-08-20
sys/console: add support for silencing output from the console

vrahane:
- Add support for silencing output from the console
- Change parse_ull_bounds to check for range and return SYS_ERANGE error


vrahane:
@ccollins476ad made changes, can you please taken a look.

vrahane:
Done, thanks @ccollins476ad

PR #1344, opened 2018-08-21
hw/bsp/nrf52-thingy: Update BSP to use nrf52xxx

andrzej-kaczmarek:


utzig:
Does the thingy52 use external flash or what is the reasoning for using encrypted flash? (apart from this doubt looks good to me as well)

andrzej-kaczmarek:
@utzig oops, I removed encrypted flash code but apparently forgot to commit it before pushing. fixed :)

andrzej-kaczmarek:
so I'll just go ahead and merge this (only `#ifdef` code for encrypted flash was removed since last approval)

PR #1346, opened 2018-08-22
hw/bsp: Refactor BSP for RuuviTag rev B

andrzej-kaczmarek:
Main purpose of this commit is to update BSP for RuuviTag to use
refactored `nrf52xxx package`. However, few extra modifications were
applied during clean up:
- Name is changed to more generic `ruuvitag_rev_b` since it should work
  with any rev B hardware. Previous name suggested it was for rev B2
  and at the same time description suggested it was for rev B3 and B4
  (latest rev is B6 at the time of writing).
- Debug and flash scripts were renamed to follow BSP name
- Added `BSP_RUUVITAG_REV_B_NUM` syscfg setting to specify revision of target
  hardware. This makes possible to apply proper default configuration
  since there were small differences between revisions (pin numbers and
  available peripherals).


andrzej-kaczmarek:
Note that for 1.5.0 we at least need to update this BSP to new MCU package - I can prepare appropriate PR if refactor is not accepted.

PR #1348, opened 2018-08-22
hw/bsp: Rename Nordic BSPs

andrzej-kaczmarek:
This renames all Nordic BSPs to use their board names (the one on sticker) instead of dev kits name.

The reason for this change is:
- to allow easily which BSP is needed for board since it's always on a   sticker unlike dev kit name
- to avoid need for rename when dev kit name changes due to board rev change (e.g. nRF52840 PDK was changed to nRF52840 DK but the board is the same except for its revision)


ccollins476ad:
I also don't see the need for this change, but I likely haven't given it enough thought.  I don't fully understand the stated reasons for the change:
> to allow easily which BSP is needed for board since it's always on a sticker unlike dev kit name
> to avoid need for rename when dev kit name changes due to board rev change (e.g. nRF52840 PDK was changed to nRF52840 DK but the board is the same except for its revision)

Aren't the PDK and DK different boards with different characteristics?  If they are indeed different, then the distinction that we have today seems like a good thing.

andrzej-kaczmarek:
@ccollins476ad both are PCA10056 and the only major difference is IC revision. boards rev 0.9-0.12 with engineering rev IC were sold as part of PDK and board rev 1.0 with final rev IC is now sold as part of DK (see [1] and [2]). so I don't think it makes sense to have `nrf52840pdk` and `nrf52840dk` as separate BSP-s and even if we do not want to change names of other BSP-s to follow board names, `nrf52840pdk` should be renamed to `nrf52840dk` since this is current name of dev kit.

btw, Zephyr uses PCA board names ;-)

[1] https://devzone.nordicsemi.com/f/nordic-q-a/35057/difference-between-nrf85-pdk-and-nrf52840-dk-both-seemingly-being-pca10056/134566#134566
[2] https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52%2Fdita%2Fnrf52%2Fcompatibility_matrix%2FnRF52840_ic_rev_comp_with_dev_hw.html&cp=2_0_2_3

sjanc:
also, it would be nice to keep this backward compatible for release or two (by using transient package feature of new newt)  if possible

andrzej-kaczmarek:
rebased to resolve conflicts

PR #1350, opened 2018-08-22
helpers for exporting manufacturing time data via sys/confg

mkiiskila:
sys/id; added device model/manufacturer fields under config branch id/
sys/config/conf_mmap; new package making it easy to export memory mapped structured data as config (e.g. from NRF UICR, STM32 OTP region etc.)

### DIFF
--- a/sys/config/conf_mmap/include/conf_mmap/conf_mmap.h
+++ b/sys/config/conf_mmap/include/conf_mmap/conf_mmap.h
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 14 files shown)


sys/config/conf_mmap/include/conf_mmap/conf_mmap.h
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * conf_mmap_src(&my_static_conf);
+ */
+
+/**
+ * Config key value mapping declaration.
+ */
+struct conf_mmap_kv {
+    const char *cmk_key;  /* key (string) */
+    uint16_t cmk_off;     /* offset from conf_mmap.cm_base */
```
ccollins476ad: Nitpick: this comment might be more clear if it were "offset of value from conf_mmap.cm_base".

PR #1353, opened 2018-08-24
Ucifi changes

paulfdietrich:
Some changes to mynewt core to prepare for Ucifi open source stack.

### DIFF
--- a/sys/console/full/src/uart_console.c
+++ b/sys/console/full/src/uart_console.c
@@ -30,8 +30,8 @@
 #include "console_priv.h"
 
 struct console_ring {
-    uint8_t head;
-    uint8_t tail;
+    uint16_t head;
+    uint16_t tail;
     uint16_t size;
     uint8_t *buf;
 };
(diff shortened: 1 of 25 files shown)


ccollins476ad:
I have the following comments:

*lwIP static routing addon:* As I mentioned, this library was rejected when I attempted to merge it into mynewt-core earlier.  The suggestion was write our own add on.  

*tu_time_advance()*: This commit contains a hack to make things work (`extern struct os_callout_list g_callout_list;` in a source file).  I would want to do this properly before merging it.

Aside from the above, it looks good to me.  Do you want to split this PR up so that parts can be merged?

sys/console/full/src/uart_console.c
```diff
@@ -68,6 +68,7 @@ uart_console_ring_pull_char(struct console_ring *cr)
     uint8_t ch;
 
     ch = cr->buf[cr->tail];
+    cr->buf[cr->tail] = 0xee;
```
ccollins476ad: Does `0xee` get written here just as a debugging aid?

paulfdietrich: oops. that was for debug. I can remove this.

ccollins476ad:
I don't believe Mynewt-ucifi work is ongoing.  Please reopen if necessary.

PR #1354, opened 2018-08-27
net/nimble: Change packages to transient

andrzej-kaczmarek:
These packages shall be removed before next release, but since we now
have transient packages let's use them for now so people see warnings
and can finally switch to proper packages.

Note that this requires latest 'newt' to work properly.


andrzej-kaczmarek:
@ccollins476ad we can add transient packages for apps which were moved to NimBLE repo and keep them for next release

PR #1363, opened 2018-08-29
sensor_shell: Use sensor_data_func for sensor_read() instead of the listener

vrahane:


hw/sensor/src/sensor_shell.c
```diff
@@ -591,7 +572,7 @@ sensor_cmd_read(char *name, sensor_type_t type, struct sensor_poll_data *spd)
 err:
     os_sem_release(&g_sensor_shell_sem);
```
ccollins476ad: I don't think this call to `os_sem_release()` is necessary (I know this was here before your change).

PR #1048, opened 2018-04-23
Test travis integration

utzig:
WIP - DO NOT MERGE

PR #1366, opened 2018-08-30
[RFC] Add Travis CI integration

michal-narajowski:
What Travis CI will do:

- `newt test all -e net/oic/test,net/ip/mn_socket/test` (These tests fail on Ubuntu 14.04 which travis uses. Help is welcome)
- Build blinky app on every bsp Mynewt supports (a list is hardcoded in a script. Some bsps require specific toolchains, so I disabled them for now. Help is welcome)
- `newt build all` on simple targets prepared for all apps in core and nimble. Most of these use nrf52840pdk bsp and a few use other bsps if required to build. Bletest app was excluded as it is very outdated and doesn't compile.

All these tests will performed on both Linux and OSX. 

I created a separate repository that contains some utility scripts and targets for Travis to use for testing.
https://github.com/michal-narajowski/mynewt-travis-ci
Some of these scripts were copied from @kaidokert. I hope he will share them under Apache License 2.0 

Original PR https://github.com/apache/mynewt-core/pull/1048 from @utzig that started the initiative. There are some good ideas about extensions to these tests.

I think there is room for improvement, as this is my first time using Travis CI.
Comments are welcome.


utzig:
The directory navigation would be simpler (IMO) like this:

`- pushd repos/apache-mynewt-core && git checkout master && popd`
`- pushd repos/apache-mynewt-nimble && git checkout master && popd`

Adding a `set -e` at the beginning of `script:` might be good.

Why `gcc5` for osx instead of just `gcc`? Also `michal-narajowski/mynewt-travis-ci` could probably move to `runtimeco/mynewt-travis-ci` or some other shared org.

michal-narajowski:
Thanks, didn't know about pushd/popd. 

Honestly, i didn't set osx to use gcc5. I took the setup for osx from @kaidokert and it worked when I tested. If you have an idea how better configure OSX or Ubuntu, then let me know.

+1 about moving `mynewt-travis-ci` to shared org.

utzig:
This looks good to start, I would suggest moving the personal repo, updating references and then I would approve and then we can do PRs for updates later

michal-narajowski:
I want to wait, till @kaidokert updates his scripts with Apache 2.0.

michal-narajowski:
@utzig about gcc5:

```
newt build -q -l info all
Error: exec: "gcc-5": executable file not found in $PATH
```

utzig:
@michal-narajowski 

And does it fail when you remove

> compiler.path.cc.DARWIN.OVERWRITE: "gcc-5"
> compiler.path.as.DARWIN.OVERWRITE: "gcc-5"

from `sim/compiler.yml`?

PR #1369, opened 2018-08-30
 kernel/os; add hal watchdog monitor

mkiiskila:
...which tries to dump system state before HAL watchdog fires.


utzig:
And who watches the watchdog monitor? ;-)

I think this makes at least some sense, in that an app might use a timer which might be on its way to trigger but watchdog gets in first, and the boot loader might not be "able" to handle the timer interrupt (I had such a problem in the past).

mkiiskila:
Adding a check to make sure watchdog timer is not too small is a good idea. I made it such that watchdog timeout should be at least 4 seconds. Allowing a little bit of breathing room before watchdog monitor kicks the system.

PR #1373, opened 2018-08-31
[DO NOT MERGE] I2C retries - HAL

ccollins476ad:
(dev list discussion: https://lists.apache.org/thread.html/bd2b854cbf7d6e0811fff33177c85fa1435dff7f1fb3c8baab6061e0@%3Cdev.mynewt.apache.org%3E)

This PR implements I2C retries at the "HAL level". This is an alternative to the "driver level" implementation here: #1375 

This PR contains the following changes:
1. Define a common set of error code for the I2C HAL.
2. Add a single member to the `hal_i2c_master_data` struct: `retries`.
3. Modify the nRFxxx HAL I2C implementation so that it retries up to `retries` times when an unexpected NACK is received.
4. Update the drivers in `apache-mynewt-core` to use I2C retries.

*Note:* The only HAL implementation modified by this PR is nRFxxx.  This PR does not break the other implementations, but they won't perform any retries until they are updated separately.

### DIFF
--- a/hw/drivers/bq27z561/src/bq27z561.c
+++ b/hw/drivers/bq27z561/src/bq27z561.c
@@ -133,6 +133,7 @@ bq27z561_rd_std_reg_byte(struct bq27z561 *dev, uint8_t reg, uint8_t *val)
     struct hal_i2c_master_data i2c;
 
     i2c.address = dev->bq27_itf.itf_addr;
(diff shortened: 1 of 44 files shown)


ccollins476ad:
Closing this in favor of the driver implementation (#1375).

PR #1375, opened 2018-08-31
I2C retries

ccollins476ad:
(dev list discussion: https://lists.apache.org/thread.html/bd2b854cbf7d6e0811fff33177c85fa1435dff7f1fb3c8baab6061e0@%3Cdev.mynewt.apache.org%3E)

This PR implements I2C retries at the "driver level".  This is an alternative to the "HAL level" implementation here: https://github.com/apache/mynewt-core/pull/1373

This PR contains the following changes:

1. Define a common set of error code for the I2C HAL.
2. Modify the nRFxxx HAL I2C implementation so that it uses the new set of error codes.
3. Define a new library: `hw/drivers/i2c/i2cn` (I2C-n, as in "n retries").
4. Update the drivers in `apache-mynewt-core` to use i2cn with retries.

*Note:* The only HAL implementation modified by this PR is nRFxxx. This PR does not break the other implementations, but they are continue to return inconsistent error codes.

### DIFF
--- a/hw/drivers/bq27z561/pkg.yml
+++ b/hw/drivers/bq27z561/pkg.yml
@@ -28,6 +28,7 @@ pkg.deps:
     - "@apache-mynewt-core/kernel/os"
     - "@apache-mynewt-core/hw/hal"
     - "@apache-mynewt-core/hw/battery"
+    - "@apache-mynewt-core/hw/util/i2cn"
     - "@apache-mynewt-core/sys/log/modlog"
 
 pkg.req_apis:
--- a/hw/drivers/bq27z561/src/bq27z561.c
+++ b/hw/drivers/bq27z561/src/bq27z561.c
@@ -25,6 +25,7 @@
 #include "bq27z561/bq27z561.h"
 #include "hal/hal_gpio.h"
 #include "hal/hal_i2c.h"
+#include "i2cn/i2cn.h"
 
 #include "battery/battery_prop.h"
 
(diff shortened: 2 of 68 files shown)


hw/drivers/i2c/i2cn/include/i2cn/i2cn.h
```diff
@@ -0,0 +1,52 @@
...
+ * failure.
+ *
+ * @param i2c_num               The index of the I2C interface to read from.
+ * @param pdata                 Additional parameters describing the read
+ *                                  operation.
+ * @param timeout               The time, in OS ticks, to wait for the MCU to
+ *                                  indicate completion of each clocked byte.
+ * @param last_op               1 if this is the final message in the
+ *                                  transaction.
+ *
```
gstein: might want to document "retries" :)

andrzej-kaczmarek:
This looks good to me and I prefer these helpers over extending HAL.
Perhaps you may just consider moving this helper package to `hw/drivers/util/i2cn` so `hw/drivers/i2c` "namespace" can be used for I2C driver layer with bus arbitration etc. (something I mentioned on dev list - I can prepare more concrete proposal and send for discussion). But since this is separate directory and package I think it can live in its current place as well.

ccollins476ad:
@andrzej-kaczmarek, that sounds reasonable.  Something that has been bugging me is that this "I2C-with-retries" thing really doesn't feel like a driver to me.  So I am wondering if it should even go in the `drivers` directory at all.  Would `hw/util/i2cn` be too top-level and intrusive?

ccollins476ad:
I have moved the library to `hw/util/i2cn`.

PR #1377, opened 2018-09-03
hw/sensor: Disable OIC sensor server by default

andrzej-kaczmarek:
It requires OIC support to be enabled so usually it needs to be anyway
manually disabled in app/target.

### DIFF
--- a/hw/sensor/syscfg.yml
+++ b/hw/sensor/syscfg.yml
@@ -25,7 +25,7 @@ syscfg.defs:
 
     SENSOR_OIC:
         description: 'Enable/Disable the OIC sensor server'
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Yes please. Im manually disabling this in all my targets and it annoys me when I see it every time :)

vrahane:
Actually, there was a reason this was enabled by default. We wanted the mynewt sensor android app to work out of box with any device that uses sensors_test app. Hence, the default ON behavior. Is there a reason for disabling this ? Maybe @sterlinghughes and @aditihilbert can chime in since it was their ask in the first place.

PR #1390, opened 2018-09-07
sys/console/full: Prevent permanent rx stall

ccollins476ad:
When the console receive buffer becomes full, the UART enters a "stalled" state.  While stalled, the UART ceases to receive bytes.  This makes sense - there is no space for any bytes if it kept receiving them.

The problem is that nothing unstalls the UART after the console buffer is drained.  Consequently, the UART is stuck in the stalled state and never receives any more data.

This PR fixes this problem as follows:
1. In the console receive event callback, check if the buffer is full before doing any processing.
2. If the buffer was full, restart the UART before returning.


sys/console/full/src/uart_console.c
```diff
@@ -230,19 +230,16 @@ uart_console_rx_char_event(struct os_event *ev)
...
+    bool full;
+
+    full = uart_console_ring_is_full(&cr_rx);
 
     /* We may have unhandled character - try it first */
     if (b >= 0) {
         ret = console_handle_char(b);
         if (ret < 0) {
-            return;
+            goto done;
```
andrzej-kaczmarek: I think in this case it does not make sense to restart RX since we did not pull any character from ring buffer so RX will be stalled again.

sys/console/full/src/uart_console.c
```diff
@@ -252,11 +249,20 @@ uart_console_rx_char_event(struct os_event *ev)
...
     }
 
     b = -1;
+
+done:
+    /*
+     * It is possible that UART RX was stalled in the meantime since we were
+     * not consuming data from ring buffer - make sure RX is started.
+     */
+    if (full) {
```
andrzej-kaczmarek: Perhaps to make it a bit smarter we can remember that RX was stalled in `uart_console_rx_char` (line 211) and only restart in such case. I'm not 100% sure if this won't create some race though...

Also, restart can be done after leaving critical section above (line 249) since we already have free space in ring buffer and can accept new data immediately.

sys/console/full/src/uart_console.c
ccollins476ad: Good point.

ccollins476ad:
Thanks, Andrzej.  I have implemented all of your suggestions.  Please take another look if you get a chance!

PR #1392, opened 2018-09-11
bugfix/c++ global constructor

mlampert:
By overwriting `__libc_init_array` the C++ constructors of global variables are not being called, leaving those unitialized.

This patch leaves the compiler generated version of `__libc_init_array` in place which properly calls all constructors of global variables and implements `_init` instead.

### DIFF
--- a/libc/baselibc/src/start.c
+++ b/libc/baselibc/src/start.c
@@ -39,6 +39,6 @@ void _start(void)
 }
 
 void
-__libc_init_array(void)
+_init(void)
 {
 }


utzig:
I don't think the compiler generates `__libc_init_array`, does it? It seems to come from newlib (or some other libc) which most cross toolchains link against. This will probably work with most targets, but would it work on any microcontroller/toolchain? Maybe implementing our own version in `baselibc` would be better?

kasjer:
I guess that empty function was added just to provided liker symbol (for some BSP) not for c++ support.
I we drop this function and hope for newlib implementation some people can face linking problem when they will not use newlib.
We could leave empty function as it is (for some bsp sake) and add new initialization function that would be called at the beginning of main (this is the way it is handled in zephyr)

utzig:
Another option: add a `syscfg` to `libc/baselibc` with a name like `BASELIBC_DEFINE_INIT_ARRAY` (sorry for the lack of inspiration!) with a default of `1` and override on BSPs that need C++ constructors.

mlampert:
This backwards though, the function `__libc_init_array`  has to be called. It is how all global constructors are called. Replacing it with an empty implementation means breaking support for C++.

If there are systems where the toolchain doesn't have a `__libc_init_array` then the startup code in the BSP shouldn't call it and use whatever their system provides in order to initialize the globals.

mlampert:
@utzig  Sorry, didn't see your suggestion of a syscfg variable - shouldn't the default be the other way around? I mean C++ is currently broken in mynewt for all gcc users.

utzig:
> shouldn't the default be the other way around? I mean C++ is currently broken in mynewt for all gcc users.

Sure, that might be better. Would be good to tell users of this change (mips and pic32 ports primarily?) through the mailing list. Which can be done later...

mkiiskila:
MCU startup code does various things. Calling __libc_init_array() would be done by BSPs/MCUs which expect to link against newlib, as that's who has this as the C++ global constructor hook.
libc_init_array() implementation also expects things from the linker script, so you might want to be on lookout for those. Note that some (most?) startup code in mynewt does not call this function.

The most common tool installation probably has happened using the instructions we have on the mynewt pages for getting the compiler for ARM. This installation comes with gcc which has newlib available also. So any symbol not found from baselibc, and referenced by rest of the code, is brought in from newlib.

I'm ok bringing this in; people whose build this will break are the ones who have startup code which calls __libc_init_array(), but who don't have newlib available for linking.
I.e. who built gcc without newlib, using BSP which uses startup code which calls this. So there might be some broken builds as a result.
PIC32 should be fine, as it doesn't call libc_init_array.

PR #1394, opened 2018-09-12
hw/drivers/lis2dh12: Add sensor notifications support to LIS2DH12 driver

andrzej-kaczmarek:
Kudos go to @kasjer - I just made few small fixes in few places and tested this :-)

### DIFF
--- a/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
+++ b/hw/drivers/sensors/lis2dh12/src/lis2dh12.c
@@ -34,6 +34,80 @@
 #include "modlog/modlog.h"
 #include "stats/stats.h"
 #include <syscfg/syscfg.h>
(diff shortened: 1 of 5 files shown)


hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -34,6 +34,80 @@
 #include "modlog/modlog.h"
 #include "stats/stats.h"
 #include <syscfg/syscfg.h>
+#include <sensor/sensor.h>
+#include <console/console.h>
```
vrahane: `#include <console/console.h>` should be removed

vrahane:
This was merged as part of https://github.com/apache/mynewt-core/pull/1520

PR #1395, opened 2018-09-13
os_arch cleanup

andrzej-kaczmarek:
There seems to be a lot of common code in `os_arch.h` files which can be moved to common header and thus make process of updating easier (if necessary). Especially API declarations should be kept in single place as otherwise they may be out of sync... and actually there are some APIs declared which are not defined/used anymore anywhere so they are also removed.


andrzej-kaczmarek:
@utzig Thanks. I'm not sure about PIC32 change as well so I removed it from this PR and will push as separate PR.

PR #1399, opened 2018-09-14
RTT console fixes

andrzej-kaczmarek:
This fixes 2 problems with RTT console:
1. output data may be corrupted
2. output data may be skipped when writing lots of data at once to RTT

1st problem is because we write data using function which can overwrite existing data in buffer - this should not be used when host is reading data since both sides may modify write pointer without sync (thus data corruption). To resolve this we use proper function which skips data instead of overwriting - this is safe.
2nd problem is because we can write data to RTT quickly, but host reads data by polling periodically thus it may not start reading data when we fill buffer. To solve this we will now wait a bit when buffer is full and retry which gives host some time to start reading data. Retry number and delay is configurable. Also we will not retry if detected that host is not reading data at all.

### DIFF
--- a/sys/console/full/src/rtt_console.c
+++ b/sys/console/full/src/rtt_console.c
@@ -35,20 +35,25 @@ static const char CR = '\r';
 int
 console_out(int character)
 {
+    os_sr_t sr;
     char c = (char)character;
 
     if (g_silence_console) {
         return c;
     }
 
+    OS_ENTER_CRITICAL(sr);
+
     if ('\n' == c) {
-        SEGGER_RTT_WriteWithOverwriteNoLock(0, &CR, 1);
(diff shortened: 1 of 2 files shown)


sys/console/full/src/rtt_console.c
```diff
@@ -35,20 +35,25 @@ static const char CR = '\r';
...
 
     if (g_silence_console) {
         return c;
     }
 
+    OS_ENTER_CRITICAL(sr);
+
     if ('\n' == c) {
-        SEGGER_RTT_WriteWithOverwriteNoLock(0, &CR, 1);
+        SEGGER_RTT_WriteNoLock(0, &c, 1);
```
rymanluk: &CR should be there instead of &c

andrzej-kaczmarek: good catch, fixed

PR #1400, opened 2018-09-14
[DO NOT MERGE] - Fix BQ27Z561 and LP5523 I2C repeated-stop operations

benmccrea:
The main part of this PR is a possible fix for an I2C failure in our nRF52 system which hangs the bus and keeps SDA low indefinitely. The failure appears to be triggered when doing a read of bq27561 or lp5523 registers, if the write operation (first part of a write-read) fails and no STOP condition gets generated. The condition can be very difficult to reproduce, though it happens with some regularity if certain specific conditions are met which are too detailed to go into here. 

The solution applied by this PR is to allow the following I2C read and STOP condition to go through even though the write has failed. 

This fix is preliminary and testing is on going, though I want to submit it now for review and comments. A more thorough solution might be found by modifying hal_i2c_master_write in the nrf52xxx HAL to place a STOP on the bus if the write operation fails when the >last_op< argument is 0.

Also, the bq27z561 i2c timeout has been reduced from 1 second to 16 milliseconds, as the long timeout was holding up processing of default queue events.

### DIFF
--- a/hw/drivers/bq27z561/src/bq27z561.c
+++ b/hw/drivers/bq27z561/src/bq27z561.c
@@ -41,6 +41,9 @@
 #define BQ27Z561_LOG(lvl_, ...)
 #endif
 
+#define BQ27Z561_I2C_TIMEOUT_TICKS \
+    ((1000 / OS_TICKS_PER_SEC) / MYNEWT_VAL(BQ27Z561_I2C_TIMEOUT_MS))
+
 static uint8_t
 bq27z561_calc_chksum(uint8_t *tmpbuf, uint8_t len)
 {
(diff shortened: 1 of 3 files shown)


hw/drivers/bq27z561/src/bq27z561.c
```diff
@@ -41,6 +41,9 @@
 #define BQ27Z561_LOG(lvl_, ...)
 #endif
 
+#define BQ27Z561_I2C_TIMEOUT_TICKS \
+    ((1000 / OS_TICKS_PER_SEC) / MYNEWT_VAL(BQ27Z561_I2C_TIMEOUT_MS))
```
ccollins476ad: Is this calculation correct?  Shouldn't it be
```
MYNEWT_VAL(BQ27Z561_I2C_TIMEOUT_MS) * 1000 / OS_TICKS_PER_SEC
```

?

benmccrea:
@ccollins476ad I looked further into this and I think my changes in this PR are just a workaround for a bug in `hal_i2c_master_write`. In the case where `last_op` is 0 and an error takes place, `EVENTS_STOPPED=0` doesn't ever happen, and I think that results in the bus being left without a stop condition.

ccollins476ad:
@benmccrea, I see what you mean, and I think that is definitely worth trying.

benmccrea:
Closing this in favor of opening a new PR with a more thorough fix.

PR #1401, opened 2018-09-15
On Newt Load, use the specified port if one is given in --extrajtagcmd options

JerrodAnderson:
If -port is specified, use that port instead.

When using 'newt load', if '--extrajtagcmd "-port 3334"' is used, the load will hang.  The hardcoded 3333 port needs to match JLink's port.
In this case, the command looks like:

Command line: -device nRF52 -speed 1000 -if SWD -port 3333 -singlerun -select USB=123456789 -port 3334

JLink uses the second declared port of 3334, while newt is listening to 3333.

Note: Port needs to be unique for parallel execution.  In the case of multiple boards attached to the computer and code that could potentially try to upload to different boards at the same time, this change is necessary.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,11 +1,11 @@
 # RELEASE NOTES
 
-04 June 2018 - Apache Mynewt v1.4.0
+21 June 2018 - Apache Mynewt v1.4.1
 
 For full release notes, please visit the
 [Apache Mynewt Wiki](https://cwiki.apache.org/confluence/display/MYNEWT/Release+Notes).
 
-This is the seventh source release of Apache Mynewt Core. For more
+This is the first bugfix source release of Apache Mynewt Core 1.4. For more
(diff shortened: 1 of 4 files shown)


JerrodAnderson:
My first PR - I'm not sure what the other 3 commits are.  Sorry if I was supposed to do something with those.

andrzej-kaczmarek:
@JerrodAnderson you created patch on 1.4 branch but PR is created on `master` so these extra commits were also pulled from 1.4 branch. you need to rebase your patch on `master` and update PR with force push (or create new one).

mkiiskila:
I'd recommend creating a function out of the digging the PORT out of EXTRA_JTAG_CMD.
Then we could use it from jlink_debug(), as well as from openocd_load()/_debug().

mkiiskila:
Changes in this PR were incorporated as part of PR#1453. Closing this one.

PR #1402, opened 2018-09-17
kernel/os: Add os_arch_in_isr() API

andrzej-kaczmarek:
Added implementation for Cortex-M.
MIPS and PIC32 already had local implementation so just moved this to public header.
RISC-V is missing.


utzig:
Added other port maintainers for suggestions!

andrzej-kaczmarek:
can we merge this since there are apparently no suggestions? :-)

PR #1406, opened 2018-09-18
hw/hal; hal_flash_isempty() also returns data from the location.

mkiiskila:
sys/flash_map; flash_area_isempty_at() also returns data from the location.

Addresses issue @utzig mentioned before. Common case for checking whether flash area is empty is followed by checking what kind of data is there.


mkiiskila:
I feel the same, the function name could be improved.

utzig:
Looks much better with the new name, please merge!

PR #1410, opened 2018-09-20
Log statistics

mkiiskila:
Adds optional statistics per-log. Stat block name comes from the name of the log.
Statistics are kept for number of log entries written, logs filtered out because of log level
settings, log entries dropped due to log collection not fetching (clearing) them on time, and errors while writing.

Note that log entries dropped from cbmem are not counted yet.


mkiiskila:
Rebased and fixed conflicts.

PR #1417, opened 2018-09-21
sys/reboot: Add info to log entries and limit writes to one per boot

ccollins476ad:
This PR introduces two changes to the reboot log.

#### 1. Add info to reboot log entries

Prior to this commit, a reboot entry contained the following information:

* reason (`rsn`)
* count (`cnt`)
* image version (`img`)

This commit adds the following:

* image hash (`hash`)
* file:line, if available
* value of PC register, if available

#### 2. Limit log writes to one per boot

Prior to this PR, some crashes would result in two entries being written to the reboot log.  For example, on a failed assert, an entry gets logged with the die file and die line.  `__assert_func()` then triggers a nonmaskable interrupt in order to generate a core, which results in a second entry being written to the reboot log.

Finally, after resetting, an additional entry would often (always?) get written to the log.

This PR changes the reboot log semantics as follows:
* When the system resets or crashes, a single entry gets written.  Subsequent faults do not generate an additional log entry.
* Upon booting, the system only records a new log entry if it detects that it did not write one for the previous reset.

### DIFF
--- a/kernel/os/pkg.yml
+++ b/kernel/os/pkg.yml
@@ -43,5 +43,8 @@ pkg.deps.BSP_SIMULATED:
 pkg.deps.OS_SYSVIEW:
     - "@apache-mynewt-core/sys/sysview"
 
+pkg.deps.OS_CRASH_LOG:
+    - "@apache-mynewt-core/sys/reboot"
+
 pkg.init:
     os_pkg_init: 0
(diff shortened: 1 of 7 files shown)


ccollins476ad:
The failure reported by travis as since been fixed.

ccollins476ad:
Yeah, I know what you mean.  Calling `snprintf` in a loop like that is certainly not the most efficient thing in the world.  I think it yields better code size than converting the hash to a string by hand, though, and speed probably doesn't matter in this case since this code is only executed once on reboot.

PR #1423, opened 2018-09-24
Fix double read on enc_flash driver

utzig:
- Update flash_area_is_empty to use the new hal_flash_isempty_no_buf call
- When enc_flash is testing for an empty flash, now it checks that the lower-level driver returned an empty flash already and avoids reading data that when it is known to be empty.


utzig:
@mkiiskila @ccollins476ad This is a fix for the blocker I mentioned before in pvt. This works for me with mcuboot updating images over enc_flash without issues (with local changes for now!). I don't know if this change breaks something else.

PR #1425, opened 2018-09-24
Don't include os/mynewt.h on apollo2

utzig:


utzig:
Already passed on travis-ci, thanks @andrzej-kaczmarek!

PR #1431, opened 2018-09-25
Reduce I2C timeout values to reduce system latency when failures take place

benmccrea:
Most I2C devices will respond in well under 1ms but currently the timeout in most drivers is specified as OS_TICKS_PER_SEC/10. This causes undue latency in recovering from I2C errors. This change adds syscfg to some drivers to specify a smaller I2C timeout of 3 OS ticks. 

Also changes BQ27Z561_LOG default value to 1, since the original default value of 0 can cause I2C errors on the device to go unnoticed.


benmccrea:
@vrahane The gas gauge has the worst clock stretching and from what I've seen it results in around 1ms of latency per I2C operation. I set it's syscfg timeout at 6 ticks just in case.

PR #1432, opened 2018-09-26
Add enable/disable SAADC channel to nordic SDK

wes3:
The current nordic SDK does not allow the enabling/disabling of a SAADC channel with calling nrfx_channel_init() and nrfx_channel_uninit(). This means that to enable/disable a channel you need to reconfigure the channel each time. These helper functions allow one to configure a SAADC channel and leave it with the configuration if it does not need to be reconfigured (which is probably a more common case).

Note that this modification was only done for any project that sets the  BSP_NRF52 syscfg value. This might be somewhat arbitrary and we might consider doing this for all NRF52 related BSP's (52840, for example).

### DIFF
--- a/hw/mcu/nordic/src/ext/nrfx/drivers/src/nrfx_saadc.c
+++ b/hw/mcu/nordic/src/ext/nrfx/drivers/src/nrfx_saadc.c
@@ -1,21 +1,21 @@
 /**
  * Copyright (c) 2015 - 2018, Nordic Semiconductor ASA
  * All rights reserved.
- * 
+ *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice, this
  *    list of conditions and the following disclaimer.
- * 
+ *
  * 2. Redistributions in binary form must reproduce the above copyright
  *    notice, this list of conditions and the following disclaimer in the
  *    documentation and/or other materials provided with the distribution.
- * 
+ *
  * 3. Neither the name of the copyright holder nor the names of its
  *    contributors may be used to endorse or promote products derived from this
  *    software without specific prior written permission.
- * 
+ *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
  * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
  * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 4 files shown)


sterlinghughes:
Cc @carlescufi -- are we submitting a patch back to the nrfx folks?

hw/mcu/nordic/src/ext/nrfx/drivers/src/nrfx_saadc.c
```diff
@@ -627,4 +627,20 @@ void nrfx_saadc_limits_set(uint8_t channel, int16_t limit_low, int16_t limit_hig
         nrf_saadc_int_enable(int_mask);
     }
 }
+
+void nrfx_enable_adc_chan(int chan, nrf_saadc_input_t pselp,
+                          nrf_saadc_input_t pseln)
+{
+    NRFX_ASSERT(m_cb.active_channels < NRF_SAADC_CHANNEL_COUNT);
+    ++m_cb.active_channels;
```
ccollins476ad: Should this function verify that the specified channel is disabled before incrementing `m_cb.active_channels`?  I don't know how important this is, but I noticed that the nrfx code performs this check:
```
    if (m_cb.psel[channel].pselp == NRF_SAADC_INPUT_DISABLED) 
    {
        ++m_cb.active_channels;
    }
```

A similar comment applies to the disable case.

andrzej-kaczmarek:
Code looks good, but I think instead of creating local modified copies we should submit PR to nrfx repo and then update from there (we can even update just single file before next version of nrfx is tagged). Otherwise I'm pretty sure we'll forget about this during next update ;-)

carlescufi:
@wes3 @sterlinghughes please submit a PR to the nrfx repo directly:
https://github.com/NordicSemiconductor/nrfx

hw/mcu/nordic/src/ext/nrfx/drivers/src/nrfx_saadc.c
wes3: Hmm... I think it is definitely a worthy comment. I did not consider this when I wrote this code but one thing I did consider was the fact that this function is not thread safe. With these small helper functions I generally do not make them thread safe nor check what you are suggesting as I want to make them as small as possible and the presumption is that the user already knows that the channel is enabled. Thinking about this, I would be more inclined to NRFX_ASSERT() if the channel was not disabled when enabling or enabled when disabling it. I think that would point out some unexpected error. A compromise (so to speak) could be to return an error in that case and let the caller decide if they want to assert but again, this is starting to go against why I wanted to add these in the first place: less time/code spent enabling/disabling a channel by not calling channel_init() and channel_uninit(). I was even considering not checking the # of active channels for sanity when incrementing or decrementing it :-)

carlescufi:
> I did not want to wait for nordic to get this in. Lets get the PR's done and when it gets into an SDK that we import we will remove it. I will be 100% sure to remember it.

Thanks for sending the PR to nrfx.

wes3:
@carlescufi  I am confused. I have not yet submitted the PR to nrfx. Going to do that in a bit.

carlescufi:
@wes3 sorry, I saw a PR about ADC in nrfx and thought it was this one. I now see it was not.

wes3:
@carlescufi No worries Carles. I did attempt to create a PR but it seems I am not allowed to do so. I obviously do not have the correct permissions.

carlescufi:
@wes3 strange, what error did you get?

wes3:
git push origin nrfx_saadc_chan_helpers
remote: Permission to NordicSemiconductor/nrfx.git denied to wes3.
fatal: unable to access 'https://github.com/NordicSemiconductor/nrfx.git/': The requested URL returned error: 403

wes3:
@carlescufi I may have cloned this incorrectly. Let me check this out.

PR #1434, opened 2018-09-27
Remove gcc-multilib dependency

utzig:
The gcc-multilib was bringing in full gcc-4.8 (including g++-4.8), so remove it and add kernel headers dep manually.

Also removed `gcc-7-multlib` from ARM target tests.


utzig:
Depends on: https://github.com/runtimeco/mynewt-travis-ci/pull/9

PR #1439, opened 2018-10-03
test/runtest: Null-terminate segments of log entry

ccollins476ad:
Overly long segments (test case name, suite name, and message) were not being null-terminated.  This caused malformed JSON to be included in the log.

### DIFF
--- a/compiler/arm-none-eabi-m4/compiler.yml
+++ b/compiler/arm-none-eabi-m4/compiler.yml
@@ -25,11 +25,11 @@ compiler.path.objdump: arm-none-eabi-objdump
 compiler.path.objsize: arm-none-eabi-size
 compiler.path.objcopy: arm-none-eabi-objcopy
 
-compiler.flags.base: -mcpu=cortex-m4 -mthumb-interwork -mthumb -Wall -Werror -fno-exceptions -ffunction-sections -fdata-sections
-compiler.flags.default: [compiler.flags.base, -O1 -ggdb]
-compiler.flags.optimized: [compiler.flags.base, -Os -ggdb]
-compiler.flags.speed: [compiler.flags.base, -O3 -ggdb]
-compiler.flags.debug: [compiler.flags.base, -Og -ggdb]
+compiler.flags.base: [-mcpu=cortex-m4, -mthumb-interwork, -mthumb, -Wall, -Werror, -fno-exceptions, -ffunction-sections, -fdata-sections]
+compiler.flags.default: [compiler.flags.base, -O1, -ggdb]
+compiler.flags.optimized: [compiler.flags.base, -Os, -ggdb]
+compiler.flags.speed: [compiler.flags.base, -O3, -ggdb]
+compiler.flags.debug: [compiler.flags.base, -Og, -ggdb]
 
 compiler.as.flags: [-x, assembler-with-cpp]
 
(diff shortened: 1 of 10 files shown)


test/runtest/src/runtest.c
```diff
@@ -165,8 +165,10 @@ runtest_log_result(const char *msg, bool passed)
     int m_len;
     int len;
 
-    /* str length of {"k":"","n":"","s":"","m":"","r":1}<token> */
-    len = 35 + strlen(runtest_token);
+    /* str length of {"k":"","n":"","s":"","m":"","r":1}<token> plus three
+     * null-terminators.
+     */
+    len = sizeof 38 + strlen(runtest_token);
```
andrzej-kaczmarek: I guess 38 is 35+3 ("plus three null-terminators"), but what 'sizeof` does here? if this is correct, then I guess it should be written in less cryptic way :)

compiler/arm-none-eabi-m4/compiler.yml
```diff
@@ -26,9 +26,9 @@ compiler.path.objsize: arm-none-eabi-size
 compiler.path.objcopy: arm-none-eabi-objcopy
 
 compiler.flags.base: -mcpu=cortex-m4 -mthumb-interwork -mthumb -Wall -Werror -fno-exceptions -ffunction-sections -fdata-sections
-compiler.flags.default: [compiler.flags.base, -O1 -ggdb]
-compiler.flags.optimized: [compiler.flags.base, -Os -ggdb]
-compiler.flags.debug: [compiler.flags.base, -Og -ggdb]
+compiler.flags.default: [compiler.flags.base, -O1, -ggdb]
+compiler.flags.optimized: [compiler.flags.base, -Os, -ggdb]
+compiler.flags.debug: [compiler.flags.base, -Og, -ggdb]
```
utzig: Why must this be done? And why not do the same for m0/m3/m7/riscv?

test/runtest/src/runtest.c
ccollins476ad: Jeeze... no that is not correct :).  That is a leftover from an earlier fix that I didn't fully remove.  I have fixed it (removed `sizeof`).

compiler/arm-none-eabi-m4/compiler.yml
ccollins476ad: I mentioned this in the commit message, though admittedly I snuck it into this PR where it doesn't really belong.  It is better to separate the compiler flags because then newt can override the optimization level of a single package.  With the flags combined into a single item, newt doesn't recognize any of them as the "-O" flag.

Yes, we should make this same change to the other compiler definitions.

test/runtest/src/runtest.c
```diff
@@ -165,34 +165,39 @@ runtest_log_result(const char *msg, bool passed)
...
+    /* str length of {"k":"","n":"","s":"","m":"","r":1}<token> plus three
+     * null-terminators.
+     */
+    len = 38 + strlen(runtest_token);
 
     /* How much of the test name can we log? */
     n_len = strlen(tu_case_name);
-    if (len + n_len >= LOG_PRINTF_MAX_ENTRY_LEN) {
-        n_len = LOG_PRINTF_MAX_ENTRY_LEN - len - 1;
+    if (len + n_len >= MYNEWT_VAL(MODLOG_MAX_PRINTF_LEN)) {
```
mkiiskila: I would've added a #define MODLOG_MAX_PRINTF_LEN MYNEWT_VAL(MODLOG_MAX_PRINTF_LEN) in the beginning of the file to shorten these lines :) But this is ok.

PR #1444, opened 2018-10-04
spiflash: Add support for multiple chips

kasjer:
SpiFlash chips are often compatible to some extent even
when they come from different vendors.
It is quite possible to have same product that will use
different spiflash even in production.
Now it is possible to specify more than one flash chip.

### DIFF
--- a/hw/drivers/flash/spiflash/src/spiflash.c
+++ b/hw/drivers/flash/spiflash/src/spiflash.c
@@ -28,18 +28,6 @@
 
 #if MYNEWT_VAL(SPIFLASH)
 
-#if MYNEWT_VAL(SPIFLASH_MANUFACTURER) == 0
-#error SPIFLASH_MANUFACTURER must be set to the correct value in bsp syscfg.yml
-#endif
-
-#if MYNEWT_VAL(SPIFLASH_MEMORY_TYPE) == 0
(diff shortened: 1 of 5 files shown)


hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -319,14 +841,93 @@ spiflash_sector_info(const struct hal_flash *hal_flash_dev, int idx,
...
+                     * Try another release power down method if found.
+                     */
+                    continue;
+                }
+                /* Something was read from flash, do not try another power up
+                 * just check if chip should be supported */
+                break;
+            }
+        }
+        for (i = 0; supported_chips[i].fc_jedec_id.ji_manufacturer != 0; ++i) {
```
utzig: So this `for` loop is reusing `i` from the external `for` loop with the same test, which is not really clear to me why it's being done. Would you mind adding a comment saying why this `supported_chips[i].fc_jedec_id.ji_manufacturer` is being used again here just to make it clearer? (unless it was not supposed to be this way!)

utzig: Oh, this is outside, got it, oops!

kasjer: There is no external loop just two loops with i.
First loop finds unique ways to wake up chips and tries to read jedec id after each try.
Second loop checks if chip is supported.

PR #1445, opened 2018-10-08
adxl345 accelerometer fixes

kasjer:
couple of issues were detected when accelerometer was added to project

- compile error in adxl345.c (due to log_register call that should be removed)
- compile error in sensor creator due to missing function end

- alternative address can be selected for accelerometer in syscfg

- shell dump command had some mismatched prints
- shell dump command was reading registers in assert call that could be compiled out and prints would display garbage

### DIFF
--- a/hw/drivers/sensors/adxl345/src/adxl345_shell.c
+++ b/hw/drivers/sensors/adxl345/src/adxl345_shell.c
@@ -135,9 +135,23 @@ adxl345_shell_cmd_read(int argc, char **argv)
     return 0;
 }
 
+static void
+dump_reg(int reg_addr, const char *reg_name, uint8_t val, int rc)
+{
+    if (rc == 0) {
+        console_printf("0x%02X (%s): 0x%02X\n", reg_addr, reg_name, val);
+    } else {
+        console_printf("0x%02X (%s): read failed rc=%d\n",
+            reg_addr, reg_name, rc);
+    }
+}
+
+#define DUMP_REG(reg, value, rc) dump_reg(ADXL345_##reg, #reg, val, rc)
+
 static int
 adxl345_shell_cmd_dump(int argc, char **argv)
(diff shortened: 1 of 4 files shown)


hw/drivers/sensors/adxl345/src/adxl345_shell.c
```diff
@@ -146,67 +147,67 @@ adxl345_shell_cmd_dump(int argc, char **argv)
...
+    rc = adxl345_read8(&g_sensor_itf, ADXL345_DATAY1, &val);
+    console_printf("0x%02X (DATAY1): 0x%02X %s\n", ADXL345_DATAY1, val, rc ? " read failed": "");
+    rc = adxl345_read8(&g_sensor_itf, ADXL345_DATAZ0, &val);
+    console_printf("0x%02X (DATAZ0): 0x%02X %s\n", ADXL345_DATAZ0, val, rc ? " read failed": "");
+    rc = adxl345_read8(&g_sensor_itf, ADXL345_DATAZ1, &val);
+    console_printf("0x%02X (DATAZ1): 0x%02X %s\n", ADXL345_DATAZ1, val, rc ? " read failed": "");
+    rc = adxl345_read8(&g_sensor_itf, ADXL345_FIFO_CTL, &val);
+    console_printf("0x%02X (FIFO_CTL): 0x%02X %s\n", ADXL345_FIFO_CTL, val, rc ? " read failed": "");
+    rc = adxl345_read8(&g_sensor_itf, ADXL345_FIFO_STATUS, &val);
+    console_printf("0x%02X (FIFO_STATUS): 0x%02X %s\n", ADXL345_FIFO_STATUS, val, rc ? " read failed": "");
```
ccollins476ad: It might be more helpful to printing the value of `rc` rather than a "success" / "failure" indication.

PR #1447, opened 2018-10-09
Use generated sysdown() function

ccollins476ad:
NOTE: This commit requires a newt tool update (a38614ebc22959231ad9597c24467e34ed0e5068).

Upon application-triggered restart, execute each package's shutdown subprocedure.  This feature is discussed here:
https://lists.apache.org/thread.html/ebd53c9b56230184650ea57e5534d5bcfbfc170b97547a1a26f100b5@%3Cdev.mynewt.apache.org%3E

### DIFF
--- a/sys/sysdown/src/sysdown.c
+++ b/sys/sysdown/src/sysdown.c
@@ -1,0 +1,162 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "os/mynewt.h"
+
+/**
+ * The NEWT_FEATURE_SYSDOWN setting is injected by new-enough versions of the
+ * newt tool.
+ */
+#if !MYNEWT_VAL(NEWT_FEATURE_SYSDOWN)
+
+int
+sysdown(int reason)
+{
+    os_system_reset();
+    return 0;
+}
+
+#else
+
+#include <assert.h>
+#include <stdio.h>
+#include <stddef.h>
+#include <limits.h>
+#include "hal/hal_system.h"
+#include "hal/hal_watchdog.h"
+
+#define SYSDOWN_TIMEOUT_TICKS   \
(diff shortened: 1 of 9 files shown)


sys/sysdown/src/sysdown.c
```diff
@@ -0,0 +1,169 @@
...
+    sysdown_complete_cb = complete_cb;
+    sysdown_complete_arg = arg;
+
+    os_callout_init(&sysdown_timer, os_eventq_dflt_get(), sysdown_timer_exp,
+                    NULL);
+    rc = os_callout_reset(&sysdown_timer, SYSDOWN_TIMEOUT_TICKS);
+    assert(rc == 0);
+
+    /* Call each configured sysdown callback. */
+    for (i = 0; sysdown_cbs[i] != NULL; i++) {
```
mkiiskila: Could it be sizeof(sysdown_cbs) / sizeof(sysdown_cbs[0])? Would save 4 bytes of text, if sysdown_cbs[] we're not terminated by a NULL...

sys/sysdown/src/sysdown.c
```diff
@@ -0,0 +1,169 @@
...
+            break;
+
+        case SYSDOWN_IN_PROGRESS:
+            OS_ENTER_CRITICAL(sr);
+            sysdown_num_in_progress++;
+            OS_EXIT_CRITICAL(sr);
+            break;
+
+        default:
+            return SYS_EUNKNOWN;
```
mkiiskila: I'm wondering if it would be better to continue calling sysdown() routines, and proceed with the shutdown?

sys/sysdown/src/sysdown.c
ccollins476ad: The reason I null-terminated the array is to handle the case where there aren't any sysdown procedures (a zero-length array being undefined).  gcc does allow zero-length arrays, but I didn't think four bytes' savings justified using an extension.

sys/sysdown/src/sysdown.c
ccollins476ad: Yes, I agree.  It doesn't make sense for sysdown to just stop like this.

sys/sysdown/src/sysdown.c
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * The SYSDOWN setting is injected by new-enough versions of the newt tool.
+ */
+#if !MYNEWT_VAL(SYSDOWN)
+
+int
+sysdown(int reason, sysdown_complete_fn *complete_cb, void *arg)
+{
+    return 0;
```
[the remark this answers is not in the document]
ccollins476ad: Yes, good catch.

sys/sysdown/src/sysdown.c
mkiiskila: That's fair. Agreed.

ccollins476ad:
Thanks, @mkiiskila.  I believe I have addressed all your comments as follows:
1. If newt doesn't support the sysdown feature, `sysdown()` just performs a system reset.
2. If a sysdown subprocedure returns an unknown error code, pretend it returned `SYSDOWN_COMPLETE` (i.e., continue shutting down).

Also, I changed the syscfg setting name from `SYSDOWN` to `NEWT_FEATURE_SYSDOWN`.  I think we can avoid some confusion in the future by clearly identifying settings that newt injected by itself.

PR #1448, opened 2018-10-09
Reduce ram usage in fcb

kasjer:
FCB when used with big flash can use a lot of RAM needed for sector information storage.
Information about sectors can be compacted so FCB will use reasonable amount of RAM.

Function introduced in flash_map can also be used in bootloader to get flash slot information.

### DIFF
--- a/fs/fcb/include/fcb/fcb.h
+++ b/fs/fcb/include/fcb/fcb.h
@@ -36,14 +36,17 @@ extern "C" {
 
 #define FCB_MAX_LEN	(CHAR_MAX | CHAR_MAX << 7) /* Max length of element */
 
+#define FCB_SECTOR_OLDEST UINT16_MAX
+
 /**
- * Entry location is pointer to area (within fcb->f_sectors), and offset
- * within that area.
+ * Entry location point to sector, and offset
+ * within that sector.
  */
 struct fcb_entry {
-    struct flash_area *fe_area;	/* ptr to area within fcb->f_sectors */
-    uint32_t fe_elem_off;	/* start of entry */
-    uint32_t fe_data_off;	/* start of data */
+    struct sector_range *fe_range;  /* ptr to area within fcb->f_ranages */
+    uint16_t fe_sector;     /* sector number in fcb flash */
+    uint16_t fe_elem_off;	/* start of entry in sector */
+    uint16_t fe_data_off;	/* start of data in sector */
     uint16_t fe_data_len;	/* size of data area */
 };
 
@@ -51,16 +54,23 @@ struct fcb {
     /* Caller of fcb_init fills this in */
     uint32_t f_magic;		/* As placed on the disk */
     uint8_t f_version;  	/* Current version number of the data */
-    uint8_t f_sector_cnt;	/* Number of elements in sector array */
     uint8_t f_scratch_cnt;	/* How many sectors should be kept empty */
-    struct flash_area *f_sectors; /* Array of sectors, must be contiguous */
+    uint8_t f_range_cnt;    /* Number of elements in range array */
+    uint16_t f_sector_cnt;	/* Number of sectors used by fcb */
+    struct sector_range *f_ranges;
 
     /* Flash circular buffer internal state */
     struct os_mutex f_mtx;	/* Locking for accessing the FCB data */
-    struct flash_area *f_oldest;
+    uint16_t f_oldest_sec;
     struct fcb_entry f_active;
     uint16_t f_active_id;
     uint8_t f_align;		/* writes to flash have to aligned to this */
+};
+
+struct fcb_sector_info {
+    struct sector_range *si_range;
+    uint32_t si_sector_offset;
+    uint16_t si_sector_in_range;
 };
 
 /**
@@ -109,7 +119,7 @@ int fcb_append_finish(struct fcb *, struct fcb_entry *append_loc);
  * loc->fe_area, loc->fe_data_off, and loc->fe_data_len as arguments.
  */
 typedef int (*fcb_walk_cb)(struct fcb_entry *loc, void *arg);
-int fcb_walk(struct fcb *, struct flash_area *, fcb_walk_cb cb, void *cb_arg);
+int fcb_walk(struct fcb *, int sector, fcb_walk_cb cb, void *cb_arg);
 int fcb_getnext(struct fcb *, struct fcb_entry *loc);
 
 /**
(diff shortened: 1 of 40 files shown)


fs/fcb/src/fcb.c
```diff
@@ -265,3 +301,183 @@ fcb_clear(struct fcb *fcb)
     }
     return rc;
 }
+
+int
+fcb_init_flash_area(struct fcb *fcb, int flash_area_id, uint8_t version)
```
ccollins476ad: Is this function needed?  I don't see it getting called anywhere, and I don't see it declared in any headers.  If it is needed, then it should be declared in a header, and the references to a "log" should be removed from the comments.

fs/fcb/src/fcb.c
```diff
@@ -265,3 +301,183 @@ fcb_clear(struct fcb *fcb)
...
+        entry->fe_range = info.si_range;
+        entry->fe_sector = (uint16_t)sector;
+    }
+
+    return rc;
+}
+
+
+int
+fcb_offset_in_sector(const struct fcb *fcb, int offset, int sector)
```
ccollins476ad: This function (`fcb_offset_in_sector`) would probably be simpler if it called `fcb_get_sector_info()`.

fs/fcb/src/fcb.c
```diff
@@ -121,10 +125,29 @@ fcb_free_sector_cnt(struct fcb *fcb)
 int
 fcb_is_empty(struct fcb *fcb)
 {
-    return (fcb->f_active.fe_area == fcb->f_oldest &&
+    return (fcb->f_active.fe_sector == fcb->f_oldest_sec &&
       fcb->f_active.fe_elem_off == sizeof(struct fcb_disk_area));
 }
 
+struct sector_range *
+fcb_get_sector_range(const struct fcb *fcb, int sector)
```
ccollins476ad: This function (`fcb_get_sector_range`) would probably be simpler if it called `fcb_get_sector_info()`.

fs/fcb/src/fcb_area_info.c
```diff
@@ -21,24 +21,28 @@
...
     int elems = 0;
     int bytes = 0;
 
-    loc.fe_area = fa;
+    fcb_get_sector_info(fcb, sector, &info);
+    loc.fe_range = info.si_range;
+    loc.fe_sector = info.si_sector_in_range + loc.fe_range->sr_first_sector;
     loc.fe_elem_off = 0;
+    /* In case caller passed oldest, get real sector number */
+    if (sector == FCB_SECTOR_OLDEST)
```
ccollins476ad: I believe there is as coding standard requirement to use `{` / `}` for all `if` statements.

sys/log/full/src/log_fcb.c
```diff
@@ -86,9 +86,10 @@ log_fcb_start_append(struct log *log, int len, struct fcb_entry *loc)
          * oldest flash area which was erased. If yes, then move watermark to
          * beginning of current oldest area.
          */
-        if ((fcb_log->fl_watermark_off >= old_fa->fa_off) &&
-            (fcb_log->fl_watermark_off < old_fa->fa_off + old_fa->fa_size)) {
-            fcb_log->fl_watermark_off = fcb->f_oldest->fa_off;
+        if (fcb_offset_in_sector(fcb, fcb_log->fl_watermark_off, old_sec)) {
+            struct fcb_sector_info si;
```
ccollins476ad: Sorry, another annoying coding style nitpick: this variable definition needs to go at the top of the function.

fs/fcb/include/fcb/fcb.h
```diff
@@ -139,17 +149,57 @@ int
...
+int fcb_get_sector_loc(const struct fcb *fcb, int sector,
+    struct fcb_entry *entry);
+
+int fcb_get_sector_offset(const struct fcb *fcb, int sector);
+int fcb_entry_get_sector_offset(const struct fcb_entry *loc);
+
+int fcb_write_to_sector(struct fcb_entry *loc, int off,
+    const void *buf, int len);
+
+int fcb_read_from_sector(struct fcb_entry *loc, int off, void *buf, int len);
```
ccollins476ad: It's tedious, but I think we should get in the habit of adding doxygen comments for all new public functions.  Otherwise, this stuff will probably never get documented.

mkiiskila:
Why not revive this function instead? https://github.com/apache/mynewt-core/blob/0_10_0_dev/hw/hal/src/flash_map.c#L98

It was used to limit the number of FCB sectors by aggregating the underlying flash sectors into bundles.

fs/fcb/include/fcb/fcb.h
```diff
@@ -36,33 +36,43 @@ extern "C" {
...
+ * Entry location point to sector, and offset
+ * within that sector.
  */
 struct fcb_entry {
-    struct flash_area *fe_area;	/* ptr to area within fcb->f_sectors */
-    uint32_t fe_elem_off;	/* start of entry */
-    uint32_t fe_data_off;	/* start of data */
+    struct sector_range *fe_range;  /* ptr to area within fcb->f_ranages */
+    uint16_t fe_sector;     /* sector number in fcb flash */
+    uint16_t fe_elem_off;	/* start of entry in sector */
```
mkiiskila: There are flash sectors > 64k in STM.

kasjer: done

fs/fcb/include/fcb/fcb.h
kasjer: done

fs/fcb/src/fcb.c
kasjer: done

fs/fcb/src/fcb.c
kasjer: done

mkiiskila:
Note that FCB does not need physical sector as the unit in it's flash_area array; when clearing out oldest used area, it uses hal_flash_erase(). That routine can erase multiple physical sectors.
Given that, I'm not sure this PR is necessary.

I created another helper routine which can be used in creating a smaller flash array when using FCB. Use it instead of flash_area_to_sectors(). See https://github.com/apache/mynewt-core/pull/1472

mkiiskila:
Note that FCB does not need physical sector as the unit in it's flash_area array; when clearing out oldest used area, it uses hal_flash_erase(). That routine can erase multiple physical sectors.
Given that, I'm not sure this PR is necessary.

I created another helper routine which can be used in creating a smaller flash array when using FCB. Use it instead of flash_area_to_sectors(). See https://github.com/apache/mynewt-core/pull/1472

mkiiskila:
Note that FCB does not need physical sector as the unit in it's flash_area array; when clearing out oldest used area, it uses hal_flash_erase(). That routine can erase multiple physical sectors.
Given that, I'm not sure this PR is necessary.

I created another helper routine which can be used in creating a smaller flash array when using FCB. Use it instead of flash_area_to_sectors(). See https://github.com/apache/mynewt-core/pull/1472

mkiiskila:
Note that FCB does not need physical sector as the unit in it's flash_area array; when clearing out oldest used area, it uses hal_flash_erase(). That routine can erase multiple physical sectors.
Given that, I'm not sure this PR is necessary.

I created another helper routine which can be used in creating a smaller flash array when using FCB. Use it instead of flash_area_to_sectors(). See https://github.com/apache/mynewt-core/pull/1472

kasjer:
Needs rebase to incorporate latest code.

PR #1449, opened 2018-10-09
Padded log timestamps

InternetofAwesome:
This is added to facilitate a parsing operation by a resource constrained device which depends on fixed length strings.
It also abuses the preprocessor to keep the inconvenience compiler-side.
I'm sorry.

### DIFF
--- a/sys/console/full/src/console_fmt.c
+++ b/sys/console/full/src/console_fmt.c
@@ -27,6 +27,15 @@
 
 #if MYNEWT_VAL(BASELIBC_PRESENT)
 
+/* This collection of macros turns an integer into a string */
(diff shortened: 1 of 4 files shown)


andrzej-kaczmarek:
I created simple PR for tinyprintf which adds support for parametrized field width (https://github.com/apache/mynewt-core/pull/1451) so we don't need to abuse preprocessor here ;-)

Also, personally I'd call syscfg values `WIDTH` instead of `PAD`, but that's just my nitpicking.

ccollins476ad:
It sounds like this change is no longer needed.  Please reopen if I have misunderstood.

PR #1450, opened 2018-10-10
sys/metrics: Fix mbuf double free

ccollins476ad:
`metrics_event_to_cbor()` frees the provided mbuf on failure.  Therefore, callers of this function must not free the mbuf a second time.

### DIFF
--- a/sys/metrics/src/metrics.c
+++ b/sys/metrics/src/metrics.c
@@ -125,7 +125,6 @@ metrics_event_end(struct metrics_event_hdr *hdr)
                 log_append_mbuf_typed(hdr->log, hdr->log_module, hdr->log_level,
                                       LOG_ETYPE_CBOR, om);
             } else {
-                os_mbuf_free_chain(om);
                 ret = -1;
             }
         } else {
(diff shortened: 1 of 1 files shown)


sys/metrics/include/metrics/metrics.h
```diff
@@ -409,6 +409,9 @@ int metrics_set_series_value(struct metrics_event_hdr *hdr, uint8_t metric,
  * for CBOR stream. This way of calling should thus be used only if all data for
  * an event were already collected.
  *
+ * This function always consumes the provided mbuf, both on success and
+ * failure.
```
andrzej-kaczmarek: this is not correct because on success provided mbuf is still valid as otherwise the result of serialization would be gone. so perhaps instead of change below we should modify this one not to consume mbuf on failure? tbh I'm not sure why I done it this way.

ccollins476ad: Heh, you're right.  That is the whole point of the function: to supply a valid mbuf.  So, I'm not sure what I was thinking...

I think this function is a bit awkward to use because the caller has to pass in an mbuf.  It would be simpler if the function returned a newly-allocated and filled mbuf on success, or NULL on failure.  However, I think I understand why you implemented it as you did - so that the caller can reserve space at the beginning of the mbuf for the `log_entry_hdr`.

OK, I agree - I have reverted my change, and modified `metrics_event_to_cbor` to not free on failure.

andrzej-kaczmarek: Yeah, I think one reason was to be able to easily reserve space for log header (otherwise I guess we'd need to prepend large mbuf and we may not have too many mbufs left) and other could be that caller can allocate for any pool (since `metrics_get_buf` allocs from separate pool, not from msys). But I should have put some comments there in first place... ;-)

PR #1451, opened 2018-10-10
tinyprintf: Add support for parametrized field width

andrzej-kaczmarek:
This adds support for field width passed as a parameter, e.g.: `printf("%0*d", width, value);`

### DIFF
--- a/libc/baselibc/src/tinyprintf.c
+++ b/libc/baselibc/src/tinyprintf.c
@@ -282,7 +282,10 @@ size_t tfp_format(FILE *putp, const char *fmt, va_list va)
             }
 
             /* Width */
-            if (ch >= '0' && ch <= '9') {
+            if (ch == '*') {
+                p.width = intarg(0, 1, &va);
+                ch = *(fmt++);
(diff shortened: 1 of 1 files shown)


libc/baselibc/src/tinyprintf.c
```diff
@@ -282,7 +282,10 @@ size_t tfp_format(FILE *putp, const char *fmt, va_list va)
             }
 
             /* Width */
-            if (ch >= '0' && ch <= '9') {
+            if (ch == '*') {
+                p.width = intarg(0, 1, &va);
```
ccollins476ad: It might be a good idea to do a range check here.  A standard-compliant `printf` allows an `int` variable width, but tinyprintf's `width` member is an `unsigned char`.  I'm *not* suggesting you change this to support the full `int` range (that doesn't seem like a worthwhile change).  Perhaps this code could clamp the width to `[0, UCHAR_MAX]`.

andrzej-kaczmarek: Yes, this is a good idea. I'll merge this one since it's already approved and address this in separate PR soon.

PR #1453, opened 2018-10-11
allow user to override TCP port gdb and jtag emulator use to communicate

mkiiskila:
This contains cherry-picked change from https://github.com/apache/mynewt-core/pull/1401. Here it is extended to be used also for 'newt debug' with Jlink, as well as use with openocd.


mkiiskila:
Now it should be possible to have 2 debug sessions open from the same machine. E.g. 
newt debug slinky_nrf52 --extrajtagcmd "-port 5431 -select usb=682148664"

PR #1456, opened 2018-10-15
hw/mcu: Allow SPI_2 for MCU_NRF52832, add SPI_3 for MCU_NRF52840.

jeremywood:
Fix config options for MCU_NRF52832 and MCU_NRF52840.

MCU_NRF52832 has 3 SPI controllers and MCU_NRF52840 has 4. The syscfg only allows for 2 and 3, respectively.


jeremywood:
@andrzej-kaczmarek Thanks, I missed that! Fixed.

PR #1470, opened 2018-10-19
fe310 fix build

kasjer:
fe310 bsp had function called _init that was used during startup.
Recent rename of __libc_init_array to _init resulted in
multiple definition.
bsp function was just renamed to SystemInit that is commonly used
in other platforms.


utzig:
@kasjer Btw, this is also required on `master`.

PR #1471, opened 2018-10-19
sys/sys: Add `DEBUG_PANIC()` macro

ccollins476ad:
The `DEBUG_PANIC()` macro triggers a crash if the `DEBUG_PANIC_ENABLED` syscfg setting is enabled.  If this setting is disabled, the macro expands to nothing.

This feature is discussed briefly here: https://lists.apache.org/thread.html/0ce981b90eb74226fd660f25ba88f4ce6804eea2943edac70bf8731e@%3Cdev.mynewt.apache.org%3E

### DIFF
--- a/kernel/os/include/os/os_fault.h
+++ b/kernel/os/include/os/os_fault.h
@@ -24,8 +24,14 @@
 extern "C" {
 #endif
 
-void __assert_func(const char *, int, const char *, const char *)
+void __assert_func(const char *file, int line, const char *func, const char *e)
     __attribute((noreturn));
+
+#if MYNEWT_VAL(OS_CRASH_FILE_LINE)
+#define OS_CRASH() __assert_func(__FILE__, __LINE__, NULL, NULL)
+#else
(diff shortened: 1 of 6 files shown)


sjanc:
quick question, how this differ from assert?

sjanc:
nevermind,  I see that those are just convenient macros around asserts

ccollins476ad:
The dev list email probably goes into more detail than anyone is interested in :).  The short answer is:

* `DEBUG_ASSERT()` does not evaluate an expression to determine success / failure; it assumes failure.  Since these failures are expected, I think it is likely that the application will always want to execute some code in the failure case, regardless of whether the macro is configured to trigger a crash.  For example: logging a message to the console when the failure is detected.

* `DEBUG_ASSERT()` is intended to be disabled in production builds, so it should also be followed by code that recovers from the failure.  `assert()` is meant to ensure that that failure that "should never happen" indeed never happen, so there is no need for recovery code.  In addition, even though `assert()` can be disabled via `NDEBUG`, I am of the opinion that it should never be disabled, as it helps eliminate a lot of possibilities when debugging faults in the field.

PR #1472, opened 2018-10-22
flash_map; add flash_area_to_subareas().

mkiiskila:
Simpler solution when memory use of FCB is of concern. Creates a flash_area array suitable for passing to FCB out of flash_map index, caller controls the number of array entries created.

### DIFF
--- a/sys/flash_map/include/flash_map/flash_map.h
+++ b/sys/flash_map/include/flash_map/flash_map.h
@@ -107,6 +107,13 @@ uint32_t flash_area_erased_val(const struct flash_area *fa);
 int flash_area_to_sectors(int idx, int *cnt, struct flash_area *ret);
 
 /*
+ * Given flash map index, return info about sectors within the area.
+ * If the physical sector count within region exceeds that, collect
+ * multiple physical sectors into same number of regions.
+ */
+int flash_area_to_subareas(int idx, int *cnt, struct flash_area *ret);
+
+/*
  * Get-next interface for obtaining info about sectors.
  * To start the get-next walk, call with *sec_id set to -1.
  */
--- a/sys/flash_map/src/flash_map.c
+++ b/sys/flash_map/src/flash_map.c
@@ -132,6 +132,73 @@ end:
 }
 
 int
+flash_area_to_subareas(int id, int *cnt, struct flash_area *ret)
+{
+    int rc;
+    int i, j;
+    const struct hal_flash *hf;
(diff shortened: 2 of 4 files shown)


sys/flash_map/include/flash_map/flash_map.h
```diff
@@ -106,6 +106,13 @@ uint32_t flash_area_erased_val(const struct flash_area *fa);
  */
 int flash_area_to_sectors(int idx, int *cnt, struct flash_area *ret);
 
+/*
+ * Given flash map index, return info about sectors within the area.
+ * If the physical sector count within region exceeds that, collect
```
kasjer: Second sentence is not that clear,

sys/flash_map/src/flash_map.c
```diff
@@ -131,6 +131,73 @@ flash_area_getnext_sector(int id, int *sec_id, struct flash_area *ret)
...
+        }
+    }
+    if (*cnt > max_cnt) {
+        min_size = fa->fa_size / max_cnt;
+    } else {
+        min_size = 0;
+    }
+
+    next = 1;
+    for (i = first_idx, j = 0; i < last_idx + 1; i++) {
```
kasjer: could be i <= last_idx

sys/flash_map/src/flash_map.c
```diff
@@ -131,6 +131,73 @@ flash_area_getnext_sector(int id, int *sec_id, struct flash_area *ret)
...
+
+    hf = hal_bsp_flash_dev(fa->fa_device_id);
+    for (i = 0; i < hf->hf_sector_cnt; i++) {
+        hf->hf_itf->hff_sector_info(hf, i, &start, &size);
+        if (start >= fa->fa_off && start < fa->fa_off + fa->fa_size) {
+            if (first_idx == -1) {
+                first_idx = i;
+            }
+            last_idx = i;
+            *cnt = *cnt + 1;
```
kasjer: looks like *cnt is assigned in line 196, so line 165 is not really needed

mkiiskila: On close inspection, realized that it is used as an intermediate variable, and used in line 168.

kasjer: Maybe adding local variable for this would make it simpler and probably would reduce code size as well.

ccollins476ad:
These changes look to be obsolete.  Please reopen as needed.

PR #1475, opened 2018-10-23
STM32F0 Support

adityaxavier:
STM32F0 MCU Support and relevant changes in stm32_common


hw/bsp/nucleo-f030r8/include/bsp/bsp_sysid.h
```diff
@@ -0,0 +1,36 @@
+/**
```
mkiiskila: This file is unnecessary, does not need to exit.

hw/mcu/stm/stm32f0xx/src/hal_flash.c
```diff
@@ -0,0 +1,133 @@
...
+
+static const struct hal_flash_funcs stm32f0_flash_funcs = {
+    .hff_read = stm32f0_flash_read,
+    .hff_write = stm32f0_flash_write,
+    .hff_erase_sector = stm32f0_flash_erase_sector,
+    .hff_sector_info = stm32f0_flash_sector_info,
+    .hff_init = stm32f0_flash_init
+};
+
+struct hal_flash stm32f0_flash_dev_;
```
mkiiskila: Change stm32f0_flash_dev_ -> stm32f0_flash_dev. Also, you should be able to initialize it here at compile-time. That would save some text space. And RAM, if you declare it as const.

hw/mcu/stm/stm32f0xx/src/hal_system_clock.c
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#include <assert.h>
+#include <hal/hal_system.h>
+#include <mcu/cortex_m0.h>
+#include "stm32f0xx_hal.h"
+#include "stm32f0xx_hal_rcc.h"
+
+void hal_system_clock_start(void) {
+    #if MYNEWT_VAL(MCU_SYSCLK_PLL_HSI)
```
mkiiskila: Indent #if to beginning of line.

hw/mcu/stm/stm32f0xx/src/hal_flash.c
adityaxavier: stm32f0_flash_dev is currently a function which is returning that struct. 
As a constant :-
`Size of Application Image: app`
`FLASH     RAM `
`     23     206 *fill*`
`    52       4 apps_blinky.a`
`    490       4 hw_bsp_nucleo-f030r8.a`
`     32       0 hw_cmsis-core.a`
`    124       0 hw_hal.a`
`   1580      36 hw_mcu_stm_stm32_common.a`
`   1740      32 hw_mcu_stm_stm32f0xx.a`
`   3601    1081 kernel_os.a`
`     50       0 libc_baselibc.a`
`    270     128 sys_flash_map.a`
`    340      12 sys_mfg.a`
`     30       5 sys_sysinit.a`
`     88       0 stm32_blinky-sysinit-app.a`
`    320       0 libgcc.a`

`objsize`
`   text	   data	    bss	    dec	    hex	filename`
`   8740	     32	   1476	  10248	   2808`

As a static struct being passed
`Size of Application Image: app`
`  FLASH     RAM `
`     27     206 *fill*`
`     52       4 apps_blinky.a`
`    490       4 hw_bsp_nucleo-f030r8.a`
`     32       0 hw_cmsis-core.a`
`    124       0 hw_hal.a`
`   1580      36 hw_mcu_stm_stm32_common.a`
`   1724      56 hw_mcu_stm_stm32f0xx.a`
`   3601    1081 kernel_os.a`
`     50       0 libc_baselibc.a`
`    270     128 sys_flash_map.a`
`    340      12 sys_mfg.a`
`     30       5 sys_sysinit.a`
`     88       0 stm32_blinky-sysinit-app.a`
`    320       0 libgcc.a`

`objsize`
`   text	   data	    bss	    dec	    hex	filename`
`   8728	     56	   1476	  10260	   2814`

Which method would you prefer.
BTW, I believe we should ideally do the same to other stm32fxx series as well for keeping them the similar.

utzig:
@adityaxavier Looks good, but before a merge is done, I have a few suggestions:

1) `Mynewt` carries its own copy of `CMSIS` so everything under `hw/mcu/stm/stm32f0xx/src/ext/Drivers/CMSIS/Include` can be removed
2) Remove those STM32Cube manuals in `chm` format
3) Remove the STM32Cube `Release_Notes.html` files
4) `hw/mcu/stm/stm32f0xx/src/ext/version.txt` has text `STM32Cube_FW_F3_V1.9.0`, please update the version (if necessary) and fix the MCU model.

adityaxavier:
Updated the STM32Cube version, as it reduced the text size by more than 100bytes (fixed the MCU model in the versions.txt as well).

PR #1483, opened 2018-10-24
Remove bsp_sysid.h which is no longer used

adityaxavier:
Removed obsolete bsp_sysid.h from various bsp.


utzig:
I am not sure about the purpose of this file, but looking at history, seems like the first one that was added was for the `ci40` BSP which is a MIPS MCU. This file might be required by the MIPS/PIC32 SDK. It should be safe to remove for `nucleo-f303k8`, `nucleo-f303re`, `stm32f3discovery` and `apollo2_evb`. I'll wait for others opinions.

PR #1484, opened 2018-10-24
Brought struct hal_flash calling in parity with other bsp

adityaxavier:
Changed from function call to return struct hal_flash to const hal_flash implementation as is in other BSP.


utzig:
The reason this failed on CI and why it was a function instead of a static struct is that `HAL_FLASH_SIZE` is determined at run-time by reading the flash registers. Not sure if it would be better to just parameterize this, but I'll take a look in a few days.

utzig:
Fixed by #1522

PR #1489, opened 2018-10-29
Nucleo-F072RB

utzig:
DO NOT MERGE

### DIFF
--- a/hw/bsp/nucleo-f072rb/bsp.yml
+++ b/hw/bsp/nucleo-f072rb/bsp.yml
@@ -1,0 +1,60 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+bsp.arch: cortex_m0
+bsp.compiler: compiler/arm-none-eabi-m0
+bsp.linkerscript:
+    - "hw/bsp/nucleo-f072rb/nucleo-f072rb.ld"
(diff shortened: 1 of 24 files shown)


hw/bsp/nucleo-f072rb/bsp.yml
```diff
@@ -0,0 +1,60 @@
...
+    - "@apache-mynewt-core/hw/mcu/stm/stm32f0xx/stm32f0xx.ld"
+bsp.downloadscript: "hw/bsp/nucleo-f072rb/nucleo-f072rb_download.sh"
+bsp.debugscript: "hw/bsp/nucleo-f072rb/nucleo-f072rb_debug.sh"
+bsp.downloadscript.WINDOWS.OVERWRITE: "hw/bsp/nucleo-f072rb/nucleo-f072rb_download.cmd"
+bsp.debugscript.WINDOWS.OVERWRITE: "hw/bsp/nucleo-f072rb/nucleo-f072rb_debug.cmd"
+
+bsp.flash_map:
+    areas:
+        # System areas.
+        FLASH_AREA_BOOTLOADER:
```
mkiiskila: Is it possible to just leave out BOOTLOADER/IMAGE_1 areas from here?

hw/bsp/nucleo-f072rb/nucleo-f072rb_debug.sh
```diff
@@ -0,0 +1,52 @@
...
+GDB_CMD_FILE=.gdb_cmds
+
+echo "Debugging" $FILE_NAME
+
+#
+# Block Ctrl-C from getting passed to openocd.
+# Exit openocd when gdb detaches.
+#
+set -m
+openocd -f board/st_nucleo_f0.cfg -c "gdb_port 3333; telnet_port 4444; debug_level 0; stm32f0x.cpu configure -event gdb-detach {resume;shutdown}" -c "$EXTRA_JTAG_CMD" -c init -c halt &
```
mkiiskila: Are the common openocd functions not working with this board?

hw/bsp/nucleo-f072rb/bsp.yml
utzig: Some dependencies require the definitions that are generated by newt, eg, flash_map requires `FLASH_AREA_IMAGE_1` to exist, also some to bootloader; removing it would require me to change my sample programs' deps...

hw/bsp/nucleo-f072rb/nucleo-f072rb_debug.sh
utzig: This was probably copy/paste from other F0 BSP; I updated it to use `st-util` which is a lot more straight-forward for STM32 targets.

utzig:
@mkiiskila I updated this with fixes, now all peripherals are tested, eg, I2C/SPI/Timers, etc. Will merge if it passes CI.

mkiiskila:
Sounds reasonable to me.

PR #1490, opened 2018-10-29
Hal flash write protection

kasjer:
In certain cases (i.e. low voltage) it may be wise to prevent
writes to flash since it may result in flash corruption.
New function hal_flash_write_protect() allows to block/unblock writes
to flash.
@InternetofAwesome please take a look if that is what you had in mind

### DIFF
--- a/hw/hal/src/hal_flash.c
+++ b/hw/hal/src/hal_flash.c
@@ -26,6 +26,8 @@
 #include "hal/hal_flash.h"
 #include "hal/hal_flash_int.h"
 
+static uint8_t protected_flash[1];
+
 int
 hal_flash_init(void)
 {
@@ -165,6 +167,10 @@ hal_flash_write(uint8_t id, uint32_t address, const void *src,
         return -1;
     }
 
+    if (protected_flash[id / 8] & (1 << (id & 7))) {
+        return SYS_EACCES;
+    }
+
     rc = hf->hf_itf->hff_write(hf, address, src, num_bytes);
     if (rc != 0) {
         return rc;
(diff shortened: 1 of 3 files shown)


hw/hal/src/hal_flash.c
```diff
@@ -165,6 +167,10 @@ hal_flash_write(uint8_t id, uint32_t address, const void *src,
         return -1;
     }
 
+    if (protected_flash[id / 8] &= (1 << (id & 7))) {
```
ccollins476ad: This looks a bit odd to me.  Did you intend to modify the `protected_flash` array here?  (i.e., should this expression use `&` rather than `&=`?).  This comment applies to a few other uses of `&=` in this file.

hw/hal/src/hal_flash.c
```diff
@@ -338,3 +352,23 @@ hal_flash_ioctl(uint8_t id, uint32_t cmd, void *args)
...
 }
+
+int
+hal_flash_write_protect(uint8_t id, uint8_t protect)
+{
+    if (NULL == hal_bsp_flash_dev(id)) {
+        return SYS_EINVAL;
+    }
+    if (id / 8 >= sizeof(protected_flash) / sizeof(protected_flash[0])) {
+        assert(0);
```
ccollins476ad: I don't think this function should assert that the ID argument is valid.  I would just let the caller assert success if desired.

hw/hal/src/hal_flash.c
kasjer: done.
Thanks @ccollins476ad I must have been brain dead when writing it.

hw/hal/src/hal_flash.c
kasjer: done

PR #1505, opened 2018-11-05
hw/drivers: Add display driver SSD1673

michal-narajowski:
This code was ported from Zephyr


hw/drivers/display/ssd1673/syscfg.yml
```diff
@@ -0,0 +1,47 @@
...
+#
+
+syscfg.defs:
+    SSD1673_OS_DEV_NAME:
+        description: 'os_dev device name for SSD1673'
+        value: '"ssd1673"'
+
+    SSD1673_BUSY_PIN:
+        description: 'BUSY pin for SSD1673'
+        value: 14
```
kasjer: I think it would be better to change value:14 to
```
        value:
        restrictions:
            - $notnull
```
and define pin in bsp instead, since those values do not make sense in general.
Same applies to other pins.

hw/drivers/display/cfb/include/display/cfb.h
```diff
@@ -42,17 +42,17 @@ enum cfb_display_param {
 };
 
 enum cfb_font_caps {
-	CFB_FONT_MONO_VPACKED		= BIT(0),
-	CFB_FONT_MONO_HPACKED		= BIT(1),
+	CFB_FONT_MONO_VPACKED		= (1UL << (0)),
```
kasjer: I guess this is macro code extracted. Parentheses make sense in macros bur here 0 could be just 0 instead of (0). Other parts of zephyr ported code do not uses extra () for constants.

hw/drivers/display/cfb/syscfg.yml
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+    CFB_FONTS_COUNT:
```
kasjer: description could be nice here

hw/drivers/display/ssd1673/src/ssd1673.c
```diff
@@ -298,36 +306,40 @@ static int ssd1673_write(const struct device *dev, const u16_t x,
...
 	}
 
-	gpio_pin_write(driver->dc, CONFIG_SSD1673_DC_PIN, 1);
+	hal_gpio_write(CONFIG_SSD1673_DC_PIN, 1);
 	/* clear unusable page */
 	if (driver->scan_mode == SSD1673_DATA_ENTRY_XDYIY) {
-		sbuf.buf = dummy_page;
-		sbuf.len = sizeof(dummy_page);
-		if (spi_write(driver->spi_dev, &driver->spi_config, &buf_set)) {
+		if (hal_spi_txrx(CONFIG_SSD1673_SPI_DEV, &dummy_page,
```
kasjer: usually & is not added in this type of usage

hw/drivers/display/cfb/syscfg.yml
```diff
@@ -0,0 +1,21 @@
...
+        value: 3
```
andrzej-kaczmarek: this should not be configurable because number of fonts available depends on what's defined in `cfb_fonts.c` so you can't just put any value here - it just has to be `3`.

hw/drivers/display/ssd1673/src/ssd1673.c
```diff
@@ -4,19 +4,18 @@
...
-#include <string.h>
-#include <device.h>
-#include <display.h>
-#include <init.h>
-#include <gpio.h>
-#include <spi.h>
+#include "os/mynewt.h"
+#include "hal/hal_gpio.h"
+#include "hal/hal_spi.h"
+#include "mcu/nrf52_hal.h"
```
andrzej-kaczmarek: this is available only for nrf52xxx so should not be here

michal-narajowski:
Thanks @kasjer and @andrzej-kaczmarek! Fixed.

PR #1507, opened 2018-11-06
Add generic serial bus driver

andrzej-kaczmarek:
This PR adds initial proposed API and implementation of generic serial bus driver.

The goal is to create abstraction for using serial buses like I2C and SPI in a generic way where bus state is managed automatically to allow access from multiple tasks and to multiple devices concurrently in an easy way [1]. It will be also good place to add special handling for known issues in existing devices if they are attached to bus so we can recover from errors [2].

[1] currently we use global mutexes which are shared via sensor/charger/etc. interface with each driver and each driver needs to lock it manually - we should have lock defined and managed automatically on bus level
[2] currently we already have some workaround for faulty device in `hal_i2c` implementation of `hw/mcu/nordic/nrf52xxx` - HAL is not a good place to workaround problems not related to particular I2C controller and this workaround will also not work when platform other than `nrf52xxx` is used

There are two kind of objects managed by bus driver:
- `bus_dev` = bus device which is a representation of I2C or SPI bus instance; should be created by MCU (or BSP) code and effectively replace existing init calls to HAL since this is done now internally by bus driver,
- `bus_node` = bus node which is a representation of an actual device connected to bus; should be created by driver or even an application in case application wants to access some "loose" object on the bus without dedicated driver.

It may be a bit counter-intuitive to use `bus_dev` for bus and not for an actual device, but it originated from `os_dev` which is used internally and honestly I don't have any better names for both so any other ideas are welcome. I eventually got used to these names after a week of coding ;-)

The setup to use bus driver for I2C controller is as follows:
1. MCU creates `i2c0` `bus_dev` which initializes hardware (similar to what `hal_xxx_init` call did previously)
2. for each device attached to bus, driver or application needs to create `bus_node` which specifies:
- `bus_dev` (by name) where node is attached
- bus configuration used for communication with each device (clock frequency for I2C, will be also data mode for SPI)
- optional quirks for device; this allows to define some known issues for a device and allow bus driver to automatically apply some workarounds in case such device is used

Application which wants to talk to one of created `bus_node`s should do as follows:
1. call `os_dev_open()` to open `bus_node` with specified name
2. on 1st open, driver can configure device before 1st use
3. `os_dev` obtained from `os_dev_open()` call can now used to talk to device using `bus_node` APIs

Each operation on `bus_node` does as follows:
1. parent `bus_dev` is locked for exclusive access
2. bus configuration is adjusted if new device requires different configuration that device previously accessed on the same bus (for I2C this probably won't be used too much since we can run all devices with the same speed, but it's easily possible to use different devices with different speeds)
3. operation is executed via appropriate HAL
4. `bus_dev` is unlocked

So effectively what users of `bus_node`s need to do is just call single `bus_node_xxx` API and parse the result. They don't need to care about current bus state. The goal is to be able to use a simple call to make most of bus accesses.

APIs are split into two main parts:
- `bus/bus.h` is what is enough to access `bus_node` as opened by `os_dev_open()` and this is what driver and application will mostly use
- `bus/bus_driver.h` is what is necessary to create `bus_dev` and `bus_node` from MCU/BSP/driver so more low-level stuff that needs to be only done once to create `os_dev`s and not required to access bus from apps
(there's also `bus/bus_debug.h` but it's just some helper for debug code, so not really useful from "outside")

Current API allows to do following operations:
1. read data from `bus_node`
2. write data to `bus_node`
3. write data from and then read data to the same `bus_node` (with bus lock held during entire transaction) which is probably the most commonly used transaction
4. lock and unlock `bus_dev` to allow combine the above simple ops into compound transactions
5. (already deprecated) obtain direct access to `bus_dev` lock

1/2/3 are available as blocking operations; non-blocking versions of these APIs will be also available and they will quite similar to blocking calls except the result will be returned via supplied callback
4 I considered to provide API to define and execute compound transactions (other than most common write->read) but abandoned this for now since it will be quite complicated and I am not sure if it will be used too much, so probably not worth the effort, but we can still add this later
5 is to provide some compatibility with existing code so we can share mutex with other existing ways of locking, at least for now

In addition to new `hw/bus` and `hw/bus/i2c` packages which implements generic bus driver and I2C bus driver, there are few other commits included which were either necessary or just useful to implement this:
1. `hal_i2c` is extended with some calls that already exist in `hal_spi` which are separate calls to enable, disable and configure interface
2. `hw/mcu/nordic/nrf52xxx` is extended to create `bus_dev`s for enabled I2C instances when `hw/bus` is included in build
3. `apps/bus_test` is provided which includes two simple implementation of bus_nodes which it can then talk to (they are running at different speeds just to see if things are reconfigured properly on-the-fly)

Finally, there are few things that are not finished yet which I'll try to update in upcoming days:
- doxygen doc is a bit sparse in some places,
- error handling is barely implemented,
- device quirks are not implemented
- SPI bus driver is missing

*UPDATE 2018-11-08:*
- SPI driver is working now
- sensors interface adjusted to use bus driver if enabled (all drivers need to support bus driver also)
- BME280 driver modified to use bus driver as an example; as a bonus it can work with both I2C and SPI, just need to create proper node using helper function
- test app can talk to BME280 via sensors interface using either I2C or SPI
- some issues fixes


apps/bus_test/bme280_node/src/bme280_node.c
```diff
@@ -0,0 +1,154 @@
...
+bme280_node_read(struct os_dev *dev, struct bme280_node_measurement *measurement)
+{
+    struct bme280_node *bme280_node = CONTAINER_OF(dev, struct bme280_node, i2c_node);
+    struct bme280_data sensor_data;
+    int rc;
+
+    assert(measurement);
+
+    rc = bme280_set_sensor_mode(BME280_FORCED_MODE, &bme280_node->bme280_dev);
+    assert(rc == 0);
```
kasjer: asserting on access may be to drastic. Maybe rc could be returned on errors.
In release build when assert is not present function returns 0 event if data is wrong.

apps/bus_test/bme280_node/src/bme280_node.c
```diff
@@ -0,0 +1,154 @@
...
+    rc = bme280_set_sensor_settings(BME280_OSR_PRESS_SEL | BME280_OSR_TEMP_SEL |
+                                    BME280_OSR_HUM_SEL | BME280_FILTER_SEL,
+                                    bme280_dev);
+    assert(rc == 0);
+}
+
+static void
+close_node_cb(struct bus_node *node)
+{
+    console_printf("%s: node %p\n", __func__, node);
```
kasjer: I guess dependency to console is needed

apps/bus_test/bme280_node/src/bme280_node.c
```diff
@@ -0,0 +1,154 @@
...
+
+struct reg_val {
+    uint8_t addr;
+    uint8_t val[32];
+};
+
+static int8_t
+bme280_read(uint8_t dev_id, uint8_t reg_addr, uint8_t *data, uint16_t len)
+{
+    struct reg_val reg;
```
kasjer: I guess bme280_read/write/delay are expected by Bosch code. I fail to see a reason for additional stack space for local buffer for registers that will be copied to Bosch provided buffer (maybe also on stack).
It's more understandable in write when register is also passed in on write command buffer.

apps/bus_test/lis2dh_node/include/lis2dh_node/lis2dh_node.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <stdint.h>
+#include "os/os_dev.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct lis2dh_node_pos {
+    uint16_t x;
```
kasjer: I think it maybe signed data

hw/bus/spi/include/bus/spi.h
```diff
@@ -0,0 +1,116 @@
...
+#define BUS_SPI_MODE_3              (HAL_SPI_MODE3)
+
+#define BUS_SPI_DATA_ORDER_LSB      (HAL_SPI_LSB_FIRST)
+#define BUS_SPI_DATA_ORDER_MSB      (HAL_SPI_MSB_FIRST)
+
+struct bus_spi_node_cfg {
+    /** General node configuration */
+    struct bus_node_cfg node_cfg;
+    /** */
+    uint8_t pin_cs;
```
kasjer: other pins are int

hw/bus/spi/include/bus/spi.h
```diff
@@ -0,0 +1,116 @@
...
+    uint8_t data_order;
+    /** SCK frequency to be used for node */
+    uint16_t freq;
+    /** Quirks to be applied for device */
+    uint16_t quirks;
+};
+
+struct bus_spi_node {
+    struct bus_node bnode;
+    uint8_t pin_cs;
```
kasjer: maybe int would be better

hw/hal/include/hal/hal_spi.h
```diff
@@ -60,6 +60,14 @@ extern "C" {
 /* Prototype for tx/rx callback */
 typedef void (*hal_spi_txrx_cb)(void *arg, int len);
 
+/** SPI controller hardware settings */
+struct hal_spi_hw_settings {
+    uint8_t pin_sck;
```
kasjer: in other places pins are ints

apps/bus_test/bme280_node/src/bme280_node.c
andrzej-kaczmarek: there's already dependency in `bus_test` app and since this is sub-package not meant to be used outside I think this is enough

andrzej-kaczmarek:
Changed:
- updated review comments
- rebased
- fixed problem with bus node creation for use with sensors (sample BME280 worked only because it did not use data from sensor_itf)

hw/bus/spi/src/spi.c
```diff
@@ -0,0 +1,225 @@
...
+
+    return 0;
+}
+
+static const struct bus_dev_ops bus_spi_ops = {
+    .enable = bus_spi_enable,
+    .configure = bus_spi_configure,
+    .read = bus_spi_read,
+    .write = bus_spi_write,
+    .disable = bus_spi_disable,
```
mkiiskila: For SPI devices combined read/write is occasionally needed. I see there's an call API matching combined read/write semantic in bus_node(), but no such thing here?

I think there's 2 options here; either add combined read/write to allow that for SPI, or make the bus_node_XXX interface such that the driver writer can take advantage of the locking/reconfig for bus-sharing done there, and then they can call hal_XXX routines of their choice while the device is locked/configured.

Including the non-blocking read/write version. Or were you going to roll those as separate interface within bus_dev_ops()?

hw/drivers/sensors/bme280/src/bme280.c
```diff
@@ -826,9 +830,21 @@ int
...
                uint8_t len)
 {
+    int rc;
+
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    struct os_dev *dev = SENSOR_ITF_GET_DEVICE(itf);
+
+    /* XXX this is only required for SPI, but apparently device has no problem
+     * with this being set also for I2C so let's leave it for now since there's
+     * no API now to figure out bus type for node
```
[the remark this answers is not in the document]
andrzej-kaczmarek: that's a good point, I did not think about it

hw/bus/src/bus.c
```diff
@@ -0,0 +1,314 @@
...
+    struct bus_node *bnode = (struct bus_node *)node;
+    struct bus_dev *bdev = bnode->parent_bus;
+    os_error_t err;
+
+    BUS_DEBUG_VERIFY_DEV(bdev);
+    BUS_DEBUG_VERIFY_NODE(bnode);
+
+    err = os_mutex_release(&bdev->lock);
+    if (err == OS_BAD_MUTEX) {
+        return SYS_EACCES;
```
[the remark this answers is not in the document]
andrzej-kaczmarek: agree, will fix

apps/bus_test/lis2dh_node/include/lis2dh_node/lis2dh_node.h
andrzej-kaczmarek: yeah, I just used this a raw data since it has no other purpose than demo ;-)

hw/bus/spi/src/spi.c
andrzej-kaczmarek: `bus_node_write_read_transact` does write-then-read to simplify perhaps one of most common transaction for peripheral devices. I also had write-and-read API prototype but removed it as I did not implement it since I don't really know device which would require it (I've only seen ones that can use it, but not require). This can be added later as a `bus_node` API and as a new op here (like `write_read`). My original idea was that this would work as expected for SPI bus and return `SYS_ENOTSUP` for I2C bus. For now, as you mentioned, one can just lock bus manually via `bus_dev_lock_by_node` (btw, I don't like this name, thinking of changing it) and use HAL if really necessary.

Once I add async APIs which can use non-blocking HAL calls I think it will require to add non-blocking versions of read, write (and read-write) ops as well.

hw/bus/src/bus.c
```diff
@@ -0,0 +1,314 @@
...
+
+#include <assert.h>
+#include <string.h>
+#include "defs/error.h"
+#include "bus/bus.h"
+#include "bus/bus_debug.h"
+#include "bus/bus_driver.h"
+
+static int
+bus_node_open_func(struct os_dev *odev, uint32_t wait, void *arg)
```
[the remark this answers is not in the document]
andrzej-kaczmarek: probably not (I don't really see how I could use those parameters here), but this is callback from `os_dev` so it needs to be like this.

hw/bus/src/bus.c
```diff
@@ -0,0 +1,314 @@
...
+
+    /* Call close callback if closing last ref */
+    if (odev->od_open_ref == 1) {
+        bnode->callbacks.close(bnode);
+    }
+
+    return 0;
+}
+
+static inline int
```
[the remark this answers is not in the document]
andrzej-kaczmarek: good point

hw/bus/src/bus.c
```diff
@@ -0,0 +1,314 @@
...
+    if (!bdev->dops->read) {
+        return SYS_ENOTSUP;
+    }
+
+    rc = bus_dev_lock_by_node(node, timeout);
+    if (rc) {
+        return rc;
+    }
+
+    rc = bus_dev_configure_for(bdev, bnode);
```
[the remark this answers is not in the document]
andrzej-kaczmarek: agree, configuring bus for device while locking it sounds reasonable.

hw/bus/src/bus.c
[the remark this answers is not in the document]
mkiiskila: Ah yes. Of course.

hw/bus/spi/src/spi.c
mkiiskila: Given that it is a public API, I'm all for changing it. bus_node_lock()/bus_node_unlock() ?

I suspect it's ok to leave the full-duplex operation out of the bus_node interface.

andrzej-kaczmarek:
Fixed latest comments.

Also had to update device creation since when adjusting other drivers it turned out that some of them need custom arg to be passed to init function - the one we have for `os_dev` is already used by bus driver so need to have some other way to pass extra arg directly to driver. This is possible via modified node creation APIs.

Finally, BME280 sample node is removed since it did not pass RAT check as it uses driver from Bosch.

If current API is good enough I'd prefer to merge this now and fix/update with subsequent PRs - I guess before release we can even adjust some APIs if necessary since they are not frozen.
I also will push some drivers adjusted to new interface in separate PR.

hw/drivers/sensors/bme280/src/bme280.c
```diff
@@ -880,9 +897,32 @@ int
...
+
+    rc = bus_node_write(dev, &addr, 1, OS_TIMEOUT_NEVER, BUS_F_NOSTOP);
+    if (rc) {
+        goto done;
+    }
+
+    rc = bus_node_simple_write(dev, payload, len);
+
+done:
+    rc = bus_dev_unlock_by_node(dev);
```
rymanluk: lost previous rc here

andrzej-kaczmarek:
@rymanluk fixed

I have some more fixes ready for this but would prefer to put them in separate PR as it becomes more and more  time consuming to rebase all of those patches :-)

PR #1508, opened 2018-11-06
kernel/os: Add helper CONTAINER_OF macro

andrzej-kaczmarek:
This is a convenience macro to calculate original pointer of "parent"
object using a pointer to one of its members. It's commonly used in
e.g. Linux kernel and Zephyr Project so it should be also useful in
Mynewt :)

For example, with following structure:
```
struct foo {
    int a;
    struct bar bar;
    int c;
}
```
Let's assume we have:
```
struct foo g_foo;
struct bar *x = &g_foo.bar;
```
And we can use `CONTAINER_OF` to calculate pointer to `g_foo` using `x`:
```
struct foo *foo = CONTAINER_OF(x, struct foo, bar);
```

### DIFF
--- a/kernel/os/include/os/util.h
+++ b/kernel/os/include/os/util.h
@@ -26,4 +26,8 @@
 #define POINTER_TO_INT(p) ((int) ((intptr_t) (p)))
 #define INT_TO_POINTER(u) ((void *) ((intptr_t) (u)))
 
+/* Helper to retrieve pointer to "parent" object in structure */
+#define CONTAINER_OF(ptr, type, field) \
+        ((type *)(((char *)(ptr)) - offsetof(type, field)))
+
 #endif


kernel/os/include/os/util.h
```diff
@@ -26,4 +26,8 @@
 #define POINTER_TO_INT(p) ((int) ((intptr_t) (p)))
 #define INT_TO_POINTER(u) ((void *) ((intptr_t) (u)))
 
+/* Helper to retrieve pointer to "parent" object in structure */
+#define CONTAINER_OF(ptr, type, field) \
```
mkiiskila: For this particular macro, I'd prefer container_of in small caps. To match offsetof(). But I'm not hung up on it.

andrzej-kaczmarek: I also considered it as a lowercase for the same reason and also this is how it's defined in Linux kernel - but since it's still not full consistent with `offsetof` (i.e. without underscore) I eventually did this as in Zephyr which has the advantage that we port some code from there so there will be less changes required :-)

PR #1510, opened 2018-11-06
shell; add ls_dev command to list OS devices.

mkiiskila:


sys/shell/src/shell_os.c
```diff
@@ -249,6 +268,13 @@ static const struct shell_cmd os_commands[] = {
         .sc_cmd_func = shell_os_reset_cmd,
 #if MYNEWT_VAL(SHELL_CMD_HELP)
         .help = &reset_help,
+#endif
+    },
+    {
+        .sc_cmd = "ls_dev",
```
ccollins476ad: Just a thought - what do you think about renaming this to `lsdev` (without the underscore)?  That more closely mimics `lshw`,`lspci`, etc. from Unix.

kernel/os/include/os/os_dev.h
```diff
@@ -253,6 +253,13 @@ int os_dev_close(struct os_dev *dev);
  */
 void os_dev_reset(void);
 
+/**
+ * Walk through all devices, calling callback for every device.
+ *
+ * @param walk_func Function to call
+ */
+void os_dev_walk(void (*walk_func)(struct os_dev *));
```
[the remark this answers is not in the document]
mkiiskila: Sure.

sys/shell/src/shell_os.c
mkiiskila: That should be ok.

PR #1512, opened 2018-11-08
sys/stats: Add missing consts

andrzej-kaczmarek:


sys/stats/full/include/stats/stats.h
```diff
@@ -33,7 +33,7 @@ struct stats_name_map {
 } __attribute__((packed));
 
 struct stats_hdr {
-    char *s_name;
+    const char *s_name;
```
mkiiskila: stats/stub has this structure too. Not sure if used, but should probably be modified to match this.

PR #1514, opened 2018-11-09
sys/stats/full: Persistent stats

ccollins476ad:
Persistent stat functionality can be enabled with the `STATS_PERSIST`
syscfg setting.  This feature is described in more detail here:
https://ponymail-vm.apache.org/_GUI_/thread.html/d14032985bc959af468f1f0300b725fb15a8a17df8c2f171c25b469c@%3Cdev.mynewt.apache.org%3E

A persistent stat group is defined using the `STATS_PERSISTED_SECT_START` macro.

A persistent stat group is registered using the `stats_persist_init()` function.

A few questionable decisions in this PR:
1. The `STATS_INC()` macros sacrifice some efficiency to ensure correctness.  Every increment includes a check against the stat group's `STAT_HDR_F_PERSISTED` flag to see if the group's flush timer needs to be scheduled.  If the user knows a stat is non-persistent, he can use `STATS_INC_RAW()` instead, which is as efficient as `STATS_INC()` was before this PR.  Maybe the opposite should have been done: `STATS_INC()` would assume a non-persistent stat group, and the user would be required to use a new set of macros for persistent stats.

2. A large buffer is allocated on the stack whenever stats are persisted (buffer size == `STATS_PERSIST_BUF_SIZE`).  At build time, we don't know the size of the largest persistent stat group, so the user probably overallocates.  This is not great, but I don't know of a better solution.  I believe there are plans for a version of `sys/config` that supports binary data.  With it, we can do away with this temporary buffer.

### DIFF
--- a/sys/stats/full/src/stats_conf.c
+++ b/sys/stats/full/src/stats_conf.c
@@ -1,0 +1,199 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "os/mynewt.h"
+
+#if MYNEWT_VAL(STATS_PERSIST)
+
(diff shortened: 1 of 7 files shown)


sys/stats/full/src/stats_conf.c
```diff
@@ -0,0 +1,199 @@
...
+    if (argc == 1) {
+        hdr = stats_group_find(argv[0]);
+        if (hdr != NULL) {
+            size = stats_size(hdr);
+            data = stats_data(hdr);
+
+            decode_len = base64_decode_len(val);
+            if (decode_len > size) {
+                DEBUG_PANIC();
+                return SYS_ENOMEM;
```
mkiiskila: SYS_ENOMEM -> OS_ENOMEM

sys/stats/full/src/stats_conf.c
```diff
@@ -0,0 +1,199 @@
...
+            }
+
+            memset(data, 0, size);
+            base64_decode(val, data);
+
+            return 0;
+        }
+    }
+    return -1;
```
mkiiskila: return OS_ENOENT;

ccollins476ad:
Thanks, Marko.  I have made the requested changes.

PR #1524, opened 2018-11-19
hw/drivers: Add bus driver support to some of existing drivers

andrzej-kaczmarek:
This updates some of drivers to support new bus driver. Old code is still there and thus they can still work just fine without bus driver - the proper way to access bus is selected automatically in build time depending if bus driver is included or not.

Each driver have now helper function to quickly create bus node so there's no need to figure out how to call `os_dev_create` properly - just use helper instead.

Modified drivers seem to work fine, however I was not able to check each and every single functionality there - just ran this on few existing device with sensors I have here and they all work fine.

### DIFF
--- a/hw/drivers/bq27z561/include/bq27z561/bq27z561.h
+++ b/hw/drivers/bq27z561/include/bq27z561/bq27z561.h
@@ -162,11 +162,13 @@ struct bq27z561_itf
     struct os_mutex *itf_lock;
 };
 
+#if !MYNEWT_VAL(BUS_DRIVER_PRESENT)
 struct bq27z561_init_arg
 {
     struct bq27z561_itf itf;
     struct os_dev *battery;
 };
+#endif
 
 /* BQ27Z561 device */
 struct bq27z561
@@ -177,8 +179,10 @@ struct bq27z561
     /* Configuration values */
     struct bq27z561_cfg bq27_cfg;
 
+#if !MYNEWT_VAL(BUS_DRIVER_PRESENT)
     /* Interface */
     struct bq27z561_itf bq27_itf;
+#endif
 
     /* Device initialization complete flag */
     uint8_t bq27_initialized;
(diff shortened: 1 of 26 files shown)


hw/bus/syscfg.yml
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+    BUS_DRIVER_PRESENT:
```
kasjer: This definition got me confused for a while, why value should not be overridden, but I think it would be beneficial to have **newt** generated value derived from package name and path that would serve this purpose.
The **newt** tool could add something like **_HW_BUS_PRESENT**.
What do you think @ccollins476ad @mkiiskila ?

andrzej-kaczmarek: I think this is a good idea, few other packages could benefit from such addition (e.g. NimBLE).
perhaps packages could define smth like features it implements and then newt would generate MYNEWT_FEATURE macro similar to MYNEWT_VAL. I'd rather not use package name as a base since we can eventually move some packages so name can change and we'll have problem then :-)

hw/drivers/chg_ctrl/adp5061/src/adp5061.c
```diff
@@ -251,20 +264,25 @@ int
...
 
     payload[0] = addr;
     for (i = 0; i < count; ++i) {
         payload[i + 1] = values[i];
     }
 
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    rc = bus_node_simple_write((struct os_dev *)&dev->a_node, payload, count + 1);
+#else
+    int i;
```
kasjer: redeclaration of i

hw/drivers/sensors/bmp280/include/bmp280/bmp280.h
```diff
@@ -22,6 +22,9 @@
 #define __bmp280_H__
 
 #include "os/mynewt.h"
+#include "bus/bus.h"
```
kasjer: missing
`#if MYNEWT_VAL(BUS_DRIVER_PRESENT)`

hw/drivers/bq27z561/include/bq27z561/bq27z561.h
```diff
@@ -162,11 +162,13 @@ struct bq27z561_itf
     struct os_mutex *itf_lock;
 };
 
+#if !MYNEWT_VAL(BUS_DRIVER_PRESENT)
```
kasjer: driver never references bus/i2c.h like other drivers

andrzej-kaczmarek:
@kasjer thanks, fixed

PR #1528, opened 2018-11-21
kernel/os: Fix os_trace APIs

andrzej-kaczmarek:
os_trace_api.h uses local #defined symbols from .c file to determine
whether traces should be enabled for given file or not. In order for
this to work properly it has to be included only after such symbols
are #defined.

However, assert.h includes os/mynewt.h and the nasty side-effect is
that we have os_everything included by including one of std library
includes.

Let's assert.h just include os_fault.h instead as this should not have
any side-effects.


andrzej-kaczmarek:
ooops, forgot to include nrf5xxxx changes...

PR #1378, opened 2018-09-03
lp5523: Fixing selftest, set_output_on and set_n_regs

mlaz:
So far I have found several issues with this driver:
* Not being able to use the driver(not even the basic stuff like simple PWM) without having auto-increment enabled;  
* Not passing selftest when D9 is enabled (also not being able to enable it);

PR #1530, opened 2018-11-22
hw/drivers/lp5523: Fixes for set_n_regs

andrzej-kaczmarek:


mlaz:
Looks good, although I was addressing this here https://github.com/apache/mynewt-core/pull/1378/files

PR #1533, opened 2018-11-26
hw/mcu/nordic: Rework workaround for unresponsive TWI

andrzej-kaczmarek:
Current workaround for unresponsive TWI will be triggered every time
there is operation timeout which has some drawbacks: it will be
applied to every operation that times out, it takes entire operation
time to recover and it will make operation fail so caller needs to
retry.

It is possible to detect this condition and apply workaround quicker
without making operation fail by using simple hack: after triggering
START task SCL line should be pulled low to make START condition and
also when address is written. We can just wait a moment to see if SCL
is pulled low during this time - if not, restart controller and try
again.

This seems to work fine with LP5523 on the bus which can be used to
easily trigger this condition by sending reset command.


andrzej-kaczmarek:
@benmccrea

PR #1536, opened 2018-11-26
hw/bus: Multiple fixes for bus driver (and related code)

andrzej-kaczmarek:


hw/bus/include/bus/bus.h
```diff
@@ -210,6 +218,17 @@ bus_node_lock(struct os_dev *node, os_time_t timeout);
 int
 bus_node_unlock(struct os_dev *node);
 
+/**
+ * Get node configured lock timeout
+ *
+ * Returns lock timeout as configured for node. If not timeout is configured for
```
mkiiskila: Spelling: If not timeout is... -> If no timeout is...

hw/bus/src/bus.c
```diff
@@ -200,7 +202,7 @@ bus_node_write_read_transact(struct os_dev *node, const void *wbuf,
         return SYS_ENOTSUP;
     }
 
-    rc = bus_node_lock(node, timeout);
+    rc = bus_node_lock(node, g_bus_node_lock_timeout);
```
mkiiskila: bus_node_get_lock_timeout()?

hw/bus/src/bus.c
```diff
@@ -172,7 +174,7 @@ bus_node_write(struct os_dev *node, const void *buf, uint16_t length,
...
```
mkiiskila: bus_node_get_lock_timeout()?

hw/bus/src/bus.c
```diff
@@ -145,7 +147,7 @@ bus_node_read(struct os_dev *node, void *buf, uint16_t length,
...
```
mkiiskila: bus_node_get_lock_timeout()?

hw/bus/src/bus.c
```diff
@@ -285,8 +292,18 @@ bus_node_unlock(struct os_dev *node)
     return 0;
 }
 
+os_time_t
+bus_node_get_lock_timeout(struct os_dev *node)
```
mkiiskila: I assume this is not the final implementation, and that bus_node_set_lock_timeout() is going to follow?

andrzej-kaczmarek: my idea here is to let user specify per-node lock timeout in node configuration and getter is to return effective lock timeout for a node, i.e. either default one (if not configured) or the one configured for a node. I think there's no reason for user to reconfigure lock timeout in runtime - by default all nodes can be locked with the same /reasonable/ timeout but perhaps some nodes should be accessed with e.g. very short lock timeout so they either can be read instantly or not at all and for this it's enough to configure this on build time. but there may be some actual scenarios which require different behaviour so comments are always welcome :)

I'll include implementation of this in next series of commits I'm working on currently.

andrzej-kaczmarek:
Thanks Marko, I'll merge this now and include fixes for those comments in my next PR with bunch of another fixes/improvements for bus driver (just so I don't modify already approved PR).

PR #1538, opened 2018-11-28
Fix nrf52 qspi erase

utzig:
I found this issue while trying to "emulate" a 8KB sector sized device on the QSPI flash (`QSPI_FLASH_SECTOR_SIZE: 8192`). While mcuboot was able to swap the images, the resulting image in slot1 is corrupted because not all sectors were erased. The current implementation seems to work by coincidence because the 2 LSb of 4096 are zero!

### DIFF
--- a/hw/mcu/nordic/nrf52xxx-compat/syscfg.yml
+++ b/hw/mcu/nordic/nrf52xxx-compat/syscfg.yml
@@ -153,7 +153,7 @@ syscfg.defs:
         value: 0
     QSPI_FLASH_PAGE_SIZE:
         description: >
-            QSPI page size. Writes can only be perfrmed to one page at a time.
+            QSPI page size. Writes can only be performed to one page at a time.
             In most cases it should be 256.
         value: 0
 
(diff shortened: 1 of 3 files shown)


utzig:
Well assuming I am right in my assessment, this fix is probably too simple, it would be better to maybe send multiple erase commands based on the chosen size of the sector size...

utzig:
Btw, just updated to this code locally:

```
int8_t erases;

erases = MYNEWT_VAL(QSPI_FLASH_SECTOR_SIZE) / 4096;
while (erases-- > 0) {
    while ((NRF_QSPI->STATUS & QSPI_STATUS_READY_Msk) == 0)
        ;
    NRF_QSPI->EVENTS_READY = 0;
    NRF_QSPI->ERASE.PTR = sector_address;
    NRF_QSPI->ERASE.LEN = NRF_QSPI_ERASE_LEN_4KB;
    NRF_QSPI->TASKS_ERASESTART = 1;
    while (NRF_QSPI->EVENTS_READY == 0)
        ;
    sector_address += 4096;
}
```

And now I can set `QSPI_FLASH_SECTOR_SIZE: 8192` and it totally works, swapping with interruptions and everything!

kasjer:
Looks even better

PR #1539, opened 2018-11-29
spiflash: Make spiflash thread safe

kasjer:
If several task would try to use spiflash at the same time they
could get wrong result and even deadlock waiting for some SPI
state that would not happen.
Adding mutex to most of the functions makes spiflash usable from
multiple tasks.
For boot loader sake spiflash can be built without locking.

### DIFF
--- a/hw/drivers/flash/spiflash/src/spiflash.c
+++ b/hw/drivers/flash/spiflash/src/spiflash.c
@@ -611,6 +611,20 @@ struct spiflash_dev spiflash_dev = {
     .flash_chip = NULL,
 };
 
+static inline void spiflash_lock(struct spiflash_dev *dev)
+{
+#if MYNEWT_VAL(OS_SCHEDULING)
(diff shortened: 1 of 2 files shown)


hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -748,6 +794,8 @@ spiflash_read(const struct hal_flash *hal_flash_dev, uint32_t addr, void *buf,
 
     dev = (struct spiflash_dev *)hal_flash_dev;
 
+    spiflash_lock(dev);
+
     err = spiflash_wait_ready(dev, 100);
```
utzig: Just do be sure, you are locking just above and again inside `spiflash_wait_ready`. I never remember the semantics of these mutexes but does this not cause a deadlock?

utzig:
```
Note that nested calls to`os_mutex_pend` are allowed but there needs to be one release per pend.
```

Answering my own question! :-P

PR #1541, opened 2018-12-03
adp5061: Make charge_control dependency optional

kasjer:
This change allows to disable all code related charge_control.
It will be possible to used ADP5061 API with alternative
charging manager, possibly reducing code size of application.

### DIFF
--- a/hw/drivers/chg_ctrl/adp5061/include/adp5061/adp5061.h
+++ b/hw/drivers/chg_ctrl/adp5061/include/adp5061/adp5061.h
@@ -24,7 +24,9 @@
 #include "os/os_dev.h"
 #include "syscfg/syscfg.h"
 #include "os/os_time.h"
+#if MYNEWT_VAL(ADP5061_USE_CHARGE_CONTROL)
 #include "charge-control/charge_control.h"
+#endif
 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
 #include "bus/bus_driver.h"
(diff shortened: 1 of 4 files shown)


hw/drivers/chg_ctrl/adp5061/include/adp5061/adp5061.h
```diff
@@ -49,10 +51,13 @@ struct adp5061_config {
 struct adp5061_dev {
 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
     struct bus_i2c_node     a_node;
+#if MYNEWT_VAL(ADP5061_USE_CHARGE_CONTROL)
+    struct charge_control   a_chg_ctrl;
+#endif
 #else
     struct os_dev           a_dev;
-#endif
     struct charge_control   a_chg_ctrl;
```
mkiiskila: So ADP5061_USE_CHARGE_CONTROL setting is only meaningful if BUS_DRIVER_PRESET is set?

kasjer: No, but a_chg_ctrl has cc_itf filed that is used for addressing, type of interface is also defined in charge_control so there is dependency to package and as you noticed, but charge control integration can be disabled, meaning that driver will not register to it.
This is result of having i2c/spi interface settings scattered in different places (sensor/charger maybe more).

PR #1545, opened 2018-12-03
adding battery polling rate 0 to stop polling

JustineKH:
Adding an option to stop battery polling by setting polling rate to 0

### DIFF
--- a/hw/battery/src/battery.c
+++ b/hw/battery/src/battery.c
@@ -487,8 +487,13 @@ battery_set_poll_rate_ms(struct os_dev *battery, uint32_t poll_rate)
 {
     struct battery *bat = (struct battery *)battery;
 
-    if ((poll_rate == 0) || (bat == NULL)) {
+    if ((bat == NULL)) {
         return -1;
+    }
+
+    if (poll_rate == 0) {
+        os_callout_stop(&battery_manager.bm_poll_callout);
+        return 0;
     }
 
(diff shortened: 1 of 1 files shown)


hw/battery/src/battery.c
```diff
@@ -487,10 +487,15 @@ battery_set_poll_rate_ms(struct os_dev *battery, uint32_t poll_rate)
 {
     struct battery *bat = (struct battery *)battery;
 
-    if ((poll_rate == 0) || (bat == NULL)) {
+    if ((bat == NULL)) {
```
ccollins476ad: Just a nitpick, but you could remove the extra set of parentheses here.

JustineKH: of course!

hw/battery/src/battery.c
```diff
@@ -487,10 +487,15 @@ battery_set_poll_rate_ms(struct os_dev *battery, uint32_t poll_rate)
...
         return -1;
     }
 
+    if (poll_rate == 0) {
+        os_callout_stop(&battery_manager.bm_poll_callout);
+        return 0;
```
ccollins476ad: This is another minor thing: I think it would be better to set `bat->b_poll_rate` to 0 before returning.  That way, the contents of the data structure reflect the actual state of the system.  If someone is debugging live or looking at a core, a non-zero value here could be misleading.

PR #1547, opened 2018-12-04
nrf52; create a linker script for compiling apps which are loaded to RAM,

mkiiskila:
instead of flash. Show a sample on how to use this with one BSP.


mkiiskila:
Agreed; it's better to add it to .rat-excludes now.

PR #1548, opened 2018-12-04
sys/log: Add pretty printing for CBOR log entries

andrzej-kaczmarek:
...and also some useful additions to CLI.

### DIFF
--- a/sys/log/full/src/log_shell.c
+++ b/sys/log/full/src/log_shell.c
@@ -214,7 +214,10 @@ shell_log_dump_cmd(int argc, char **argv)
     struct log *log;
     struct log_offset log_offset;
     bool last = false;
+    bool list_only;
     int rc;
+
+    list_only = ((argc > 1) && !strcmp(argv[1], "-l"));
 
     log = NULL;
     do {
@@ -224,6 +227,11 @@ shell_log_dump_cmd(int argc, char **argv)
         }
 
         if (log->l_log->log_type == LOG_TYPE_STREAM) {
+            continue;
+        }
+
+        if (list_only) {
+            console_printf("%s\n", log->l_name);
             continue;
         }
 
--- a/sys/log/full/src/log_shell.c
+++ b/sys/log/full/src/log_shell.c
@@ -58,26 +58,31 @@ shell_log_dump_entry(struct log *log, struct log_offset *log_offset,
     }
     data[rc] = 0;
 
+    console_printf("[%llu] ", ueh->ue_ts);
+
 #if MYNEWT_VAL(LOG_VERSION) <= 2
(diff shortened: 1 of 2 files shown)


sys/log/full/src/log_shell.c
```diff
@@ -99,6 +242,15 @@ shell_log_dump_all_cmd(int argc, char **argv)
             continue;
         }
 
+        if (list_only) {
+            console_printf("%s\n", log->l_name);
+            continue;
+        }
+
+        if ((argc > 1) &&  strcmp(log->l_name, argv[1])) {
```
kasjer: extra space after &&

kasjer: searching for next log could stop after one that was requested was dumped

sys/log/full/src/log_shell.c
```diff
@@ -45,49 +167,70 @@ shell_log_dump_entry(struct log *log, struct log_offset *log_offset,
...
 #else
     switch (ueh->ue_etype) {
     case LOG_ETYPE_STRING:
-        console_printf("[%llu] %s\n", ueh->ue_ts, data);
+        console_write(data, strlen(data));
+        break;
+    case LOG_ETYPE_CBOR:
+        log_shell_cbor_reader_init(&cbor_reader, log, dptr, len);
+        cbor_parser_init(&cbor_reader.r, 0, &cbor_parser, &cbor_value);
+        cbor_value_to_pretty(stdout, &cbor_value);
```
kasjer: printing to stdout would end up on console as other output, right?

sys/log/full/src/log_shell.c
andrzej-kaczmarek: agree, will fix

sys/log/full/src/log_shell.c
andrzej-kaczmarek: correct

sys/log/full/src/log_shell.c
```diff
@@ -36,6 +36,128 @@
...
+                          int src_offset, size_t len)
+{
+    struct log_shell_cbor_reader *cbr = (struct log_shell_cbor_reader *)d;
+    uint8_t buf[16];
+    int chunk_len;
+    int offset;
+
+    offset = 0;
+
+    while (offset < len) {
```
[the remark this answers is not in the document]
andrzej-kaczmarek: I do not know in advance how long the data is and also do not want to allocate huge buffers on stack. perhaps current buffer can be made a bit larger though, not sure exactly where this is used in tinycbor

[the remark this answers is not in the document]
andrzej-kaczmarek: ah, you're right - reading in chunks makes sense for memcmp but for memcpy we can read at once :-)

PR #1550, opened 2018-12-05
spiflash bus driver support

kasjer:
Code to enable bus driver in spiflash.

To use this bus package must be included and hal should add spiflash device with
call to bus_spi_node_create() with correct SPI parameters.

### DIFF
--- a/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
+++ b/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
@@ -25,6 +25,9 @@
 #endif
 #include <hal/hal_flash_int.h>
 #include <hal/hal_spi.h>
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+#include <bus/drivers/spi_common.h>
+#endif
 
 #ifdef __cplusplus
 extern "C" {
@@ -32,10 +35,14 @@ extern "C" {
 
 struct spiflash_dev {
     struct hal_flash hal;
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    struct bus_spi_node dev;
+#else
     struct hal_spi_settings spi_settings;
     int spi_num;
     void *spi_cfg;                  /** Low-level MCU SPI config */
     int ss_pin;
+#endif
     uint16_t sector_size;
     uint16_t page_size;
     /* Array of supported flash chips */
@@ -108,6 +115,11 @@ void spiflash_power_down(struct spiflash_dev *dev);
 void spiflash_release_power_down(struct spiflash_dev *dev);
 
 int spiflash_auto_power_down_set(struct spiflash_dev *dev, uint32_t timeout_ms);
+
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+int spiflash_create_spi_dev(struct bus_spi_node *node, const char *name,
+                            const struct bus_spi_node_cfg *spi_cfg);
+#endif
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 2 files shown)


hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -581,6 +585,9 @@ static const struct hal_flash_funcs spiflash_flash_funcs = {
     .hff_init         = spiflash_init,
 };
 
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+#endif
```
andrzej-kaczmarek: looks like some leftover

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -628,13 +636,23 @@ static inline void spiflash_unlock(struct spiflash_dev *dev)
 static inline void
 spiflash_cs_activate(struct spiflash_dev *dev)
 {
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    struct bus_spi_node *node = (struct bus_spi_node *)&dev->dev;
+    hal_gpio_write(node->pin_cs, 0);
```
andrzej-kaczmarek: this is redundant since `spiflash_cs_activate` is not used (?) with bus driver as CS line is controlled automatically so perhaps it should be ifdef'd-out

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -722,15 +753,21 @@ uint8_t
 spiflash_read_status(struct spiflash_dev *dev)
 {
     uint8_t val;
+    const uint8_t cmd = SPIFLASH_READ_STATUS_REGISTER;
 
     spiflash_lock(dev);
```
andrzej-kaczmarek: do we still need this lock for bus driver?

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -644,11 +662,15 @@ spiflash_power_down(struct spiflash_dev *dev)
...
 
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    bus_node_simple_write(&dev->dev.bnode.odev, cmd, sizeof(cmd));
```
andrzej-kaczmarek: use `(struct os_dev *)&dev->dev` - code outside bus driver should not access contents of bus node object and they are defined in a way that such cast is always possible by design

hw/drivers/flash/spiflash/include/spiflash/spiflash.h
```diff
@@ -25,17 +25,24 @@
 #endif
 #include <hal/hal_flash_int.h>
 #include <hal/hal_spi.h>
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+#include <bus/spi.h>
```
andrzej-kaczmarek: you could rebase this on top of https://github.com/apache/mynewt-core/pull/1568/ and just include `bus/drivers/spi_common.h` here and also add dependency to `pkg.yml`:
```
pkg.deps.BUS_DRIVER_PRESENT:
    - "@apache-mynewt-core/hw/bus/drivers/spi_common"
```

hw/drivers/flash/spiflash/src/spiflash.c
kasjer: done

hw/drivers/flash/spiflash/src/spiflash.c
kasjer: lock removed for bus driver

hw/drivers/flash/spiflash/src/spiflash.c
kasjer: done

PR #1555, opened 2018-12-06
Battery - Optional Start Delay

brolan-juul:
Adding an extra parameter to potentially delay the first event.


brolan-juul:
I just created a function for battery_set_poll_rate_ms_delay()

PR #1557, opened 2018-12-07
flash_loader; app and gdb macros to facilite populating external flash

mkiiskila:
via MCU & programmer.

### DIFF
--- a/apps/flash_loader/src/fl.c
+++ b/apps/flash_loader/src/fl.c
@@ -1,0 +1,210 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


apps/flash_loader/src/fl.c
```diff
@@ -0,0 +1,210 @@
...
+
+int
+main(int argc, char **argv)
+{
+    int rc;
+
+    hal_bsp_init();
+    flash_map_init();
+
+    fl_data = malloc(MYNEWT_VAL(FLASH_LOADER_DL_SZ));
```
kasjer: malloc? really for this?

mkiiskila: Yeah. gdb itself seems to need it. Otherwise it complains if I try to use string variables in it's macros (ie. the filename there). Initially I had this as a regular global array.

PR #1564, opened 2018-12-13
lis2dw12 improve self test

kasjer:
- Format adjusted to match mynewt style.

- Self works correctly when FIFO is enabled before self test starts.


jacobrosenthal:
Adding @mlaz as he should have some actual hardware to spin it on once (which I dont atm), but otherwise code looks good to me.

mlaz:
@jacobrosenthal I do and I was keeping my eye on this PR, waiting for it to pass on the IC tests before testing and reviewing.

kasjer:
@vrahane could you be more specific which other registers (apart from CTRL1-CTRL6 and FIFO_CTRL) you have in mind?

PR #1566, opened 2018-12-14
lis2dh12 add self test procedure

kasjer:
This adds self-test procedure.

It also fixes multi-byte write over I2C that was not incrementing register address.


jacobrosenthal:
Link or app note? All I could go off of was the datasheets recommended self test and there wasnt any explanation why any changes from that were necessary.

kasjer:
@jacobrosenthal here is a link to the application note, I'm not sure how persistent it will stay so I did not put it in the code: https://www.st.com/resource/en/application_note/dm00365457.pdf

PR #1568, opened 2018-12-17
hw/bus: Simplify bus driver creation and usage

andrzej-kaczmarek:
This reorganizes `hw/bus` packages structure and includes few fixes to make bus driver creation and usage easier.

To use bus driver it's now enough to only include `hw/bus/drivers/(i2c|spi)_common` package dependency and include `bus/drivers/(i2c|spi)_common.h` header file and all other dependencies should be automatically pulled in.

Also, creating bus node now calls generic `init` function first which then calls `init` function assigned by bus device driver. This allows to easily swap drivers for each bus device without need to modify any code which uses bus nodes, i.e. there may be two I2C bus drivers (HAL and something optimized) and the same call to create node will works properly no matter which bus device this node is attached to.

### DIFF
--- a/hw/battery/include/battery/battery_drv.h
+++ b/hw/battery/include/battery/battery_drv.h
@@ -30,9 +30,8 @@
 
 #include "os/mynewt.h"
 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
-#include "bus/bus_driver.h"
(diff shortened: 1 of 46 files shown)


andrzej-kaczmarek:
Thanks, Marko! I'll address comment about SPI in next PR with bus driver updates :-)

PR #1575, opened 2018-12-19
hw/bus: Add I2C bus driver for nRF52xxx TWIM (DMA)

andrzej-kaczmarek:
This adds optimized I2C bus driver which can be used with nRF52xxx MCUs instead of default `i2c_hal` driver.

The difference here is that it uses TWIM instead of TWI which means it uses DMA for transfer in the background and is synchronized by semaphore instead of using busy loops. This allows other tasks to execute while transfer is ongoing.

Driver can be automatically enabled with new syscfg in MCU package:
    `MCU_BUS_DRIVER_I2C_USE_TWIM: 1`

The only limitation here compared to `i2c_hal` driver is that it's not possible to make a read operation with `NOSTOP` flag. This is due to missing shortcut in TWIM peripheral to trigger `TASK_SUSPEND` on `EVENT_LASTRX`. It would be possible to workaround this with PPI, however I do not think read with NOSTOP is really that useful - `i2c_hal` does support this properly and should be used instead if needed.

Workarounds implemented in `hal_i2c` are also present in driver as it looks like TWIM also suffers from the same lockup issue. For debugging purposes stats were added which count how many times each workaround has been applied.

Note: this was created since adding TWIM support to I2C HAL requires also using semaphores there for synchronization which should not be really used in HAL (see discussion in https://github.com/apache/mynewt-core/pull/1559) but there's no problem using them in bus driver since it works on OS level.

### DIFF
--- a/hw/bus/drivers/i2c_nrf52_twim/src/i2c_nrf52_twim.c
+++ b/hw/bus/drivers/i2c_nrf52_twim/src/i2c_nrf52_twim.c
@@ -1,0 +1,659 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <assert.h>
+#include "defs/error.h"
+#include "hal/hal_gpio.h"
+#include "bus/bus.h"
+#include "bus/bus_debug.h"
+#include "bus/bus_driver.h"
+#include "bus/drivers/i2c_common.h"
+#include "mcu/nrf52_hal.h"
+#include "nrfx.h"
+
+#define TWIM_GPIO_PIN_CNF \
+    ((GPIO_PIN_CNF_SENSE_Disabled << GPIO_PIN_CNF_SENSE_Pos) |  \
+     (GPIO_PIN_CNF_DRIVE_S0D1 << GPIO_PIN_CNF_DRIVE_Pos) |      \
+     (GPIO_PIN_CNF_PULL_Pullup << GPIO_PIN_CNF_PULL_Pos) |      \
+     (GPIO_PIN_CNF_INPUT_Connect << GPIO_PIN_CNF_INPUT_Pos) |   \
+     (GPIO_PIN_CNF_DIR_Input << GPIO_PIN_CNF_DIR_Pos))
+
+#if MYNEWT_VAL(I2C_NRF52_TWIM_STAT)
+STATS_SECT_START(twim_stats_section)
+    STATS_SECT_ENTRY(sda_lo_err)        /* SDA pulled low on r/w */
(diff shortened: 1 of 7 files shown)


hw/bus/drivers/i2c_nrf52_twim/src/i2c_nrf52_twim.c
```diff
@@ -0,0 +1,659 @@
...
+    }
+
+    if (nrf_twim->EVENTS_ERROR) {
+        nrf_twim->EVENTS_ERROR = 0;
+    }
+
+    dd->errorsrc = nrf_twim->ERRORSRC;
+    nrf_twim->ERRORSRC = dd->errorsrc;
+
+    os_sem_release(&twim_devs_data[dev->cfg.i2c_num].sem);
```
kasjer: dd could be used instead of array access

hw/bus/drivers/i2c_nrf52_twim/src/i2c_nrf52_twim.c
```diff
@@ -0,0 +1,659 @@
...
+        STATS_INC(dd->stats, sda_lo_err_nrecov);
+    }
+#endif
+}
+
+static void
+nrf_twim_start_task(NRF_TWIM_Type *nrf_twim, struct twim_dev_data *dd,
+                    __O uint32_t *task_start, __IO uint32_t *event_last)
+{
+    static const int max_attempt = 2;
```
kasjer: static here does not change anything

hw/bus/drivers/i2c_nrf52_twim/src/i2c_nrf52_twim.c
```diff
@@ -0,0 +1,659 @@
...
+     * releases SDA line mid-ack which looks like premature STOP condition on bus
+     * followed by STOP condition from TWIM controller. After this sequence,
+     * TWIM controller needs to be reset in order to work properly.
+     *
+     * To workaround this problem we can just check if SCL line is pulled low
+     * after triggering start task as this indicates some activity and means
+     * that controller is responsive.
+     */
+
+    do {
```
kasjer: maybe while (1) at the beginning instead of do {} while (1)

hw/bus/drivers/i2c_nrf52_twim/src/i2c_nrf52_twim.c
```diff
@@ -0,0 +1,659 @@
...
+    nrf_twim->EVENTS_STOPPED = 0;
+    nrf_twim->EVENTS_ERROR = 0;
+    nrf_twim->EVENTS_SUSPENDED = 0;
+    nrf_twim->EVENTS_RXSTARTED = 0;
+    nrf_twim->TASKS_RESUME = 1;
+
+    nrf_twim_start_task(nrf_twim, dd, &nrf_twim->TASKS_STARTRX,
+                        &nrf_twim->EVENTS_LASTRX);
+
+    rc = os_sem_pend(&twim_devs_data[dev->cfg.i2c_num].sem, timeout);
```
kasjer: dd could be used instead of array access

hw/bus/drivers/i2c_nrf52_twim/src/i2c_nrf52_twim.c
```diff
@@ -0,0 +1,659 @@
...
+    nrf_twim->EVENTS_ERROR = 0;
+    nrf_twim->EVENTS_STOPPED = 0;
+    nrf_twim->EVENTS_SUSPENDED = 0;
+    nrf_twim->EVENTS_TXSTARTED = 0;
+    nrf_twim->TASKS_RESUME = 1;
+
+    nrf_twim_start_task(nrf_twim, dd, &nrf_twim->TASKS_STARTTX,
+                        &nrf_twim->EVENTS_LASTTX);
+
+    rc = os_sem_pend(&twim_devs_data[dev->cfg.i2c_num].sem, timeout);
```
kasjer: dd instead of array access could be used here

andrzej-kaczmarek:
Thanks, @kasjer - I pushed fixes in separate commit.

PR #1586, opened 2019-01-03
Use stlink tooling for stm32l4

utzig:
As reported previously, openocd 0.10.0 and git master have changed the script name about one year ago which forces users to have to edit the configuration files according to the openocd version they have
installed. This switches to more friendly st-link based tooling which has more regular releases and does auto-detect the stm32 line MCU in use.


utzig:
Better to be merged after #1585 

@alvarop ^

PR #1587, opened 2019-01-04
sys/log: Fix endianess in CBOR pretty-printing

andrzej-kaczmarek:


andrzej-kaczmarek:
travis crashed on one job...

PR #1588, opened 2019-01-04
NRF52840 i2c pin config handles Port1

FabienLep:


hw/mcu/nordic/nrf52xxx/src/hal_i2c.c
```diff
@@ -297,8 +297,22 @@ hal_i2c_init(uint8_t i2c_num, void *usercfg)
     scl_port = HAL_GPIO_PORT(cfg->scl_pin);
     sda_port = HAL_GPIO_PORT(cfg->sda_pin);
 
-    scl_port->PIN_CNF[cfg->scl_pin] = NRF52_SCL_PIN_CONF;
+    /* Handle pin configuration for nrf52840 on port 1*/
+#if MYNEWT_VAL(BSP_NRF52840)
+    if (sda_port == NRF_P1) {
+        sda_port->PIN_CNF[cfg->sda_pin-32] = NRF52_SDA_PIN_CONF;
```
andrzej-kaczmarek: there's no need for any branching or conditional compilation here, just use `HAL_GPIO_INDEX` macro to calculate pin index

FabienLep: Good catch didn't see that one. Changed.

andrzej-kaczmarek:
@FabienLep latest change looks good, however your PR also includes commit than changes repository version to 1.5.0 which is not ok. I suspect you created fix on `1.5.0` and then created PR for `master` branch. You need create fix on `master` branch and the you can force push the result to the same branch on GitHub to update this PR.

FabienLep:
@andrzej-kaczmarek , updated branch with changes from master. tell me if that is ok

PR #1589, opened 2019-01-07
hw/util/i2cn: Add write-read transaction support

andrzej-kaczmarek:
New `i2cn_master_write_read_transact` should be used for reading from common sensors instead of separate write and read as it guarantees that in case of read failure write will be done before retrying. Without retry, it may happen that after writing register address, read fails after some bytes are already left and register pointer inside sensor already moved to new location. Reading again, without write, would then result in invalid registers being read.

Note: there are few more sensors to fix, I'll try to update soon

### DIFF
--- a/hw/drivers/sensors/lps33thw/src/lps33thw.c
+++ b/hw/drivers/sensors/lps33thw/src/lps33thw.c
@@ -361,34 +361,20 @@ lps33thw_i2c_get_regs(struct sensor_itf *itf, uint8_t reg, uint8_t size,
 
     rc = bus_node_simple_write_read_transact(odev, &reg, 1, buffer, size);
 #else
-    struct hal_i2c_master_data data_struct = {
+    struct hal_i2c_master_data wdata = {
         .address = itf->si_addr,
         .len = 1,
         .buffer = &reg
     };
+    struct hal_i2c_master_data rdata = {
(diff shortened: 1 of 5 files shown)


hw/drivers/sensors/lps33thw/src/lps33thw.c
```diff
@@ -361,34 +361,20 @@ lps33thw_i2c_get_regs(struct sensor_itf *itf, uint8_t reg, uint8_t size,
...
-    data_struct.len = size;
-    data_struct.buffer = buffer;
-    rc = i2cn_master_read(itf->si_num, &data_struct,
-                          (MYNEWT_VAL(LPS33THW_I2C_TIMEOUT_TICKS)) * size, 1,
-                          MYNEWT_VAL(LPS33THW_I2C_RETRIES));
-
-    if (rc) {
-        LPS33THW_LOG(ERROR, "Failed to read from 0x%02X:0x%02X\n",
-                    itf->si_addr, reg);
-        STATS_INC(g_lps33thwstats, read_errors);
```
ccollins476ad: Should these log / stat operations be preserved?

andrzej-kaczmarek: yes, not sure why I removed it here (they are present in other drivers)

PR #1590, opened 2019-01-08
Adding Flash Write Functionality

brolan-juul:
Adding flash write functionality.
Adding new function as old function had incorrect naming. This is done so we aren't breaking any APIs.


hw/drivers/bq27z561/src/bq27z561.c
```diff
@@ -587,6 +587,104 @@ bq27x561_rd_flash(struct bq27z561 *dev, uint16_t addr, uint8_t *buf, int buflen)
...
+    struct os_dev *odev = &dev->dev.dev;
+#else
+    struct hal_i2c_master_data i2c;
+#endif
+
+    if ((buflen == 0) || (!buf) || (buflen > BQ27Z561_MAX_FLASH_RW_LEN)) {
+        return BQ27Z561_ERR_INV_PARAMS;
+    }
+
+    if ((addr < BQ27Z561_FLASH_BEG_ADDR) || (addr > BQ27Z561_FLASH_END_ADDR)) {
```
mkiiskila: if ((addr < BQ27Z561_FLASH_BEG_ADDR) || (addr + buflen > BQ27Z561_FLASH_END_ADDR)) {

hw/drivers/bq27z561/src/bq27z561.c
```diff
@@ -587,6 +587,104 @@ bq27x561_rd_flash(struct bq27z561 *dev, uint16_t addr, uint8_t *buf, int buflen)
...
+    i2c.buffer = tmpbuf;
+    rc = i2cn_master_write(dev->bq27_itf.itf_num, &i2c, MYNEWT_VAL(BQ27Z561_I2C_TIMEOUT_TICKS), 1,
+                           MYNEWT_VAL(BQ27Z561_I2C_RETRIES));
+    if (rc != 0) {
+        BQ27Z561_LOG(ERROR, "I2C reg read (wr) failed 0x%02X\n", tmpbuf[0]);
+        ret = BQ27Z561_ERR_I2C_ERR;
+    }
+
+err:
+    bq27z561_itf_unlock(&dev->bq27_itf);
```
mkiiskila: Calls to unlock device in error path when BUS_DRIVER_PRESENT happen before goto err;. Calls to unlock device in error path when BUS_DRIVER_PRESENT is not set happen after err: label.

I would suggest making the logic to be the same for both cases. My preference would be to do the unlocking after the err: label.

hw/drivers/bq27z561/src/bq27z561.c
brolan-juul: It's a bit unclear to me how this out of bounds check was originally intended to work. BQ27Z561_FLASH_END_ADDR ends at 0x4600. The last flash data register is located at 0x4526 with a register length of a byte.

Based on this information, we should never be out of bounds unless our buflen is large.

I'm happy putting this in there but I think the intended effect will be negligible.

hw/drivers/bq27z561/src/bq27z561.c
brolan-juul: Agreed. Updated.

PR #1591, opened 2019-01-09
 sys/mfg: 2.0 support

ccollins476ad:
This commit replaces 1.0 mfgimage support with support for version 2.0. For more information regarding mfgimage 2.0, see the documentation at: `docs/os/modules/mfg/mfg.rst`.

Also see this email thread: https://lists.apache.org/thread.html/4185891ef587a778722a39fe5a1ae78bfd0dc7ab7e9348a7ea147df7@%3Cdev.mynewt.apache.org%3E


ccollins476ad:
Thanks, @utzig!  I'm a bit leery of adding those null pointer checks because this code gets included in the boot loader.  The boot loader size is already right at the edge.  I think I will keep it as is for now, and we can improve the error checking later if it makes sense to do so.

PR #1593, opened 2019-01-11
console newtmgr coexistance

kasjer:
When newtmgr is used over serial port it is easy to have errors when some task prints on serial console during newtmgr communication (for example during coredump download).

It happens due to lack of locks in console code.

This PR adds locking functionality to console and updates shell_nlip to actually lock console during packet transmission.
Without this locking transmitted packets byte could interleave with logs and that would lead to communication failure.

### DIFF
--- a/sys/console/full/syscfg.yml
+++ b/sys/console/full/syscfg.yml
@@ -93,6 +93,10 @@ syscfg.defs:
             affect device performance due to more frequent polling.
         value: 250
 
+    CONSOLE_DEFAULT_LOCK_TIMEOUT:
+        description: 'Default timeout (in ms) for console_lock() function used in console_write.'
+        value: 1000
+
     CONSOLE_SYSINIT_STAGE:
         description: >
             Sysinit stage for console functionality.
--- a/sys/console/minimal/src/console.c
+++ b/sys/console/minimal/src/console.c
@@ -57,9 +57,10 @@ static uint8_t cur, end;
 static struct os_eventq *avail_queue;
 static struct os_eventq *lines_queue;
 bool g_silence_console;
+static struct os_mutex console_write_lock;
 
 int __attribute__((weak))
-console_out(int c)
+console_out_nolock(int c)
 {
     return c;
 }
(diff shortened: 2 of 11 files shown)


sys/console/full/syscfg.yml
```diff
@@ -93,6 +93,10 @@ syscfg.defs:
             affect device performance due to more frequent polling.
         value: 250
 
+    CONSOLE_DEFAULT_LOCK_TIMEOUT:
+        description: 'Default timeout (int ticks) for console_lock() function.'
+        value: 10
```
ccollins476ad: 10 is <=100ms on most systems. This feels a bit short to me.  Console output is slow, and I can imagine this timeout expiring when multiple tasks want to write.  Do you think something longer, such as 1 second, would be better?

Also, I would not use OS ticks are the units here.  Since OS ticks vary by platform, the default is not consistent.  I would use milliseconds here, and define a `CONSOLE_LOCK_TIMEOUT_TICKS` macro in the code.

Finally, this is a bit of a nitpick, but I don't think this setting name should contain `DEFAULT`.  It is not possible to override this setting at runtime, so it is more than just a default; it is the setting value itself!

sys/console/minimal/src/console.c
```diff
@@ -70,16 +71,73 @@ console_echo(int on)
...
+        goto end;
+    }
+
+    rc = os_mutex_pend(&console_write_lock, timeout);
+    if (rc == OS_NOT_STARTED) {
+        /* No need to block before system start make it OK */
+        rc = OS_OK;
+    }
+
+    end:
```
ccollins476ad: Nitpick: for consistency, this `end:` label should not be indented.

sys/console/minimal/src/console.c
```diff
@@ -70,16 +71,73 @@ console_echo(int on)
...
+{
+    int rc = OS_OK;
+
+    if (os_arch_in_isr()) {
+        goto end;
+    }
+
+    rc = os_mutex_release(&console_write_lock);
+    assert(rc == OS_OK || rc == OS_NOT_STARTED);
+    end:
```
ccollins476ad: Nitpick: for consistency, this `end:` label should not be indented.

sys/console/minimal/syscfg.yml
```diff
@@ -50,6 +50,10 @@ syscfg.defs:
         description: 'Console UART device.'
         value: '"uart0"'
 
+    CONSOLE_DEFAULT_LOCK_TIMEOUT:
+        description: 'Default timeout (int ticks) for console_lock() function.'
+        value: 10
```
ccollins476ad: The comments about the sibling syscfg setting apply here.

sys/shell/src/shell_nlip.c
```diff
@@ -156,6 +156,10 @@ shell_nlip_mtx(struct os_mbuf *m)
     nwritten = 0;
     off = 0;
 
+    rc = console_lock(OS_TICKS_PER_SEC);
+    if (rc != OS_OK) {
+        goto err;
+    }
     /* Start a packet */
     console_write(pkt_seq, sizeof(pkt_seq));
```
ccollins476ad: These calls to `console_write()` lock the mutex a second time.  This is legal, but I find it a bit confusing.  I don't know if it is worth adding a `console_write_nolock()` function, though.

sys/console/minimal/src/console.c
kasjer: done

sys/console/minimal/src/console.c
kasjer: done

sys/console/full/syscfg.yml
kasjer: Changed to ms.
CONSOLE_DEFAULT_LOCK_TIMEOUT left since it specifies value used in locking console from console_write and console_out functions that do not provide timeout.

PR #1600, opened 2019-01-16
Only erase FLASH_BANK_2 when available

alvarop:
Some parts in the STM32L4xx family do not have a FLASH_BANK_2.
(The STM32L432 for example)

This results in a compile error. Added #define to prevent this.

### DIFF
--- a/hw/mcu/stm/stm32l4xx/src/hal_flash.c
+++ b/hw/mcu/stm/stm32l4xx/src/hal_flash.c
@@ -36,11 +36,16 @@ stm32_mcu_flash_erase_sector(const struct hal_flash *dev, uint32_t sector_addres
 
     if (!(sector_address & (_FLASH_SECTOR_SIZE - 1))) {
         eraseinit.TypeErase = FLASH_TYPEERASE_PAGES;
+#ifdef FLASH_BANK_2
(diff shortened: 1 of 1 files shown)


hw/mcu/stm/stm32l4xx/src/hal_flash.c
```diff
@@ -38,9 +38,12 @@ stm32_mcu_flash_erase_sector(const struct hal_flash *dev, uint32_t sector_addres
         eraseinit.TypeErase = FLASH_TYPEERASE_PAGES;
         if ((sector_address - dev->hf_base_addr) < (_FLASH_SIZE / 2)) {
```
utzig: Is it still OK to test against `_FLASH_SIZE / 2` when there's only a single bank? Won't this stop the second half of the flash from being erased? I think maybe it would be better to make `eraseinit.Banks = FLASH_BANK_1;` before the test and then test if we're in the 2nd half and change the value to `FLASH_BANK_2` (inside the #ifdef). What do you think?

alvarop: Good call! Didn't even think of that. IO'll take a look

alvarop:
Fixed and squashed the commits :D

PR #1601, opened 2019-01-16
Adding hal_nvreg.c for STM32L4xx devices.

alvarop:
The STM32L4xx family has 32 32-bit RTC backup registers.
These can be used to store data that persists after a system reset.

Tested on the STM32L432KC.

### DIFF
--- a/hw/mcu/stm/stm32_common/src/hal_nvreg.c
+++ b/hw/mcu/stm/stm32_common/src/hal_nvreg.c
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <hal/hal_nvreg.h>
+#include <mcu/stm32_hal.h>
+
+#if defined RTC_BACKUP_SUPPORT
+#define HAL_NVREG_MAX (RTC_BKP_NUMBER)
+#else
+#define HAL_NVREG_MAX (0)
+#endif
+
+/* RTC backup registers are 32-bits wide */
+#define HAL_NVREG_WIDTH_BYTES (4)
+
+#if defined RTC_BACKUP_SUPPORT
+static volatile uint32_t *regs[HAL_NVREG_MAX] = {
+#if HAL_NVREG_MAX > 0
+    &RTC->BKP0R,
+    &RTC->BKP1R,
+    &RTC->BKP2R,
+    &RTC->BKP3R,
+    &RTC->BKP4R,
+#endif /* HAL_NVREG_MAX > 0 */
+#if HAL_NVREG_MAX > 5
+    &RTC->BKP5R,
+    &RTC->BKP6R,
+    &RTC->BKP7R,
+    &RTC->BKP8R,
+    &RTC->BKP9R,
+    &RTC->BKP10R,
+    &RTC->BKP11R,
+    &RTC->BKP12R,
+    &RTC->BKP13R,
+    &RTC->BKP14R,
+    &RTC->BKP15R,
+#endif /* HAL_NVREG_MAX > 5 */
+#if HAL_NVREG_MAX > 16
+    &RTC->BKP16R,
+    &RTC->BKP17R,
+    &RTC->BKP18R,
+    &RTC->BKP19R,
+#endif /* HAL_NVREG_MAX > 16 */
+#if HAL_NVREG_MAX > 20
+    &RTC->BKP20R,
+    &RTC->BKP21R,
+    &RTC->BKP22R,
+    &RTC->BKP23R,
+    &RTC->BKP24R,
+    &RTC->BKP25R,
+    &RTC->BKP26R,
+    &RTC->BKP27R,
+    &RTC->BKP28R,
+    &RTC->BKP29R,
+    &RTC->BKP30R,
+    &RTC->BKP31R,
+#endif /* HAL_NVREG_MAX > 20 */
+};
+#endif /* RTC_BACKUP_SUPPORT */
+
+void
+hal_nvreg_write(unsigned int reg, uint32_t val)
+{
+#if defined RTC_BACKUP_SUPPORT
+    if (reg < HAL_NVREG_MAX) {
+        HAL_PWR_EnableBkUpAccess();
+        *regs[reg] = val;
+        HAL_PWR_DisableBkUpAccess();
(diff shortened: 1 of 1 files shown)


hw/mcu/stm/stm32l4xx/src/hal_nvreg.c
```diff
@@ -0,0 +1,102 @@
...
+    &RTC->BKP28R,
+    &RTC->BKP29R,
+    &RTC->BKP30R,
+    &RTC->BKP31R
+};
+
+void
+hal_nvreg_write(unsigned int reg, uint32_t val)
+{
+    if(reg < HAL_NVREG_MAX)
```
utzig: Please fix the style, a space between `if (` and open braces at the end of same line.

hw/mcu/stm/stm32l4xx/src/hal_nvreg.c
```diff
@@ -0,0 +1,102 @@
...
+{
+    uint32_t val = 0;
+
+    if(reg < HAL_NVREG_MAX)
+    {
+        HAL_PWR_EnableBkUpAccess();
+        val = *regs[reg];
+        HAL_PWR_DisableBkUpAccess();
+    }
+
```
utzig: remove a few of those empty lines.

utzig:
This looks OK, I am gonna merge after the few nits are addressed. Do you have any idea if this is available on other MCU models/lines? Maybe it can be moved to `stm32_common` later.

hw/mcu/stm/stm32l4xx/src/hal_nvreg.c
```diff
@@ -0,0 +1,102 @@
...
+
+
+    return val;
+}
+
+unsigned int hal_nvreg_get_num_regs(void)
+{
```
utzig: the function names should be on a new line after the return type

mkiiskila:
These registers are present in other STM32s, and these registers are exactly the ones I thought we should use as nvreg with STM32 :). So I agree Fabio, this code should move to stm32_common.

utzig:
@alvarop Please also rebase this on latest master to fix the travis issues.

alvarop:
> These registers are present in other STM32s, and these registers are exactly the ones I thought we should use as nvreg with STM32 :). So I agree Fabio, this code should move to stm32_common.

I did a quick survey of all the stm32xxxxxx.h files and, unfortunately, there are some devices that don't have the RTC backup registers (19/138 don't have any. This includes some from the stm32f0 and stm32f1 families).
The STM32L4xx family all has them, but some only have 0-4.

I added a #ifndef check so it will #error when the device does not support it.

utzig:
@alvarop Please squash your commits; also you still have braces below `if` instead of at the end of the same line.

alvarop:
Will do. Having to re-learn git after my screwups :D

alvarop:
Should be squashed and fixed now!

hw/mcu/stm/stm32_common/src/hal_nvreg.c
```diff
@@ -0,0 +1,78 @@
...
+#include <hal/hal_nvreg.h>
+#include <mcu/stm32_hal.h>
+
+/* Check header file for RTC backup register support */
+#ifndef RTC_BKP0R
+#error This device does not have any RTC backup registers.
+#endif
+
+/* There are 5 RTC backup registers on all stm32l4xx devices */
+#define HAL_NVREG_MAX (5)
```
utzig: This might end up being MCU specific so could be moved to each `stm32_hal.h`...

utzig: Ah, also the comment uses `stm32l4xx` which is not correct after the move.

alvarop:
Figured out a way to do it in file without having to specify the mcu. Let me know what you think!

hw/mcu/stm/stm32_common/src/hal_nvreg.c
```diff
@@ -0,0 +1,118 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <mcu/cortex_m4.h>
+#include <hal/hal_nvreg.h>
+#include <mcu/stm32_hal.h>
+
+/* Use header files to determine number of RTC Backup registers */
+#if defined RTC_BKP31R
```
utzig: There is already a macro named `RTC_BKP_NUMBER`.

hw/mcu/stm/stm32_common/src/hal_nvreg.c
```diff
@@ -0,0 +1,118 @@
...
+    &RTC->BKP12R,
+    &RTC->BKP13R,
+    &RTC->BKP14R,
+    &RTC->BKP15R,
+#if HAL_NVREG_MAX > 16
+    &RTC->BKP16R,
+    &RTC->BKP17R,
+    &RTC->BKP18R,
+    &RTC->BKP19R,
+    &RTC->BKP20R,
```
utzig: You can add `#if HAL_NVREG_MAX > 20` because many devices have that amount of registers.

hw/mcu/stm/stm32_common/src/hal_nvreg.c
alvarop: Don't know how I missed this!

hw/mcu/stm/stm32_common/src/hal_nvreg.c
alvarop: Changed it to match what they themselves have in the rtc header files.
For example: https://github.com/apache/mynewt-core/blob/master/hw/mcu/stm/stm32l4xx/src/ext/Drivers/STM32L4xx_HAL_Driver/Inc/stm32l4xx_ll_rtc.h#L562

alvarop:
Updated!
I don't really understand all the travis failing stuff. If there's anything I can do about that, let me know!

hw/mcu/stm/stm32_common/src/hal_nvreg.c
```diff
@@ -0,0 +1,110 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <hal/hal_nvreg.h>
+#include <mcu/stm32_hal.h>
+
+#if defined RTC_BACKUP_SUPPORT
```
utzig: I think you can use this define to protect the whole file, because it's failing on the F1 BSPs on Travis.

utzig: Otherwise you could as well use `#if !MYNEWT_VAL(MCU_STM32F1)` if the above suggestion doesn't work!

alvarop: Ah yes, I'm finally understanding.
The bluepill one has RTC_BKP_NUMBER defined since it has some other backup register thing.
I'll use RTC_BACKUP_SUPPORT as you suggested.

alvarop:
Woah, this Travis.ci thing is super neat! I didn't know it could work with embedded c stuff. I'll definitely have to set this up for my own projects in the future.

I had the top define flipped, so it was setting HAL_NVREG_MAX to (0) when RTC_BACKUP_SUPPORT was there :man_facepalming:. I also added it in the main functions since HAL_PWR_EnableBkUpAccess() wont be there when there's no support.

alvarop:
Is there a way to run these tests locally so I don't spam everyone and github with emails?

alvarop:
Woo! Looks like the last error left is not from my changes :D

utzig:
> Is there a way to run these tests locally so I don't spam everyone and github with emails?

It's not possible without having to make a lot of changes first!

PR #1605, opened 2019-01-17
hw/bus: Add power management for bus devices

andrzej-kaczmarek:
This adds some power management capabilities to bus devices. By default, it just allows to disable bus device by calling `os_dev_suspend` and enable it by calling `os_dev_resume`.

Extra pm capabilities can be enabled with `BUS_PM: 1` which adds automatic pm mode. It can be enabled in runtime at any time and allows to control enabled state of bus device automatically depending on its inactivity, i.e. it will enable bus device before transaction and will disable it when not needed. This allows for simpler app and avoids duplicating code unnecessarily.

### DIFF
--- a/hw/bus/src/bus.c
+++ b/hw/bus/src/bus.c
@@ -343,6 +343,16 @@ bus_node_lock(struct os_dev *node, os_time_t timeout)
         return 0;
     }
 
+    /*
+     * Configuration is done on 1st lock so in case we need to configure device
+     * on nested lock it means that most likely bus device was locked for one
+     * node and then access is done on another node which is not correct.
+     */
+    if (os_mutex_get_level(&bdev->lock) != 1) {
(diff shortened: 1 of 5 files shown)


hw/bus/src/bus.c
```diff
@@ -125,6 +137,29 @@ bus_dev_resume_func(struct os_dev *odev)
...
 
+#if MYNEWT_VAL(BUS_PM)
+static void
+bus_dev_inactivity_tmo_func(struct os_event *ev)
+{
+    struct bus_dev *bdev = (struct bus_dev *)ev->ev_arg;
+    int rc;
+
+    /* Just in case PM was changed while timer was running */
+    if (bdev->pm_mode != BUS_PM_MODE_AUTO) {
```
sjanc: should this be under bdev->lock ?

PR #1608, opened 2019-01-18
hw/drivers/spiflash: Add auto power down

andrzej-kaczmarek:
This adds new API which allows to enable auto power down for SPI flash.
Application can define timeout after which power down command will be
sent to flash. It will be automatically release from power down for any
new request from application.

Note: There should be some delay between sending power down release
command and using flash, but the required delay was so small on flashes
I tested that it does not really matter. If required at some point, we
should add this to SPI flash configuration matrix.

### DIFF
--- a/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
+++ b/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
@@ -45,6 +45,13 @@ struct spiflash_dev {
 #if MYNEWT_VAL(OS_SCHEDULING)
     struct os_mutex lock;
 #endif
+#if MYNEWT_VAL(SPIFLASH_AUTO_POWER_DOWN)
+#if MYNEWT_VAL(OS_SCHEDULING)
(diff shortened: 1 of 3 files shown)


hw/drivers/flash/spiflash/include/spiflash/spiflash.h
```diff
@@ -44,6 +44,9 @@ struct spiflash_dev {
     const struct spiflash_chip *flash_chip;
 #if MYNEWT_VAL(OS_SCHEDULING)
     struct os_mutex lock;
+    struct os_callout apd_tmo_co;   /* Auto power down timeout callout */
+    os_time_t apd_tmo;              /* Auto power down timeout value (ticks) */
+    bool pd_active;                 /* Power down active */
```
kasjer: I could be beneficial to have pd_active visible even without scheduling. Then it could be used to automatically wake up flash on each access as it is done right now with this new feature.

andrzej-kaczmarek:
@kasjer thanks, fixed!

PR #1610, opened 2019-01-22
mgmt/imgmgr: Accept slot in `test` and `confirm` CLI commands

ccollins476ad:
Prior to PR: the `imgr test` and (optionally) the `imgr confirm` commands required their argument to be the hash of the image to switch to.

Now: the argument can be either 1) image slot index, or 2) image hash.

Rationale: Allow for a consistent and simple image swap process during manufacturing.  The process should not depend on the hash of the secondary image.


mgmt/imgmgr/src/imgmgr_cli.c
```diff
@@ -114,34 +130,62 @@ imgr_cli_hash_parse(const char *arg, int *out_slot)
...
+static int
+imgr_cli_slot_or_hash_parse(const char *arg, int *out_slot)
+{
+    int rc;
+
+    /* First, parse argument as a slot number.  Parts of the system assume slot
+     * is 0 or 1; enforce those bounds here.
+     */
+    *out_slot = parse_ll_bounds(arg, 0, 1, &rc);
+    if (rc == 0) {
```
utzig: checking `rc == SYS_EOK` would probably be better here in the long run...

ccollins476ad: Thanks, @utzig.  It might just be a pet peeve of mine, but I really don't like named "OK" return codes.  To me, such constants look like errors at first glance, and it takes a second look to understand what is actually happening.

I think most of the code in core just uses `0` to signify success, so if it is OK with you, I would like to keep this as it is.

utzig: Yup, I prefer `0` as non-error return value as well (unless it's a bool return); but maybe not having `*_EOK` would be better than...

PR #1613, opened 2019-01-25
Fix disconnect during OTA by spacing out erases

saulpw:
* Erasing the entire flash image size at once can take significant time, causing a bluetooth disconnect or significant battery sag.  Instead, we erase immediately prior to crossing a sector while writing the image.

* We could check for empty to increase efficiency.  However, for simplicity and consistency we will always erase lazily.

* wrap with `MYNEWT_VAL(IMGMGR_LAZY_ERASE)`.  Set IMGMGR_LAZY_ERASE to 0 to disable (default, same as existing behavior), or 1 to enable.

### DIFF
--- a/mgmt/imgmgr/src/imgmgr.c
+++ b/mgmt/imgmgr/src/imgmgr.c
@@ -127,6 +127,11 @@ static struct {
     /** Hash of image data; used for resumption of a partial upload. */
     uint8_t data_sha_len;
     uint8_t data_sha[IMGMGR_DATA_SHA_LEN];
+#if MYNEWT_VAL(IMGMGR_LAZY_ERASE)
+    int sector_id;
+    uint32_t sector_end;
+#endif
+
 } imgr_state;
 
 static imgr_upload_fn *imgr_upload_cb;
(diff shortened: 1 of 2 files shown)


mgmt/imgmgr/src/imgmgr.c
```diff
@@ -564,19 +606,22 @@ imgr_upload_inspect(const struct imgr_upload_req *req,
...
         }
 
+#if MYNEWT_VAL(IMGMGR_LAZY_ERASE) == 0
         rc = flash_area_open(action->area_id, &fa);
         if (rc) {
             *errstr = imgmgr_err_str_flash_open_failed;
             return MGMT_ERR_EUNKNOWN;
         }
 
+        bool empty;
```
andrzej-kaczmarek: variables should be declared at beginning of function. you can either `#ifdef` it there or alternatively use `(void)empty;` so compiler won't complain - I'm fine with either way of doing this.

mgmt/imgmgr/syscfg.yml
```diff
@@ -30,6 +30,11 @@ syscfg.defs:
             The maximum amount of image or core data that can fit in a
             single NMP message
         value: 512
+    IMGMGR_LAZY_ERASE:
+        description: >
+            During a firmware upgrade, erase flash a sector at a time
+            prior to writing to it, rather than all at once at start
+        value: 1
```
andrzej-kaczmarek: since this changes current behavior it should be disabled by default

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -739,19 +784,32 @@ imgr_upload(struct mgmt_cbuf *cb)
...
     /* Write the image data to flash. */
     if (rc == 0 && req.data_len != 0) {
-        rc = flash_area_write(fa, req.off, req.img_data, action.write_bytes);
-        if (rc != 0) {
+#if MYNEWT_VAL(IMGMGR_LAZY_ERASE)
+        /* erase as we cross sector boundaries */
+        if (imgr_erase_if_needed(fa, req.off, action.write_bytes) != 0) {
+            rc = MGMT_ERR_EUNKNOWN;
+            errstr = imgmgr_err_str_flash_erase_failed;
+        } else
```
andrzej-kaczmarek: such `if`s split by `#ifdef` are quite confusing so I'd prefer to avoid this by either using `goto` or perhaps just moving this into its own `if (rc == 0 && req.data_len != 0)` and `#ifdef` as a whole

saulpw: Ironically we had a `goto` in the first place and I argued that we should match the surrounding code style and remove it.  Looks like @jerobi was right in the first place :)  Fixed now.

PR #1629, opened 2019-02-01
hw/mcu/nrf52: Add option to use GPIOTE port event

andrzej-kaczmarek:
Current implementation of interrupts in `hal_gpio` uses `GPIOTE` pin events which allows for easier and more accurate handling of multiple pins but in this mode HFCLK is always running which increases power consumption.

Another option is to use `GPIOTE` port event which does not require HFCLK but has some limitations like less accurate detection, level state sensing only and uses single event (or-ed) for all pins.

This patch adds syscfg option which enables `hal_gpio` to use `GPIOTE` port event for handling interrupts. This involves some extra logic which is required to translate single event into events for specific pin.

To enable set `MCU_GPIO_USE_PORT_EVENT: 1`.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
@@ -64,10 +64,21 @@
 /* GPIO interrupts */
 #define HAL_GPIO_MAX_IRQ        8
 
+#if MYNEWT_VAL(MCU_GPIO_USE_PORT_EVENT)
+#define HAL_GPIO_SENSE_TRIG_NONE    0x00 /* just 0 */
+#define HAL_GPIO_SENSE_TRIG_BOTH    0x01 /* something else than both below */
+#define HAL_GPIO_SENSE_TRIG_HIGH    0x02 /* GPIO_PIN_CNF_SENSE_High */
+#define HAL_GPIO_SENSE_TRIG_LOW     0x03 /* GPIO_PIN_CNF_SENSE_Low */
+#endif
+
 /* Storage for GPIO callbacks. */
 struct hal_gpio_irq {
     hal_gpio_irq_handler_t func;
     void *arg;
+#if MYNEWT_VAL(MCU_GPIO_USE_PORT_EVENT)
+    int pin;
+    uint8_t sense_trig;
+#endif
 };
 
 static struct hal_gpio_irq hal_gpio_irqs[HAL_GPIO_MAX_IRQ];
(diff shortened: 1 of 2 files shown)


hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
```diff
@@ -381,11 +521,35 @@ hal_gpio_irq_enable(int pin)
...
+#if MYNEWT_VAL(MCU_GPIO_USE_PORT_EVENT)
+    nrf_gpio = HAL_GPIO_PORT(pin);
+    pin_index = HAL_GPIO_INDEX(pin);
+
+    nrf_gpio->PIN_CNF[pin_index] &= ~GPIO_PIN_CNF_SENSE_Msk;
+
+    for (i = 0; i < HAL_GPIO_MAX_IRQ; i++) {
+        if (hal_gpio_irqs[i].sense_trig != HAL_GPIO_SENSE_TRIG_NONE) {
+            sense_enabled = true;
+            continue;
```
wes3: I think this should be a break here; not a continue.

hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
```diff
@@ -207,17 +218,83 @@ hal_gpio_toggle(int pin)
...
+
+        if (!sense_trig) {
+            continue;
+        }
+
+        /*
+         * A little hack... SENSE values are 0x02 for high and 0x03 for low, so
+         * bit #0 is the opposite of state which triggers interrupt.
+         */
+        pin_state = gpio_state & (1 << hal_gpio_irqs[i].pin);
```
kasjer: For 52840 64 bits from gpio_state will not fit in 32 bits of pin_state

hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
andrzej-kaczmarek: indeed :-)

hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
```diff
@@ -394,7 +394,7 @@ hal_gpio_irq_init(int pin, hal_gpio_irq_handler_t handler, void *arg,
     hal_gpio_init_in(pin, pull);
 
 #if MYNEWT_VAL(MCU_GPIO_USE_PORT_EVENT)
-    conf = 0;
+    (void)conf;
```
kasjer: It could be possible to use if instead of #if then we would have syntax checking for 0 and 1 at once and possibly no need to (void) unused variables.
Not that I object to what has been done, just a thought for the future.

andrzej-kaczmarek:
squashed

PR #1630, opened 2019-02-04
Hal gpio deinit

ygaklk:
Current implementation does not allow to uninitialize a pin which has been previously initialized.
An uninitialized pin means a pin in state "unconnected", or at least in the same state after a boot (not input, not output, not analog...).

It is useful in some cases to be able to revert a `hal_gpio_init_*()` call.

I created the prototype in the `hal/hal_gpio.h`.
Then, I have implemented this into 2 different MCU references (stm32 + nrf51/52) as the first step, but it is not blocking if your MCU doesn't implement it, as long as you don't call the `hal_gpio_deinit()`.

We can discuss about it with pleasure if you have other questions.

### DIFF
--- a/hw/hal/include/hal/hal_gpio.h
+++ b/hw/hal/include/hal/hal_gpio.h
@@ -102,6 +102,15 @@ int hal_gpio_init_in(int pin, hal_gpio_pull_t pull);
 int hal_gpio_init_out(int pin, int val);
 
 /**
+ * Deinitialize the specified pin to revert the previous initialization
+ *
+ * @param pin Pin number to unset
+ *
+ * @return int  0: no error; -1 otherwise.
+ */
+int hal_gpio_deinit(int pin);
+
+/**
  * Write a value (either high or low) to the specified pin.
  *
  * @param pin Pin to set
--- a/hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
@@ -151,6 +151,27 @@ hal_gpio_init_out(int pin, int val)
 }
 
 /**
+ * Deinitialize the specified pin to revert the previous initialization
+ *
+ * @param pin Pin number to unset
+ *
+ * @return int  0: no error; -1 otherwise.
+ */
+int
+hal_gpio_deinit(int pin)
+{
+    uint32_t conf;
+    NRF_GPIO_Type *port;
+
+    conf = GPIO_PIN_CNF_INPUT_Disconnect << GPIO_PIN_CNF_INPUT_Pos;
+    port = HAL_GPIO_PORT(pin);
+    port->PIN_CNF[pin] = conf;
+    port->DIRCLR = HAL_GPIO_MASK(pin);
+
+    return 0;
+}
+
+/**
  * gpio write
  *
  * Write a value (either high or low) to the specified pin.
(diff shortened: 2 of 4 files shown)


mkiiskila:
This seems like a reasonable addition to me.

Could you add stubs for other MCUs also? And check the travis build failures. Also nrf52xxx-compat package is gone, so you'll want to rebase this PR.

utzig:
> It is useful in some cases to be able to revert a hal_gpio_init_*() call.

Could you provide a few examples?

hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
```diff
@@ -139,6 +139,27 @@ hal_gpio_init_out(int pin, int val)
...
+ */
+int
+hal_gpio_deinit(int pin)
+{
+    uint32_t conf;
+    NRF_GPIO_Type *port;
+
+    conf = GPIO_PIN_CNF_INPUT_Disconnect << GPIO_PIN_CNF_INPUT_Pos;
+    port = HAL_GPIO_PORT(pin);
+    port->PIN_CNF[pin] = conf;
```
andrzej-kaczmarek: you need to use `HAL_GPIO_INDEX` here for this to work properly on nRF52840

hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
```diff
@@ -139,6 +139,27 @@ hal_gpio_init_out(int pin, int val)
...
+    port->DIRCLR = HAL_GPIO_MASK(pin);
```
andrzej-kaczmarek: this is probably redundant since direction is already set above in `PIN_CNF`

hw/hal/include/hal/hal_gpio.h
```diff
@@ -101,6 +101,15 @@ int hal_gpio_init_in(int pin, hal_gpio_pull_t pull);
  */
 int hal_gpio_init_out(int pin, int val);
 
+/**
+ * Deinitialize the specified pin to revert the previous initialization
```
andrzej-kaczmarek: I think this should say _to revert to default configuration_ as _previous initialization_ may suggest this will revert to previous call to `hal_gpio_init`

hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
ygaklk: Correct, fixed by: https://github.com/apache/mynewt-core/pull/1630/commits/36004f1985f4d34758cf5204a84ea29adfc0fae6

hw/mcu/nordic/nrf52xxx/src/hal_gpio.c
ygaklk: Agree, I have made a quick copy/paste of the init_out. Fixed by: https://github.com/apache/mynewt-core/pull/1630/commits/11a9c211abac17dde7e4ef663169ee94e92c6473

hw/hal/include/hal/hal_gpio.h
ygaklk: Agree, changed by: https://github.com/apache/mynewt-core/pull/1630/commits/9acf80b398984ac1164c9cb7fc387f28e1b275ee

ygaklk:
Code rebased on top of master, and nrf52xxx-compat deleted now.

PR #1635, opened 2019-02-06
Small fixes for C++ compilation.

mlampert:


kernel/os/include/os/os_mutex.h
```diff
@@ -119,7 +119,7 @@ os_error_t os_mutex_pend(struct os_mutex *mu, os_time_t timeout);
  */
 static inline os_error_t os_mutex_get_level(struct os_mutex *mu)
 {
-    return mu->mu_level;
+    return (os_error_t)mu->mu_level;
```
andrzej-kaczmarek: returned type should be changed (probably it was just c&p from other functions) - this type cast does not make sense

sjanc:
os_mutex_get_level issue has already been fixed in master,  you'll have to rebase your PR before it can go in

PR #1638, opened 2019-02-08
[TRNG] Add K64F driver

utzig:
Improvements to trng_test app and driver for K64F.

### DIFF
--- a/hw/drivers/trng/trng_k64f/src/trng_k64f.c
+++ b/hw/drivers/trng/trng_k64f/src/trng_k64f.c
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <string.h>
+#include "fsl_rnga.h"
+#include "trng/trng.h"
+#include "trng_k64f/trng_k64f.h"
+
+static uint8_t rng_cache[ MYNEWT_VAL(K64F_TRNG_CACHE_LEN) ];
+static uint16_t rng_cache_out;
+static uint16_t rng_cache_in;
+static struct os_mutex rng_cache_mu;
+static os_stack_t *pstack;
+static bool running;
+static struct os_eventq rng_evtq;
+
+#define RNGA_POLLER_PRIO (8)
+#define RNGA_POLLER_STACK_SIZE OS_STACK_ALIGN(64)
+static struct os_task poller_task;
+
+static inline void
+k64f_rnga_start(void)
+{
+    struct os_event evt;
+
+    RNGA_SetMode(RNG, kRNGA_ModeNormal);
+    running = true;
+
+    evt.ev_queued = 0;
+    evt.ev_arg = NULL;
+    (void)os_eventq_put(&rng_evtq, &evt);
+}
+
+static inline void
+k64f_rnga_stop(void)
+{
+   RNGA_SetMode(RNG, kRNGA_ModeSleep);
+   running = false;
+}
+
+static size_t
+k64f_trng_read(struct trng_dev *trng, void *ptr, size_t size)
+{
+    size_t num_read;
+
+    os_mutex_pend(&rng_cache_mu, OS_TIMEOUT_NEVER);
+
+    if (rng_cache_out <= rng_cache_in) {
+        size = min(size, rng_cache_in - rng_cache_out);
+        memcpy(ptr, &rng_cache[rng_cache_out], size);
+        num_read = size;
+    } else if (rng_cache_out + size <= sizeof(rng_cache)) {
+        memcpy(ptr, &rng_cache[rng_cache_out], size);
+        num_read = size;
+    } else {
+        num_read = sizeof(rng_cache) - rng_cache_out;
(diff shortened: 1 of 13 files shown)


hw/drivers/trng/trng_k64f/src/trng_k64f.c
```diff
@@ -0,0 +1,175 @@
...
+rnga_poller_handler(void *arg)
+{
+    int8_t i;
+    uint8_t data[4];
+    int rc;
+
+    while (1) {
+        if (running) {
+            rc = RNGA_GetRandomData(RNG, data, sizeof(uint32_t));
+            if (rc == 0) {
```
kasjer: it probably should be (rc == kStatus_Success)

hw/drivers/trng/trng_k64f/src/trng_k64f.c
```diff
@@ -0,0 +1,175 @@
...
+                for (i = 0; i < 4; i++) {
+                    rng_cache[rng_cache_in++] = data[i];
```
kasjer: there is possible time between rng_cache_in get incremented and checked against buffer size. During that time value rng_cache_in is invalid (out of range). If during this time k64f_trng_read() reads it in line 58 it will copy last byte that was not just generated.

hw/drivers/trng/trng_k64f/src/trng_k64f.c
```diff
@@ -0,0 +1,175 @@
...
+                    }
+
+                    if ((rng_cache_in + 1) % sizeof(rng_cache) == rng_cache_out) {
+                        k64f_rnga_stop();
+                        break;
+                    }
+                }
+            }
+        }
+        os_time_delay(1);
```
kasjer: this may be causing platform to wake from sleep quite often (possibly) for no reason. Maybe some other form of waiting could be used instead.

hw/drivers/trng/trng_k64f/src/trng_k64f.c
```diff
@@ -0,0 +1,175 @@
...
+
+static size_t
+k64f_trng_read(struct trng_dev *trng, void *ptr, size_t size)
+{
+    os_sr_t sr;
+    size_t num_read;
+
+    OS_ENTER_CRITICAL(sr);
+
+    if (rng_cache_out <= rng_cache_in) {
```
kasjer: in case of no data in cache lines 59-61 could be skipped instead, not that it matters much

hw/drivers/trng/trng_k64f/syscfg.yml
```diff
@@ -0,0 +1,22 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+    K64F_TRNG_CACHE_LEN:
+        description: 'Internal cache length, shall be power of 2'
```
kasjer: I can't see any reason in code for this power of 2 requirement, but maybe it is more future proof this way.

hw/drivers/trng/trng_k64f/src/trng_k64f.c
```diff
@@ -0,0 +1,175 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <string.h>
+#include "mcu/cmsis_nvic.h"
```
kasjer: cmsis_nvic.h looks strange, but I did not check if it is needed, ignore this comment if it's required

hw/drivers/trng/trng_k64f/src/trng_k64f.c
```diff
@@ -0,0 +1,175 @@
...
+
+    if (!(dev->od_flags & OS_DEV_F_STATUS_OPEN)) {
+        rng_cache_out = 0;
+        rng_cache_in = 0;
+
+        RNGA_Init(RNG);
+
+        k64f_rnga_start();
+
+        os_task_init(&poller_task, "rnga_poller", rnga_poller_handler, NULL,
```
kasjer: os_dev_close could be called on dev at some point, then flags could have OS_DEV_F_STATUS_OPEN cleared but task would be already running on second call os_dev_open; then os_task_init could be called second time and that may not be a good idea.

hw/drivers/trng/trng_k64f/src/trng_k64f.c
utzig: removed

hw/drivers/trng/trng_k64f/src/trng_k64f.c
utzig: I moved the task initialization to _init instead of _open.

hw/drivers/trng/trng_k64f/src/trng_k64f.c
utzig: done

hw/drivers/trng/trng_k64f/src/trng_k64f.c
utzig: Good idea, when not running the poller now waits on an eventq.

hw/drivers/trng/trng_k64f/src/trng_k64f.c
utzig: This is a good one! `k64f_trng_read` was disabling interrupts because it was based on nrf52/stm32 which generate ints when there is RNG data (and are also affected by this bug). I removed the interrupt disabling code and added a simple mutex based access to `rng_cache_*`.

hw/drivers/trng/trng_k64f/src/trng_k64f.c
```diff
@@ -0,0 +1,192 @@
...
+        num_read = size;
+    } else if (rng_cache_out + size <= sizeof(rng_cache)) {
+        memcpy(ptr, &rng_cache[rng_cache_out], size);
+        num_read = size;
+    } else {
+        num_read = sizeof(rng_cache) - rng_cache_out;
+        memcpy(ptr, &rng_cache[rng_cache_out], num_read);
+
+        size -= num_read;
+        ptr += num_read;
```
mkiiskila: I avoid doing math with void *, as it's not portable.

PR #1639, opened 2019-02-11
hw/mcu/nordic: Fix hal_system_start

andrzej-kaczmarek:
While current implementation of hal_system_start works fine now, it is
possible that compiler will "optimize" [1] code in a way that it will
not work properly because we first set stack pointer to new location
and then execute some more code which uses local variables.

To make sure this always works fine, let's use inline assembly here to
set stack pointer and then jump to image.

[1] while working with other Cortex-M MCU it happened to me that just
    before jumping to image, GCC inserted "ldmia sp! {r3, lr}" which
    basically attempted a read outside RAM area since stack pointer was
    already at the very end of RAM area.

### DIFF
--- a/hw/mcu/nordic/nrf51xxx/src/hal_system_start.c
+++ b/hw/mcu/nordic/nrf51xxx/src/hal_system_start.c
@@ -27,25 +27,18 @@
  *
  * @param hdr                   The header for the image to boot.
  */
-void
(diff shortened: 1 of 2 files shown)


andrzej-kaczmarek:
updated with more optimized version with only 3 asm instructions required (kudos to @kasjer for idea)

PR #1641, opened 2019-02-12
CoAP; cleanups for observing resources when using transport(s) with connections

mkiiskila:
- observer cleanup when BLE connection is going away was running (potentially) in wrong task context
- add a generic mechanism for init/cleanup notifications for connections when using coap
- allow more flexibility for transport to report whether particular endpoint is running over a connection or if it is connectionless

### DIFF
--- a/net/oic/src/port/mynewt/conns.c
+++ b/net/oic/src/port/mynewt/conns.c
@@ -75,7 +75,7 @@ oc_conn_created(struct oc_conn_ev *oce)
 {
     OC_LOG(DEBUG, "oc_conn_created: ");
     OC_LOG_ENDPOINT(LOG_LEVEL_DEBUG, &oce->oce_oe);
-    oce->oce_type = OC_ENDPOINT_CONN_EV_CLOSE;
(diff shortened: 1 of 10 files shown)


net/oic/src/port/mynewt/conns.c
```diff
@@ -0,0 +1,109 @@
...
+    OS_EXIT_CRITICAL(sr);
+    os_eventq_put(oc_evq_get(), &oc_conn_cb_ev);
+}
+
+void
+oc_conn_created(struct oc_conn_ev *oce)
+{
+    OC_LOG(DEBUG, "oc_conn_created: ");
+    OC_LOG_ENDPOINT(LOG_LEVEL_DEBUG, &oce->oce_oe);
+    oce->oce_type = OC_ENDPOINT_CONN_EV_CLOSE;
```
utzig: should not this be `oce->oce_type = OC_ENDPOINT_CONN_EV_OPEN;` ?

mkiiskila: Yes, it should be.

PR #1642, opened 2019-02-13
kernel/os: Track most recent event in debug builds

ccollins476ad:
Add a new syscfg setting: `OS_EVENTQ_DEBUG`.  When this setting is enabled, each eventq remembers its most recently processed event.

This feature is only useful when debugging with gdb.

### DIFF
--- a/kernel/os/include/os/os_eventq.h
+++ b/kernel/os/include/os/os_eventq.h
@@ -71,10 +71,13 @@ struct os_eventq {
      */
     struct os_task *evq_task;
 
+#if MYNEWT_VAL(OS_EVENTQ_DEBUG)
+    /** Most recently processed event. */
+    struct os_event *evq_prev;
+#endif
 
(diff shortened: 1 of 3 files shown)


kernel/os/include/os/os_eventq.h
```diff
@@ -71,11 +71,14 @@ struct os_eventq {
      */
     struct os_task *evq_task;
 
+#if MYNEWT_VAL(OS_EVENTQ_DEBUG)
+    /** Most reently procesed event. */
```
utzig: typo `reently`

utzig: btw, `procesed` is also a typo!

ccollins476ad: That's pretty funny :).  Thanks for catching those.  I have fixed and force pushed.

PR #1643, opened 2019-02-14
[FRDM-K64F] Increase size of bootloader slot

utzig:
This makes the bootloader 40kb (from 24kb) to have enough space to do encrypted image testing, and reduces slot 0/1 sizes from 480KB to 472KB which does not seem very significant.

Also the linker script size of the bootloader was correctly adjusted to what is defined on the BSP avoiding non-caught errors when the bootloader size is big enough to overflow the configured space.


utzig:
@mkiiskila The main issue that need fixing was the size of the bootloader in the linker script, my bootloader was overflowing and it was being caught automatically, but since I am doing tests with encrypted images on the K64F I would like to propose already that the slots' sizes are changed to avoid having to maintain local changes...

PR #1645, opened 2019-02-14
NRF52 Timer - Achieve Low Power When Off

brolan-juul:
This PR implements a suggested work around to achieve low power with timer off.

**NRF52832 Errata v1.9 - 3.19 [78] TIMER: High current consumption when using
timer STOP task only**

**Symptoms**
Increased current consumption when the timer has been running and the STOP task is used to stop it.

**Conditions**
The timer has been running (after triggering a START task) and then it is stopped using a STOP task only.

**Consequences**
Increased current consumption.

**Workaround**
Use the SHUTDOWN task after the STOP task or instead of the STOP task.


brolan-juul:
@wes3 Agreed. I think the clear timer makes sense.

PR #1646, opened 2019-02-15
kernel/os: Add helper to calculate array size

andrzej-kaczmarek:
No more sizeof()/sizeof()...


utzig:
@andrzej-kaczmarek CI failed because there are redefinitions:

```
2019/02/15 08:54:51.791 [WARNING] Transient package hw/bsp/nrf52840pdk used, update configuration to use linked package instead (@apache-mynewt-core/hw/bsp/nordic_pca10056)
Error: In file included from repos/apache-mynewt-nimble/apps/blemesh_models_example_1/src/main.c:41:0:
repos/apache-mynewt-nimble/nimble/host/mesh/include/mesh/glue.h:130:0: error: "ARRAY_SIZE" redefined [-Werror]
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
```

PR #1649, opened 2019-02-18
Add initial crypto driver framework

utzig:
This is a proposal for a crypto framework using HW acceleration in MCUs that have the support.

It was tested with the added app (crypto_test) as well as enc_flash + encrypted images on NXP K64F and Nrf52840-PDK.

Still a few TODOS before this is complete for reviews:

1) Some error handling in the main driver needs finishing
2) Expand doxygen comments
3) STM32 driver must be finished (remove IRQ handling because it's slower)

PS: there are two commits which are not part of the PR but are helping me on my tests, will remove them soon... (UPDATE: one commit!)


hw/drivers/crypto/crypto_stm32/src/crypto_stm32.c
```diff
@@ -0,0 +1,150 @@
...
+    case CRYPTO_ALGO_AES_256_ECB:
+        g_hcryp.Init.KeySize = CRYP_KEYSIZE_256B;
+        break;
+#endif
+    default:
+        return -1;
+    }
+
+    g_hcryp.Init.pKey = (uint8_t *)key;
+    assert(HAL_CRYP_Init(&g_hcryp) == HAL_OK);
```
mkiiskila: This will call HAL_CRYP_Init() only if assert() is compiled in. Should have the assert() on a separate line.

hw/drivers/crypto/crypto_stm32/src/crypto_stm32.c
```diff
@@ -0,0 +1,150 @@
...
+    assert(crypto);
+
+    //LOCK_CRYPTO_DEV(crypto);
+
+    if (!(dev->od_flags & OS_DEV_F_STATUS_OPEN)) {
+        __HAL_RCC_CRYP_CLK_ENABLE();
+
+        g_hcryp.Instance = CRYP;
+        g_hcryp.Init.DataType = CRYP_DATATYPE_8B;
+        HAL_CRYP_Init(&g_hcryp);
```
mkiiskila: Do you need to call this, if every encrypt operation also calls it?

hw/drivers/crypto/crypto_k64f/src/arch/cortex_m4/mmcau_aes_functions.s
```diff
@@ -0,0 +1,941 @@
+#*******************************************************************************
```
mkiiskila: I assume license is ok for us to include this?

hw/bsp/nucleo-f767zi/pkg.yml
```diff
@@ -27,7 +27,8 @@ pkg.keywords:
     - stm32f7
     - nucleo
 
-pkg.cflags: -DSTM32F767xx
+      # pkg.cflags: -DSTM32F767xx
+pkg.cflags: -DSTM32F777xx
```
mkiiskila: Was this change intentional? A bit unexpected.

utzig: STM32 is not finished, this is here because F777 supports CRYP and other parts don't. Once this is properly finished I will drop this commit.

hw/drivers/crypto/include/crypto/aes_alt.h
```diff
@@ -0,0 +1,55 @@
...
+#include <crypto/crypto.h>
+
+typedef struct mbedtls_aes_context
+{
+    struct crypto_dev *crypto;
+    uint8_t key[AES_MAX_KEY_LEN];
+    uint8_t keylen;
+} mbedtls_aes_context;
+
+typedef struct {} mbedtls_aes_xts_context;
```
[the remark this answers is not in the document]
utzig: Was a leftover and removed.

hw/drivers/crypto/crypto_k64f/src/arch/cortex_m4/mmcau_aes_functions.s
utzig: Yes, BSD, it's missing the LICENSE header, I will add it or LICENSE file.

utzig:
@mkiiskila Most issues were already addressed, ended up being a overhaul in the API but more extendable now IMO, although can still be clean up. STM32 will be done next.

hw/drivers/crypto/crypto_k64f/src/arch/cortex_m4/mmcau_aes_functions.s
utzig: Actually not so simple, but I'll review it and comment back here.

utzig:
@mkiiskila STM32 should be OK now (apart from locking); I removed the NXP library because it seems to be Categoy-X (thread here: https://mail-archives.apache.org/mod_mbox/www-legal-discuss/201902.mbox/browser) and added a C implementation with https://github.com/apache/mynewt-core/pull/1649/commits/f0326e539c254034a0bb444d55c9393b15414cdc (does not yet expand keys above AES-128); I will add a global locking per driver now; also considering adding a few syscfg for the low-level driver to export the functionality it supports (to allow link time removal of some symbols) and syscfgs for a library/app declare functionality it needs (or disable what it does not, so it can be link time removed from build).

utzig:
Ops, correct ML link: https://mail-archives.apache.org/mod_mbox/www-legal-discuss/201902.mbox/%3C0551171b-8d4b-4064-84b9-9ad0030be810%40www.fastmail.com%3E and commit https://github.com/apache/mynewt-core/commit/52f9682c833e5653b9092e1e70eddf20fa82cce0

mkiiskila:
Thanks for going through the process of verifying license!
I don't think we need to complete the individual drivers with all the features, as long as the API seems like something we can start with :) We probably should merge it earlier, and then let the API soak for a release or two.

hw/drivers/crypto/src/mbedtls_aes_alt.c
```diff
@@ -0,0 +1,66 @@
...
+#include "crypto/crypto.h"
+#include "crypto/aes_alt.h"
+
+void
+mbedtls_aes_init(mbedtls_aes_context *ctx)
+{
+    memset(ctx, 0, sizeof(*ctx));
+    ctx->crypto = (struct crypto_dev *) os_dev_open("crypto", OS_TIMEOUT_NEVER,
+            NULL);
+}
```
mkiiskila: I probably would assert() here if ctx->crypto is NULL. User might've forgotten to create the crypto device. Although the system would crash when the first crypto operation is done. So not absolutely necessary.

hw/drivers/crypto/include/crypto/crypto.h
```diff
@@ -0,0 +1,221 @@
...
+ *
+ * @var crypto_interface::decrypt
+ * decrypt is a crypto_op_func_t pointer to the decryption routine
+ *
+ * @var crypto_interface::has_algomode
+ * caps exports the capabilities of the HW driver. see TODO
+ */
+struct crypto_interface {
+    crypto_op_func_t encrypt;
+    crypto_op_func_t decrypt;
```
mkiiskila: I have few things which I want  here, and probably should phrase these explicitly. Sorry for the late response, I had to gather my thoughts a bit.

1) I want to be able decrypt/encrypt data in-place. I don't want to spend cycles copying data from temporary buffers, when doing crypto. I can see some of the drivers appear to be able to do it, but I did not go through all
of them. Thanks for addressing so many architectures with this, BTW.

2) I want to be able to use discontiguous buffers. Specifically, I want to be able to add routines which decrypt/encrypt data within mbuf chain without having to move data to a temporary buffer for encryption/decryption, and then having to copy it back. I'd be ok with the interface being even iovec[], although state structure would've been best. The former allows for datagrams carried in mbuf chains, while the latter would allow me to use this with byte streams, like TCP.

utzig: > I want to be able decrypt/encrypt data in-place. I don't want to spend cycles copying data from temporary buffers, when doing crypto. I can see some of the drivers appear to be able to do it, but I did not go through all of them.

You mean doing `encrypt_aes_ecb(buf, buf, len)`? The only mode that was not working was CTR, but I fixed it yesterday. Documentation might still mention it need separate buffers though.

> I want to be able to use discontiguous buffers. Specifically, I want to be able to add routines which decrypt/encrypt data within mbuf chain without having to move data to a temporary buffer for encryption/decryption, and then having to copy it back. I'd be ok with the interface being even iovec[], although state structure would've been best. The former allows for datagrams carried in mbuf chains, while the latter would allow me to use this with byte streams, like TCP.

Good idea, I will add functions like `crypto_algo+mode_(en|de)cryptv(xxx)` that accept `iovec` instead of `inbuf/outbuf`. The tree seems to have cbor and lwip `iovec` definitions; should be using one of those or should I add a new definition? (`crypto_iovec`?, but being similarly defined will accept the others as well...)

mkiiskila: Thanks! I'd leave it up to you whether to define your own, or reuse the struct from other package. I probably would have a separate one, so there's no dependency to those packages :)

utzig: I think all issues were tackled and some extra bugs for in-place operation were slashed. Doxygen should be updated as well. I added some syscfg to compile out non-required functionality, must still find a way to document it.

mkiiskila:
This seems like a reasonable start with this. I'd be ok starting with this.

PR #1651, opened 2019-02-19
Check if ARRAY_SIZE is already defined

vrahane:
-  Check if ARRAY_SIZE is already defined before defining in the OS

### DIFF
--- a/kernel/os/include/os/util.h
+++ b/kernel/os/include/os/util.h
@@ -31,7 +31,8 @@
         ((type *)(((char *)(ptr)) - offsetof(type, field)))
 
 /* Helper to calculate number of elements in array */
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
not sure why do we need this? other code should simply not try to redefine this macro and use common definition from OS...

utzig:
@andrzej-kaczmarek what if I have 3rd party libraries around that define it?

andrzej-kaczmarek:
then they should not define common macro unconditionally, i.e. libraries should have `#ifdef` instead ;-)

I just wondered whether we have an actual problem with such 3rd party code which we need to workaround, but since this is already merged and is basically harmless let's just leave as is.

PR #1654, opened 2019-02-20
nrf52: Make target MCU and LFCLK source selection as a choice

andrzej-kaczmarek:
This is proposed change to use choice settings to select target MCU and LFCLK source for nRF52xxx BSP-s. It requires new `newt` feature implemented here: https://github.com/apache/mynewt-newt/pull/270. The advantage here is that we do not need separate values for each option and `newt` will take care of proper value being set for those settings (and there will be only one value). The disadvantage is that it breaks compatibility with earlier `newt` version and possible with some external code that uses those settings.


sterlinghughes:
This would require a new version of newt to build *-core, correct?  We should probably upgrade the satisfies dependency for newt.

andrzej-kaczmarek:
dependency for newt will be updated for next release
we probably do not do this for master since newt version does not change (perhaps we should bump patchver for such cases?)

sjanc:
master is expected to be used with master newt
as mentioned, if needed deps will be updated when next release is done

PR #1655, opened 2019-02-21
kernel/os: Fix mem corruption on 0-entry mempool

ccollins476ad:
This fixes a memory corruption bug.  When a mempool was initialized with 0 elements, its 0-length buffer would still be written to.  The bad write occurred when setting the last element's `next` pointer to NULL here:

```
    /* Last one in the list should be NULL */
    SLIST_NEXT(block_ptr, mb_next) = NULL;
```

The fix is to only perform this final write if there is at least one element in the pool.

### DIFF
--- a/kernel/os/src/os_mempool.c
+++ b/kernel/os/src/os_mempool.c
@@ -121,6 +121,7 @@ os_mempool_init_internal(struct os_mempool *mp, uint16_t blocks,
                          uint8_t flags)
 {
     int true_block_size;
+    int i;
     uint8_t *block_addr;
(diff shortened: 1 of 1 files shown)


ccollins476ad:
I don't think a zero element mempool is useful.  You really can't do anything with it :).  But I think an application can end up with one nonetheless, particularly when a mempool's size is defined as a syscfg setting (that is how I encountered this).

I think it is good to allow a zero element mempool as long as it isn't too difficult.  Such a mempool behaves as you'd expect: any attempt to allocate from it just returns null.

PR #1664, opened 2019-02-27
spiflash: Improve erase sectors

rymanluk:
Each flash has known time which is needed to erase sector.
Having this in mind we can wait a bit before reading status after sector
erase.

Note: This could be done smarter along with adjusting longest time which
might be needed. For now we choose some "default" value

### DIFF
--- a/hw/drivers/flash/spiflash/src/spiflash.c
+++ b/hw/drivers/flash/spiflash/src/spiflash.c
@@ -876,6 +876,8 @@ spiflash_wait_ready(struct spiflash_dev *dev, uint32_t timeout_ms)
     spiflash_lock(dev);
 
     while (!spiflash_device_ready(dev)) {
+        /* If not ready let's give it 10ms */
+        os_time_delay(os_time_ms_to_ticks32(10));
         if (os_time_get() > exp_time) {
             rc = -1;
             goto err;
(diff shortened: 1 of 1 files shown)


hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -1033,6 +1035,11 @@ spiflash_erase_sector(const struct hal_flash *hal_flash_dev,
     spiflash_cs_deactivate(dev);
 #endif
 
+    /* Usually erase sector takes around 40ms, let's wait here that time.
+     * TODO: Do it configurable per flash.
```
andrzej-kaczmarek: so why not just add syscfg for this now? :-)

rymanluk: because I believe correct fix should be to add defines per all the supported flash and not give it to the user to set. Those values are in the specifications.

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -876,6 +876,8 @@ spiflash_wait_ready(struct spiflash_dev *dev, uint32_t timeout_ms)
     spiflash_lock(dev);
 
     while (!spiflash_device_ready(dev)) {
+        /* If not ready let's give it 10ms */
+        os_time_delay(os_time_ms_to_ticks32(10));
         if (os_time_get() > exp_time) {
```
wes3: I have a question about this 10msec delay here. The spiflash_wait_ready() function is called by called that does not do erases. Waiting 10msecs seems like it could seriously impact flash read/write times, especially if you are doing small reads and writes. Was this change added to the code? I think I would only do something like this when erasing as opposed to reading/writing flash. Looks like there is a change below to that effect...

PR #1670, opened 2019-03-01
Implement low-power support in the LoRa stack

amrbekhit:
Use of timers in the LoRa stack has been split into critical and non-critial timers. Non-critical timers (such as retry delays) are now implemented using the os_cputime functions. The
os_cputime needs to be configured to use a lower power time in the BSP (such as is done by the BLE stack via the BLE_LP_CLOCK syscfg). Critical timers are only used to time the opening of
the RX windows. These continue to use the hal_timer functions, but require that the user implements two callback functions (lora_low_power_enter and lora_low_power_exit) to turn the
LORA_MAC_TIMER off and on. This is necessary as turning a timer on is a hardware-specific operation that may require that the caller pass in a custom configuration struct.

~~This patch also modifies the nRF timer code to use the HFXO manager found in hw/mcu/nordic/nrf52xxx/src/nrf5x_clock.c. This allows multiple different timers to be turned on and off
independently and manager will automatically turn the nRF HFXO on and off as required. This is necessary in order for the LoRa and BLE stacks to be able to run side by side.~~ **EDIT:** [As requested](https://github.com/apache/mynewt-core/pull/1670#pullrequestreview-224000758), this has now been removed and will be pushed as a separate pull request.


net/lora/node/src/mac/LoRaMac.c
```diff
@@ -853,7 +858,7 @@ lora_mac_process_radio_tx(struct os_event *ev)
                 timeout += (uint32_t)((RxWindow2Config.tsymbol * 1000) *
                     RxWindow2Config.WindowTimeout);
             }
-            hal_timer_start_at(&g_lora_mac_data.rtx_timer, curTime + timeout);
+            os_cputime_timer_relative(&g_lora_mac_data.rtx_timer, curTime + timeout);
         }
```
wes3: Same comment as above

net/lora/node/src/mac/LoRaMac.c
```diff
@@ -1973,8 +1981,8 @@ ScheduleTx(void)
 
             // Send later - prepare timer
             LoRaMacState |= LORAMAC_TX_DELAYED;
-            hal_timer_stop(&TxDelayedTimer);
-            hal_timer_start(&TxDelayedTimer, duty_cycle_time_off);
+            os_cputime_timer_stop(&TxDelayedTimer);
+            os_cputime_timer_relative(&TxDelayedTimer, duty_cycle_time_off);
```
wes3: I guess duty_cycle_time_off is in microseconds here. I would double check that to be sure.

amrbekhit:
> I might have missed a commit but I do not see lora_low_power_exit() and enter().

This something I tried to discuss on the mailing list https://lists.apache.org/thread.html/48451519a2a6990dfa98dea84ee4cf8e6c8b407e9220f152e9ab49a2@%3Cdev.mynewt.apache.org%3E

Basically, the hal_timer functions don't provide a way for libraries to turn a timer on in a hardware agnostic fashion. You have to call the hal_timer_init function with a hardware specific parameter. Therefore, I declared the callbacks lora_low_power_enter/exit() that need to be implemented by the user in order to turn the timer on and off.

hw/drivers/lora/src/utilities.c
```diff
@@ -17,25 +17,19 @@
...
-#include "os/mynewt.h"
-#include "lora/utilities.h"
+#include "os/os.h"
 
 uint32_t
 timer_get_current_time(void)
 {
-    return hal_timer_read(MYNEWT_VAL(LORA_MAC_TIMER_NUM));
+    /* Convert the OS time ticks to seconds, and then to us*/
+    return os_cputime_ticks_to_usecs(os_cputime_get32());
```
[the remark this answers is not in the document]
amrbekhit: As you say, in the original implementation, `timer_get_current_time` directly returned the raw timer value. However, the timer value was hard-coded to run at 1MHz, giving a 1us time per tick.

https://github.com/apache/mynewt-core/blob/b886f237076f8206ffc1c8f0b451ae61abd023b4/net/lora/node/src/mac/LoRaMac.c#L50

The code also runs on the assumption that the timer_get_current_time value returns a number of us. This can be seen, for example, in the definition of `LORA_JOIN_REQ_RAND_DELAY`, which is in milliseconds:

https://github.com/apache/mynewt-core/blob/f5769ec3e2caf365e5a1036a300fe943a6feb67c/net/lora/node/syscfg.yml#L78

It is then used here (note the hard-coded multiplication by 1000, to convert it to us):

https://github.com/apache/mynewt-core/blob/b886f237076f8206ffc1c8f0b451ae61abd023b4/net/lora/node/src/mac/LoRaMac.c#L726

As such, my modifications to the timer_xxx functions maintain the us time unit.

net/lora/node/src/mac/LoRaMac.c
amrbekhit: It is in microseconds. As you can see from the original code, TxDelayedTimer is initialised using `LORA_MAC_TIMER_NUM`, which is configured at 1MHz:

https://github.com/apache/mynewt-core/blob/10fd2c3c2358accf9bba5a569e133da5ec7b24c1/net/lora/node/src/mac/LoRaMac.c#L1977

net/lora/node/src/mac/LoRaMac.c
```diff
@@ -833,7 +838,7 @@ lora_mac_process_radio_tx(struct os_event *ev)
     if (LM_F_NODE_ACK_REQ()) {
         getPhy.Attribute = PHY_ACK_TIMEOUT;
         phyParam = RegionGetPhyParam(LoRaMacRegion, &getPhy);
-        hal_timer_start_at(&g_lora_mac_data.rtx_timer, curTime +
+        os_cputime_timer_relative(&g_lora_mac_data.rtx_timer, curTime +
```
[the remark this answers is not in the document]
amrbekhit: Yes, this is a mistake, probably due to regex find/replace.

net/lora/node/src/mac/LoRaMac.c
amrbekhit: Response above.

wes3:
And another comment. Not sure how to handle the lora enter low power. Quite a few of the drivers require a hw specific config and this is usually handled in the bsp or in the architecture specific portion of drivers. So handling that in the lora hw specific driver seems reasonable.

amrbekhit:
Hi @wes3 as requested I've removed the HFXO changes to hal_system.c and hal_timer.c and will add these as a separate pull request.

I've made some changes to the low power callbacks - since the LoRa code can turn the timer off itself using `hal_timer_deinit`, I've removed the `lora_low_power_enter` callback function and renamed `lora_low_power_exit` to `lora_bsp_enable_mac_timer`. BSPs now only need to implement `lora_bsp_enable_mac_timer`.

I've also modified the b-l072z-lrwan1 and telee02 BSPs to implement `lora_bsp_enable_mac_timer`. Travis CI now builds find and this pull request passes all checks.

Hopefully this can now be merged.

wes3:
Reviewed the changes and merged. Sorry for the delay.

PR #1674, opened 2019-03-05
sys/config: Add API for conf_load_one()

vrahane:
- Earlier we would load all configuration values together, this API gives us a way to load just one value.
- conf_load() still loads everything the way it used to.

### DIFF
--- a/sys/config/include/config/config.h
+++ b/sys/config/include/config/config.h
@@ -190,9 +190,10 @@ int conf_load(void);
 
 /**
  * Load configuration from a specific registered persistence source.
- * Handlers will be called for configuration subtrees for
+ * Handlers will be called for configuration subtree for
  * encountered values.
  *
+ * @param name of the configuration subtree.
  * @return 0 on success, non-zero on failure.
  */
 int conf_load_one(char *name);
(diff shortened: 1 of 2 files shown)


sys/config/src/config_store.c
```diff
@@ -67,7 +67,38 @@ conf_dst_register(struct conf_store *cs)
 static void
 conf_load_cb(char *name, char *val, void *cb_arg)
 {
-    conf_set_value(name, val);
+    if ((cb_arg && !strcmp((char*)cb_arg, name)) ||
```
mkiiskila: if (!cb_arg || !strcmp(cb_arg, name)) {

sys/config/include/config/config.h
```diff
@@ -188,6 +188,15 @@ int conf_register(struct conf_handler *cf);
  */
 int conf_load(void);
 
+/**
+ * Load configuration from a specific registered persistence source.
+ * Handlers will be called for configuration subtrees for
+ * encountered values.
+ *
+ * @return 0 on success, non-zero on failure.
+ */
```
mkiiskila: Comment copied from function above, should be updated to describe this function.

Also, note the existence of conf_get_stored_value(), which also can be used in finding a value associated with a specific name.

vrahane: Yeah, that could be used but we wanted the callbacks getting called as well. Hence, the additional API.

sys/config/src/config_store.c
vrahane: Thanks, that's a neat way, did not think about it.

PR #1679, opened 2019-03-07
sys/config: commit for all fcbs together only once

vrahane:
@mkiiskila I am trying to understand this. Can you please confirm that this can be indeed done and would not cause any issues.

### DIFF
--- a/sys/config/src/config_store.c
+++ b/sys/config/src/config_store.c
@@ -91,9 +91,6 @@ conf_load_one(char *name)
     conf_loading = true;
     SLIST_FOREACH(cs, &conf_load_srcs, cs_next) {
         cs->cs_itf->csi_load(cs, conf_load_cb, name);
-        if (SLIST_NEXT(cs, cs_next)) {
-            conf_commit(name);
-        }
     }
     conf_loading = false;
     conf_unlock();


mkiiskila:
No, this is not ok, and would cause issues. Here is what this would specifically break:

We currently can load manufacturing config from one config source, and runtime config from another.
And we can have a config variable which says whether user/app config will be stored on config save.

Thinking is that you start in manufacturing mode, with no config present. Then you enter serial number, per devices keys etc. Stuff you cannot change at runtime, but need to configure at factory. Then, as the last thing, you set a config variable saying 'not in manufacturing mode'.

Select config handlers then check this variable before accepting new config values.
Manufacturing mode is exited at commit handler. Not when config value is set, but at commit time. 
This because the order in which config variables are loaded from a source is not deterministic. And can even change, thanks to compression.

mkiiskila:
Ah, scratch that comment. The change you're proposing is for conf_load_one() only. Yeah, that should be ok :)

ccollins476ad:
Nonetheless, I appreciate the explanation.  Now I understand why `conf_load()` calls `conf_commit()`  multiple times.

vrahane:
Thanks @mkiiskila that definitely clears up things. conf_load_one() is only for runtime configs.

vrahane:
Thanks @mkiiskila that definitely clears up things. conf_load_one() is only for runtime configs.

PR #1680, opened 2019-03-07
app/pwm_test improvements and pwm_nrf52 fixes

rymanluk:
This PR brings pwm_shell and couple of fixes to pwm_nrf52

### DIFF
--- a/apps/pwm_test/src/pwm_shell.c
+++ b/apps/pwm_test/src/pwm_shell.c
@@ -1,0 +1,455 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "os/mynewt.h"
+#include "os/os_dev.h"
+#include "console/console.h"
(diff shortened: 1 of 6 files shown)


hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
```diff
@@ -297,7 +297,7 @@ nrf52_pwm_open(struct os_dev *odev, uint32_t wait, void *arg)
     inst_id = dev->pwm_instance_id;
 
     if (instances[inst_id].in_use) {
-        return (EINVAL);
+        return OS_EINVAL;
```
andrzej-kaczmarek: OS_EBUSY

apps/pwm_test/src/pwm_shell.c
```diff
@@ -0,0 +1,438 @@
...
+store_pwm(char *pwm_name, struct pwm_dev *dev)
+{
+    int i;
+
+    if (pwm_name[0] == 's') {
+        /* in case of spwmx*/
+        i = atoi(pwm_name + 4);
+    } else {
+        i = atoi(pwm_name + 3);
+    }
```
andrzej-kaczmarek: you can have both `pwmx` and `spwmx` devices at the same time so this won't work well in such case

rymanluk: ah you right

hw/drivers/pwm/pwm_nrf52/src/pwm_nrf52.c
rymanluk: agree

apps/pwm_test/src/pwm_shell.c
```diff
@@ -0,0 +1,455 @@
...
+{
+    int i;
+
+    if (pwm_name[0] == 's') {
+        /* in case of spwmx*/
+        i = atoi(pwm_name + 4);
+        /* starting from 4 we have soft pwms in the table */
+        i += 4;
+    } else {
+        i = atoi(pwm_name + 3);
```
andrzej-kaczmarek: so `pwm4` (if created by any BSP/MCU) will effectively overwrite `spwm0` :-)
anyway, this is a test app so I do not think it really matters, but probably to support any number of different kind of PWMs with any names it would be better to just have a lookup table

rymanluk: yeah this is test code only and I was lazy. I think this is already too much :D

PR #1692, opened 2019-03-15
hw/bus: Allow to use nonblock SPI operations

rymanluk:


hw/bus/drivers/spi_common/include/bus/drivers/spi_common.h
```diff
@@ -41,7 +41,9 @@ struct bus_spi_dev_cfg {
 struct bus_spi_dev {
     struct bus_dev bdev;
     struct bus_spi_dev_cfg cfg;
-
+#if MYNEWT_VAL(BUS_SPI_NOBLOCK)
+    struct os_sem sem;
+#endif
```
andrzej-kaczmarek: this should be put in `spi_hal` since it's specific to that driver
I think the best way to do this which is consistent with the way nodes are implemented would be to add `struct bus_spi_hal_dev` which "extends" this struct and adds new field. one thing that would need to be modified is nrf52 mcu package so it uses proper struct when creating dev object, but it's ok since we did not have release yet and can change things easily.

rymanluk: agreed and updated according to this

InternetofAwesome:
LGTM!

PR #1693, opened 2019-03-15
Bookmarks optimization for FCB-backed logs

ccollins476ad:
Bookmarks are an optimization to speed up lookups in FCB-backed logs.  The concept is simple: maintain a set of flash area+offset pairs corresponding to recently found log entries.  When we perform a log lookup, the walk starts from the bookmark closest to our desired entry rather than from the beginning of the log.

Bookmarks are stored in a circular buffer in the fcb_log object.  Each time the log is walked, the starting point of the walk is added to the set of bookmarks.

FCB rotation invalidates all bookmarks.

By default, this optimization is disabled.  To enable it, set the `LOG_FCB_BOOKMARKS` syscfg setting to 1.  Each FCB-backed log must then be configured with bookmark storage at runtime.

### DIFF
--- a/fs/fcb/src/fcb_bmark.c
+++ b/fs/fcb/src/fcb_bmark.c
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "os/mynewt.h"
+
+#if MYNEWT_VAL(LOG_FCB_BOOKMARKS)
+
+#include "fcb/fcb.h"
+#include "fcb_priv.h"
+#include "string.h"
+
+void
+fcb_log_init_bmarks(struct fcb_log *fcb_log,
+                    struct fcb_log_bmark *buf, int bmark_count)
+{
+    fcb_log->fl_bset = (struct fcb_log_bset) {
+        .fls_bmarks = buf,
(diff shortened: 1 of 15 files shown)


sterlinghughes:
Is this functionality useful for fcb2 as well?

ccollins476ad:
> Is this functionality useful for fcb2 as well?

It probably would be.  I have not looked at fcb2 much, so I can't say for sure that this feature is applicable.  One prerequisite is to write an fcb2-backed log implementation.  Currently, fcb2 is not supported by `sys/log/full`.

If fcb2 is an improvement over fcb, then we should create a log implementation.  When we do that, we can port the bookmark code.

fs/fcb/src/fcb_bmark.c
```diff
@@ -0,0 +1,98 @@
...
+#include "fcb_priv.h"
+#include "string.h"
+
+void
+fcb_log_init_bmarks(struct fcb_log *fcb_log,
+                    struct fcb_log_bmark *buf, int bmark_count)
+{
+    fcb_log->fl_bset = (struct fcb_log_bset) {
+        .fls_bmarks = buf,
+        .fls_cap = bmark_count,
```
wes3: I am making this comment just so I am sure I understand this, but would you ever expect a buffer to not be sized based on bmark_count? I do not think this is a big deal but I was wondering if the code should check to see if buf is sized appropriately (based on the bmark_count). I guess if it did that it could calculate the # of bookmarks. No need to change anything here btw.

fs/fcb/src/fcb_bmark.c
```diff
@@ -0,0 +1,98 @@
...
+    uint32_t diff;
+    int i;
+
+    min_diff = UINT32_MAX;
+    closest = NULL;
+
+    for (i = 0; i < fcb_log->fl_bset.fls_size; i++) {
+        bmark = &fcb_log->fl_bset.fls_bmarks[i];
+        if (bmark->flb_index <= index) {
+            diff = index - bmark->flb_index;
```
wes3: The presumption here is that this index does not wrap correct (as it is a straight compare)?

fs/fcb/src/fcb_bmark.c
ccollins476ad: I do think that is a legitimate danger.  The size of `buf` may not match `bmark_count`.  I don't think there is a good way around this, though.  This is just a limitation of C - you can't pass an array to a function, you can only pass a pointer.  So it is always the user's responsibility to specify the correct size as a separate argument.

fs/fcb/src/fcb_bmark.c
ccollins476ad: That is a good point.  This seems to be an issue throughout the logging code.  We should address it in a separate PR!

PR #1700, opened 2019-03-19
added image upload callbacks to imgmgr

JustineKH:
Added `start` `stop` and `complete` callbacks for imgmgr

### DIFF
--- a/mgmt/imgmgr/include/imgmgr/imgmgr.h
+++ b/mgmt/imgmgr/include/imgmgr/imgmgr.h
@@ -44,6 +44,17 @@ extern "C" {
 #define IMGMGR_STATE_F_ACTIVE           0x04
 #define IMGMGR_STATE_F_PERMANENT        0x08
 
+/** @brief Generic callback function for events */
+typedef void (*imgrmgr_dfu_cb)(void);
+
+/** Callback function pointers */
+typedef struct
+{
+    imgrmgr_dfu_cb dfu_started_cb;
+    imgrmgr_dfu_cb dfu_stopped_cb;
+    imgrmgr_dfu_cb dfu_complete_cb;
+} imgmgr_dfu_callbacks_t;
+
 /** @typedef imgr_upload_fn
  * @brief Application callback that is executed when an image upload request is
  * received.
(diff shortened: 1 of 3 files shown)


mgmt/imgmgr/include/imgmgr/imgmgr.h
```diff
@@ -111,7 +122,10 @@ int imgmgr_state_slot_in_use(int slot);
 int imgmgr_state_set_pending(int slot, int permanent);
 int imgmgr_state_confirm(void);
 int imgmgr_find_best_area_id(void);
-
+int imgmgr_register_callbacks(const imgmgr_dfu_callbacks_t *cb_struct);
+void imgr_dfu_stopped(void);
```
rymanluk: was it intention to expose the `img_dfu_foo()`? If so it should have prefix `imgmgr`. If it was not the intention then those functions should be `static`

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -844,13 +853,47 @@ imgr_upload(struct mgmt_cbuf *cb)
     return imgr_upload_good_rsp(cb);
 }
 
+void
+imgr_dfu_stopped(void)
+{
+    if (imgmgr_dfu_callbacks_fn->dfu_stopped_cb != NULL)
```
rymanluk: brackles here and below are not according to https://github.com/apache/mynewt-core/blob/master/CODING_STANDARDS.md

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -844,13 +853,47 @@ imgr_upload(struct mgmt_cbuf *cb)
...
+}
+
 void
 imgr_set_upload_cb(imgr_upload_fn *cb, void *arg)
 {
     imgr_upload_cb = cb;
     imgr_upload_arg = arg;
 }
 
+int
```
rymanluk: maybe there is no need to return `int` here since we never fail.

wes3:
Another comment: the pointer imgmgr_dfu_callbacks_fn should be checked for NULL before accessing. If someone does not call the register callbacks function the code will access a NULL pointer.

PR #1701, opened 2019-03-19
drivers/lp5523: change multi byte instruction read write to single

JustineKH:
I was seeing an issue with the 2-byte instruction read and write. Single byte transfers fixed the issue.

### DIFF
--- a/hw/drivers/led/lp5523/src/lp5523.c
+++ b/hw/drivers/led/lp5523/src/lp5523.c
@@ -424,7 +424,6 @@ lp5523_get_output_reg(struct led_itf *itf, enum lp5523_output_registers addr,
     return lp5523_get_reg(itf, addr + (output - 1), value);
 }
 
-
 int
 lp5523_set_engine_reg(struct led_itf *itf, enum lp5523_engine_registers addr,
     uint8_t engine, uint8_t value)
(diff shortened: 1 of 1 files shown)


hw/drivers/led/lp5523/src/lp5523.c
```diff
@@ -635,20 +634,32 @@ lp5523_get_engine_mapping(struct led_itf *itf, uint8_t engine,
 static int
 lp5523_set_pr_instruction(struct led_itf *itf, uint8_t addr, uint16_t *ins)
 {
-    uint8_t mem[2] = { ((*ins) >> 8) & 0x00ff, (*ins) & 0x00ff };
+    int rc;
+
+    uint8_t byte1[1] = { ((*ins) >> 8) & 0x00ff };
```
kasjer: changing mem[2] to byte[1] byte2[1] is a little bit extreme and does improve anything I guess

hw/drivers/led/lp5523/src/lp5523.c
```diff
@@ -635,20 +634,32 @@ lp5523_get_engine_mapping(struct led_itf *itf, uint8_t engine,
...
 {
     int rc;
-    uint8_t mem[2];
+    uint8_t byte1[1];
+    uint8_t byte2[1];
+
+    addr = addr << 1;
 
-    rc = lp5523_get_n_regs(itf, LP5523_PROGRAM_MEMORY + (addr << 1), mem, 2);
+    rc = lp5523_get_n_regs(itf, LP5523_PROGRAM_MEMORY + (addr), byte1, 1);
```
kasjer: without shift operator () are not needed any more

hw/drivers/led/lp5523/src/lp5523.c
```diff
@@ -658,15 +669,18 @@ lp5523_verify_pr_instruction(struct led_itf *itf, uint8_t addr,
...
     int rc;
-    uint8_t mem[2];
+    uint8_t byte1[1];
+    uint8_t byte2[1];
     uint16_t ver;
 
-    rc = lp5523_get_n_regs(itf, LP5523_PROGRAM_MEMORY + (addr << 1), mem, 2);
+    addr = addr << 1;
+    rc = lp5523_get_n_regs(itf, LP5523_PROGRAM_MEMORY + (addr), byte1, 1);
+    rc |= lp5523_get_n_regs(itf, LP5523_PROGRAM_MEMORY + (addr + 1), byte2, 1);
```
kasjer: I guess lp5525_get_n_regs() does not return bit mask so this shortcut could be omitted. Same for other places.

hw/drivers/led/lp5523/src/lp5523.c
vrahane: @kasjer you mean doesn't, correct ?

vrahane:
@JustineKH , I agree with @kasjer's comments, can you please address those. Also, going forward can you please add a prefix for your commits and PR, not a very long one, just containing two words or less. Something like "drivers/lp5523: change multi byte instruction read write to single".

JustineKH:
@kasjer @vrahane I think I have covered all requests, please take another look when you get a chance.

PR #1703, opened 2019-03-20
nrf52: Add debugger override syscfg

vrahane:
- Add MCU_DEBUG_IGNORE_BKPT syscfg
- To be set if one wants to ignore debugger check

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_system.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_system.c
@@ -47,7 +47,9 @@ hal_system_reset(void)
             /*
              * If debugger is attached, breakpoint here.
              */
+#if !MYNEWT_VAL(MCU_DEBUG_IGNORE_BKPT)
             asm("bkpt");
+#endif
         }
         NVIC_SystemReset();
     }
@@ -56,11 +58,7 @@ hal_system_reset(void)
 int
 hal_debugger_connected(void)
 {
-#if MYNEWT_VAL(NRF52_HAL_DEBUGGER_CHK)
     return CoreDebug->DHCSR & CoreDebug_DHCSR_C_DEBUGEN_Msk;
-#else
-    return 0;
-#endif
 }
 
 /**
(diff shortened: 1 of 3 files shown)


vrahane:
@mkiiskila, @wes3 and @InternetofAwesome do you think this would be a good enough solution ?

andrzej-kaczmarek:
what is the purpose of this setting?

vrahane:
The MCU goes into a debug mode which it does not come out of and gets stuck on a breakpoint on doing a reset when a debugger is connected. There is no way to override this behavior. Hence, the setting just to skip this check and move on with a reset.

sterlinghughes:
I don't think this is the right way to do this, as it changes the behaviour of hal_debugger_connected() to be wrong.  I'd rather apply to the setting to not issue asm(bkpt), then change the behavior of the function to be wrong.

vrahane:
I spoke to @andrzej-kaczmarek in a DM, he was fine with adding MCU_DEBUG_IGNORE_DEBUGGER so that we can use it for other MCUs as well. I actually do not mind any implementation but with `asm("bkpt")` being called at various places, it seemed fine to use `hal_debugger_connected()`, so we have one central place for it.

vrahane:
@sterlinghughes I made the changes and tried it out. Works fine. I'll update the PR.

benmccrea:
@vrahane Hey, I'm late to the game here as this is merged, but I have a concern about this implementation. When the debugger is enabled the nRF52 will keep running off the HFXO, making for significantly higher current usage. Is it possible for it to kick itself out of debug mode, like `CoreDebug->DHCSR &= ~CoreDebug_DHCSR_C_DEBUGEN_Msk`? It seems like that would be preferable to just ignoring the asm("bkpt").

andrzej-kaczmarek:
@benmccrea it's not possible - `C_DEBUGEN` bit is not writable by core

vrahane:
@andrzej-kaczmarek We are not doing that, that bit gets reset on reboot. We just synced up on it. Also, you might be wrong about it not being writable, From the ARM reference manual:
`When C_DEBUGEN is set to 0, the processor ignores the values of all other bits in
this register.`

andrzej-kaczmarek:
@vrahane `This bit can only be written by the DAP, it ignores writes from software.` (_ARMv7-M Architecture Reference Manual, C1.6.2_)

PR #1704, opened 2019-03-20
compiler: Disable DSP extension for Cortex-M33

andrzej-kaczmarek:
DSP extension is optional and thus may not be supported by some MCUs.
Let's make this package generic one and disable DSP extension by
default - those who want to use DSP extension in their BSP can create
own compiler package with proper settings.


andrzej-kaczmarek:
@mkiiskila that was my first idea to resolve this, but `newt` does not support conditional expressions in `compiler.flags` so this was the easy and reasonable way. once `newt` is updated, I'm fine with syscfg setting.

PR #1706, opened 2019-03-20
Update CMSIS to 4.5.0

utzig:
Tested this on multiple cm0, cm3, cm4 and cm7 based MCUs, builds and run OK. Most changes are for v8m though, so might be interesting to test in one of those targets. Might be better to wait to merge after a new release (or not)...


andrzej-kaczmarek:
M33 builds and seems to work fine

utzig:
@andrzej-kaczmarek That was fast! :-P

utzig:
It's funny that CMSIS changed the license to Apache and every file includes a proper license header but RAT is not able to parse it...

sjanc:
cause rat assumes licence is first entry in file... this tool is less sophisticated than scripted grep  :)

for now lets just update rat exludes file, we can report this to rat too, so that it is smarter with matching licence and eg add support for SPDX too

utzig:
@sjanc The .rat-excludes file is already OK; the CI just doesn't use it.

mkiiskila:
You can report issues with RAT here https://issues.apache.org/jira/projects/RAT/ :)

PR #1707, opened 2019-03-20
spiflash performance improvements

kasjer:
- Wait times after erase and program are not configurable as specified in spiflash datasheets.
this should optimize Read Status Register command after writes to actual timing requirements.

- hal_flash interface was extended to allow erases using block size instead of sector size only

- spiflash erase functions are not public and can be used by application

- waiting time should work better for bootloader where os_time_delay is not used any more.

### DIFF
--- a/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
+++ b/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
@@ -109,8 +109,6 @@ struct spiflash_chip {
 #define FLASH_CAPACITY_16MBIT       0x15
 #define FLASH_CAPACITY_32MBIT       0x16
 
-int spiflash_init(const struct hal_flash *dev);
-
 void spiflash_power_down(struct spiflash_dev *dev);
 void spiflash_release_power_down(struct spiflash_dev *dev);
 
--- a/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
+++ b/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
@@ -89,6 +89,9 @@ extern struct spiflash_dev spiflash_dev;
 #define SPIFLASH_WRITE_ENABLE               0x06
 #define SPIFLASH_FAST_READ                  0x0B
 #define SPIFLASH_SECTOR_ERASE               0x20
+#define SPIFLASH_BLOCK_ERASE_32KB           MYNEWT_VAL(SPIFLASH_BLOCK_ERASE_32BK)
+#define SPIFLASH_BLOCK_ERASE_64KB           MYNEWT_VAL(SPIFLASH_BLOCK_ERASE_64BK)
+#define SPIFLASH_CHIP_ERASE                 0x60
 #define SPIFLASH_DEEP_POWER_DOWN            0xB9
 #define SPIFLASH_RELEASE_POWER_DOWN         0xAB
 #define SPIFLASH_READ_MANUFACTURER_ID       0x90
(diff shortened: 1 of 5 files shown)


hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -860,36 +860,59 @@ spiflash_read_status(struct spiflash_dev *dev)
...
+
 bool
 spiflash_device_ready(struct spiflash_dev *dev)
 {
     return !(spiflash_read_status(dev) & SPIFLASH_STATUS_BUSY);
 }
 
 int
-spiflash_wait_ready(struct spiflash_dev *dev, uint32_t timeout_ms)
+spiflash_wait_ready_till(struct spiflash_dev *dev, uint32_t usecs,
```
rymanluk: I would use  `timeout_us` instead of `usec`

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -860,36 +860,59 @@ spiflash_read_status(struct spiflash_dev *dev)
...
 {
-    uint32_t ticks;
-    os_time_t exp_time;
-    os_time_ms_to_ticks(timeout_ms, &ticks);
-    exp_time = os_time_get() + ticks;
-    int rc = 0;
+    int rc = -1;
+    uint32_t limit = os_cputime_get32() + usecs;
+
+    if (step_us < 5) {
```
rymanluk: some comment about this 5 would be nice :)

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -860,36 +860,59 @@ spiflash_read_status(struct spiflash_dev *dev)
...
-err:
     spiflash_unlock(dev);
 
     return rc;
 }
 
+int
+spiflash_wait_ready(struct spiflash_dev *dev, uint32_t timeout_ms)
+{
+    return spiflash_wait_ready_till(dev, timeout_ms * 1000, timeout_ms * 10);
```
rymanluk: Why it is `timeout_ms * 10`?

hw/drivers/flash/spiflash/src/spiflash.c
rymanluk: btw, seems that this function could be `static`

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -1143,6 +1155,72 @@ spiflash_sector_erase(struct spiflash_dev *dev, uint32_t addr)
...
+{
+    int rc = 0;
+
+    if (address == 0 && size == dev->hal.hf_size) {
+        return spiflash_chip_erase(dev);
+    }
+    address &= ~0xFFFU;
+    while (size) {
+        if ((address & 0xFFFFU) == 0 && (size >= 0x10000)) {
+            /* 64 KB erase if possible */
```
rymanluk: are there flash which could not support block erase? If so maybe we should have MYNEWT_VAL which would enable block erase.

hw/drivers/flash/spiflash/include/spiflash/spiflash.h
```diff
@@ -33,6 +33,24 @@
...
+ */
+struct spiflash_time_spec {
+    uint32_t typical;
+    uint32_t maximum;
+};
+
+struct spiflash_characteristics {
+    struct spiflash_time_spec tse;  /* Sector erase time (4KB) */
+    struct spiflash_time_spec tbe1; /* Block erase time (32KB) */
+    struct spiflash_time_spec tbe2; /* Block erase time (64KB) */
```
wes3: I have not looked at spi flashes lately but these sector/block sizes seem specific. What if there is a block that is much bigger or something? Are these things generally linear and one could simply take a 32KB sector and multiply it by sector size/32KB? Or maybe we can get things from the specific spi flash itself (meaning have calls in to the specific spi flash driver code)?

hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -985,18 +1013,23 @@ hal_spiflash_write(const struct hal_flash *hal_flash_dev, uint32_t addr,
...
 
     spiflash_lock(dev);
 
-    while (len) {
-        if (spiflash_wait_ready(dev, 100) != 0) {
-            rc = -1;
-            goto err;
-        }
+    if (spiflash_wait_ready(dev, 100) != 0) {
+        rc = -1;
```
wes3: Should this 100 magic number be either a definition or mynewt val?

PR #1708, opened 2019-03-20
imgrmgr: dfu callback fixes

JustineKH:
Added check for NULL imgmgr_dfu_callbacks_fn pointer
Added/renamed imgmgr callouts

### DIFF
--- a/mgmt/imgmgr/src/imgmgr.c
+++ b/mgmt/imgmgr/src/imgmgr.c
@@ -856,8 +856,7 @@ end:
 void
 imgmgr_dfu_stopped(void)
 {
-    if (imgmgr_dfu_callbacks_fn && imgmgr_dfu_callbacks_fn->dfu_stopped_cb != NULL)
-    {
+    if (imgmgr_dfu_callbacks_fn && imgmgr_dfu_callbacks_fn->dfu_stopped_cb) {
         imgmgr_dfu_callbacks_fn->dfu_stopped_cb();
     }
 }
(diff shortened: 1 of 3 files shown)


mgmt/imgmgr/src/imgmgr.c
```diff
@@ -854,29 +854,38 @@ imgr_upload(struct mgmt_cbuf *cb)
 }
 
 void
-imgr_dfu_stopped(void)
+imgmgr_dfu_stopped(void)
 {
-    if (imgmgr_dfu_callbacks_fn->dfu_stopped_cb != NULL)
+    if (imgmgr_dfu_callbacks_fn && imgmgr_dfu_callbacks_fn->dfu_stopped_cb != NULL)
```
kasjer: this statements checks two pointers, one is compared to NULL other is treated as boolean value, both are correct but using two different styles in one line is strange.

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -854,29 +854,38 @@ imgr_upload(struct mgmt_cbuf *cb)
...
     }
 }
 
 void
-imgr_dfu_started(void)
+imgmgr_dfu_started(void)
 {
-    if (imgmgr_dfu_callbacks_fn->dfu_started_cb != NULL)
+    if (imgmgr_dfu_callbacks_fn && imgmgr_dfu_callbacks_fn->dfu_started_cb != NULL)
     {
```
kasjer: I know this is was not changed in this commit, but since "if" line is changed maybe adjusting opening { placement could also be corrected. Same for other places that slipped through previous review.

mgmt/imgmgr/src/imgmgr.c
JustineKH: I do agree

mgmt/imgmgr/src/imgmgr.c
JustineKH: Yes, this has been fixed now

PR #1714, opened 2019-03-22
logs_nmgr: add handlers for log module get/set

vrahane:


vrahane:
newtmgr tool support for this will be added later

sys/log/common/include/log_common/log_common.h
```diff
@@ -96,6 +96,7 @@ struct log;
 #define LOGS_NMGR_OP_LEVEL_LIST   	(4)
 #define LOGS_NMGR_OP_LOGS_LIST    	(5)
 #define LOGS_NMGR_OP_SET_WATERMARK	(6)
+#define LOGS_NMGR_OP_MODLEVEL	        (8)
```
ccollins476ad: What happened to 7? :)

vrahane: oops, sorry about that. I had two separate IDs earlier and I deleted the 7 instead of the 8.

PR #1715, opened 2019-03-22
time/timepersist; allow user to persist device time.

mkiiskila:
This is to help platforms which don't have RTC retaining time over resets.
Offer 2 options initially, sys/config (works for all, remembers time from long time back) or nvreg (will work for some architectures).

### DIFF
--- a/time/timepersist/src/time_persist_nvreg.c
+++ b/time/timepersist/src/time_persist_nvreg.c
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <os/mynewt.h>
+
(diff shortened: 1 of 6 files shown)


time/timepersist/src/time_persist_nvreg.c
```diff
@@ -0,0 +1,75 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <os/mynewt.h>
+
+#if MYNEWT_VAL(TIMEPERSIST_NVREG_INDEX) >= 0
+#if MYNEWT_VAL(TIMEPERSIST_SYS_CONFIG)
+#error "Can't have both methods of persisting"
+#endif
```
utzig: Is this required if you already added a restriction on the syscfg?

time/timepersist/syscfg.yml
```diff
@@ -0,0 +1,49 @@
...
+#
+    TIMEPERSIST_NVREG_INDEX:
+        description: >
+            Index of retained register to use (using hal_nvreg) for storing
+            lower 32 bits of wallclock time.
+        value: -1
+    TIMEPERSIST_SYS_CONFIG:
+        description: >
+            Use sys/config to persist wallclock time.
+        value: 0
```
andrzej-kaczmarek: you could use choice for this, i.e.:
```
TIMEPERSIST_STORE:
    description: foo
    value:
    choices:
        - nvreg
        - config
```
then you can check in code which one is selected like this:
```
#if MYNEWT_VAL_CHOICE(TIMEPERSIST_STORE, config)
#endif
```
and you don't have to worry if both are enabled because it's not possible

time/timepersist/src/time_persist_nvreg.c
mkiiskila: Strictly speaking no. I've had mixed success previously when writing restrictions.
But I guess I can remove this check, as the build would fail eventually due to duplicate symbols, if both ended up defined at the same time :)

time/timepersist/syscfg.yml
mkiiskila: Appealing, but I'd still need a restriction to make sure that the nvreg index was set to something. So I think I'll keep the current method.

mkiiskila:
Changed the time to be persisted using sysdown in controlled shutdown.

PR #1716, opened 2019-03-22
net/oic; if resource handler fails a request, don't honor observer option.

mkiiskila:
Previously it would only disregard the observer option if the request was rejected within common RX processing. Now also prevent observer being added if the resource handler rejects the request.

### DIFF
--- a/net/oic/src/api/oc_ri.c
+++ b/net/oic/src/api/oc_ri.c
@@ -577,8 +577,7 @@ oc_ri_invoke_coap_entity_handler(struct coap_packet_rx *request,
     response_buffer.response_length = 0;
     response_buffer.code = oc_status_code(OC_STATUS_METHOD_NOT_ALLOWED);
     success = false;
-  }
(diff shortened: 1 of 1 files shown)


net/oic/src/api/oc_ri.c
```diff
@@ -587,6 +586,8 @@ oc_ri_invoke_coap_entity_handler(struct coap_packet_rx *request,
     response_buffer.response_length = 0;
     response_buffer.code = oc_status_code(OC_STATUS_FORBIDDEN);
     success = false;
+  } else if (response_buffer.code >= BAD_REQUEST_4_00) {
```
utzig: I think it would be more idiomatic by writing `response_buffer.code >= oc_status_code(OC_STATUS_BAD_REQUEST)`

mkiiskila: Yeah, I did consider it. However, given that the error code mappings are present and done within this very file, I decided that it was worth it.

PR #1719, opened 2019-03-25
spiflash: Fix timeout calculation

kasjer:
Time calculation did work correctly on non-1MHz cpu timer.

### DIFF
--- a/hw/drivers/flash/spiflash/src/spiflash.c
+++ b/hw/drivers/flash/spiflash/src/spiflash.c
@@ -920,7 +920,7 @@ spiflash_wait_ready_till(struct spiflash_dev *dev, uint32_t timeout_us,
     uint32_t step_us)
 {
     int rc = -1;
-    uint32_t limit = os_cputime_get32() + timeout_us;
+    uint32_t limit;
 
     if (dev->ready) {
         return 0;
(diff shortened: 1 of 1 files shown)


hw/drivers/flash/spiflash/src/spiflash.c
```diff
@@ -1130,12 +1131,13 @@ spiflash_execute_erase(struct spiflash_dev *dev, const uint8_t *buf,
...
     dev->ready = false;
 
-    start_time = os_cputime_ticks_to_usecs(os_cputime_get32());
+    start_time = os_cputime_get32();
     /* Wait typical erase time before starting polling for ready */
     spiflash_delay_us(delay_spec->typical);
 
     /* Poll status ready for remaining time */
-    wait_time_us = start_time + delay_spec->maximum - os_cputime_get32();
+    wait_time_us = delay_spec->maximum -
```
wes3: Is there any chance that the time has advanced such that this calculation "goes negative" and thus ends up with a super high wait time?

Changes look good otherwise.

kasjer: Your are right, it could happen when delay times were very short, specially when maximum time was shorter then typical time.

wes3: I think I might be a bit confused with the force push but will make comments on what I think the changes were. For some reason I do not like the typecast of the unit32_t timeout to an int32_t to check if it is negative. I would have made sure that wait_time_us did not go negative. If anything, I would check the input timeout_us to see if it is too large but keep it uint32_t (not a check for negative).

kasjer: I'm sorry about force push, but change is simple enough.

PR #1724, opened 2019-03-26
kernel/os: MSYS memory pool not in mempool list

wes3:
Change the order of os_msys_init() and os_mempool_module_init() so that memory pools allocated by msys show up in the mempool list. Removed the initialization of g_os_mempool_list as it is initialized when the module init function is called. Looking at the code I wonder if the order of where things are done in os_pkg_init() might be an issue. For example, os_dev_initialize_all() is called which will call os device initialization functions if their init stage is OS_DEV_INIT_KERNEL.


wes3:
Yes, that is what I am worried about.

PR #1729, opened 2019-03-28
[STM32] Enable timer0 for all BSPs

utzig:


utzig:
@sjanc @kasjer This standardizes `TIMER_0: 1`, unfortunately I have not tested each one, and will only be able to on Monday.

PR #1731, opened 2019-03-29
spiflash: Add simple cache

kasjer:
Reading 1 byte from spiflash generates 5 bytes of SPI traffic.
Reading 8 bytes on by one is 40 bytes on SPI.
Reading 8 bytes at once is 12 bytes on SPI.
This change improves reading time when there are a lot of small
read requests.
This can be useful if FCB is stored on external spiflash.
Tested on 440KB FCB that stored CBOR data
fcb_walk time in ms
no cache:   53625     100%
 4 bytes:   41570      77%
 8 bytes:   38695      72%
12 bytes:   36695      68%
16 bytes:   37031      69%
20 bytes:   36875      68%

### DIFF
--- a/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
+++ b/hw/drivers/flash/spiflash/include/spiflash/spiflash.h
@@ -79,6 +79,10 @@ struct spiflash_dev {
 #endif
     bool pd_active;                 /* Power down active */
 #endif
+#if MYNEWT_VAL(SPIFLASH_CACHE_SIZE)
+    uint32_t cached_addr;
+    uint8_t cache[MYNEWT_VAL(SPIFLASH_CACHE_SIZE)];
+#endif
 };
 
 extern struct spiflash_dev spiflash_dev;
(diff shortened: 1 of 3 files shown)


mkiiskila:
I like the idea, and I want this kind of functionality in. Given the cost of external SPI flash access, this is very valuable.

Few comments:
You need to handle the case of writes hitting the cache.
Most of the FCB accesses indeed walk forward in offset, so the way you're serving data from cache gets most of the benefits. I'd like to also see it serve data from cache if the trailing part of the read hits there, not just the beginning.

One thought I got was the code probably would look 'nicer' if you had the hal_spiflash_read() routine serve data from cache (if available), and issue reads to hal_spiflash_actual_read() (?) to fill in the missing pieces. And maybe follow the same model for write case. But this part is just a suggestion :)

kasjer:
When I wrote this I was not sure what will be gained so I did not bother to make it work for all cases.
All writes and erases invalidate the cache, two one-liners to make sure data is consistent.
I tried to leave original code as it was when cache is not enabled so nothing is broken.

PR #1737, opened 2019-04-02
mgmt/imgmgr: Allow upgrade-only requests

ccollins476ad:
This commit adds an additional boolean field to the "image upload" request: `upgrade`.  If this field is true, the firmware will only accept an upload attempt if the new image's version number is greater than that of the currently running image (using a semver-compatible comparision).

If the `upgrade` field is false or missing, no version check is performed.

The firmware only processes the `upgrade` field on the first image chunk.  It is harmless, but pointless, for a client to specify `upgrade` in any subsequent chunks.

Note: This PR has siblings:
* newtmgr: https://github.com/apache/mynewt-newtmgr/pull/120
* mcumgr: https://github.com/apache/mynewt-mcumgr/pull/25


sterlinghughes:
can we please make the corresponding change to mcumgr as well to keep them in synch?

ccollins476ad:
Thanks, @sterlinghughes.  I have submitted a PR to mcumgr (https://github.com/apache/mynewt-mcumgr/pull/25).

PR #1739, opened 2019-04-03
kernel/os_msys: Improve finding mpool

rymanluk:
This PR does two things
1) When it search for mpool for given data size it pick mempool with a databuf_len closes to requested datasize. With this Mynewt tries to me smart and not used mempools with big block size for small data 

2) When user requests msys mbuf with data size 0, Mynewt assumes that data might be big, therefore it will assign mempool with biggest block size. It is done because if there would be msys_2 available with small block size for e.g. some headers as in https://github.com/apache/mynewt-nimble/pull/380, those small msys blocks would be consumed by  others e.g. OIC, where datasize 0 is requested for the reason that it is hard to estimate how big mbuf is needed.

### DIFF
--- a/kernel/os/src/os_msys.c
+++ b/kernel/os/src/os_msys.c
@@ -87,6 +87,13 @@ os_msys_reset(void)
 }
 
 static struct os_mbuf_pool *
+os_msys_find_biggest_pool(void)
+{
+    /* Mempools are sorted by the blocksize, so just return last one */
+    return STAILQ_LAST(&g_msys_pool_list, os_mbuf_pool, omp_next);
+}
+
+static struct os_mbuf_pool *
 os_msys_find_pool(uint16_t dsize)
 {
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
it would work even better if we just sort pools by buffer size when registering them

rymanluk:
thanks @wes3 , I saw the @andrzej-kaczmarek  comment too and like the idea. Will do it. 

btw, wondering why oic tests failes.. need to look into it.

rymanluk:
improved according to @andrzej-kaczmarek  comment

andrzej-kaczmarek:
so if pools re now sorted you do not need to loop over entire queue to find largest one - it will be the last one

rymanluk:
@andrzej-kaczmarek  yeah, check the force pushed version two minutes before your comment :)   funny I was just writing this change while you were writing this comment.

PR #1741, opened 2019-04-04
Add initial  support for Dialog DA1469x MCU family

andrzej-kaczmarek:
This adds initial support for Dialog DA1469x MCU family and BSP for DA1469x Pro Development Kit. More about the MCU itself here: https://www.dialog-semiconductor.com/products/da1469x-product-family

Supported features:
- all HALs
- ADC (both GP and ΣΔ) also with DMA support
- PWM driver
- TRNG driver
- BLE controller (via separate PR on NimBLE)

Port is fully functional, however it is not complete and will be extended over time (in particular, power consumption is not optimized at all).

Flashing script currently is provided for Linux/OSX only and uses `apps/flash_loader` (has to be build separately) to write image.

Kudos to @kasjer @mkiiskila and @rymanluk for implementing many parts of this PR.


utzig:
Time to update the gcc toolchain on travis... ;-)

PR #1743, opened 2019-04-04
Additional Wake Up / Reset Reason

brolan-juul:
Obtaining more information on reasons why it is resetting. Adding wake up from system off.

### DIFF
--- a/hw/hal/include/hal/hal_system.h
+++ b/hw/hal/include/hal/hal_system.h
@@ -66,10 +66,10 @@ enum hal_reset_reason {
     HAL_RESET_SOFT = 4,
     /** Low supply voltage */
     HAL_RESET_BROWNOUT = 5,
-    /** Power on */
-    HAL_RESET_POWER_OFF = 6,
     /** Restart due to user request */
-    HAL_RESET_REQUESTED = 7,
+    HAL_RESET_REQUESTED = 6,
+    /** System Off, wakeup on external interrupt*/
+    HAL_RESET_SYS_OFF_INT = 7,
 };
 
 /**
--- a/hw/hal/include/hal/hal_system.h
+++ b/hw/hal/include/hal/hal_system.h
@@ -66,8 +66,10 @@ enum hal_reset_reason {
     HAL_RESET_SOFT = 4,
     /** Low supply voltage */
     HAL_RESET_BROWNOUT = 5,
+    /** Power on */
+    HAL_RESET_POWER_OFF = 6,
     /** Restart due to user request */
-    HAL_RESET_REQUESTED = 6,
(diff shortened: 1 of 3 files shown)


hw/hal/include/hal/hal_system.h
```diff
@@ -66,8 +66,10 @@ enum hal_reset_reason {
     HAL_RESET_SOFT = 4,
     /** Low supply voltage */
     HAL_RESET_BROWNOUT = 5,
+    /** Power on */
+    HAL_RESET_POWER_OFF = 6,
     /** Restart due to user request */
-    HAL_RESET_REQUESTED = 6,
+    HAL_RESET_REQUESTED = 7,
```
sjanc: any reason to change HAL_RESET_REQUESTED value? why not just add POWER_OFF at the end of enum?

hw/hal/include/hal/hal_system.h
```diff
@@ -66,8 +66,10 @@ enum hal_reset_reason {
     HAL_RESET_SOFT = 4,
     /** Low supply voltage */
     HAL_RESET_BROWNOUT = 5,
+    /** Power on */
```
kasjer: I would change it to HAL_RESET_SYSTEM_OFF or such. Now we have comment stating Power on (which is double confusing:  first see comment on HAL_RESET_POR, second name is HAL_RESET_POWER_OFF)

hw/mcu/nordic/nrf52xxx/src/hal_reset_cause.c
```diff
@@ -37,6 +37,8 @@ hal_reset_cause(void)
         reason = HAL_RESET_SOFT;
     } else if (reg & POWER_RESETREAS_RESETPIN_Msk) {
         reason = HAL_RESET_PIN;
+    } else if(reg & POWER_RESETREAS_OFF_Msk) {
```
kasjer: space is missing afater if

hw/hal/include/hal/hal_system.h
brolan-juul: Good point. Moving it back to original lineup.

hw/hal/include/hal/hal_system.h
brolan-juul: Updated comment to better reflect as to what is happening and to reduce confusion. 

`    /** System Off, wakeup on external interrupt*/`

hw/mcu/nordic/nrf52xxx/src/hal_reset_cause.c
brolan-juul: Good catch. Fixed.

brolan-juul:
@kasjer Noted. Will get to it soon.

PR #1750, opened 2019-04-08
LED_2 is used by the blemesh example application 

iandmorris:
Unfortunately there is only 1 LED that can be controlled by a GPIO on the DA1469x development kit boards (daughter plus motherboard). Therefore LED_2 is defined to to use the same GPIO as LED_1, which will be ok as long as they are not both used by the same application...


iandmorris:
Good comment, i've updated to use an unused gpio that is brought out to a header allowing easy attachment of an external LED.

PR #1752, opened 2019-04-10
mbedtls; add a routine to add AAD data incrementally.

mkiiskila:


sterlinghughes:
Is this a core mbedtls function or added by us?  If added by us, we should find a way to not have it in core distro (or upstream fix), so that we can maintain / upgrade mbedtls more easily.

mkiiskila:
There's a PR against upstream mbedtls now. Includes this change as well as the previously added GCM related routine. See https://github.com/ARMmbed/mbedtls/pull/2583

utzig:
Btw, the CI did not pass for upstream, I'm not sure they will review it in this state.

utzig:
The reason the CI failed it that `mbedtls_gcm_setkey_noalloc` has an argument named `cipher_info` and in the doxygen header you wrote `\param cipher    The 128-bit block cipher to use.`. Both names should match, and fixing this should enable it to pass...

PR #1760, opened 2019-04-11
boot: Init watchdog

kasjer:
Watchdog is used during image swap unconditionally.
If watchdog is not initialized kicking it may result in
undesired behaviour.

### DIFF
--- a/apps/boot/src/boot.c
+++ b/apps/boot/src/boot.c
@@ -29,6 +29,7 @@
 #include <console/console.h>
 #include "bootutil/image.h"
 #include "bootutil/bootutil.h"
+#include "hal/hal_watchdog.h"
 
 #define BOOT_AREA_DESC_MAX  (256)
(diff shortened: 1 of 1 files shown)


utzig:
When `OS_SCHEDULING` is enabled, the watchdog is already initialized by the idle thread, so this avoids double initialization. I don't remember why `WATCHDOG_INTERVAL` is tested but boot serial seems to always tickle, so it might be a bug, although one that has never shown up!

kasjer:
`OS_SCHEDULING` is set to 0 in **syscfg.yml** for this app

utzig:
As it should, but there were circumstances where we were testing with it set (one can always change the target to set it), although it ends up in redundancy but better to be safe... Btw, I am not suggesting you do the same here, I think it's OK, I was just replying to wills question.

wes3:
Fabio, were you going to approve this so it gets merged? I was hoping someone with more bootloader experience would approve it.

PR #1763, opened 2019-04-12
Replace bootutils with mcuboot

sjanc:
'newt test all' doesn't work due to issues on mcuboot side and need to be fixed before this PR can go in

Initial idea is that core master is using latest mcuboot release (not master), and for core release we stick to latest available version (eg 1.3.0)

There is also a question about newt create-image -2 option... should we change default behavior for newt and always use version 2 now?


apps/boot/pkg.yml
```diff
@@ -18,18 +18,5 @@
...
-
-pkg.deps:
-    - "@apache-mynewt-core/boot/bootutil"
-    - "@apache-mynewt-core/kernel/os"
-    - "@apache-mynewt-core/sys/console/stub"
-    - "@apache-mynewt-core/sys/log/stub"
-
-pkg.deps.BOOT_SERIAL:
-    - "@apache-mynewt-core/boot/boot_serial"
+pkg.type: transient
```
rymanluk: Maybe we should remove it at all as it is in the end bit different bootloader.

utzig:
> There is also a question about newt create-image -2 option... should we change default behavior for newt and always use version 2 now?

After the switch, definitely yes, otherwise `create-image` would be generating incompatible images! ;-)

sjanc:
yes, after switch of course :)

PR #1765, opened 2019-04-12
hw/mcu/dialog: Move common code in I2C to helper

andrzej-kaczmarek:


hw/mcu/dialog/da1469x/src/hal_i2c.c
```diff
@@ -210,40 +217,14 @@ hal_i2c_init(uint8_t i2c_num, void *usercfg)
...
 
     cfg.frequency = da1469x_cfg->frequency;
     rc = i2c_config(i2c, &cfg);
     if (rc) {
         return rc;
     }
 
-    i2c->regs->I2C_INTR_MASK_REG = 0x0000;
-
-    NVIC_EnableIRQ(i2c->irqn);
```
rymanluk: looks like now interrupt is enabled inside `i2c_init_hw()` before `i2c_config()` is called and this does not seems to be right. On the other side we support only MASTER role so it should be acceptable.

PR #1767, opened 2019-04-12
Fix copying CI targets

utzig:
CI currently expects target tree inside targets directory and uses a hardcoded path. The targets directory might not exist on some of the repos tested, so create it first here. CI will be updated to use the repo slug to determine which repo's targets it should build.


utzig:
Related to JuulLabs-OSS/mynewt-travis-ci#48

PR #1769, opened 2019-04-15
da1469x: Disable PD_COM and PD_PER when not used

andrzej-kaczmarek:


hw/mcu/dialog/da1469x/src/hal_gpio.c
```diff
@@ -74,6 +76,9 @@ static struct hal_gpio_irq hal_gpio_irqs[HAL_GPIO_MAX_IRQ];
 
 static uint32_t hal_gpio_latch_state[2];
 
+static uint8_t g_mcu_gpio_retained_num;
+static struct da1469x_retreg g_mcu_gpio_retained[ MYNEWT_VAL(MCU_GPIO_RETAINABLE_NUM) ];
```
kasjer: Unusual spaces inside []

PR #1770, opened 2019-04-16
sys/console: start pkt should set g_is_output_nlip

vrahane:


vrahane:
it also gets used in `handle_nlip()`, we cannot get rid of it, I tested it out. We also have to track `\n`

ccollins476ad:
I only see it being used in `console_write()`, so I think it can be made into a normal local variable (non static).

One other minor thing - I would really prefer to see a brief comment above the
```
    if (cnt >= 3 && str[1] == CONSOLE_NLIP_PKT_START1 &&
```
line.  It seems weird that we check the first two bytes for "DATA" packets, but check bytes 2 and 3 for "START" packets (I know why this is the case, but I doubt anyone else does!)

PR #1778, opened 2019-04-22
sys/log/full: Allow for limiting size of newtmgr log response

wes3:
Add a syscfg value (LOG_NMGR_MAX_RSP_LEN) which will limit the size of a newtmgr log response. If the response exceeds this limit and there is only one entry in the response the entry type is set to string and the message "error: entry too large (xxx bytes)" is encoded as in the message portion of the response. If the response contains multiple entries the code is unchanged (the additional entry is not added and the encode function returns OS_ENOMEM).

Note that the previous code had a hard-coded limit of 400 bytes for a newtmgr log response but this was not applied to a response with a single log entry in it. This fix applies the same limit to a response with one or more entries in it.

### DIFF
--- a/sys/log/full/src/log_nmgr.c
+++ b/sys/log/full/src/log_nmgr.c
@@ -79,6 +79,8 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
     uint8_t data[128];
     int rc;
     int rsp_len;
+    uint8_t etype;
+    bool too_long;
     CborError g_err = CborNoError;
     struct log_encode_data *ed = log_offset->lo_arg;
     CborEncoder rsp;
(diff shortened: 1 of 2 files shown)


sys/log/full/src/log_nmgr.c
```diff
@@ -220,18 +247,29 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
...
      */
     g_err |= cbor_encoder_create_indef_byte_string(&rsp, &str_encoder);
-    for (off = 0; off < len && !g_err; ) {
-        rc = log_read_body(log, dptr, data, off, sizeof(data));
-        if (rc < 0) {
-            g_err |= 1;
-            break;
-        }
+    if (too_long) {
+        data[0] = 0;
```
ccollins476ad: This line (`data[0] = 0;`) is not necessary, is it?  `sprintf()` always null-terminates the destination string, so I think this can be removed.  This same comment applies to the deprecated `LOG_VERSION` case below.

wes3: Yeah, I do this always with string stuff when I do not need to. Addressed the issues. Thanks for the review!

PR #1780, opened 2019-04-25
LIS2DW12 Self Test

brolan-juul:
Adding ability to check individual axis to see which one is broken.


brolan-juul:
That is not a bad idea. I was wondering how it could be done to avoid API changes.

PR #1782, opened 2019-04-25
sys/log: Add setting for maximum log entry size when writing.

wes3:
This commit adds the ability to set a maximum log entry for a given log. By default, the log maximum entry length is set to 0. A value of 0 for this field means that no check should be made for maximum log entry length. If set to a non-zero value, the log write routine(s) will check the log body length (does nt include the size of the log entry header) and if it exceeds the maximum, OS_ENOMEM will be returned and the too_long log stat will be incremented. A new API was added named log_set_max_entry_len() which allows the user to set the maximum entry length.

### DIFF
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -69,6 +69,7 @@ STATS_NAME_START(logs)
   STATS_NAME(logs, drops)
   STATS_NAME(logs, errs)
   STATS_NAME(logs, lost)
+  STATS_NAME(logs, too_long)
 STATS_NAME_END(logs)
 #endif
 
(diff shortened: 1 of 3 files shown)


sys/log/full/src/log.c
```diff
@@ -569,6 +599,21 @@ log_append_mbuf_typed_no_free(struct log *log, uint8_t module, uint8_t level,
...
+     * Check that the log body length is less than the maximum entry. This code
+     * may appear a bit odd in that it checks that the length is greater than
+     * a log entry header length. The reason for this check is to ensure any
+     * error handling of this case to be the same as it was before the
+     * maximum entry length was checked.
+     */
+    len = os_mbuf_len(om);
+    if (len > LOG_ENTRY_HDR_SIZE) {
+        rc = log_chk_max_entry_len(log, len - LOG_ENTRY_HDR_SIZE);
+        if (rc) {
```
rymanluk: nitpick: once `if (rc )` is used, other time ` if(rc != OS_OK)` ;)

wes3: Fixed and pushed.

PR #1783, opened 2019-04-26
Modified the nRF5x timer and system init code to use the nRF HFXO manager

amrbekhit:
This allows the timer code to interoperate with other code that uses the HFXO (such as BLE) such that the HFXO only enabled when required and turned off when not in use, saving power.

This is necessary in order for the LoRa low power changes implemented [in this pull request](https://github.com/apache/mynewt-core/pull/1670) to have an effect.

### DIFF
--- a/hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
+++ b/hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
@@ -39,6 +39,7 @@ nrf51_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf51_clock_hfxo_refcnt < 0xff);
     if (nrf51_clock_hfxo_refcnt == 0) {
+        NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
         NRF_CLOCK->TASKS_HFCLKSTART = 1;
         started = 1;
     }
@@ -46,6 +47,15 @@ nrf51_clock_hfxo_request(void)
     __HAL_ENABLE_INTERRUPTS(ctx);
 
     return started;
+}
+
+/**
+ * Wait for the HFXO clock to be turned on.
+ */
+void
+nrf51_clock_hfxo_wait_until_started(void)
+{
+    while (NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
 }
 
 /**
--- a/hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
+++ b/hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
@@ -39,23 +39,17 @@ nrf51_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf51_clock_hfxo_refcnt < 0xff);
     if (nrf51_clock_hfxo_refcnt == 0) {
+        /* Make sure that the HFXO has not previously been started manually */
+        assert(NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
         NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
         NRF_CLOCK->TASKS_HFCLKSTART = 1;
+        while (NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
         started = 1;
     }
     ++nrf51_clock_hfxo_refcnt;
     __HAL_ENABLE_INTERRUPTS(ctx);
 
     return started;
-}
-
-/**
- * Wait for the HFXO clock to be turned on.
- */
-void
-nrf51_clock_hfxo_wait_until_started(void)
-{
-    while (NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
 }
 
 /**
--- a/hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
+++ b/hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
@@ -40,7 +40,8 @@ nrf51_clock_hfxo_request(void)
     assert(nrf51_clock_hfxo_refcnt < 0xff);
     if (nrf51_clock_hfxo_refcnt == 0) {
         /* Make sure that the HFXO has not previously been started manually */
-        assert(NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
+        assert((NRF_CLOCK->HFCLKSTAT & CLOCK_HFCLKSTAT_STATE_Msk) !=
+            (CLOCK_HFCLKSTAT_STATE_Running << CLOCK_HFCLKSTAT_STATE_Pos));
         NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
         NRF_CLOCK->TASKS_HFCLKSTART = 1;
         while (NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
--- a/hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
+++ b/hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
@@ -39,6 +39,7 @@ nrf52_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf52_clock_hfxo_refcnt < 0xff);
     if (nrf52_clock_hfxo_refcnt == 0) {
+        NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
         NRF_CLOCK->TASKS_HFCLKSTART = 1;
         started = 1;
     }
@@ -46,6 +47,15 @@ nrf52_clock_hfxo_request(void)
     __HAL_ENABLE_INTERRUPTS(ctx);
 
     return started;
+}
+
+/**
+ * Wait for the HFXO clock to be turned on.
+ */
+void
+nrf52_clock_hfxo_wait_until_started(void)
+{
+    while (NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
 }
 
 /**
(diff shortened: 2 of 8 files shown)


amrbekhit:
> code looks good, but I think we should ensure that hfxo started after we requested it since original code did this, otherwise in some corner case we may change actual behavior here. an additional API like `nrf52_clock_hfxo_wait_until_started` would be just fine to do this.

This has now been implemented. Please review the latest commit.

andrzej-kaczmarek:
this pulls in a lot of unrelated changes, you probably need to rebase and push again

amrbekhit:
> this pulls in a lot of unrelated changes, you probably need to rebase and push again

I've tried to rebase and push again - the modified files are now correct, but the 30+ commits remain. Not sure how best to fix this - I can close this pull request and create a new clean one if that's better.

amrbekhit:
> I have a concern with the nrf52_clock_hfxo_wait_until_started() code. If you look at the original code it checked to see if HFCLK was already started and, if so, did not wait to see if EVENTS_HFCLKSTART gets set. I do not know what the chip would do if HFCLK was already started; would it set EVENTS_HFCLKSTART again after it is cleared? If it does not this code would loop infinitely. I do understand that you are probably assuming that if the counter is zero it means that the clock has not been started and there is no need to check it. I think I would check to see if the clock is already started before waiting forever. Would welcome other opinions about this.

Why don't we just include the wait for the clock to start in the original `nrf5x_clock_hfxo_request` functions? The original code in the hal_system.c files does exactly that. It seems reasonable that you might always want to wait for the HFXO to start up before using it. Is there any benefit to *not* waiting?

andrzej-kaczmarek:
> > this pulls in a lot of unrelated changes, you probably need to rebase and push again
> 
> I've tried to rebase and push again - the modified files are now correct, but the 30+ commits remain. Not sure how best to fix this - I can close this pull request and create a new clean one if that's better.

if you still have other commits after rebase then it means you did not do proper rebase - closing and creating new one won't fix it since you will push the same commits again.

if you do not know how to do proper rebase, just start from scratch by applying your changes on top of latest master (latest in git, not in newt) and then do force push to github.

amrbekhit:
> > > this pulls in a lot of unrelated changes, you probably need to rebase and push again
> > 
> > 
> > I've tried to rebase and push again - the modified files are now correct, but the 30+ commits remain. Not sure how best to fix this - I can close this pull request and create a new clean one if that's better.
> 
> if you still have other commits after rebase then it means you did not do proper rebase - closing and creating new one won't fix it since you will push the same commits again.
> 
> if you do not know how to do proper rebase, just start from scratch by applying your changes on top of latest master (latest in git, not in newt) and then do force push to github.

I've managed to properly rebase it now.

mlaz:
> > I have a concern with the nrf52_clock_hfxo_wait_until_started() code. If you look at the original code it checked to see if HFCLK was already started and, if so, did not wait to see if EVENTS_HFCLKSTART gets set. I do not know what the chip would do if HFCLK was already started; would it set EVENTS_HFCLKSTART again after it is cleared? If it does not this code would loop infinitely. I do understand that you are probably assuming that if the counter is zero it means that the clock has not been started and there is no need to check it. I think I would check to see if the clock is already started before waiting forever. Would welcome other opinions about this.
> 
> Why don't we just include the wait for the clock to start in the original `nrf5x_clock_hfxo_request` functions? The original code in the hal_system.c files does exactly that. It seems reasonable that you might always want to wait for the HFXO to start up before using it. Is there any benefit to _not_ waiting?

Hi @amrbekhit, that seems reasonable, I wonder if could there be a case on which we'd want to do something in between(i.e. setting some registers), which would actually result in less idle time? Anyway there should be no problem having it all on the same function. Your rebased code still doesn't feature any checking on whether HFCLK was already started before blocking. Can you please do those changes?

amrbekhit:
> > > I have a concern with the nrf52_clock_hfxo_wait_until_started() code. If you look at the original code it checked to see if HFCLK was already started and, if so, did not wait to see if EVENTS_HFCLKSTART gets set. I do not know what the chip would do if HFCLK was already started; would it set EVENTS_HFCLKSTART again after it is cleared? If it does not this code would loop infinitely. I do understand that you are probably assuming that if the counter is zero it means that the clock has not been started and there is no need to check it. I think I would check to see if the clock is already started before waiting forever. Would welcome other opinions about this.
> > 
> > 
> > Why don't we just include the wait for the clock to start in the original `nrf5x_clock_hfxo_request` functions? The original code in the hal_system.c files does exactly that. It seems reasonable that you might always want to wait for the HFXO to start up before using it. Is there any benefit to _not_ waiting?
> 
> Hi @amrbekhit, that seems reasonable, I wonder if could there be a case on which we'd want to do something in between(i.e. setting some registers), which would actually result in less idle time? Not sure. Anyway there should be no problem having it all on the same function.

In theory, you could turn it on and then go away and do something else, but then we would need to use the interrupt the notify the starting code that it is ready. Seeing as none of the existing code seems to need to do this, I would argue let's keep it simple for now.

amrbekhit:
> The rebased code still doesn't feature any checking on whether HFCLK was already started before blocking. Can you please do those changes?

This has now been implemented. The functions now do the following:
- If the counter is 0 but the HFXO is running, this means that the user has mistakenly started the HFXO manually themselves. The assert should warn the user that this a mistake as they should only ever start the HFXO via these functions, otherwise the count will be out of sync.
- The functions now block until the HFXO is started.

hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
```diff
@@ -39,7 +39,11 @@ nrf51_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf51_clock_hfxo_refcnt < 0xff);
     if (nrf51_clock_hfxo_refcnt == 0) {
+        /* Make sure that the HFXO has not previously been started manually */
+        assert(NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
```
mlaz: According to the  [documentation](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Fperipheral_interface.html&cp=3_1_0_14&anchor=concept_twz_bpz_vr), EVENTS_HFCLKSTARTED is an event register thus it signals a state change and not a state, this means its value won't change back to zero when the clock is stopped. Documentation also states that after an event(in this case HFCLK stating) this register's value won't be changed by the peripheral until the firmware sets it to 0. So, to stay safe I believe we should be testing HFCLKSTAT, which is a state register, similarly to how we do it on the original implementation.

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
```diff
@@ -39,7 +39,11 @@ nrf52_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf52_clock_hfxo_refcnt < 0xff);
     if (nrf52_clock_hfxo_refcnt == 0) {
+        /* Make sure that the HFXO has not previously been started manually */
+        assert(NRF_CLOCK->EVENTS_HFCLKSTARTED == 0);
```
mlaz: Please refer to my previous comment nrf51_clock.c.

hw/mcu/nordic/nrf51xxx/src/nrf51_clock.c
mlaz: The link I documentation provided is the nRF52(which is in html format, easier to refer to a specific section) although, on the sections regarding of the subject we are discussing (event registers), the text is exactly the same.

amrbekhit: Ah yes, good catch. This has been corrected now.

adityaxavier:
Hi, any update on this? Wanted to utilise the same for LORA

amrbekhit:
I forgot that this hadn't been merged yet.

@wes3 @mlaz @andrzej-kaczmarek are you guys happy to merge this?

mlaz:
Hi @amrbekhit, yes I'll be happy to merge this but I would feel comfortable doing it after proper testing. I am waiting for a mail delivery for a nordic power profiler kit so I can test it, sorry about this but I am having some problems with fedex.

amrbekhit:
> Hi @amrbekhit, yes I'll be happy to merge this but I would feel comfortable doing it after proper testing. I am waiting for a mail delivery for a nordic power profiler kit so I can test it, sorry about this but I am having some problems with fedex.

Ah OK - thanks for the update. Good luck

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
```diff
@@ -39,7 +39,12 @@ nrf52_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf52_clock_hfxo_refcnt < 0xff);
     if (nrf52_clock_hfxo_refcnt == 0) {
+        /* Make sure that the HFXO has not previously been started manually */
+        assert((NRF_CLOCK->HFCLKSTAT & CLOCK_HFCLKSTAT_STATE_Msk) !=
+            (CLOCK_HFCLKSTAT_STATE_Running << CLOCK_HFCLKSTAT_STATE_Pos));
```
mlaz: When I compile a blinky and try to run it on an nrf52840 board it breaks on this assert.
I just looked at some [nrfx code](https://github.com/NordicSemiconductor/nrfx/blob/master/hal/nrf_clock.h#L505) which verifies whether hfclk is running and I found a few differences. @amrbekhit Can you change and verify this please?

mlaz: also I tried this out and it worked https://github.com/mlaz/mynewt-core/pull/new/nrf52_lp_timer

ccollins476ad:
Closing this as development seems to have stalled.  Please reopen if this is still being worked on.

PR #1784, opened 2019-04-26
[STM32F4]: Add black_vet6 board

kasjer:
Support for cheap board with STM32F407VE (168MHz, 192KB Ram and 512 KBFlash)
and external 2MB of spiflash.

### DIFF
--- a/hw/bsp/black_vet6/boot-stm32f4ve.ld
+++ b/hw/bsp/black_vet6/boot-stm32f4ve.ld
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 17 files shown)


.rat-excludes
```diff
@@ -383,6 +383,11 @@ stm32f1xx_hal_conf.h
 system_stm32f1xx.c
 startup_stm32f103xb.s
 
+# Black VET6 BSP - BSD License
+stm32f4xx_hal_conf.h
+system_stm32f4xx.c
+startup_STM32F40x.s
+
```
utzig: This files are already in `.rat-excludes`. Note that CI does not apply it, just informs user of files without licenses (excluded or not!). Leaving here should not be a problem...

kasjer: Thanks, I did not think clear.
I took those files from some other bsp so it was kind of obvious they should be handle already.
Removed.

PR #1790, opened 2019-04-30
da1469x LED_3

iandmorris:
Added definition for LED_3 (used by at least one ble mesh application). Assigned to unused pin on DA1469x dev board which is brought out to a header. An an external LED will be have to be wired to it as only one LED is available on the board.


andrzej-kaczmarek:
@iandmorris could you please change commit messages to have title of about ~50 characters followed by description (if necessary) wrapped at 72 characters? when you put everything in single line, it looks really bad when looking through history using e.g. `git log` or `gitk`.

PR #1793, opened 2019-05-02
hw: Moving software devices creation code from BSP to pkg_init.

mlaz:
Recently all peripheral device creation code was moved from hw/bsp/board/src/hal_bsp.c to hw/mcu/manuf/variant/src/mcu_periph.c. This prevents replicating code throughout BSPs.
This PR moves soft peripheral(i.e. soft pwm, uart bb) code to hw/mcu.

### DIFF
--- a/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
+++ b/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
@@ -30,25 +30,6 @@
 #include "mcu/nrf52_periph.h"
 #include "bsp/bsp.h"
 #include "defs/sections.h"
-#if MYNEWT_VAL(SOFT_PWM)
-#include "pwm/pwm.h"
-#include "soft_pwm/soft_pwm.h"
-#endif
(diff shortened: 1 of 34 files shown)


andrzej-kaczmarek:
soft PWM and UART BB were not moved to MCU package because they are not MCU peripherals so it's up to BSP to provide those. I did not remove them from any BSP just so existing functionality is retained but tbh I think we do not need soft peripherals in standard BSP-s unless e.g. MCU lacks certain peripheral implemented in hardware. and also for this reason I did not add any soft peripheral to Dialog's BSP.

mlaz:
Completely agree, in fact I suppose we can move this code from the hal_bsp.c files into (optional) sysinit package initialization functions on those drivers' packages while maintaining the mynewt vals which include the packages on the BSP dependencies(not to break compatibility for now).

mlaz:
Now both devices are created on sysinit, UARTBB_0 and SOFT_PWM macros are kept in the BSPs' syscfg so we don't break compatibility until next release. UART bitbang was a added a baud rate restriction to its syscfg.

PR #1794, opened 2019-05-03
Add MCU_PRE_BKPT_CB() syscfg

vrahane:
- It should get called before asm("bkpt") gets called

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_system.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_system.c
@@ -21,6 +21,10 @@
 #include "hal/hal_system.h"
 #include "nrf.h"
 
+#ifdef MYNEWT_VAL_MCU_PRE_BKPT_CB
+extern void MYNEWT_VAL(MCU_PRE_BKPT_CB)(void);
+#endif
+
 /**
  * Function called at startup. Called after BSS and .data initialized but
  * prior to the _start function.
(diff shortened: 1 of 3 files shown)


andrzej-kaczmarek:
syscfg is defined in nrf package, but also used in `arch/cortex_m4` - I don't think this is a good idea.

also I agree with @ccollins476ad that syscfg should not be named after `bkpt` since it has nothing to do with `bktp` instruction in general. a better names imo would be e.g. `MCU_PRE_RESET_CALLBACK` and `OS_ASSERT_CALLBACK`.

and my last concern is that this specifies name of a function so basically this injects piece of source code which I do not really like - a better idea would be to have a predefined callback which application should define if this option is enabled.

vrahane:
I just merged it and your comment popped up. Will address your comments soon.

ccollins476ad:
@andrzej-kaczmarek, if I understand, you are suggesting that the application configure a callback at runtime.  I agree that injecting code via syscfg is a bit messy, but I think it is appropriate for this case.  I believe this PR implements a "safety feature" - the specified code must run every time the device resets.  If the reset occurs before the application can configure the callback, then that is not good.

andrzej-kaczmarek:
@ccollins476ad by "predefined callback" I meant function declaration which is called here and should be implemented by application if this option is enabled. and you have a good point why we should not use an actual callback here instead.

ccollins476ad:
@andrzej-kaczmarek makes sense.

PR #1802, opened 2019-05-08
hw/mcu/dialog: Enable DCDC only when VBAT is OK

kasjer:
Enabling DCDC converter when battery level is low results
in instant reboot even when VBUS is present.
This would make it impossible to charge battery if it drops
below critical level.


hw/mcu/dialog/da1469x/src/system_da1469x.c
```diff
@@ -110,7 +110,13 @@ void SystemInit(void)
     /* Initialize and configure power rails */
     da1469x_prail_initialize();
 #if MYNEWT_VAL(MCU_DCDC_ENABLE)
-    da1469x_prail_dcdc_enable();
+    /*
+     * Enabling DCDC while VBAT is below 2.5 V renders system unstable even
+     * if VBUS is available. Enable DCDC only if VBAT is above minimal value.
+     */
+    if (CRG_TOP->ANA_STATUS_REG & CRG_TOP_ANA_STATUS_REG_COMP_VBAT_HIGH_Msk) {
+        da1469x_prail_dcdc_enable();
```
andrzej-kaczmarek: this also stores dcdc registers in retregs to be restored when going out of deep sleep so probably what we should do is to split this call to be able to configure dcdc and enable/disable depending on VBAT state

PR #1807, opened 2019-05-10
Software breakpoints and asserts

kasjer:
Motivation behind this PR

- When assert fails debugger stops inside __assert_func (not at assert() call) which may
  seem like little problem but when system crashes inside __assert_func finding out
  what happened can be time consuming.
- os_assert_cb() is meant to be called before debugger stops code, in some cases
  __assert_func never reaches that point since if OS_PRINT_ASSERT or log_reboot can stack.
  If os_assert_cb() is meant to be used to prevent time critical hardware protection it should
  be called before.
- Once there is hal_debug_break() definition os_fault could be standard.

### DIFF
--- a/hw/util/break_hook_example/pkg.yml
+++ b/hw/util/break_hook_example/pkg.yml
@@ -1,0 +1,37 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: hw/util/break_hook_example
+pkg.description: >
+    Break hook example.
+    This package defines HAL_BREAK_HOOK syscfg value which informs HAL_DEBUG_BREAK to
+    use user provided callback before execution stops.
+    Example prints some log. Production code hook should not do any printing it should
+    do minimum to leave platform in safe state before CPU stops.
+
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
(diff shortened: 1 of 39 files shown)


utzig:
@kasjer Need to add `#include <syscfg/syscfg.h>` to those modified `cortex_mX.h` headers where you added `MYNEWT_VAL(MCU_DEBUG_IGNORE_BKPT)`.

hw/hal/include/hal/hal_system.h
```diff
@@ -52,6 +54,17 @@ void hal_system_restart(void *img_start) __attribute((noreturn));
...
+#if MYNEWT_VAL(HAL_BREAK_HOOK)
+/* User defined function to be called before code is stopped in debugger */
+void hal_break_hook(void);
+#else
+#define hal_break_hook()
+#endif
+
+#ifndef HAL_DEBUG_BREAK
+#define HAL_DEBUG_BREAK() do { if (hal_debugger_connected()) { hal_break_hook(); hal_debug_break(); }} while (0)
+#endif
```
wes3: Not that there is anything wrong with this but I do not like this "syntax". Very hard to read. Why is this just not a function? Something else that I might have missed in the other commits: is there any expectation that HAL_DEBUG_BREAK is defined with something else?

kasjer: @wes3  similar pattern was use in some files, but I'm not going to force this. I could drop () for once, or change it to static inline.
Macro with ifdef makes is easier to completely drop it if someone wants to like -DHAL_DEBUG_BREAK or such.

kernel/os/src/arch/cortex_m0/os_fault.c
```diff
@@ -121,12 +121,7 @@ __assert_func(const char *file, int line, const char *func, const char *e)
...
 #endif
-    if (hal_debugger_connected()) {
-       /*
-        * If debugger is attached, breakpoint before the trap.
-        */
-       asm("bkpt");
-    }
+
     SCB->ICSR = SCB_ICSR_NMIPENDSET_Msk;
     /* Exception happens right away. Next line not executed. */
```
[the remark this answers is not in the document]
kasjer: I was not part of the change, I will add it so it does not raises any doubts.

hw/util/break_hook_example/pkg.yml
```diff
@@ -0,0 +1,36 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: hw/util/break_hook_example
+pkg.description: >
+    Break hook example.
+    This package defines HAL_BREAK_HOOK syscfg value which informs HAL_DEBUG_BREAK to
+    use user provided before break hook that could modify do some critical hardware
```
vrahane: Was just briefly reading through the PR, did you mean to say `user provided callback` ?

PR #1809, opened 2019-05-11
Shell newtmgr bridge

ccollins476ad:
Implement a new newtmgr command: `shell exec`.  This command allows shell commands to be executed via newtmgr.

This PR requires a sibling change to the newtmgr tool: https://github.com/apache/mynewt-newtmgr/pull/126

### DIFF
--- a/util/streamer/src/streamer_mbuf.c
+++ b/util/streamer/src/streamer_mbuf.c
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "os/mynewt.h"
+#include "streamer/streamer.h"
+
+static int
+streamer_mbuf_write(struct streamer *streamer, const void *src, size_t len)
+{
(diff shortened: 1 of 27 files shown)


util/streamer/src/streamer_mbuf.c
```diff
@@ -0,0 +1,87 @@
...
+    struct streamer_mbuf *sm;
+    int rc;
+
+    if (len > UINT16_MAX) {
+        return SYS_EINVAL;
+    }
+
+    sm = (struct streamer_mbuf *)streamer;
+    rc = os_mbuf_append(sm->om, src, len);
+    if (rc != 0) {
```
wes3: Not sure this matters, but I was wondering what would happen if sm->om was NULL. Looking at os_mbuf_append() it wil return EINVAL in this case. Looks like we will lose that error in this case (gets converted to SYS_ENOMEM). Should this code just return the error code returned by os_mbuf_append()?

util/streamer/src/streamer_mbuf.c
```diff
@@ -0,0 +1,87 @@
...
+
+static const struct streamer_cfg streamer_cfg_mbuf = {
+    .write_cb = streamer_mbuf_write,
+    .vprintf_cb = streamer_mbuf_vprintf,
+};
+
+void
+streamer_mbuf_new(struct streamer_mbuf *sm, struct os_mbuf *om)
+{
+    *sm = (struct streamer_mbuf) {
```
wes3: Does it matter if om is NULL when this is called? I see that msys_new call checks that there will not be a NULL passed here...

wes3:
Reviewed the code. Only a couple of comments to look at.

util/streamer/src/streamer_mbuf.c
ccollins476ad: `os_mbuf_append()` uses a different set of error codes (`OS_E[...]`), so we can't just return that.  I am not strongly opposed to checking if `sm->om` is null here, but I think it would be better to just do this check once when the streamer gets initialized (I see you have made a similar comment below!).

However, I do think we need something that converts from `OS_E` codes to `SYS_E` codes.  Other parts of the code would benefit from such a thing.  So I will add it as a separate commit in this PR.

util/streamer/src/streamer_mbuf.c
ccollins476ad: I have pushed a new commit which modifies this function to check if its arguments are null.

wes3: cool

PR #1811, opened 2019-05-13
kernel/os/arch: Fix HardFault_IRQn usage

andrzej-kaczmarek:
Seems like hardcoded -13 for HardFault_IRQn are just some leftovers in
Cortex-M code, we should use proper symbol here.


utzig:
@andrzej-kaczmarek It seems that STM32Cube for F4 and F7 that we have bundled don't define this symbol name for some reason. At least the F4 we have is very old, maybe the newer versions will have this defined.

utzig:
Latest STM32Cube for F4 also does not have this symbol defined so maybe for CM4 and CM7 this should still use `-13`?

andrzej-kaczmarek:
@utzig @kasjer @wes3 
so since ST is broken with this PR, what about something like this: https://github.com/apache/mynewt-core/pull/1813

PR #1813, opened 2019-05-14
kernel/os/arch: Reset all handlers to os_default_irq_asm

andrzej-kaczmarek:
Instead of leaving some (possibly) unused interrupt handlers set to
default values set by startup code, let's just set all of them to
our os_default_irq_asm.

This allows to avoid magic numbers like "-13" for HardFault_IRQn which
for some reason is not defined for some ST MCUs.

### DIFF
--- a/kernel/os/src/arch/cortex_m0/os_arch_arm.c
+++ b/kernel/os/src/arch/cortex_m0/os_arch_arm.c
@@ -201,21 +201,21 @@ os_arch_os_init(void)
             NVIC->IP[i] = -1;
         }
 
+        /*
+         * Install default interrupt handler for all interrupts except Reset,
+         * which'll print out system state at the time of the interrupt, and
+         * few other regs which should help in trying to figure out what went
+         * wrong.
+         */
+        for (i = -NVIC_USER_IRQ_OFFSET + 2;
+             i < NVIC_NUM_VECTORS - NVIC_USER_IRQ_OFFSET; i++) {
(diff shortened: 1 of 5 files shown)


kernel/os/src/arch/cortex_m0/os_arch_arm.c
```diff
@@ -201,22 +201,22 @@ os_arch_os_init(void)
...
-        NVIC_SetVector(HardFault_IRQn, (uint32_t)os_default_irq_asm);
-        NVIC_SetVector(-13, (uint32_t)os_default_irq_asm); /* Hardfault */
-        for (i = 0; i < NVIC_NUM_VECTORS - NVIC_USER_IRQ_OFFSET; i++) {
+        for (i = -NVIC_USER_IRQ_OFFSET + 2;
+             i < NVIC_NUM_VECTORS - NVIC_USER_IRQ_OFFSET; i++) {
             NVIC_SetVector(i, (uint32_t)os_default_irq_asm);
         }
 
+        /* Install our system interrupt handlers */
+        NVIC_SetVector(SVCall_IRQn, (uint32_t)SVC_Handler);
```
utzig: This is called `SVC_IRQn` for some STM32 families (annoying, I know!)

utzig: Use `SVC_IRQ_NUMBER` here!

PR #1814, opened 2019-05-14
Dev bst

bstbud:
sensors/bma253: to support FIFO streaming and double tap interrupt with low power mode

sensors/bmp388: newly added driver


utzig:
@bstbud This needs to be rebased on `master`, also maybe would make more sense to squash those commits, no?

bstbud:
agree that it's better squashed

bstbud:
close as this PR is not complaint with the guidelines

PR #1818, opened 2019-05-15
sensors/bma253: sensor driver with the following changes

bstbud:
- sensor data read uses burst read to get all axes

  - from register read, change param last_op to 0 for i2cn_master_write

  - dump registers after DT event notification setup

  - reuse get_registers() to simplify get_register()

  - more sophiscated fifo read using fifo frame counter info

  - add a flag to control the bus read write tracing

  - added lpm support with config arbitration for interrupts and data streaming

  - refactor struct int_status to bma253_int_stat_t for optimized efficiency and size (struct from 23 bytes to 4 bytes)

  - added bma253_clear_fifo

  - fix insufficient delay issue after register_write

  - spi support was added

  - added a log level filter and fine tuned some logs

  - changed configuration parameters for data streaming and tapping interrupts

  - code clean up: remove some of the comments

  - unset ev_enabled at unset_notification() api hook

  - clear registered_mask after double tap interrupt fire, so that future notificaiton is possible

  - invalidate data and add necessary delays when needed before reading data so that data get is valid

  - add a delay needed immediately after changing power mode

  - readme about the driver added

### DIFF
--- a/hw/mcu/ambiq/apollo2/src/hal_system_start.c
+++ b/hw/mcu/ambiq/apollo2/src/hal_system_start.c
@@ -30,11 +30,9 @@
 void
 hal_system_start(void *img_start)
 {
-    typedef void jump_fn(void);
-
     uint32_t base0entry;
     uint32_t jump_addr;
-    jump_fn *fn;
+    __attribute__((noreturn)) void (*fn)(void);
 
     /* First word contains initial MSP value. */
     __set_MSP(*(uint32_t *)img_start);
@@ -42,7 +40,7 @@ hal_system_start(void *img_start)
     /* Second word contains address of entry point (Reset_Handler). */
     base0entry = *(uint32_t *)(img_start + 4);
     jump_addr = base0entry;
-    fn = (jump_fn *)jump_addr;
+    fn = (void *)jump_addr;
 
     /* Jump to image. */
     fn();
--- a/hw/mcu/sifive/fe310/src/hal_system_start.c
+++ b/hw/mcu/sifive/fe310/src/hal_system_start.c
@@ -27,9 +27,7 @@
 void
 hal_system_start(void *img_start)
 {
-    typedef void jump_fn(void);
-
-    jump_fn *fn = (jump_fn *)img_start;
+    __attribute__((noreturn)) void (*fn)(void) = img_start;
 
     /* Jump to image. */
     fn();
--- a/hw/mcu/stm/stm32_common/src/hal_system_start.c
+++ b/hw/mcu/stm/stm32_common/src/hal_system_start.c
@@ -31,11 +31,9 @@
 void
 hal_system_start(void *img_start)
 {
-    typedef void jump_fn(void);
-
     uint32_t base0entry;
     uint32_t jump_addr;
-    jump_fn *fn;
+    __attribute__((noreturn)) void (*fn)(void);
 
     /* First word contains initial MSP value. */
     __set_MSP(*(uint32_t *)img_start);
@@ -43,7 +41,7 @@ hal_system_start(void *img_start)
     /* Second word contains address of entry point (Reset_Handler). */
     base0entry = *(uint32_t *)(img_start + 4);
     jump_addr = base0entry;
-    fn = (jump_fn *)jump_addr;
+    fn = (void *)jump_addr;
 
     STM32_HAL_FLASH_REMAP();
 


hw/drivers/sensors/bma253/src/bma253_shell.c
```diff
@@ -256,10 +271,11 @@ bma253_stream_read_cmd(struct bma253 * bma253,
         return EINVAL;
     }
 
-    return bma253_stream_read(bma253,
-                              bma253_stream_read_cb,
-                              &ctx,
-                              0);
+    return bma253_stream_read(&bma253->sensor,   //zg
+            SENSOR_TYPE_ACCELEROMETER,   //zg
```
vrahane: As per the mynewt coding style guide, we should not be using C++ comments. Can you please change it to C style comments. `/*    */`

hw/drivers/sensors/bma253/src/bma253_shell.c
```diff
@@ -256,10 +271,11 @@ bma253_stream_read_cmd(struct bma253 * bma253,
         return EINVAL;
     }
 
-    return bma253_stream_read(bma253,
-                              bma253_stream_read_cb,
-                              &ctx,
-                              0);
+    return bma253_stream_read(&bma253->sensor,   //zg
```
vrahane: Can you please remove `zg` from here ?

hw/drivers/sensors/bma253/syscfg.yml
```diff
@@ -20,7 +20,7 @@
 syscfg.defs:
     BMA253_CLI:
         description: 'Enable shell support for BMA253'
-        value: 0
+        value: 1
```
vrahane: CLI should be disabled by default

hw/drivers/sensors/bma253/syscfg.yml
```diff
@@ -47,7 +47,7 @@ syscfg.defs:
         value: -1
     BMA253_LOG:
         description: 'Enable BMA253 logging'
-        value: 0
+        value: 1
```
vrahane: Disable logging by default, should be enabled only if needed

hw/sensor/creator/src/sensor_creator.c
```diff
@@ -299,17 +299,30 @@ static struct sensor_itf i2c_0_itf_ms40 = {
 };
 #endif
 
-
 #if MYNEWT_VAL(I2C_0) && MYNEWT_VAL(BMA253_OFB)
-static struct sensor_itf i2c_0_itf_lis = {
```
vrahane: Why is there a need to change this to SPI if the driver is working over I2C. We want to keep it as I2C.

hw/sensor/creator/src/sensor_creator.c
```diff
@@ -299,17 +299,30 @@ static struct sensor_itf i2c_0_itf_ms40 = {
...
-            MYNEWT_VAL(BMA2XX_INT_CFG_ACTIVE)}
+        { 12, MYNEWT_VAL(BMA253_INT_PIN_DEVICE),
+            MYNEWT_VAL(BMA253_INT_CFG_ACTIVE)},
+        { 24, MYNEWT_VAL(BMA253_INT2_PIN_DEVICE),
+            MYNEWT_VAL(BMA253_INT_CFG_ACTIVE)}
+    },
+};
+#endif
+
+#if MYNEWT_VAL(SPI_0_MASTER) && MYNEWT_VAL(BMA253_OFB)
```
vrahane: Do not defined both interfaces here as that is going to cause confusion, specially with the board has both I2C and SPI enabled, this is going to break. So, just select one, either SPI or I2C. I think the best way is to keep it as I2C.

hw/sensor/creator/src/sensor_creator.c
```diff
@@ -796,20 +809,23 @@ config_bma253_sensor(void)
...
     cfg.offset_z_g = 0.0;
-    cfg.power_mode = BMA253_POWER_MODE_NORMAL;
-    cfg.sleep_duration = BMA253_SLEEP_DURATION_0_5_MS;
+    /* options: BMA253_POWER_MODE_SUSPEND, BMA253_POWER_MODE_NORMAL, BMA253_POWER_MODE_LPM_1, */
+    cfg.power_mode = BMA253_POWER_MODE_SUSPEND;
+    cfg.sleep_duration = BMA253_SLEEP_DURATION_10_MS;
     cfg.sensor_mask = SENSOR_TYPE_ACCELEROMETER;
-    cfg.read_mode = BMA253_READ_M_POLL;
+    /* options: BMA253_READ_M_STREAM, BMA253_READ_M_POLL */
+    cfg.read_mode = BMA253_READ_M_STREAM;
```
vrahane: We generally keep the sensors in the most basic operating mode which is polling based. We do not want to be detecting taps by default.

hw/drivers/sensors/bma253/src/bma253.c
```diff
@@ -2714,65 +2987,114 @@ bma253_set_fifo_cfg(const struct bma253 * bma253,
...
+    enum bma253_g_range g_range = bma253->cfg.g_range;
+
+    struct sensor *sensor = &bma253->sensor;
+
+    rc = get_accel_scale(bma253, g_range, &accel_scale);
+    if (0 != rc) {
+        return rc;
     }
 
+    //zg: to improve
```
vrahane: Can you please remove it.

hw/drivers/sensors/bma253/src/bma253.c
```diff
@@ -2714,65 +2987,114 @@ bma253_set_fifo_cfg(const struct bma253 * bma253,
...
+            //force the frame counter to be the max
+            ff_frm_cnt = SPEC_MAX_FIFO_DEPTH;
+        }
+    } else {
+        return rc;
+    }
+
+    if (ff_frm_cnt > 0) {
+    } else {
+        //exit mildly
```
vrahane: C++ style comments to C style comments.

hw/drivers/sensors/bma253/src/bma253.c
```diff
@@ -2714,65 +2987,114 @@ bma253_set_fifo_cfg(const struct bma253 * bma253,
...
     return set_register((struct bma253 *)bma253, REG_ADDR_FIFO_CONFIG_1, data);
 }
 
-int
-bma253_get_fifo(const struct bma253 * bma253,
-                enum bma253_g_range g_range,
-                enum fifo_data fifo_data,
-                struct accel_data * accel_data)
+
+struct sensor_data_subscriber_info {
```
vrahane: This structure is not needed. There is a reason we have:
```
 644 struct sensor_read_ctx {
 645     sensor_data_func_t user_func;
 646     void *user_arg;
 647 };  
```
We want to be able to abstract app from the driver. Please let it be that way.

hw/drivers/sensors/bma253/src/bma253_shell.c
nanguoguangzi: will do, thanks for comments

hw/drivers/sensors/bma253/syscfg.yml
nanguoguangzi: agree

hw/drivers/sensors/bma253/syscfg.yml
nanguoguangzi: will do

hw/drivers/sensors/bma253/include/bma253/bma253.h
```diff
@@ -197,8 +201,29 @@ struct bma253_private_driver_data {
     uint8_t int_num;
     uint8_t int_route;
     uint8_t int_ref_cnt;
+
+
+    //zg: check for potential improvement
+    uint8_t fifo_buf[31 * 6];
```
vrahane: There is no need for the driver to define the FIFO buffer. The app can define it and the driver can just have a ptr to it and fill it up. `hw/drivers/sensors/lis2dh12/src/lis2dh12.c` has a `lis2dh12_stream_read` function which you can use as reference. It's quite necessary for it to remain that way or we might have issues with completing abstracting the application from the driver.

hw/drivers/sensors/bma253/src/bma253.c
```diff
@@ -30,30 +30,56 @@
...
 #endif
 
+
+#define BMA253_LOG_LEVEL_THRESH LOG_LEVEL_WARN
+
 #if MYNEWT_VAL(BMA253_LOG)
 
 #define BMA253_LOG(lvl_, ...) \
-    MODLOG_ ## lvl_(MYNEWT_VAL(BMA253_LOG_MODULE), __VA_ARGS__)
+    if ((LOG_LEVEL_ ## lvl_) >= BMA253_LOG_LEVEL_THRESH) { MODLOG_ ## lvl_(MYNEWT_VAL(BMA253_LOG_MODULE), __VA_ARGS__); }
```
vrahane: Please do not change this. There is a reason it is that way, we need to be able to have a different log level than `WARN`

hw/drivers/sensors/bma253/src/bma253.c
```diff
@@ -795,10 +1004,34 @@ bma253_set_power_settings(const struct bma253 * bma253,
...
+        BMA253_DRV_CHECK_RC(rc);
+    }
+
     rc = set_register((struct bma253 *)bma253, REG_ADDR_PMU_LPW, data[0]);
     if (rc != 0) {
         return rc;
     }
+    ((struct bma253 *)bma253)->power = power_settings->power_mode;
+
+    //immediately after changing power mode, a delay is needed
```
vrahane: Please convert all C++ comments to C comments

hw/drivers/sensors/bma253/src/bma253.c
vrahane: Can you please check how we handle FIFOs in `hw/drivers/sensors/lis2dh12/src/lis2dh12.c`

hw/drivers/sensors/bma253/src/bma253.c
```diff
@@ -2942,14 +3266,16 @@ reset_and_recfg(struct bma253 * bma253)
         return rc;
     }
 
-    fifo_cfg.fifo_mode = FIFO_MODE_BYPASS;
+    fifo_cfg.fifo_mode = FIFO_MODE_STREAM;
```
vrahane: Do not change this to `STREAM`, the basic should always be polling mode and hence the mode should be set to `BYPASS`

hw/sensor/creator/src/sensor_creator.c
nanguoguangzi: will do

hw/drivers/sensors/bma253/src/bma253.c
nanguoguangzi: ok

hw/drivers/sensors/bma253/src/bma253.c
nanguoguangzi: agree, we did not notice there was a similar structure needed, thanks for the info, and we will reuse sensor_read_ctx.

hw/drivers/sensors/bma253/include/bma253/bma253.h
nanguoguangzi: Hello @vrahane 

Thanks for the info, actually, this buf is not for filling up users's data (structured data), this is the buffer to hold the raw data (binary and non-structured) from the FIFO and to take care of the worst sense scenario (when fifo is full). 

Otherwise, all other reviews are agreed upon and fixed/changed correspondingly.

hw/drivers/sensors/bma253/src/bma253.c
nanguoguangzi: changed

hw/drivers/sensors/bma253/src/bma253.c
nanguoguangzi: Hello
this is a small optimization we tried to make over it.
in polling mode, it does not hurt to set the fifo mode to STREAM, because the data is polled from the data registers and it's not affected by the fifo settings.
in stream mode, we need the fifo_mode to be set to FIFO_MODE_STREAM, and if it's already setup in this way (in reset_and_recfg), we don't need to setup it again at time of stream_read, this saves a bit of time for the driver.

But I agree that this is not huge optimization, so if you think it's necessary to keep the original setting, we can do that.

bstbud:
Hello @vrahane 
We have closed this PR and changed the driver according to the comments above, and a new PR (#1820) with the changes incorporated has been open and this will be closed.

vrahane:
@bstbud So, you do not need to close the PRs and open new ones, you can just address the current one. It's fine for now but for something in the future.

PR #1820, opened 2019-05-15
sensors/bma253: sensor driver with the following changes

bstbud:
- sensor data read uses burst read to get all axes
  - from register read, change param last_op to 0 for i2cn_master_write
  - dump registers after DT event notification setup
  - reuse get_registers() to simplify get_register()
  - more sophiscated fifo read using fifo frame counter info
  - add a flag to control the bus read write tracing
  - added lpm support with config arbitration for interrupts and data streaming
  - refactor struct int_status to bma253_int_stat_t for optimized efficiency and size (struct from 23 bytes to 4 bytes)
  - added bma253_clear_fifo
  - fix insufficient delay issue after register_write
  - spi support was added
  - added a log level filter and fine tuned some logs
  - changed configuration parameters for data streaming and tapping interrupts
  - code clean up: remove some of the comments
  - unset ev_enabled at unset_notification() api hook
  - clear registered_mask after double tap interrupt fire, so that future notificaiton is possible
  - invalidate data and add necessary delays when needed before reading data so that data get is valid
  - add a delay needed immediately after changing power mode
  - readme about the driver added
  - make comments in code complaint with coding style, remove c++ comments
  - use struct sensor_read_ctx to replace struct sensor_data_subscriber_info

### DIFF
--- a/hw/drivers/sensors/bma253/README.md
+++ b/hw/drivers/sensors/bma253/README.md
@@ -1,0 +1,64 @@
+<!--
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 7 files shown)


vrahane:
@bstbud Can you please address the PR based on the review that @kasjer did and remove the API which wraps the cputime delay API.

Also, on line `3263` `    fifo_cfg.fifo_mode = FIFO_MODE_STREAM;` should be `BYPASS` since we want it to default to POLLING mode by default, so the FIFOs are bypassed.

nanguoguangzi:
all the changes per the reviews from @vrahane  and @kasjer are incorporated, including the ones mentioned above (FIFO_MODE_BYPASS), and also tested. 

thanks.

PR #1822, opened 2019-05-15
Check for existing stats instance before inserting in the global stats registry

benmccrea:
Adds a check for a statistics instance in `g_stats_registry` with the same address as the one being added (previously duplicates were being rejected only if the name is the same). 

Multiple sensors may use the same driver but some drivers only implement one instance of statistics. This can lead to a loop in the `g_stats_registry` linked list, because currently `stats_register_internal()` assumes that statistics with different names will use unique statistics. 

The problem arises when a statistics instance is registered twice under different names. The actual instance is the same, so the second time it is registered the linked list insertion will cause `s_next` of the entry to point to itself.

### DIFF
--- a/sys/stats/full/src/stats.c
+++ b/sys/stats/full/src/stats.c
@@ -115,7 +115,7 @@ stats_register_internal(const char *name, struct stats_hdr *shdr)
      * is already registered.
      */
     STAILQ_FOREACH(cur, &g_stats_registry, s_next) {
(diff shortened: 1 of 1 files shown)


vrahane:
I think this fix is needed since a loop in the linked list needs to be detected. But the real problem here is now only one stat would be enabled. So, it would be better to define stats in the driver structure instead of the source file as a global. That way there are two separate globals for stats since there would be two different driver structures.

PR #1824, opened 2019-05-16
da1469x-dk-pro; 'mon reset' does not appear to reset the MCU

mkiiskila:
...completely at the end of download.


gfreese-juul:
I've verified this workaround works on actual hardware.

PR #1831, opened 2019-05-20
Bump go requirement to 1.11

utzig:
The usage of RSA that allows for keys of any sizes requires go 1.11 to work.


utzig:
This might have to applied to nimble as well; and any others?

PR #1838, opened 2019-05-23
[Feature]: Main Task OS Sanity Check

brolan-juul:
Added ability to include `os_sanity` check with `main task`.
Developers will have to create their os_sanity check-in in their application.

### DIFF
--- a/kernel/os/src/os.c
+++ b/kernel/os/src/os.c
@@ -235,8 +235,7 @@ os_init(int (*main_fn)(int argc, char **arg))
     if (main_fn) {
         err = os_task_init(&g_os_main_task, "main", os_main, main_fn,
                    OS_MAIN_TASK_PRIO,
-                   (OS_MAIN_TASK_TIMER_TICKS == 0) ? \
-                        OS_WAIT_FOREVER : OS_MAIN_TASK_TIMER_TICKS,
+                   (OS_MAIN_TASK_TIMER_TICKS == 0) ? OS_WAIT_FOREVER : OS_MAIN_TASK_TIMER_TICKS,
                    g_os_main_stack, OS_STACK_ALIGN(OS_MAIN_STACK_SIZE));
         assert(err == 0);
     }
--- a/kernel/os/src/os.c
+++ b/kernel/os/src/os.c
@@ -45,8 +45,14 @@ OS_TASK_STACK_DEFINE(g_idle_task_stack, OS_IDLE_STACK_SIZE);
 
 uint32_t g_os_idle_ctr;
 
-static struct os_task os_main_task;
-OS_TASK_STACK_DEFINE(os_main_stack, OS_MAIN_STACK_SIZE);
+struct os_task g_os_main_task;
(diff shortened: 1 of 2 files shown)


kernel/os/src/os.c
```diff
@@ -45,8 +45,14 @@ OS_TASK_STACK_DEFINE(g_idle_task_stack, OS_IDLE_STACK_SIZE);
...
 
-static struct os_task os_main_task;
-OS_TASK_STACK_DEFINE(os_main_stack, OS_MAIN_STACK_SIZE);
+struct os_task g_os_main_task;
+OS_TASK_STACK_DEFINE(g_os_main_stack, OS_MAIN_STACK_SIZE);
+
+/*
+ * Double the interval timer to allow proper timer check-in.
+ */
+#define OS_MAIN_TASK_TIMER_TICKS \
```
andrzej-kaczmarek: this should be called smth like `OS_MAIN_TASK_SANITY_ITVL_TICKS`

kernel/os/src/os.c
```diff
@@ -45,8 +45,14 @@ OS_TASK_STACK_DEFINE(g_idle_task_stack, OS_IDLE_STACK_SIZE);
 
 uint32_t g_os_idle_ctr;
 
-static struct os_task os_main_task;
-OS_TASK_STACK_DEFINE(os_main_stack, OS_MAIN_STACK_SIZE);
+struct os_task g_os_main_task;
+OS_TASK_STACK_DEFINE(g_os_main_stack, OS_MAIN_STACK_SIZE);
+
+/*
+ * Double the interval timer to allow proper timer check-in.
```
andrzej-kaczmarek: I looked quickly through os_sanity code and honestly have no idea why need to double the interval

kernel/os/src/os.c
brolan-juul: Agreed. That would be a more descriptive name.

kernel/os/src/os.c
brolan-juul: Removing then need for this. Thanks!

kernel/os/src/os.c
```diff
@@ -235,8 +232,8 @@ os_init(int (*main_fn)(int argc, char **arg))
     if (main_fn) {
         err = os_task_init(&g_os_main_task, "main", os_main, main_fn,
                    OS_MAIN_TASK_PRIO,
-                   (OS_MAIN_TASK_TIMER_TICKS == 0) ? \
-                        OS_WAIT_FOREVER : OS_MAIN_TASK_TIMER_TICKS,
+                   (OS_MAIN_TASK_SANITY_ITVL_TICKS == 0) ? \
+                        OS_WAIT_FOREVER : OS_MAIN_TASK_SANITY_ITVL_TICKS,
```
wes3: No need for \ here btw. Should work without it. Looks fine otherwise.

brolan-juul: I spaced it to two lines to keep it under 80 characters per line tho I'd have to admit, I haven't read the style guide for this project. I'll remove the `\`.

PR #1839, opened 2019-05-23
Fix/i2c scan

pshah111:
1. Fixed a bug in i2c_check_tx_is_empty function in hal_i2c.c file.
2. The i2c_scan command fails to detect a sensor on i2c lines when it uses i2c read function. Additionally, i2c_scan will lock up the i2c bus. 
Fix: updated the probe function to write to i2c device instead of reading from the i2c device.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_i2c.c
+++ b/hw/mcu/dialog/da1469x/src/hal_i2c.c
@@ -238,7 +238,7 @@ i2c_check_tx_is_not_full(const struct da1469x_hal_i2c *i2c)
 static bool
 i2c_check_tx_is_empty(const struct da1469x_hal_i2c *i2c)
 {
-    return i2c->regs->I2C_STATUS_REG & I2C_I2C_STATUS_REG_TFNF_Msk;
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
there are few problems with this PR:
it pushes changes we already have (reset workaround), thus there's conflict
it does not conform to our coding style
it does not explain the changes

hw/mcu/dialog/da1469x/src/hal_i2c.c
```diff
@@ -442,12 +442,13 @@ hal_i2c_master_read(uint8_t i2c_num, struct hal_i2c_master_data *pdata,
...
 
-    rx.address = address;
-    rx.buffer = &buf;
-    rx.len = 1;
+    tx.address = address;
+    tx.buffer = &buf;
+    tx.len = 1;
 
-    return hal_i2c_master_read(i2c_num, &rx, timo, 1);
+    /* Reading before writing causes i2c bus to hang up sometimes. */
```
benmccrea: I think this comment should explain more clearly why hal_i2c_master_write() is used instead of a read. Something like,  'Using a write because a read results in ... <whatever the issue with read was>'

PR #1844, opened 2019-05-28
sensors: mpu6050 add flexibility

kasjer:
Allow sensor to be placed at non I2C_0 interface.

### DIFF
--- a/hw/sensor/creator/src/sensor_creator.c
+++ b/hw/sensor/creator/src/sensor_creator.c
@@ -258,10 +258,10 @@ static struct sensor_itf i2c_0_itf_lsm = {
 };
 #endif
 
-#if MYNEWT_VAL(I2C_0) && MYNEWT_VAL(MPU6050_OFB)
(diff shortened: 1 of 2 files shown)


hw/sensor/creator/src/sensor_creator.c
```diff
@@ -258,10 +258,10 @@ static struct sensor_itf i2c_0_itf_lsm = {
 };
 #endif
 
-#if MYNEWT_VAL(I2C_0) && MYNEWT_VAL(MPU6050_OFB)
```
vrahane: We still need to check for `MYNEWT_VAL(I2C_0)` or whichever I2C we use because of the way we define the BSPs,  we do not have a common I2C or SPI setting which suggests that I2C or SPI is enabled and `.si_type = SENSOR_ITF_I2C` has the interface defined accordingly. Although, I think this is going to change when @andrzej-kaczmarek opens a PR for the bus driver changes.

PR #1854, opened 2019-06-01
nrfx: add 380kbps to nrf52840_bitfields.c for twi and twim

ccollins476ad:
nrfx used to be customized with this change - support for TWI/TWIM 380kbps.  This change was lost when nrfx was recently upgraded.

This PR adds the change back to nrfx.

We should find a better long term solution, but for now, let's just get the change back in.


mlaz:
LGTM

sjanc:
any reasons to not upstream this?

PR #1855, opened 2019-06-01
Trying to add NRF52811 support

alvarop:
Updated TWI definitions to use new struct.
Added various NRF52811 specific defines (and sometimes NRF52810!)

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/syscfg.yml
+++ b/hw/mcu/nordic/nrf52xxx/syscfg.yml
@@ -24,6 +24,8 @@ syscfg.defs:
         restrictions:
             - $notnull
         choices:
+            - nRF52810
+            - nRF52811
             - nRF52832
             - nRF52840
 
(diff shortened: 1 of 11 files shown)


hw/mcu/nordic/nrf52xxx/syscfg.yml
```diff
@@ -449,6 +451,10 @@ syscfg.defs:
             - "!XTAL_RC"
         deprecated: 1
 
+syscfg.vals.MCU_NRF52810:
+    MCU_TARGET: nRF52810
+syscfg.vals.MCU_NRF52811:
+    MCU_TARGET: nRF52811
```
andrzej-kaczmarek: vals below are for backwards compatibility with already existing syscfg definitions so this has no effect since there are no `MCU_NRF52810`/`MCU_NRF52811` defined

mlaz: Well that one didn't catch my eye. Fixed on https://github.com/apache/mynewt-core/pull/1856

PR #1862, opened 2019-06-03
Fix warning on pointer to packed struct member

utzig:
When passing a pointer to a packed struct member, gcc-9 warns about its inability to do proper alignment. This fixes the issue by using pointer to variables and copying to the struct members.

See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51628


utzig:
This problem might exist on many other locations as well, but here I am fixing for lps33hw because it's a sensor I am using at the moment...

PR #1864, opened 2019-06-04
mgmt/imgmgr: Log image management events

ccollins476ad:
This PR adds logging for four image management events.

It will be a nuisance to change the format of these log entries in the future, so we should try to get them right the first time.  Please don't hold back in critiquing the log entry formats!

All events are CBOR-encoded:
```
upstart:
    When: upon receiving an upload request with an offset of 0.
    Structure:
    {
        "ev": "upstart",
        "rc": <mgmt-error-code (int)>
    }

updone:
    When: upon receiving an upload request containing the final chunk
          of an image OR a failed upload request with a non-zero offset.
    Structure:
    {
        "ev": "updone",
        "rc": <mgmt-error-code (int)>
        "hs": <image-hash (byte-string)> (only present on success)
    }

pend:
    When: upon receiving a non-permanent `set-pending` request.
    Structure:
    {
        "ev": "pend",
        "rc": <mgmt-error-code (int)>,
        "hs": <image-hash (byte-string)>
    }

conf:
    When: upon receiving a `confirm` request OR a permanent
          `set-pending` request.
    Structure:
    {
        "ev": "conf",
        "rc": <mgmt-error-code (int)>,
        "hs": <image-hash (byte-string)> (only present for `set-pending`)
    }
```

Here is a log snippet after an upload-test-reset-confirm cycle:
```
     44518     1559419790601524us |     CUSTOM (176)         INFO (1)   cbor {"ev":"upstart","rc":0}
     46008     1559419842390600us |     CUSTOM (176)         INFO (1)   cbor {"ev":"updone","hs":"ZEiw0Ttss5oSc/FUldJB1mFiw/wed+Tt1MBr8uo9aK8=","rc":0}
     46085     1559419899070308us |     CUSTOM (176)         INFO (1)   cbor {"ev":"pend","hs":"ZEiw0Ttss5oSc/FUldJB1mFiw/wed+Tt1MBr8uo9aK8=","rc":0}
     46109     1559419964396732us |     CUSTOM (176)         INFO (1)   cbor {"ev":"conf","rc":0}
```

Image management logging is disabled by default.  To enable it:

1. Override `IMGMGR_LOG_LVL` with a value <= 1 (`LOG_LEVEL_INFO`)
2. Map the `IMGMGR_LOG_MOD` module to a log, e.g.,
```
    rc = modlog_register(MYNEWT_VAL(IMGMGR_LOG_MOD), &my_log,
                         LOG_LEVEL_INFO, NULL);
    assert(rc == 0);
```

### DIFF
--- a/mgmt/imgmgr/src/imgmgr.c
+++ b/mgmt/imgmgr/src/imgmgr.c
@@ -743,6 +743,45 @@ imgr_upload_good_rsp(struct mgmt_cbuf *cb)
     return 0;
 }
 
+/**
+ * Logs an upload request if necessary.
+ *
+ * @param is_first              Whether the request includes the first chunk of
+ *                                  the image.
+ * @param is_last               Whether the request includes the last chunk of
+ *                                  the image.
+ * @param status                The result of processing the upload request
+ *                                  (MGMT_ERR code).
+ *
+ * @return                      0 on success; nonzero on failure.
+ */
+static int
+imgr_upload_log(bool is_first, bool is_last, int status)
+{
+    uint8_t hash[IMGMGR_HASH_LEN];
+    const uint8_t *hashp;
+    int rc;
+
+    if (is_first) {
+        return imgmgr_log_upload_start(status);
+    }
+
+    if (is_last || status != 0) {
+        /* Log the image hash if we know it. */
+        rc = imgr_read_info(1, NULL, hash, NULL);
+        if (rc != 0) {
+            hashp = NULL;
+        } else {
+            hashp = hash;
+        }
+
+        return imgmgr_log_upload_done(status, hashp);
+    }
+
+    /* Nothing to log. */
+    return 0;
+}
+
 static int
 imgr_upload(struct mgmt_cbuf *cb)
 {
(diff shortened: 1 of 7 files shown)


mkiiskila:
I prefer shorter key names, if possible. How about replacing "status" with "rc" (return code)?

ccollins476ad:
Thanks @mkiiskila, that is a good point.  I have shortened the strings and updated the PR description with the new spec.

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -743,6 +743,45 @@ imgr_upload_good_rsp(struct mgmt_cbuf *cb)
...
+    uint8_t hash[IMGMGR_HASH_LEN];
+    const uint8_t *hashp;
+    int rc;
+
+    if (is_first) {
+        return imgmgr_log_upload_start(status);
+    }
+
+    /* Log the image hash if we know it. */
+    rc = imgr_read_info(1, NULL, hash, NULL);
```
mkiiskila: I would recommend moving this code behind the if-statement where you decide that this is going to log something. imgr_read_info() does quite a bit of flash activity, so it is a bit heavy to run for every data chunk.

ccollins476ad: Thanks @mkiiskila, that is a good point.  I have changed the function as follows:
```
static int
imgr_upload_log(bool is_first, bool is_last, int status)
{
    uint8_t hash[IMGMGR_HASH_LEN];
    const uint8_t *hashp;
    int rc;

    if (is_first) {
        return imgmgr_log_upload_start(status);
    }

    if (is_last || status != 0) {
        /* Log the image hash if we know it. */
        rc = imgr_read_info(1, NULL, hash, NULL);
        if (rc != 0) {
            hashp = NULL;
        } else {
            hashp = hash;
        }

        return imgmgr_log_upload_done(status, hashp);
    }

    /* Nothing to log. */
    return 0;
}
```

PR #1867, opened 2019-06-06
Add back K380 speed to nrf52xxx hal_i2c

benmccrea:
Adding back the TWI_FREQUENCY_FREQUENCY_K380 clock choice which was accidentally removed during the recent upgrade to nrfx 1.7.1.


sterlinghughes:
is this a patch to NRFX?  If so, can we upstream with Nordic team to avoid in future? @carlescufi

hw/mcu/nordic/src/ext/nrfx/mdk/nrf52_bitfields.h
```diff
@@ -11169,6 +11169,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TWI_FREQUENCY_FREQUENCY_Msk (0xFFFFFFFFUL << TWI_FREQUENCY_FREQUENCY_Pos) /*!< Bit mask of FREQUENCY field. */
 #define TWI_FREQUENCY_FREQUENCY_K100 (0x01980000UL) /*!< 100 kbps */
 #define TWI_FREQUENCY_FREQUENCY_K250 (0x04000000UL) /*!< 250 kbps */
+#define TWI_FREQUENCY_FREQUENCY_K380 (0x06147ae9UL) /*!< 380 kbps */
```
carlescufi: @anangl should we add this in nrfx?

anangl: No. Such value is not mentioned in the Product Specification, so it's not officially supported. Thus, we can't add it to nrfx (MDK actually).

benmccrea: No, I don't think it's worth the effort. It's as feature, not a bug. Can we just merge this? Next time a NRFX update takes place we need to do better review.

sterlinghughes: I think we should add this to i2c_nrf52_twim.c, rather than maintain a custom patched NRFX library.  I don't see usage of this anywhere but in that driver, is that correct?

andrzej-kaczmarek: we can just use hardcoded (+comment) value in both hal_i2c and i2c_nrf52_twim instead of using symbolic name

PR #1874, opened 2019-06-13
bsp/dialog_da1469x-dk-pro; fix download for bootloader on MacOS.

mkiiskila:
Not sure what happened, but LANG is not doing this for me. LC_CTYPE makes it do the right thing.
I could've sworn that this was working before...


mkiiskila:
I was wondering if I should leave that one in as well. Should do no harm.

PR #1876, opened 2019-06-13
[RFC] Add cryptographic HASH driver

utzig:
This adds initial support for the HASH HW in K64F and STM32.

### DIFF
--- a/hw/drivers/hash/hash_k64f/include/hash_context.h
+++ b/hw/drivers/hash/hash_k64f/include/hash_context.h
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef __HASH_CONTEXT_H__
+#define __HASH_CONTEXT_H__
+
+#include "hash/hash.h"
+
+/*
+ * SHA-224 is not supported supported on K64F
+ */
+struct hash_sha224_context {
+    void *dev;
+};
+
+/*
+ * SHA-256 is implemented.
+ */
+struct hash_sha256_context {
+    /* Store device pointer */
+    void *dev;
+    /* Total length of the hashed data */
+    uint32_t len;
+    /* Current context output of the hash */
+    unsigned int output[SHA256_DIGEST_LEN >> 2];
+    /* Last block added, used to pad on finish() */
+    uint8_t pad[SHA256_BLOCK_LEN];
+    /* Amount added to pad, used on finish() */
+    uint8_t remain;
+};
+
+#endif /* __HASH_CONTEXT_H__ */
--- a/hw/drivers/hash/hash_stm32/src/hash_stm32.c
+++ b/hw/drivers/hash/hash_stm32/src/hash_stm32.c
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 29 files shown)


utzig:
TIL: it's interesting that the driver interface table should always be the minimum amount of functions required to achieve the functionality that a user may need, because the linker cannot garbage collect the non-used functions from the table.

In this proposal I have one interface function called `process` that does the same as calls to `start + update + finish`. Although I have not used `process` it still uses 200 bytes (about the same as the other 3 functions together). Since this can be trivially done in the upper layer, I will remove later.

hw/drivers/hash/hash_k64f/include/hash_context.h
```diff
@@ -0,0 +1,59 @@
...
+/*
+ * The following HASH algos are not implemented or not
+ * supported on K64F (SHA-224 is not supported)
+ */
+
+struct hash_md5_context {
+    void *dev;
+};
+
+struct hash_sha1_context {
```
mkiiskila: Maybe just drop MD5 and SHA1? I don't think ppl are using these for anything anymore.

hw/drivers/hash/hash_stm32/syscfg.yml
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
```
mkiiskila: Maybe no need to add this file?

hw/drivers/hash/hash_stm32/src/hash_stm32.c
```diff
@@ -0,0 +1,190 @@
...
+
+static int
+stm32_hash_dev_open(struct os_dev *dev, uint32_t wait, void *arg)
+{
+    struct hash_dev *hash;
+
+    hash = (struct hash_dev *)dev;
+    assert(hash);
+
+    /* XXX Not reentrant? */
```
mkiiskila: For use with OS, we'll need this to be reentrant. Even if this was not made thread-safe right from start, it should be added soon after.

The use case most benefitting from this, bootloader, has only one thread, so this code is useful even as is.

hw/drivers/hash/include/hash/hash.h
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @var hash_interface::has_support
+ * has_support is used to inquire about which algos are natively
+ * supported
+ */
+struct hash_interface {
+    hash_start_op_func_t start;
+    hash_update_op_func_t update;
+    hash_finish_op_func_t finish;
+    hash_support_func_t has_support;
```
mkiiskila: I would make the capabilities field a uint within struct hash_dev instead. Have an inline function to inspect it. So we ever run out of bits for capabilities, it can turned into a function.
I think we'd end up with less text space use that way.

hw/drivers/hash/include/hash/sha256_alt.h
```diff
@@ -0,0 +1,64 @@
...
+#if MYNEWT_VAL(MBEDTLS_SHA256_ALT)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <hash/hash.h>
+
+typedef struct mbedtls_sha256_context
+{
```
mkiiskila: coding style; open parentheses should be in same line as struct name.

hw/drivers/hash/src/hash.c
```diff
@@ -0,0 +1,144 @@
...
+}
+
+int
+hash_custom_start(struct hash_dev *hash, void *ctx, uint16_t algo)
+{
+    if (!hash_has_support(hash, algo)) {
+        return -1;
+    }
+
+    if (hash->interface.start == NULL) {
```
mkiiskila: Does it make sense to have this check done at runtime? I think I'd rather have the code jump to NULL pointer here and crash, rather than pay the performance penalty :)

utzig: done

hw/drivers/hash/include/hash/sha256_alt.h
utzig: done

hw/drivers/hash/include/hash/hash.h
utzig: done

hw/drivers/hash/hash_stm32/syscfg.yml
utzig: removed

hw/drivers/hash/hash_stm32/src/hash_stm32.c
utzig: I added a concurrency test to the app, it still requires that calls are done in the correct order (start -> update -> finish) which might be something to fix, and it only opens the driver once (`os_dev_open`); maybe opening the device multiple times should also be checked later...

hw/drivers/hash/hash_k64f/include/hash_context.h
utzig: done

hw/drivers/hash/hash_stm32/src/hash_stm32.c
mkiiskila: I think 'start->update->finish' requirement is ok.  Use of mutex is good here. I don't think any of the SHA users currently expect the calls to return with an error, making them wait instead is ok. Also having that assert() for the failure paths is good.

PR #1881, opened 2019-06-17
hal_i2c: Adding back in TWI 380kbps

JustineKH:
380kbps was removed in commit d92e0127b8761cae2048c9eb8eb2cd5223bf9e84. 
The bit fields for 380kbps (TWI and TWIM) got added back in in commit 53a34eb4344f47685e469f69f1a5e4ff2382ced2, but not the check in hal_i2c_init() and hal_i2c_config() .
The TWI bitfield got removed at some point after that.
Adding back in TWI bitfield, and hal_i2c check for 380kbps.


sterlinghughes:
Hi justine, see #1867 -- can we please keep the changes isolated to hal_i2c.c, so we can avoid having a modified nrfx?

PR #1884, opened 2019-06-18
stm32/bsp: Fix build for sensorhub

kasjer:
- AHBPrescTable had different type then in includes.
- APBPrescTable was missing.
- Flash sector defintion was not updated as other BSP were moved to
  have common code base.

### DIFF
--- a/hw/bsp/sensorhub/src/hal_bsp.c
+++ b/hw/bsp/sensorhub/src/hal_bsp.c
@@ -31,6 +31,26 @@
 #include <mcu/stm32f4_bsp.h>
 #include "bsp/bsp.h"
 
+const uint32_t stm32_flash_sectors[] = {
+    0x08000000,     /* 16kB */
+    0x08004000,     /* 16kB */
+    0x08008000,     /* 16kB */
+    0x0800c000,     /* 16kB */
+    0x08010000,     /* 64kB */
+    0x08020000,     /* 128kB */
+    0x08040000,     /* 128kB */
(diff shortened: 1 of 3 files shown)


hw/bsp/sensorhub/src/hal_bsp.c
```diff
@@ -69,7 +90,7 @@ hal_bsp_flash_dev(uint8_t id)
     if (id != 0) {
         return NULL;
     }
-    return &stm32f4_flash_dev;
```
utzig: I did not expect this to still exist, does this BSP build?

kasjer: blinky builds with this BSP and changes in this PR.
Not tested though.

utzig: A grep returns only `extern const struct hal_flash stm32f4_flash_dev;` in `hw/mcu/stm/stm32f4xx/include/mcu/stm32f4_bsp.h`, weird!

utzig: Btw, do you have any idea what board this is?

kasjer: I have no clue, I spent few minutes to find, and failed

utzig: Same here! ;-) I assume it might be some custom board contributed way back that might not be available anymore...

PR #1885, opened 2019-06-18
kernel/os; fix core dumps & crash_restore_regs for M33.

mkiiskila:


kernel/os/src/arch/cortex_m33/os_fault.c
```diff
@@ -96,27 +96,20 @@ trap_to_coredump(struct trap_frame *tf, struct coredump_regs *regs)
     regs->r12 = tf->ef->r12;
     /*
      * SP just before exception for the coredump.
-     * See ARMv7-M Architecture Ref Manual, sections B1.5.6 - B1.5.8
+     * See ARMv8-M Architecture Ref Manual, section E2.1.236
```
andrzej-kaczmarek: this reference does not seem to be correct (it points to some pseudocode reference)

mkiiskila: The pseudocode shows what gets pushed to stack in exception. So it is relevant :)

mkiiskila:
Good point about M4, I fixed the M4 for FP active case.

PR #1887, opened 2019-06-19
STM32F4 system startup unification

utzig:
This unifies the STM32F4 series clocking into a single syscfg based clock file. It was tested on pretty much every supported board (apart from sensorhub which is not being configured, and using reset config).


kasjer:
so far looks fine. benchmark stuff look out of place though.

utzig:
@kasjer I can open another PR just for the coremark app. I will move the clock settings to a mcu file where syscfgs are used for the values, and add a fallback for BSPs that need custom clock settings when no method was chosen (HSE/LSE/HSI/LSI/MSI).

utzig:
@kasjer Now it's closer to finished, I am just finishing testing a few boards. Could you check the black vet6 still is ok? Also I will remove coremark next...

kasjer:
@utzig black_vet6 works correct with this PR, clock settings are OK

PR #1889, opened 2019-06-19
[console] fix race condition in rtt_console_write_ch

saulpw:
- if multiple tasks write to rtt console when it is not present,
rtt_console_retries_left can be decremented below zero, leading to
near-infinite retry.  This blocks the calling task for a few years.


saulpw:
Took us a few hours to debug, so hopefully we can save someone else the effort.  Go Team Open Source! :)

PR #1891, opened 2019-06-24
trng driver implemented with software backend

mkiiskila:
Mostly for use from unit tests, could also be utilized with BSPs which don't have access to HW RNG.


utzig:
@mkiiskila LGTM. I am curious, not sure how the hmac hash construction works, but wouldn't it be possible that when you add the `pid` many times as entropy, if the pid was a sequence of bits, for simplicity something like `0x11111111`, that it could fail the read test?

utzig:
@mkiiskila oops, more correctly in the question above would be something like `0x55555555`, `0xAAAAAAAA`, and so on and so forth...

mkiiskila:
I believe it would be very difficult to get the read test to fail; albeit not impossible. The probability is very, very low. The output will feel seemingly random even with limited amount of initial entropy.

What is more of a problem is predictable output; I did add comment warning folks regarding this if they use this for real.

PR #1895, opened 2019-06-25
Olimex E407: use standard openocd stm32f4x target

utzig:
This BSP has a customized open-ocd target which seems to be a relic from a time when open-ocd had no support for STM32F407 (or not good?). Using the standard stm32f4x target of open-ocd works just fine here when using one of the latest versions.


mkiiskila:
I don't remember what the reason was for the special openocd script. Might've had something to do with different versions of openocd at the time.

PR #1896, opened 2019-06-25
native; enable os_cputime for native BSP. Make task priority for

mkiiskila:
hal_timer configurable. This allows writing unittests for subsystems which use os_cputime.

This breaks nimble controller unittest. There's going to be a PR against that repo fixing that.

### DIFF
--- a/hw/mcu/native/syscfg.yml
+++ b/hw/mcu/native/syscfg.yml
@@ -65,4 +65,8 @@ syscfg.defs:
     MCU_UART_POLLER_PRIO:
         description: 'Priority of native UART poller task.'
         type: task_priority
+        value: 1
+    MCU_TIMER_POLLER_PRIO:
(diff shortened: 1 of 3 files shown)


hw/mcu/native/syscfg.yml
```diff
@@ -63,6 +63,10 @@ syscfg.defs:
         restrictions:
             - "!MCU_FLASH_STYLE_ST"
     MCU_UART_POLLER_PRIO:
+        description: 'Priority of native UART poller task.'
+        type: task_priority
+        value: 1
+    MCU_TIMER_POLLER_PRIO:
```
andrzej-kaczmarek: description should be fixed (c&p)

mkiiskila: Yes. Thanks for spotting this.

Issue #916, opened 2018-03-14
os_sem: possible token overflow


sdalu:
When releasing a semaphore `os_sem_release`, token count is incremented, but no check is done to handle the case where it will overflow

mlaz:
This was discussed on https://github.com/apache/mynewt-core/pull/1897, we concluded overflowing uint16 in this case was very unlikely.

PR #1897, opened 2019-06-25
os_sem: Preventing token overflow.

mlaz:
This PR addresses https://github.com/apache/mynewt-core/issues/916

### DIFF
--- a/kernel/os/src/os_sem.c
+++ b/kernel/os/src/os_sem.c
@@ -42,6 +42,7 @@ os_sem_init(struct os_sem *sem, uint16_t tokens)
         goto done;
     }
 
+    sem->init_tokens = tokens;
     sem->sem_tokens = tokens;
     SLIST_FIRST(&sem->sem_head) = NULL;
 
@@ -74,6 +75,12 @@ os_sem_release(struct os_sem *sem)
         ret = OS_INVALID_PARM;
         goto done;
     }
+
+    if (sem->sem_tokens >= sem->init_tokens) {
+        ret = OS_EINVAL;
+        goto done;
+    }
+
 
     /* Get current task */
     resched = 0;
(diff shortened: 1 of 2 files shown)


kernel/os/src/os_sem.c
```diff
@@ -42,6 +42,7 @@ os_sem_init(struct os_sem *sem, uint16_t tokens)
         goto done;
     }
 
+    sem->max_tokens = tokens;
```
andrzej-kaczmarek: `tokens` parameter is an initial value, not a max value
we can only check for an integer overflow on `sem_tokens`

mlaz: @andrzej-kaczmarek agreed, but wouldn't the problem also be to pass the initial value? probably more likely to happen then to OF an uint16_t.

andrzej-kaczmarek: it's ok to pass any initial value, it depends on the purpose of semaphore.
and as you said, it's unlikely to overflow a 16-bit counter in a semaphore so most likely this is why we do not care about it :-)

wes3: If the limit was a uint8_t we would have checked for token overflow. If you are overflowing a 16-bit token for a semaphore you are not using it correctly :-) Or there is something very, very wrong. I would even consider asserting if the tokens would overflow as it simply should never happen.

mlaz: Ok, so I'll probably drop this

utzig: I would think that `tokens` is the number of resources I want to protect. Let's say I can allow 4 devices access some resource (I might have 4 caches, etc), and then one after one 4 of them  `_pend` which does not allow any more to use until some of them `_release`. But say for some reason (bug ;-) someone that had a lock releases it two times. Now we have a counter that allow 5 consumers which probably means a data race. This could happen with open/closing devices, etc. So this is not how the API is supposed to be?

andrzej-kaczmarek: @utzig imposing a limit on semaphore count based on initial value would make it useless as a signal (which is most common semaphore usage in core repo right now)

PR #1898, opened 2019-06-25
fix lis2ds12_set_wake_up_dur

gonccalo:
Shifting after and'ing caused the result to be always 0. Only `val` should be shifted to be aligned with `LIS2DS12_WAKE_DUR_DUR` mask.


mlaz:
LGTM

PR #1907, opened 2019-07-04
Changing the device closed by battery_adc_close

gonccalo:
This function tries to close `bat_adc->dev` when it should be closing `bat_adc->adc_dev` that corresponds to the underlying ADC device.
In my case calling this function caused a recursion loop and crashed mynewt.


mlaz:
Looks good to me.

PR #1921, opened 2019-07-20
Fix a few issues in da1469x_charger shell

benmccrea:
The da1469x_charger shell was mostly complete, but the register read/write functionality wasn't finished. This PR completes that functionality and removes some redundant help string data.


kasjer:
@benmccrea read and write of charger register was available only in new style of commands while it was not handled in compat mode by mistake.
Help strings were part of newt style of shell.

All compat/non-compat issues are handled in #1924

benmccrea:
@kasjer Ok, I'm checking out your new PR now.

PR #1927, opened 2019-07-23
Add Dialog OTP driver and associated scripts

agross-korg:


andrzej-kaczmarek:
I don't think we need new HAL for this - we had similar case some time ago with TRNG and decided to make it a driver. I don't know if it makes sense to make it like ADC, PWM or TRNG which have common API or just driver dedicated for DA1469x, depends on how common it is in other devices.

agross-korg:
That's fair.  I suppose we could wait for a second PROM like device to come along before doing a HAL for it.

sterlinghughes:
Agree with @andrzej-kaczmarek .  Also, style-wise, braces should be on the same line as the if () statement.

agross-korg:
old habits die hard.  I am respinning.

test/spiflash_stress_test/src/spiflash_stress_test.c
```diff
@@ -0,0 +1,179 @@
...
+/*
+ * Initialize the package. Only called from sysinit().
+ */
+void
+spiflash_stress_test_init(void)
+{
+    os_sem_init(&sem, 0);
+    runtest_init_task(flash_test_task1, MYNEWT_VAL(OS_MAIN_TASK_PRIO) + 1);
+    runtest_init_task(flash_test_task1, MYNEWT_VAL(OS_MAIN_TASK_PRIO) + 2);
+    runtest_init_task(flash_test_task1, MYNEWT_VAL(OS_MAIN_TASK_PRIO) + 3);
```
vrahane: I think you can use SPIFLASH_STRESS_TEST_TASK_COUNT above as well, that way SPIFLASH_STRESS_TEST_TASK_COUNT can be used to instrument any number of tasks. You can even make it a syscfg then :-)

test/spiflash_stress_test/src/spiflash_stress_test.c
```diff
@@ -0,0 +1,179 @@
...
+{
+    if (argc > 1 && (!strcmp(argv[1], "?") || !strcmp(argv[1], "help"))) {
+        console_printf("Commands Available\n");
+        console_printf("start\n");
+        return 0;
+    }
+
+    os_sem_release(&sem);
+    os_sem_release(&sem);
+    os_sem_release(&sem);
```
vrahane: SPIFLASH_STRESS_TEST_TASK_COUNT does not get used while releasing semaphores.

test/spiflash_stress_test/src/spiflash_stress_test.c
```diff
@@ -0,0 +1,179 @@
...
+                      prio, OS_WAIT_FOREVER, stack,
+                      MYNEWT_VAL(SPIFLASH_STRESS_TEST_STACK_SIZE));
+    assert(rc == 0);
+
+#endif
+    runtest_next_task_idx++;
+    return task;
+}
+
+const uint8_t pattern[] = "1234567890 We choose to go to the moon in this decade and do the other things, not because they are easy, but because they are hard.";
```
vrahane: Our coding style suggests code size per line should be 79.

test/spiflash_stress_test/src/spiflash_stress_test.c
```diff
@@ -0,0 +1,179 @@
...
+    { MYNEWT_VAL(SPIFLASH_STRESS_TEST_FLASH_AREA_ID), 0x02000, 0x06000, 7, 12 },
+    { MYNEWT_VAL(SPIFLASH_STRESS_TEST_FLASH_AREA_ID), 0x08000, 0x08000, 13, 13 },
+};
+
+
+struct os_task *
+runtest_init_task(os_task_func_t task_func, uint8_t prio)
+{
+    struct os_task *task = NULL;
+#if 1
```
vrahane: Are you planning on removing this later ? Doesn't really serve a purpose if it's 1 :-)

agross-korg:
Identical changes were already merged.  Abandoning this PR.

PR #1930, opened 2019-07-25
Fix driver for BMP388 sensor

kasjer:
This fixes few small issues

### DIFF
--- a/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
+++ b/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
@@ -382,7 +382,7 @@ struct bmp388_cfg {
     uint8_t int_enable_type  : 2;
     uint8_t int_pp_od   : 1;
     uint8_t int_latched : 1;
-    uint8_t int_active_low  : 1;
+    uint8_t int_active_level  : 1;
 
     /* Power mode */
     uint8_t power_mode     : 4;
@@ -861,21 +861,21 @@ int bmp388_get_latched_int(struct bmp388 *bmp388, uint8_t *en);
 * Sets whether interrupts are active high or low
 *
 * @param bmp388 The device
-* @param low Value to set (0 = active high, 1 = active low)
+* @param level Value to set (1 = active high, 0 = active low)
 *
 * @return 0 on success, non-zero on failure
 */
-int bmp388_set_int_active_low(struct bmp388 *bmp388, uint8_t low);
+int bmp388_set_int_active_level(struct bmp388 *bmp388, uint8_t level);
 
 /**
 * Gets whether interrupts are active high or low
 *
 * @param bmp388 The device
-* @param low Ptr to store value (0 = active high, 1 = active low)
+* @param level Ptr to store value (1 = active high, 0 = active low)
 *
 * @return 0 on success, non-zero on failure
 */
-int bmp388_get_int_active_low(struct bmp388 *bmp388, uint8_t *low);
+int bmp388_get_int_active_low(struct bmp388 *bmp388, uint8_t *level);
 
 /**
 * Set filter config
--- a/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
+++ b/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
@@ -457,18 +457,18 @@ enum bmp3_intf {
 };
 
 /*!
-* @brief bmp3 sensor structure which comprises of temperature and pressure
+* @brief bmp3 sensor data which comprises temperature and pressure
 * data.
 */
 struct bmp3_data {
-    /*! Compensated temperature */
-    int64_t temperature;
-    /*! Compensated pressure */
-    uint64_t pressure;
+    /*! Compensated temperature in 0.01 degC*/
+    int16_t temperature;
+    /*! Compensated pressure in 0.01 Pa */
+    uint32_t pressure;
 };
 
 /*!
-* @brief bmp3 sensor structure which comprises of uncompensated temperature
+* @brief bmp388 sensor structure comprises uncompensated temperature
 * and pressure data.
 */
 struct bmp3_uncomp_data {
@@ -498,7 +498,7 @@ struct bmp3_reg_calib_data {
     int16_t par_p9;
     int8_t par_p10;
     int8_t par_p11;
-    int64_t t_lin;
+    int32_t t_lin;
 };
 
 /*!
@@ -725,7 +725,8 @@ int8_t bmp388_set_forced_mode_with_osr(struct sensor_itf *itf, struct bmp3_dev *
 */
 int8_t bmp3_init(struct sensor_itf *itf, struct bmp3_dev *dev);
 
-int8_t bmp388_get_sensor_data(struct sensor_itf *itf, struct bmp3_dev *dev, struct bmp3_data *sensor_data);
+int8_t bmp388_get_sensor_data(struct sensor_itf *itf, struct bmp3_dev *dev,
+                              struct bmp3_data *sensor_data);
 
 /*!
 * @brief This API sets the power control(pressure enable and
--- a/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
+++ b/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
@@ -427,7 +427,10 @@ STATS_SECT_END
 
 struct bmp388 {
 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
-    struct bus_i2c_node i2c_node;
+    union {
+        struct bus_i2c_node i2c_node;
+        struct bus_spi_node spi_node;
+    };
 #else
     struct os_dev dev;
 #endif
--- a/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
+++ b/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
@@ -397,8 +397,6 @@ struct bmp388_cfg {
 
 /* Used to track interrupt state to wake any present waiters */
 struct bmp388_int {
-    /* Synchronize access to this structure */
-    os_sr_t lock;
     /* Sleep waiting for an interrupt to occur */
     struct os_sem wait;
     /* Is the interrupt currently active */
--- a/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
+++ b/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
@@ -423,25 +423,6 @@ STATS_SECT_START(bmp388_stat_section)
     STATS_SECT_ENTRY(read_errors)
 STATS_SECT_END
 
-struct bmp388 {
-#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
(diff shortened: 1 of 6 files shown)


kasjer:
> I am a bit concerned about you removing the lock from the int_status structure. Also, the driver calls should be using the sensor_itf and not the os_dev or the driver as the primary argument. Reason for that is to make the calls abstract-able. If you look at the other drivers, we have done that in the past and it works well for the app.
> Rest looks good.

Don't be concern about lock, the way it was used was wrong across all sensors. Please take a look at non-sensor usage of critical sections. The way it was used did not have any negative impact simply because it was not used recursively as critical section could be.

I totally disagree with second opinion, I don't see any reason to use sensor_itf for anything apart calls to non bus driver version of i2c and spi functions. Other functions are strictly connected to bmp388 struct and I don't see a point of passing mostly useless pointer that has to be somehow up-cast to get access to fields connected to pressure sensor stuff.

vrahane:
@kasjer
For 1: I think you are right, `sr` should always be on the stack. We should change it for other drivers as well.
For 2: `sensor_itf` was supposed to be an interface that needs to get passed around to different driver public functions. All the other drivers do it the same way. What do we get by changing that paradigm now, I do not understand. Is it causing any issues for you that you want to move away from that approach ?

vrahane:
Example of usage of the `sensor_itf`: When we use it in an app which has different modules, assume one of them is power modes.
If we wanted to set the power mode using the driver function, say `bmp388_set_power_mode(itf, mode)` from the app, the `itf` is shared in the system and is generally accessible by other app packages. Are all hardware device drivers accessible from the different app packages, nope they are not. So, that is the reason `sensor_itf` is used. Please let me know if you need further . clarification.

andrzej-kaczmarek:
I agree with @kasjer on this. `sensor_itf` does not make sense anymore since `os_dev` is an universal handle for object in the system and should be used in new code. iirc we had the same discussion when bus driver was created and instead of creating new "bus interface" the decision was to go with `os_dev`. device structs are created by placing `os_dev` as a first member (or `bus_node` which does the same) thus other packages can safely case from `os_dev` to their specialized functions. it makes interfaces in the system more consistent if you can have single handle for the same object in all packages and `os_dev` is just the right interface.

kasjer:
To use `bmp388_set_power_mode()` code must include correct header that has bmp388 definition that can be used.
To get itf from user code you need to open device (but then you have pointer to bmp388 first) or you can get sensor struct first, then you have to get sensor_itf from there using `SENSOR_GET_ITF()`, but there is also `SENSOR_GET_DEVICE()` that can get device.
There are only two functions that can work on generic sensor_itf (`sensor_itf_lock()` `sensor_itf_unlock()`) that are empty in bus driver build. No other function that has itf parameter will work correctly if it will be called with correct itf taken from other device.

vrahane:
@andrzej-kaczmarek while developing the BUS driver, the thought was that `os_dev` will have a generic `itf` structure at some point and we would just replace `sensor_itf` with `os_dev_itf`. 

BTW, it seems like you trying to do extra work just to get rid of `sensor_itf` for the `BUS_DRIVER`.

```
 112 bool static inline
 113 itf_is_spi(struct bmp388 *bmp388)
 114 {      
 115 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
 116     return bmp388->node_is_spi;
 117 #else
 118     return bmp388->sensor.s_itf.si_type == SENSOR_ITF_SPI;
 119 #endif
 120 }  
```
The above code suggests you are making generic code non-generic. 

Also, in the code below, you actually are using `node_is_spi` to determine whether the node is SPI or not, what is the problem in using the `sensor_itf` which is far more generic. If you want to move away from `sensor_itf` and use `os_dev`, can you please start a thread to get rid of `sensor_itf` and add a generic mechanism for interface to `os_dev` which would be far more useful.

```
 431 int8_t
 432 bmp388_readlen(struct bmp388 *bmp388, uint8_t reg, uint8_t *buffer,
 433                uint16_t len)
 434 {
 435     int rc;
 436 
 437 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
 438     uint8_t reg_buf[2] = { reg };
 439     uint16_t wlen = 1;
 440 
 441     if (itf_is_spi(bmp388)) {
 442         reg_buf[0] |= BMP388_SPI_READ_CMD_BIT;
 443         wlen = 2;
 444     }
 445     rc = bus_node_simple_write_read_transact((struct os_dev *)bmp388, &reg_buf,
 446                                              wlen, buffer, len);
 447 #else
 448     struct sensor_itf *itf = &bmp388->sensor.s_itf;
 449 
 450     rc = sensor_itf_lock(itf, MYNEWT_VAL(BMP388_ITF_LOCK_TMO));
 451     if (rc) {
 452         return rc;
 453     }
 454 
 455     if (itf->si_type == SENSOR_ITF_I2C) {
 456         rc = bmp388_i2c_readlen(itf, reg, buffer, len);
 457     } else {
 458         rc = bmp388_spi_readlen(itf, reg, buffer, len);
 459     }
 460 
 461     sensor_itf_unlock(itf);
 462 #endif
 463 
 464     return rc;
 465 }
```

kasjer:
> @andrzej-kaczmarek while developing the BUS driver, the thought was that `os_dev` will have a generic `itf` structure at some point and we would just replace `sensor_itf` with `os_dev_itf`.
> 
> @kasjer BTW, it seems like you trying to do extra work just to get rid of `sensor_itf` for the `BUS_DRIVER`.
> 
> ```
>  112 bool static inline
>  113 itf_is_spi(struct bmp388 *bmp388)
>  114 {      
>  115 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
>  116     return bmp388->node_is_spi;
>  117 #else
>  118     return bmp388->sensor.s_itf.si_type == SENSOR_ITF_SPI;
>  119 #endif
>  120 }  
> ```
> 
> The above code suggests you are making generic code non-generic.

@vrahane Code suggest that there is difference in how you can determine spi/i2c when bus driver is enable and when is not.

> 
> Also, in the code below, you actually are using `node_is_spi` to determine whether the node is SPI or not, what is the problem in using the `sensor_itf` which is far more generic. If you want to move away from `sensor_itf` and use `os_dev`, can you please start a thread to get rid of `sensor_itf` and add a generic mechanism for interface to `os_dev` which would be far more useful.

@vrahane For bus driver build sensor_itf does not have information that can be used to switch between spi/i2c so simply speaking sensor_itf for bus driver build could not possibly by used for that.

> 
> ```
>  431 int8_t
>  432 bmp388_readlen(struct bmp388 *bmp388, uint8_t reg, uint8_t *buffer,
>  433                uint16_t len)
>  434 {
>  435     int rc;
>  436 
>  437 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
>  438     uint8_t reg_buf[2] = { reg };
>  439     uint16_t wlen = 1;
>  440 
>  441     if (itf_is_spi(bmp388)) {
>  442         reg_buf[0] |= BMP388_SPI_READ_CMD_BIT;
>  443         wlen = 2;
>  444     }
>  445     rc = bus_node_simple_write_read_transact((struct os_dev *)bmp388, &reg_buf,
>  446                                              wlen, buffer, len);
>  447 #else
>  448     struct sensor_itf *itf = &bmp388->sensor.s_itf;
>  449 
>  450     rc = sensor_itf_lock(itf, MYNEWT_VAL(BMP388_ITF_LOCK_TMO));
>  451     if (rc) {
>  452         return rc;
>  453     }
>  454 
>  455     if (itf->si_type == SENSOR_ITF_I2C) {
>  456         rc = bmp388_i2c_readlen(itf, reg, buffer, len);
>  457     } else {
>  458         rc = bmp388_spi_readlen(itf, reg, buffer, len);
>  459     }
>  460 
>  461     sensor_itf_unlock(itf);
>  462 #endif
>  463 
>  464     return rc;
>  465 }
> ```

andrzej-kaczmarek:
> @andrzej-kaczmarek while developing the BUS driver, the thought was that `os_dev` will have a generic `itf` structure at some point and we would just replace `sensor_itf` with `os_dev_itf`.

I do not really recall having such thought while developing bus driver.

> BTW, it seems like you trying to do extra work just to get rid of `sensor_itf` for the `BUS_DRIVER`.
> 
> ```
>  112 bool static inline
>  113 itf_is_spi(struct bmp388 *bmp388)
>  114 {      
>  115 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
>  116     return bmp388->node_is_spi;
>  117 #else
>  118     return bmp388->sensor.s_itf.si_type == SENSOR_ITF_SPI;
>  119 #endif
>  120 }  
> ```
> 
> The above code suggests you are making generic code non-generic.
> 
> Also, in the code below, you actually are using `node_is_spi` to determine whether the node is SPI or not, what is the problem in using the `sensor_itf` which is far more generic. If you want to move away from `sensor_itf` and use `os_dev`, can you please start a thread to get rid of `sensor_itf` and add a generic mechanism for interface to `os_dev` which would be far more useful.
> 
> ```
>  431 int8_t
>  432 bmp388_readlen(struct bmp388 *bmp388, uint8_t reg, uint8_t *buffer,
>  433                uint16_t len)
>  434 {
>  435     int rc;
>  436 
>  437 #if MYNEWT_VAL(BUS_DRIVER_PRESENT)
>  438     uint8_t reg_buf[2] = { reg };
>  439     uint16_t wlen = 1;
>  440 
>  441     if (itf_is_spi(bmp388)) {
>  442         reg_buf[0] |= BMP388_SPI_READ_CMD_BIT;
>  443         wlen = 2;
>  444     }
>  445     rc = bus_node_simple_write_read_transact((struct os_dev *)bmp388, &reg_buf,
>  446                                              wlen, buffer, len);
>  447 #else
>  448     struct sensor_itf *itf = &bmp388->sensor.s_itf;
>  449 
>  450     rc = sensor_itf_lock(itf, MYNEWT_VAL(BMP388_ITF_LOCK_TMO));
>  451     if (rc) {
>  452         return rc;
>  453     }
>  454 
>  455     if (itf->si_type == SENSOR_ITF_I2C) {
>  456         rc = bmp388_i2c_readlen(itf, reg, buffer, len);
>  457     } else {
>  458         rc = bmp388_spi_readlen(itf, reg, buffer, len);
>  459     }
>  460 
>  461     sensor_itf_unlock(itf);
>  462 #endif
>  463 
>  464     return rc;
>  465 }
> ```

You are missing the point here. The issue was whether we should pass `sensor_itf` or `os_dev`/driver-struct to driver APIs and I think this should be the latter because the same pointer is valid for driver, bus driver and `os_dev` which makes things really simple. And It's event better when using driver-struct explicitly (as @kasjer did) because you cannot pass incompatible parameter there which is easily possible with `sensor_itf`.

I do not really even care if we have `sensor_itf` or not, but since you asked... IMO this struct is redundant now because data stored in this struct are private to all drivers and can be as well moved there (just like `node_is_spi` is). And by private I mean they are not used outside driver so I don't see any benefit of having such "far more generic" struct which effectively forces each driver to keep information it may not even need, because e.g. not all drivers need the same number of interrupts and most of them likely do not even need one.

PR #1932, opened 2019-07-31
log: Add configurability to watermark storage

JustineKH:
Added `LOG_PERSIST_WATERMARK` for configurability of persisted watermark log values.
Added `LOG_REBOOT_INIT_WATERMARK_UNK` for configurability of setting watermark values to a determined "unknown" value -- `LOG_INIT_WATERMARK_VAL` -- instead of setting it to the first log entry. @ccollins476ad

### DIFF
--- a/sys/log/full/src/log_fcb.c
+++ b/sys/log/full/src/log_fcb.c
@@ -532,19 +532,28 @@ log_fcb_registered(struct log *log)
 {
 #if MYNEWT_VAL(LOG_STORAGE_WATERMARK)
     struct fcb_log *fl;
+#if MYNEWT_VAL(LOG_PERSIST_WATERMARK)
     struct fcb *fcb;
     struct fcb_entry loc;
+#endif
 
     fl = (struct fcb_log *)log->l_arg;
+
+#if MYNEWT_VAL(LOG_PERSIST_WATERMARK)
     fcb = &fl->fl_fcb;
 
     /* Set watermark to first element */
     memset(&loc, 0, sizeof(loc));
+
     if (fcb_getnext(fcb, &loc)) {
         fl->fl_watermark_off = loc.fe_area->fa_off + loc.fe_elem_off;
     } else {
         fl->fl_watermark_off = fcb->f_oldest->fa_off;
     }
+#else
+    /* Initialize watermark to designated unknown value*/
+    fl->fl_watermark_off = 0xffffffff;
+#endif
 #endif
     return 0;
 }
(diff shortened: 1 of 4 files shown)


sys/log/full/src/log_fcb.c
```diff
@@ -532,19 +532,30 @@ log_fcb_registered(struct log *log)
...
+#if !MYNEWT_VAL(LOG_REBOOT_INIT_WATERMARK_UNK)
     fcb = &fl->fl_fcb;
 
     /* Set watermark to first element */
     memset(&loc, 0, sizeof(loc));
+#endif
+
+#if MYNEWT_VAL(LOG_REBOOT_INIT_WATERMARK_UNK)
+    /* Initialize watermark to designated unknown value*/
+    fl->fl_watermark_off = (uint32_t)MYNEWT_VAL(LOG_INIT_WATERMARK_VAL);
```
ccollins476ad: No need to cast on assignment.

sys/log/full/src/log_fcb.c
```diff
@@ -603,9 +614,14 @@ log_fcb_storage_info(struct log *log, struct log_storage_info *info)
 #if MYNEWT_VAL(LOG_STORAGE_WATERMARK)
     /* Calculate used size */
     fa_used = el_max - fl->fl_watermark_off;
-    if ((int32_t)fa_used < 0) {
+
+    if (fl->fl_watermark_off == MYNEWT_VAL(LOG_INIT_WATERMARK_VAL)){
+        fa_used = (uint32_t)MYNEWT_VAL(LOG_INIT_WATERMARK_VAL);
```
ccollins476ad: No need to cast on assignment.

sys/log/full/syscfg.yml
```diff
@@ -115,6 +115,25 @@ syscfg.defs:
...
+        restrictions:
+            - LOG_PERSIST_WATERMARK == 0
+            - LOG_STORAGE_WATERMARK
+
+    LOG_INIT_WATERMARK_VAL:
+        description: >
+            Value to initialize watermarks to
+            to indicate unknown watermarks
+        value: 0xFFFFFFFF
+
```
ccollins476ad: Personally, I think this is too much configurability.  The ability to configure behavior is good, but it comes at a cost: extra complexity.  If I can't think of a case where someone would ever want to override a setting, I think it is best to remove the setting.

* `LOG_PERSIST_WATERMARK`: Makes sense.
* `LOG_REBOOT_INIT_WATERMARK_UNK`: I think this naturally follows from the persist setting.  I just can't think of a case where someone would want a) persist=1 && unk=1, or b) persist=0 && unk=0.
* `LOG_INIT_WATERMARK_VAL`: Would someone really ever want to use a value other than 0xffffffff here?

So my suggestion is to remove the second two settings.  If persistence is disabled, always initialize the watermark to 0xffffffff.

sys/log/full/src/log_fcb.c
```diff
@@ -532,19 +532,30 @@ log_fcb_registered(struct log *log)
...
+    fl->fl_watermark_off = (uint32_t)MYNEWT_VAL(LOG_INIT_WATERMARK_VAL);
+#else
     if (fcb_getnext(fcb, &loc)) {
         fl->fl_watermark_off = loc.fe_area->fa_off + loc.fe_elem_off;
     } else {
         fl->fl_watermark_off = fcb->f_oldest->fa_off;
     }
+#endif
 #endif
     return 0;
```
ccollins476ad: This function (`log_fcb_registered()`) is starting to get a bit confusing with all the `#if` / `#endif` directives.  Can you combine the bottom half into a single `#if` / `#else`?

PR #1933, opened 2019-07-31
Logging Header Image Hash

brolan-juul:
LOG_VERSION 3:
- Carry arbitrary data up to 4 bytes long per log.
- Flag to indicate what data is being stored in the reserve field.
- First 4 bytes of the payload will contain the image hash.

Currently only setup for the first four bytes of the image hash.

### DIFF
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -33,6 +33,8 @@
 #include "shell/shell.h"
 #endif
 
+#include "imgmgr/imgmgr.h"
+
 struct log_info g_log_info;
 
 static STAILQ_HEAD(, log) g_log_list = STAILQ_HEAD_INITIALIZER(g_log_list);
@@ -492,6 +494,15 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
 #else
     assert(etype == LOG_ETYPE_STRING);
 #endif
+#if MYNEWT_VAL(LOG_VERSION) > 3
+
+#if MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
+    uint8_t hash[IMGMGR_HASH_LEN];
+    imgr_read_info(0, NULL, hash, NULL);
+    memcpy(ue->ue_rsvd, hash, 4);
+#endif // MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
+
+#endif // MYNEWT_VAL(LOG_VERSION) > 3
 
 err:
     return (rc);
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -499,7 +499,7 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
 #if MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
     uint8_t hash[IMGMGR_HASH_LEN];
 
-    ue->ue_flags = LOG_FLAGS_LOG_IMG_HASH_e;
+    ue->ue_flags &= LOG_FLAGS_LOG_IMG_HASH;
 
     imgr_read_info(0, NULL, hash, NULL);
     memcpy(ue->ue_rsvd, hash, 4);
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -498,6 +498,9 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
 
 #if MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
     uint8_t hash[IMGMGR_HASH_LEN];
+
+    ue->ue_flags = LOG_FLAGS_LOG_IMG_HASH_e;
+
     imgr_read_info(0, NULL, hash, NULL);
     memcpy(ue->ue_rsvd, hash, 4);
 #endif // MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -491,20 +491,11 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
     ue->ue_index = idx;
 #if MYNEWT_VAL(LOG_VERSION) > 2
     ue->ue_etype = etype;
+#if MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
+    ue->ue_flag &= LOG_FLAGS_LOG_IMG_HASH;
+#endif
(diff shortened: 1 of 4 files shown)


sys/log/full/src/log.c
```diff
@@ -492,6 +494,18 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
...
+#if MYNEWT_VAL(LOG_VERSION) > 3
+
+#if MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
+    uint8_t hash[IMGMGR_HASH_LEN];
+
+    ue->ue_flags = LOG_FLAGS_LOG_IMG_HASH_e;
+
+    imgr_read_info(0, NULL, hash, NULL);
+    memcpy(ue->ue_rsvd, hash, 4);
+#endif // MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
```
vrahane: Our coding_style says we do not allow C++ comments. Please remove those or change them into C style comments using `/* */`.

sys/log/full/src/log_nmgr.c
```diff
@@ -265,6 +273,14 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
         }
     }
     g_err |= cbor_encoder_close_container(&rsp, &str_encoder);
+
+#if MYNEWT_VAL(LOG_VERSION) > 3
+    g_err |= cbor_encode_text_stringz(&rsp, "flag");
+    g_err |= cbor_encode_uint(&rsp, ueh->ue_flags);
+    g_err |= cbor_encode_text_stringz(&rsp, "res");
+    g_err |= cbor_encode_byte_string(&rsp, ueh->ue_rsvd, 4);
+#endif // MYNEWT_VAL(LOG_VERSION) > 3
```
vrahane: Our coding_style says we do not allow C++ comments. Please remove those or change them into C style comments using /* */.

sys/log/full/src/log_nmgr.c
```diff
@@ -173,6 +173,14 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
         off += rc;
     }
     g_err |= cbor_encoder_close_container(&rsp, &str_encoder);
+
+#if MYNEWT_VAL(LOG_VERSION) > 3
+    g_err |= cbor_encode_text_stringz(&rsp, "flag");
+    g_err |= cbor_encode_uint(&rsp, ueh->ue_flags);
+    g_err |= cbor_encode_text_stringz(&rsp, "res");
+    g_err |= cbor_encode_byte_string(&rsp, ueh->ue_rsvd, 4);
+#endif // MYNEWT_VAL(LOG_VERSION) > 3
```
vrahane: Our coding_style says we do not allow C++ comments. Please remove those or change them into C style comments using /* */.

sys/log/full/syscfg.yml
```diff
@@ -32,6 +32,12 @@ syscfg.defs:
         description: 'Limits what level log messages are compiled in.'
         value: 0
 
+    LOG_FLAGS_LOG_IMG_HASH:
+        description: >
+            Enable logging of the first 4 bytes of the image hash. 0 - disable; 
+            1 - enable.
+        value: 0
```
vrahane: I think this should be enabled by default, also the `LOG_VERSION` should be bumped up in the syscfg.yml.

sys/log/full/src/log.c
```diff
@@ -492,6 +494,18 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
 #else
     assert(etype == LOG_ETYPE_STRING);
 #endif
+#if MYNEWT_VAL(LOG_VERSION) > 3
+
+#if MYNEWT_VAL(LOG_FLAGS_LOG_IMG_HASH)
+    uint8_t hash[IMGMGR_HASH_LEN];
+
+    ue->ue_flags = LOG_FLAGS_LOG_IMG_HASH_e;
```
vrahane: ue->ue_flags should be set by left shifting `1` by `LOG_FLAGS_LOG_IMG_HASH_e` and logically anding it with the flags field.
Eg: `ue->ue_flags &= (1 << LOG_FLAGS_LOG_IMG_HASH_e)`
What we generally do is define the flags such that they are already left shifted by 1 and are not enums.

Eg:
Definition:
```
typedef enum {
    LOG_FLAGS_LOG_IMG_HASH_e = 1,
    LOG_FLAGS_MAX_e,
}log_flags_e;
```
would be
```
#define LOG_FLAGS_LOG_IMG_HASH           (1 << 0)
#define LOG_FLAGS_LOG_FUTURE_YYY1      (1 << 1)  **/* Just an example, do not add this */**
#define LOG_FLAGS_LOG_FUTURE_YYY2     (1 << 2)  **/* Just an example, do not add this */**
```
And so on. That way you do not have to do the shifting in the code and the above code snippet would be:
`ue->ue_flags &= LOG_FLAGS_LOG_IMG_HASH`

sys/log/full/src/log_nmgr.c
brolan-juul: Removed.

sys/log/full/src/log.c
brolan-juul: Removed.

sys/log/full/src/log_nmgr.c
brolan-juul: Removed.

sys/log/full/syscfg.yml
brolan-juul: Bumped up in LOG_VERSION = 4 and enabling image hash logging by default.

sys/log/full/src/log.c
brolan-juul: Moved to bit shift #define as opposed to enums.

sys/log/full/src/log_nmgr.c
```diff
@@ -148,6 +150,9 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
         return MGMT_ERR_ECORRUPT;
     }
 
+    g_err |= cbor_encode_text_stringz(&rsp, "flag");
+    g_err |= cbor_encode_uint(&rsp, ueh->ue_flag);
```
ccollins476ad: Sorry for the late comment.  In my opinion, we should not include flags in the newtmgr response.  This just makes the response bigger without providing additional information.  The presence / absence of the image hash is what the client cares about, and it can determine that without seeing the flags.

brolan-juul: Got it. Removing it.

PR #1936, opened 2019-08-01
Define logs in packages' syscfg files

ccollins476ad:
Define logs in packages' syscfg files

Defining logs in syscfg files provides a few benefits:

1. Ability to set log level per module at build time.  Setting a global log level of 0 (`LOG_LEVEL=0`) might produce an image that is too big.  Instead, we can enable debug logging only for the modules we are interested in.

2. Easier to visualize system log configuration (see below).

3. Log modules can be remapped at build time in case of conflicts.

The following procedure was used for choosing module and level syscfg names:

1. If the package already defines these settings, keep them unchanged.
2. Otherwise, define:

* `<package-name>_LOG_MOD`
* `<package-name>_LOG_LVL`

To see the system-wide log configuration, use either of these commands:

```
newt target logcfg brief <target-name>
newt target logcfg show <target-name>
```

Example output:
```
newt target logcfg brief slinky-nordic_pca10056

Brief log config for targets/slinky-nordic_pca10056:
           LOG | MODULE   | LEVEL
---------------+----------+--------------
      DFLT_LOG | 0        | 1 (INFO)
    IMGMGR_LOG | 176      | 0 (DEBUG)
```

```
newt target logcfg show slinky-nordic_pca10056

Log config for targets/slinky-nordic_pca10056:
DFLT_LOG:
    Package: @apache-mynewt-core/sys/log/common
    Module:  0                [DFLT_LOG_MOD]
    Level:   1 (INFO)         [DFLT_LOG_LVL]

IMGMGR_LOG:
    Package: @apache-mynewt-core/mgmt/imgmgr
    Module:  176              [IMGMGR_LOG_MOD]
    Level:   0 (DEBUG)        [IMGMGR_LOG_LVL]
```

### DIFF
--- a/apps/bleprph_oic/src/gatt_svr.c
+++ b/apps/bleprph_oic/src/gatt_svr.c
@@ -31,13 +31,13 @@ gatt_svr_register_cb(struct ble_gatt_register_ctxt *ctxt, void *arg)
 
     switch (ctxt->op) {
     case BLE_GATT_REGISTER_OP_SVC:
-        MODLOG_DFLT(DEBUG, "registered service %s with handle=%d\n",
+        DFLT_LOG_DEBUG("registered service %s with handle=%d\n",
                     ble_uuid_to_str(ctxt->svc.svc_def->uuid, buf),
                     ctxt->svc.handle);
         break;
 
     case BLE_GATT_REGISTER_OP_CHR:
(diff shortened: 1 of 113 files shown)


ccollins476ad:
Thanks for the reviews, all.

@kasjer, I have added the missing periods.

Also, I noticed there were two packages I missed:
* hw/drivers/sensors/dps368
* sys/mfg

PR #1939, opened 2019-08-01
kernel/os: Catch returns from tasks

kasjer:
New mynewt user may not noticed that tasks function should not
return which is common in other systems.
If user tasks do return for arm and riscv platform PC will be
set to 0 and that will soon result in hard fault exception.
In such case, even if hard fault handler is called, further
stack corruption can happen since SP will reach top and will
point to memory that do not belong to stack any more.
Inexperienced or new myenwt user can find it difficult to
track this simple problem.

This adds minimal changes to arm and riscv architecture code
that can help catch such problems easily.

### DIFF
--- a/kernel/os/src/arch/cortex_m0/os_arch_arm.c
+++ b/kernel/os/src/arch/cortex_m0/os_arch_arm.c
@@ -144,6 +144,16 @@ os_arch_in_critical(void)
     return (isr_ctx & 1);
 }
 
+static void
+os_arch_task_return_handler(void)
+{
+    /*
+     * If you are stuck here it means that task finished by
+     * simple return which is not supported.
+     */
+    while (1);
+}
+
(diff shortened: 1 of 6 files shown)


kernel/os/src/arch/cortex_m0/os_arch_arm.c
```diff
@@ -144,6 +144,16 @@ os_arch_in_critical(void)
     return (isr_ctx & 1);
 }
 
+static void
+os_arch_task_return_handler(void)
+{
+    /*
+     * If you are stack here it means that task finished by
```
mlaz: did you mean "stuck" ?

kasjer: yes, I forced push new one, thanks

PR #1940, opened 2019-08-02
sys/log: Make sure watermark is updated for each log after read

andrzej-kaczmarek:
Watermark should be updated after entries are properly encoded for each
log.


andrzej-kaczmarek:
@JustineKH

andrzej-kaczmarek:
@ccollins476ad indeed, it stores an offset. iirc (although not 100% sure now) the reason for this was to be able to calculate usage easily which is not that easy with index.

PR #1943, opened 2019-08-03
Add image hash optionally as part of log entry

vrahane:
- Add image hash optionally as part of log entry
- It is optional and only honored if the syscfg is set which would
  in-turn set the flag
- Rename LOG_ENTRY_HDR_SIZE to LOG_BASE_ENTRY_HDR_SIZE
- Add settings for logging image hash
- Add some code in log_nmgr to return image hash
  if LOG_FLAGS_IMAGE_HASH is enabled
- Add support for logging image hash for all mediums fcb, cbmem,
  fcb_slot1, console and shell
- For mbuf append functions remove the requirement to have them as
  constants
- Add imgr_get_current_hash() and use it

Before the `log_entry_header` was 
```
struct log_entry_hdr {
    int64_t ue_ts;
    uint32_t ue_index;
    uint8_t ue_module;
    uint8_t ue_level;
    uint8_t ue_etype;
}__attribute__((__packed__));
```
With this PR it has been changed to: 
```
struct log_entry_hdr {
    int64_t ue_ts;
    uint32_t ue_index;
    uint8_t ue_module;
    uint8_t ue_level;
    uint8_t ue_etype:4;
    uint8_t ue_flags:4;
    uint8_t ue_imghash[4];
}__attribute__((__packed__));
```
It is backwards compatible, meaning the log entries which were written using the old log entry headers are still valid. We make use of the 4 bits from the `ue_etype` field to define `ue_flags` which would then be used to specify what next 4 bytes are going to be interpreted as. `img_hash` in this case.

Also added a unit test to test the log entries with image hash `sys/log/full/selftest/align1_imghash`


vrahane:
This PR spawned from https://github.com/apache/mynewt-core/pull/1933 by @brolan-juul but has a lot more changes.

boot/stub/pkg.yml
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: boot/stub
+pkg.description: The stub package defines stub functions for the bootlloader api
```
ccollins476ad: Typo: extra `l`.

boot/stub/src/bootutil.c
```diff
@@ -0,0 +1,43 @@
...
+int boot_set_confirmed(void)
+{
+    return SYS_EIO;
+}
+
+int
+split_go(int loader_slot, int split_slot, void **entry)
+{
+    return SYS_EIO;
+}
```
ccollins476ad: For the rest of these bootutil functions (`boot_set_pending()`, `boot_set_confirmed()`, and `split_go()`), I recommend returning `SYS_ENOTSUP` rather than `SYS_EIO`.  I think "not supported" more accurately indicates that a stub package is being used (also, I believe other stub packages use this error code).

mgmt/imgmgr/src/imgmgr.c
```diff
@@ -301,6 +301,111 @@ imgr_read_info(int image_slot, struct image_version *ver, uint8_t *hash,
...
+ * @param hashlen Length of hash to return
+ *
+ * Returns -2 if either of the argument is 0 or NULL
+ * Returns -1 if area is not readable
+ * Returns 0 if image in slot is ok
+ * Returns 1 if there is not a full image
+ * Returns 2 if slot is empty
+ */
+int
+imgr_get_current_hash(uint8_t *hash, uint16_t hashlen)
```
ccollins476ad: Can this function (`imgr_get_current_hash()`) just call `imgr_read_info()` rather than duplicate all its code?

mgmt/imgmgr/syscfg.yml
```diff
@@ -50,6 +50,9 @@ syscfg.defs:
     IMGMGR_LOG_LVL:
         description: 'Minimum level for the image manager structured log.'
         value: 15  # Log disabled by default.
+    IMGMGR_DUMMY_HDR:
+        description: 'Returns dummy image header data for imgr functions'
```
ccollins476ad: I think this setting needs a bit more explanation.  Maybe add "Useful when there are no images present, e.g., unit tests."

sys/log/full/selftest/util/src/log_test_util.c
```diff
@@ -130,18 +130,20 @@ ltu_walk_verify(struct log *log, struct log_offset *log_offset,
     struct os_mbuf *om;
     char data[128];
     int dlen;
+    uint16_t ext_len;
```
ccollins476ad: Bit of a nitpick, but I feel like this variable is misnamed.  It contains the full header length, not just the extra length.  I would expect it to be called `hdr_len`.

sys/log/full/src/log.c
```diff
@@ -648,15 +690,13 @@ log_append_mbuf_typed_no_free(struct log *log, uint8_t module, uint8_t level,
      * maximum entry length was checked.
      */
     len = os_mbuf_len(om);
```
ccollins476ad: I realize you did not add this code, but this line is wrong.  It should use `OS_MBUF_PKTLEN()` instead of `os_mbuf_len()`.  I.e., we care about the length of the entire chain, not just the first buffer.

I don't think this needs to be fixed in this PR, but we should create an issue or a new PR soon.

sys/log/full/src/log.c
```diff
@@ -1005,3 +1070,21 @@ log_set_max_entry_len(struct log *log, uint16_t max_entry_len)
...
+    hdr->ue_flags |= LOG_FLAGS_IMG_HASH;
+
+    /* We have to account for LOG_IMG_HASHLEN bytes of hash */
+    rc = imgr_get_current_hash(hdr->ue_imghash, LOG_IMG_HASHLEN);
+    if (rc) {
+        return rc;
+    }
+
+#endif
+    return SYS_ENOTSUP;
```
ccollins476ad: If `LOG_FLAGS_IMAGE_HASH` is not enabled, I think this function should populate the hash field with a known value (all zeros? all 0xff?).  Otherwise, if a device writes some logs with the hash, then gets upgraded to an image that does not support this feature, a log read will yield entries with indeterminate hashes.

sys/log/full/src/log_cbmem.c
```diff
@@ -22,91 +22,108 @@
...
+    hdr_len = log_hdr_len((struct log_entry_hdr *)om->om_data);
+    
+    os_mbuf_pullup(om, hdr_len);
+    
+    memcpy(&hdr, om->om_data, hdr_len);
+
+    os_mbuf_adj(&m, hdr_len);
+
+    return log_cbmem_append_mbuf_body(log, &hdr, &m);
+}
```
ccollins476ad: This function (`log_cbmem_append_mbuf()`) is pretty tricky.  I think it could benefit from some comments.  In my opinion, the following points deserve explanation:
* Why copy into a temporary mbuf? (So that the original is not affected by the `os_mbuf_adj()` call).
* Why is it OK to cast `om->om_data` to a `struct log_entry_hdr *`? (Because this struct is packed, so alignment is not required).

sys/log/full/src/log_fcb.c
```diff
@@ -190,47 +190,25 @@ log_fcb_start_append(struct log *log, int len, struct fcb_entry *loc)
...
 {
     uint8_t mod;
+    uint32_t size;
+
+    size = ext_len;
 
     /* Assume power-of-two alignment for faster modulo calculation. */
     assert((align & (align - 1)) == 0);
+  
+    size += LOG_BASE_ENTRY_HDR_SIZE;
```
ccollins476ad: Sort of a nitpick, but it seems extra complicated to assign `size` in two statements.  Why not just `size = LOG_BASE_ENTRY_HDR_SIZE + ext_len`?

sys/log/full/src/log_fcb.c
```diff
@@ -239,45 +217,60 @@ log_fcb_hdr_body_bytes(uint8_t align)
...
+    ext_len = 0;
 
     if (fcb->f_align > LOG_FCB_MAX_ALIGN) {
         return SYS_ENOTSUP;
     }
 
-    rc = log_fcb_start_append(log, sizeof *hdr + body_len, &loc);
+    if (hdr->ue_flags & LOG_FLAGS_IMG_HASH) {
+        ext_len = LOG_IMG_HASHLEN;
+    }
```
ccollins476ad: I guess this is another nitpick, but `ext_len` seems like the wrong value to be calculating here.  Every time this variable gets used, it just gets added to `LOG_BASE_ENTRY_HDR_SIZE`.  Rather than manually calculate `ext_len` here, why not just do `hdr_len = log_hdr_len(hdr);`, and use `hdr_len` everywhere in this function.  

`log_fcb_hdr_body_bytes()` could also be modified to take the header length (rather than `ext_len`).  That feels more natural to me.

sys/log/full/src/log_fcb.c
```diff
@@ -364,48 +383,31 @@ log_fcb_append_mbuf(struct log *log, const struct os_mbuf *om)
...
 
-    rc = fcb_append_finish(fcb, &loc);
-    if (rc != 0) {
-        return rc;
-    }
+    os_mbuf_adj(&m, hdr_len);
 
-    return 0;
+    return log_fcb_append_mbuf_body(log, &hdr, &m);
 }
```
ccollins476ad: As with log_cbmem_append_mbuf():

* Should use OS_MBUF_PKTLEN()
* Some comments would help a lot!

sys/log/full/src/log_fcb.c
```diff
@@ -699,18 +701,24 @@ log_fcb_copy_entry(struct log *log, struct fcb_entry *entry,
...
-    if (rc != sizeof(ueh)) {
+    rc = log_fcb_read(log, entry, &ueh, 0, LOG_BASE_ENTRY_HDR_SIZE);
+    if (rc != LOG_BASE_ENTRY_HDR_SIZE) {
         goto err;
     }
 
-    dlen = min(entry->fe_data_len, LOG_PRINTF_MAX_ENTRY_LEN + sizeof(ueh));
+    if (ueh.ue_flags & LOG_FLAGS_IMG_HASH) {
+        ext_len = LOG_IMG_HASHLEN;
+    }
```
ccollins476ad: Probably better to use `hdr_len = log_hdr_len(&ueh);`.  Then this code doesn't have to get updated if we add a new optional field to the header.

sys/log/full/src/log_fcb_slot1.c
```diff
@@ -49,30 +49,45 @@ static bool g_slot1_locked;
...
+        return SYS_ENOMEM;
+    }
+
+    memcpy(&m, om, sizeof m);
+
+    os_mbuf_pullup(om, LOG_BASE_ENTRY_HDR_SIZE);
+
+    m.om_data = m.om_data + log_hdr_len((struct log_entry_hdr *)om->om_data);
+
+    return log_fcb_slot1_append_mbuf_body(log, (struct log_entry_hdr *)om->om_data, &m);
```
ccollins476ad: As with `log_cbmem_append_mbuf()` and `log_fcb_append_mbuf()`:
* Should use `OS_MBUF_PKTLEN()`
* Some comments would help a lot!

sys/log/full/src/log_nmgr.c
```diff
@@ -164,6 +164,10 @@ log_nmgr_encode_entry(struct log *log, struct log_offset *log_offset,
      */
     g_err |= cbor_encoder_create_indef_byte_string(&rsp, &str_encoder);
     for (off = 0; off < len && !g_err; ) {
+        /* 
+         * Continue subsequent reads after inserting image hash after the 
+         * first read. 
+         */
```
ccollins476ad: Is this comment accurate?  Maybe it is just out of date.  The image hash doesn't get inserted until after the body.

sys/log/full/src/log_cbmem.c
```diff
@@ -22,91 +22,108 @@
...
 
+static int
+log_cbmem_append_mbuf(struct log *log, struct os_mbuf *om)
+{
+    struct os_mbuf m;
+    uint16_t mlen;
+    uint16_t hdr_len;
+    struct log_entry_hdr hdr;
+
+    mlen = os_mbuf_len(om);
```
ccollins476ad: This should use OS_MBUF_PKTLEN() rather than os_mbuf_len(). We want to check the length of the entire mbuf chain (to see if the subsequent pull up will succeed), not the length of the first buffer in the chain.

sys/log/full/include/log/log.h
```diff
@@ -79,12 +79,12 @@ typedef int (*lh_read_mbuf_func_t)(struct log *, void *dptr, struct os_mbuf *om,
...
-                                     const struct log_entry_hdr *hdr,
+                                     struct log_entry_hdr *hdr,
                                      const void *body, int body_len);
-typedef int (*lh_append_mbuf_func_t)(struct log *, const struct os_mbuf *om);
+typedef int (*lh_append_mbuf_func_t)(struct log *, struct os_mbuf *om);
 typedef int (*lh_append_mbuf_body_func_t)(struct log *log,
-                                          const struct log_entry_hdr *hdr,
-                                          const struct os_mbuf *om);
+                                          struct log_entry_hdr *hdr,
+                                          struct os_mbuf *om);
```
ccollins476ad: Is it necessary to remove `const` from all the header parameters in the log API?  I understand why it had to be removed from the mbuf parameters (so that the log functions can call `os_mbuf_pullup()`), but I think the headers could remain const.  I would be reluctant to change the log API more than is necessary.

boot/stub/src/bootutil.c
```diff
@@ -0,0 +1,43 @@
...
+ * under the License.
+ */
+
+#include <defs/error.h>
+
+int boot_current_slot;
+
+int boot_swap_type(void)
+{
+    return SYS_EIO;
```
ccollins476ad: `boot_swap_type()` does not return an error code - it just returns a `BOOT_SWAP_TYPE_[...]` constant.  It probably makes the most sense for this function to return `BOOT_SWAP_TYPE_NONE`.

sys/log/full/src/log_cbmem.c
vrahane: Not applicable here.

sys/log/full/src/log_fcb_slot1.c
vrahane: Not applicable here.

sys/log/full/src/log_fcb.c
vrahane: Not applicable here.

sys/log/full/src/log.c
vrahane: Not applicable here.

mgmt/imgmgr/src/imgmgr.c
vrahane: Ok, that seems a fine enough ask, I actually wanted to just read the exact number of bytes and not read more bytes from the flash than needed but I guess reading does not take that many cycles, so I will just reuse the code we already have. Thanks.

sys/log/full/include/log/log.h
vrahane: Agreed, changed, I was reading the image hash in the respective append functions and hence, the log entry header could not be made const. Since that has changed and only get's called from `append_prepare()`, I can just change it back to be const.

vrahane:
@ccollins476ad I combined all the commits into a single one and addressed the review both verbal and mentioned above. Can you please take a look again and let me know if this is fine.

sys/log/full/include/log/log.h
```diff
@@ -260,11 +268,11 @@ struct log *log_find(const char *name);
  * @brief Writes the raw contents of a flat buffer to the specified log.
  *
  * NOTE: The flat buffer must have an initial padding of length
- * `LOG_ENTRY_HDR_SIZE`.  This padding is *not* reflected in the specified
+ * `LOG_BASE_ENTRY_HDR_SIZE`.  This padding is *not* reflected in the specified
```
ccollins476ad: I believe this comment is no longer accurate.  The buffer must contain enough padding to accommodate a full log entry header (which is more than `LOG_BASE_ENTRY_HDR_SIZE` if this new feature is enabled). Would you   mind changing this comment to reflect that?  I think it would be fine to just replace `LOG_ENTRY_HDR_SIZE` with `header-size` or something like that.

I only ask because these "non-body" functions have a confusing enough API as it is!

This comment applies to the doxygen comments for all the non-body functions in this file.

sys/log/full/selftest/align1_imghash/pkg.yml
```diff
@@ -0,0 +1,35 @@
...
+pkg.deps: 
+    - "@apache-mynewt-core/sys/console/stub"
+    - "@apache-mynewt-core/sys/log/full"
+    - "@apache-mynewt-core/sys/log/full/selftest/util"
+    - "@apache-mynewt-core/test/testutil"
+    - "@apache-mynewt-core/mgmt/newtmgr"
+    - "@apache-mynewt-core/boot/stub"
+
+pkg.apis:
+    - bootloader
```
ccollins476ad: I believe this API (`bootloader`) should be specified in `boot/stub`, not in this unit test.

PR #1945, opened 2019-08-05
sys/log: log_fcb_set_watermark to use log_fcb_walk

JustineKH:
Changing the way `log_fcb_set_watermark` starts the search for the read log. Now using `log_fcb_walk` to reduce the amount of time searching for log. @ccollins476ad


ccollins476ad:
The code looks great.  Just a few things:

1. There is a conflict in `log_nmgr.c`.  It is just a whitespace issue, so it should be easy to fix.
2. Can you please combine the commits into a single commit with a descriptive message?

PR #1947, opened 2019-08-06
hw/drivers/lps33thw:Added support to FIFO

mariotesi:
hw/drivers/lps33thw:Remove duplicated bitmask
    
        The F_FTH bitmask already defined and
        duplicated in CTRL_REG3

Signed-off-by: tesimari <mario.tesi@st.com>


vrahane:
Thanks for doing this Mario, however, I think the changes should be put in the earlier PR itself. You do not need a separate PR for this.

mariotesi:
@ vrahane the PR is related to lps33thw driver already merged in master, I rewrote the  message replacing lps33thw in place of wrong device lps33wh, sorry for mistake (those devices have almost the same name and I got confused)

vrahane:
@mariotesi So, you are actually fixing a bug, can you please update the comment above ?

mariotesi:
ok, done

PR #1976, opened 2019-08-30
hw/mcu/dialog; add da1469x_flash_setup()

mkiiskila:
This can be used in scenario where board does not have configuration script yet setting
up QSPI, but you need access to flash. E.g.. from flash_loader.

### DIFF
--- a/hw/mcu/dialog/da1469x/include/mcu/da1469x_hal.h
+++ b/hw/mcu/dialog/da1469x/include/mcu/da1469x_hal.h
@@ -68,6 +68,22 @@ struct da1469x_hal_spi_cfg {
     int8_t pin_ss;
 };
 
+/**
+ * Allow QSPI and the connected flash to be setup.
+ * This allows proper operation when configation script is not
+ * executed/present yet.
+ *
+ * @param dev    Pointer to device
+ * @param bursta BURSTCMDA_REG register setting
+ * @param burstb BURSTCMDB_REG register setting
+ * @param cmds   Array of byte sized commands sent over SPI to flash
+ * @param len    Number of command bytes
+ *
+ * @return 0 on success, non-zero on failure.
+ */
+void da1469x_flash_setup(const struct hal_flash *dev, uint32_t bursta,
+                         uint32_t burstb, uint8_t *cmds, int len);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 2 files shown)


hw/mcu/dialog/da1469x/include/mcu/da1469x_hal.h
```diff
@@ -68,6 +68,22 @@ struct da1469x_hal_spi_cfg {
...
+ * @param dev    Pointer to device
+ * @param bursta BURSTCMDA_REG register setting
+ * @param burstb BURSTCMDB_REG register setting
+ * @param cmds   Array of byte sized commands sent over SPI to flash
+ * @param len    Number of command bytes
+ *
+ * @return 0 on success, non-zero on failure.
+ */
+void da1469x_flash_setup(const struct hal_flash *dev, uint32_t bursta,
+                         uint32_t burstb, uint8_t *cmds, int len);
```
kasjer: it would not hurt to have cmds const

hw/mcu/dialog/da1469x/src/hal_flash.c
```diff
@@ -307,6 +307,28 @@ da1469x_qspi_erase_sector(const struct hal_flash *dev, uint32_t sector_address)
     return 0;
 }
 
+void
+da1469x_flash_setup(const struct hal_flash *dev, uint32_t bursta,
+                    uint32_t burstb, uint8_t *cmds, int len)
+{
+    int i;
+
+    da1469x_qspi_mode_manual(NULL);
```
kasjer: while da1469x_qspi_mode_manual does not use it's argument for now, it is strange to pass NULL while dev that is in this function is not used.

hw/mcu/dialog/da1469x/include/mcu/da1469x_hal.h
```diff
@@ -68,6 +68,22 @@ struct da1469x_hal_spi_cfg {
...
 
+/**
+ * Allow QSPI and the connected flash to be setup.
+ * This allows proper operation when configation script is not
+ * executed/present yet.
+ *
+ * @param dev    Pointer to device
+ * @param bursta BURSTCMDA_REG register setting
+ * @param burstb BURSTCMDB_REG register setting
+ * @param cmds   Array of byte sized commands sent over SPI to flash
```
kasjer: technically speaking looking in the the code only one command is sent (with multiply bytes).

PR #1980, opened 2019-09-03
stm32 hal_timer small fixes

kasjer:
- Prescaler was not correctly handled resulting in slightly wrong time calculation
- using TIM2 (which can be 32 bits on some MCU) could not work correctly
- redundant code removed

### DIFF
--- a/hw/mcu/stm/stm32_common/src/hal_timer.c
+++ b/hw/mcu/stm/stm32_common/src/hal_timer.c
@@ -94,7 +94,7 @@ stm32_tmr_cbs(struct stm32_hal_tmr *tmr)
     }
     ht = TAILQ_FIRST(&tmr->sht_timers);
     if (ht) {
-        tmr->sht_regs->CCR1 = ht->expiry;
+        tmr->sht_regs->CCR1 = ht->expiry & 0xFFFFU;
     } else {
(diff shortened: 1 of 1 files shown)


hw/mcu/stm/stm32_common/src/hal_timer.c
```diff
@@ -94,9 +94,8 @@ stm32_tmr_cbs(struct stm32_hal_tmr *tmr)
     }
     ht = TAILQ_FIRST(&tmr->sht_timers);
     if (ht) {
-        tmr->sht_regs->CCR1 = ht->expiry;
+        tmr->sht_regs->CCR1 = ht->expiry & 0xFFFFU;
```
utzig: I wonder if this is required. The `CCR1` register is still 32-bits, which for some models (F1 only?) will just write expiry on the 16 lower bits and ignore high word.

kasjer: This is needed for timers that are 32 bits, because of what is in line 511
```init.Period = 0xffff;```
All timer regardless if they are 32 or 16 bits are limited by this line. Then for 32 bits counters writing CCR1 value from expiry without masking will result in values grater then 0xFFFF and compare will never find those values. So this is needed.
It would be nice to use 32 bits timers at their full scale, but then code would need to program 16 and 32 bit register differently.

PR #1984, opened 2019-09-04
da1469x: enable switching to rcx clock source

JustineKH:
Changing the code to configure for different low power clock sources. Originally, it was assumed that the XTAL32K would be on board. However, this is not necessarily true. The change here is to make switching to the XTAL32K configurable. Also adding a configurable setting to switch to the RCX if desired. @andrzej-kaczmarek @kasjer

### DIFF
--- a/hw/mcu/dialog/da1469x/src/da1469x_lpclk.c
+++ b/hw/mcu/dialog/da1469x/src/da1469x_lpclk.c
@@ -32,6 +32,7 @@ bool g_mcu_lpclk_available;
 
 static da1469x_lpclk_cb *g_da1469x_lpclk_cmac_cb;
 
+#if MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, XTAL32K)
 static void
 da1469x_lpclk_settle_tmr_cb(void *arg)
 {
@@ -43,6 +44,7 @@ da1469x_lpclk_settle_tmr_cb(void *arg)
         g_da1469x_lpclk_cmac_cb();
     }
 }
+#endif
 
 void
 da1469x_lpclk_register_cmac_cb(da1469x_lpclk_cb *cb)
@@ -57,11 +59,15 @@ da1469x_lpclk_register_cmac_cb(da1469x_lpclk_cb *cb)
 void
 da1469x_lpclk_init(void)
 {
+#if MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, XTAL32K)
     static struct hal_timer lpclk_settle_tmr;
-
     da1469x_clock_lp_xtal32k_enable();
-
     os_cputime_timer_init(&lpclk_settle_tmr, da1469x_lpclk_settle_tmr_cb, NULL);
     os_cputime_timer_relative(&lpclk_settle_tmr,
                               MYNEWT_VAL(MCU_CLOCK_XTAL32K_SETTLE_TIME_MS) * 1000);
+#endif
+#if MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, RCX)
+    da1469x_clock_lp_rcx_enable();
+    da1469x_clock_lp_rcx_switch();
+#endif
 }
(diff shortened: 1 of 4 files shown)


hw/mcu/dialog/da1469x/syscfg.yml
```diff
@@ -45,6 +45,19 @@ syscfg.defs:
...
         value: 2000
 
+    MCU_CLOCK_RCX_SWITCH:
+        description: Enable if switching to RCX
+        value: 0
+
+    MCU_CLOCK_RCX_SETTLE_TIME_MS:
+        description: >
+          Specifies settling time [ms] of RCX.
+        value: 1000
```
wes3: Is the settling time of this clock really 1 second or is it 1ms? This could be a copy and paste error from the 32KHz xtal code (as that says 8000 ms for settling time and 8 seconds seems very long).

hw/mcu/dialog/da1469x/src/da1469x_lpclk.c
```diff
@@ -57,11 +63,18 @@ da1469x_lpclk_register_cmac_cb(da1469x_lpclk_cb *cb)
...
     os_cputime_timer_relative(&lpclk_settle_tmr,
                               MYNEWT_VAL(MCU_CLOCK_XTAL32K_SETTLE_TIME_MS) * 1000);
+#endif
+#if MYNEWT_VAL(MCU_CLOCK_RCX_SWITCH)
+    static struct hal_timer lpclk_settle_tmr;
+    da1469x_clock_lp_rcx_enable();
+    os_cputime_timer_init(&lpclk_settle_tmr, da1469x_lpclk_settle_tmr_cb, NULL);
+    os_cputime_timer_relative(&lpclk_settle_tmr,
+                              MYNEWT_VAL(MCU_CLOCK_RCX_SETTLE_TIME_MS) * 1000);
+#endif
```
wes3: I presume that either MCU_CLOCK_RCX_SWITCH or MCU_CLOCK_XTAL32K should be defined. If so, I think there should be some error if both are set to 0. Not sure if there is somewhere but if not it should be added.

hw/mcu/dialog/da1469x/syscfg.yml
```diff
@@ -45,6 +45,19 @@ syscfg.defs:
             Specifies settling time [us] of 32M crystal oscillator.
         value: 2000
 
+    MCU_CLOCK_RCX_SWITCH:
+        description: Enable if switching to RCX
+        value: 0
```
andrzej-kaczmarek: we should have a single syscfg to select lpclk source - the same as in nrf5x

hw/mcu/dialog/da1469x/src/da1469x_lpclk.c
```diff
@@ -57,11 +64,18 @@ da1469x_lpclk_register_cmac_cb(da1469x_lpclk_cb *cb)
...
     os_cputime_timer_init(&lpclk_settle_tmr, da1469x_lpclk_settle_tmr_cb, NULL);
     os_cputime_timer_relative(&lpclk_settle_tmr,
                               MYNEWT_VAL(MCU_CLOCK_XTAL32K_SETTLE_TIME_MS) * 1000);
+#endif
+#if MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, RCX)
+    static struct hal_timer lpclk_settle_tmr;
+    da1469x_clock_lp_rcx_enable();
+    os_cputime_timer_init(&lpclk_settle_tmr, da1469x_lpclk_settle_tmr_cb, NULL);
+    os_cputime_timer_relative(&lpclk_settle_tmr,
+                              MYNEWT_VAL(MCU_CLOCK_RCX_SETTLE_TIME_MS) * 1000);
```
andrzej-kaczmarek: I'd just switch to RCX immediately after XTAL32M is settled - we need then to add RCX calibration code, but more changes are required since RCX has different frequency than XTAL32K and our lpm code is not ready for this so RCX will not be reliable anyway, need to see what changes are required

PR #1985, opened 2019-09-04
Initialize QSPI_BURSTCMDx_REG registers in hal_flash_init()

benmccrea:
The QSPI_BURSTCMDA_REG and  QSPI_BURSTCMDB_REG registers must be configured with specific values for the target flash type. The da1469x will attempt to configure these registers from OTP or from a header prepended to the flash boot image. In the absence of either of these, the register values must be configured by initialization code (typically this would be code loaded to RAM). 

This PR adds syscfgs QSPI_BURSTCMDA_INIT_VAL and QSPI_BURSTCMDB_INIT_VAL which
can be set to enable initialization of the corresponding registers by hal_flash_init(). When the syscfg values are set to the default value, -1, the register initialization will not be performed.   

For background see Sec. 6.7 BOOTING in the DA1469x datasheet.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_flash.c
+++ b/hw/mcu/dialog/da1469x/src/hal_flash.c
@@ -382,5 +382,11 @@ da1469x_hff_sector_info(const struct hal_flash *dev, int idx,
 static int
 da1469x_hff_init(const struct hal_flash *dev)
 {
+#if MYNEWT_VAL(MCU_QSPIC_BURSTCMDA_INIT_VAL) >= 0
+    QSPIC->QSPIC_BURSTCMDA_REG = MYNEWT_VAL(MCU_QSPIC_BURSTCMDA_INIT_VAL);
+#endif
+#if MYNEWT_VAL(MCU_QSPIC_BURSTCMDB_INIT_VAL) >= 0
+    QSPIC->QSPIC_BURSTCMDB_REG = MYNEWT_VAL(MCU_QSPIC_BURSTCMDB_INIT_VAL);
+#endif
     return 0;
 }
--- a/hw/mcu/dialog/da1469x/syscfg.yml
+++ b/hw/mcu/dialog/da1469x/syscfg.yml
@@ -31,7 +31,7 @@ syscfg.defs:
         description: >
             Specifies whether or not to enable SIMO DC-DC Converter.
         value: 0
-    
+
     MCU_GPIO_RETAINABLE_NUM:
         description: >
             Number of GPIO pins which can have its state retained in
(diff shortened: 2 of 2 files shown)


hw/mcu/dialog/da1469x/syscfg.yml
```diff
@@ -113,6 +113,16 @@ syscfg.defs:
             data before writing, i.e. each such transfer will be split into
             chunks of this size. Buffer is created on stack.
         value: 128
+    QSPI_BURSTCMDA_INIT_VAL:
```
andrzej-kaczmarek: since this is specific to this particular MCU family amd not QSPI in general I'd call these settings e.g. `MCU_QSPIC_BURSTCMDA...`

hw/mcu/dialog/da1469x/syscfg.yml
```diff
@@ -113,6 +113,16 @@ syscfg.defs:
...
+        description: >
+            Initialization value for QSPI_BURSTCMDA_REG. The default value
+            is compatible with Macronix flash devices.
```
kasjer: In fact this is compatible with range of devices including Macronix, Winbond, and Gigadevice.
Macronix really wants to have A5 (10100101) while Winbond and Gigadevice want to have xx10xxxx pattern that is included in A5.

hw/mcu/dialog/da1469x/src/hal_flash.c
```diff
@@ -382,5 +382,7 @@ da1469x_hff_sector_info(const struct hal_flash *dev, int idx,
 static int
 da1469x_hff_init(const struct hal_flash *dev)
 {
+    QSPIC->QSPIC_BURSTCMDA_REG = MYNEWT_VAL(QSPI_BURSTCMDA_INIT_VAL);
```
kasjer: there is potentially danger of changing those two values if code is already running from QSPI.
If for example for some reason it was running in single or dual mode, and function wants to change it to quad mode (or other way round) any interrupt between those two instructions could lead to crash. First register changes instruction that should be used while second set's up other vital parameters.
To be on the safe side code should be running from RAM with interrupts disabled.

hw/mcu/dialog/da1469x/syscfg.yml
benmccrea: Done.

hw/mcu/dialog/da1469x/src/hal_flash.c
benmccrea: Ok, those are good points. It sounds like this really should be done only from flash_loader running in RAM, which is the only place right now where this is an issue. I'll close this and open a new PR.

benmccrea: On second thought, I'll close this PR and open another.

benmccrea:
Closing. This fix should probably be done in the flash_loader app, which typically is running from RAM.

benmccrea:
Re-opening. The fix now applies only when RAM_RESIDENT is 1 and interrupts are disabled during the register updated.

benmccrea:
@kasjer I've been testing this with settings to use single mode (0x3, 0x0 as you suggested) and this appears to be a good solution, since we need to ensure flash_loader won't fail (including verify) when programming flash for the first time.

Are you OK to approve and merge this PR?

benmccrea:
It looks like the repo settings are preventing me from reopening because I did a force-push on the branch. I'll try opening a new PR.

PR #1986, opened 2019-09-04
Fix include issue in da1469x_charger driver

benmccrea:
The include of DA1469xAB.h results in a file not found when da1469x alternate mcu types are used. The issue can be resolved by including mcu.h instead of DA1469xAB.h.


andrzej-kaczmarek:
this was already fixed in https://github.com/apache/mynewt-core/pull/1970

benmccrea:
Fix is already added.

PR #1992, opened 2019-09-07
hw/mcu/dialog: syscfgs for QSPIC_BURSTCMDA_REG, QSPIC_BURSTCMDB_REG

benmccrea:
The QSPI_BURSTCMDA_REG and QSPI_BURSTCMDB_REG registers must be configured with specific values for the target flash type. The da1469x will attempt to configure these registers from OTP or from a header prepended to the flash boot image. In the absence of either of these, these register values should be configurable by initialization code. Typically this would be code like apps/flash_loader which runs using RAM_RESIDENT and requires flash access. 

This PR adds syscfgs MCU_QSPI_BURSTCMDA_INIT_VAL and MCU_QSPI_BURSTCMDB_INIT_VAL which can be used to optionally initialize the corresponding registers in hal_flash_init(). The register initialization is done only if the syscfg value is non-empty. 

Restrictions: RAM_RESIDENT must be 1. 

Example (configure QSPIC to single mode):
MCU_QSPIC_BURSTCMDA_INIT_VAL: 0x3
MCU_QSPIC_BURSTCMDB_INIT_VAL: 0x0

For background see Sec. 6.7 BOOTING in the DA1469x datasheet.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_flash.c
+++ b/hw/mcu/dialog/da1469x/src/hal_flash.c
@@ -379,8 +379,36 @@ da1469x_hff_sector_info(const struct hal_flash *dev, int idx,
     return 0;
 }
 
-static int
-da1469x_hff_init(const struct hal_flash *dev)
+#if MYNEWT_VAL(MCU_QSPIC_APP_CFG)
+static sec_text_ram_core void
+da1469x_hff_mcu_custom_init(const struct hal_flash *dev)
 {
+    uint32_t primask;
+    uint32_t ctrlmode_reg = QSPIC->QSPIC_CTRLMODE_REG;
+    __HAL_DISABLE_INTERRUPTS(primask);
+    da1469x_qspi_mode_manual(dev);
+#if defined (MYNEWT_VAL_MCU_QSPIC_BURSTCMDA_INIT_VAL)
+    QSPIC->QSPIC_BURSTCMDA_REG = MYNEWT_VAL(MCU_QSPIC_BURSTCMDA_INIT_VAL);
+#endif
+#if defined (MYNEWT_VAL_MCU_QSPIC_BURSTCMDB_INIT_VAL)
+    QSPIC->QSPIC_BURSTCMDB_REG = MYNEWT_VAL(MCU_QSPIC_BURSTCMDB_INIT_VAL);
+#endif
+#if defined (MYNEWT_VAL_MCU_QSPIC_CTRLMODE_INIT_VAL)
+    QSPIC->QSPIC_CTRLMODE_REG = MYNEWT_VAL(MCU_QSPIC_CTRLMODE_INIT_VAL);
+#endif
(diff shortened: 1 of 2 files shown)


benmccrea:
@kasjer @andrzej-kaczmarek Opening this to replace #1985 (there was an issue due to force-push on the branch that wouldn't let me re-open the same PR). I removed disabling interrupts since this gets called from hal_flash_init(), so presumably flash is not in use yet. Also, it's only usable from RAM_RESIDENT applications so there should be no concern about corrupting cache operations.

benmccrea:
@agross The QSPIC_BURSTCMDx registers are kind of complex and there might be other situations that arise in the future where a RAM resident tool needs to modify the burst operation before running. Maybe we can make these defines later after things are settled but right now I would prefer to have the ability to override these registers so that flash_loader and other RAM resident tools can get the QSPI into the necessary state to do their operations.

agross:
@benmccrea Wrong agross 😉

benmccrea:
@kasjer Do you think there is a better way to incorporate this fix? In our testing we're still finding that we need QSPIC_BURSTCMDA_REG set to 0x3 which sets QSPIC_INST (Instruction Value for Incremental Burst or Single read access) to 03H. I do think that da1469x_hff_init() should place QSPI in a mode that works for all situations and this currently isn't the case. It also needs to avoid setting QSPIC_BURSTCMDx_REGs if they have already been set by the ROM bootloader using the OTP or product header values.

benmccrea:
@andrzej-kaczmarek We've been using this code for a while but I'm concerned it might not be the best idea to change these registers after QSPI has already come up. The hardware has a state machine for initialization and I can't find anything in the datasheet about how to force a re-initialization. There is a QSPI_INIT bit in SYS_CTRL_REG, "Enables QSPI initialization after wakeup" but I think that only has to do with re-init after flash wakes from power-down. 

These concerns of mine stem from the fact that we are still experiencing problems that might be related to QSPI operation. Do you have any thoughts about whether changing these registers from a RAM app during bsp init could cause issues?
--

hw/mcu/dialog/da1469x/src/hal_flash.c
```diff
@@ -382,5 +382,11 @@ da1469x_hff_sector_info(const struct hal_flash *dev, int idx,
 static int
 da1469x_hff_init(const struct hal_flash *dev)
 {
+#if defined (MYNEWT_VAL_MCU_QSPIC_BURSTCMDA_INIT_VAL)
```
mkiiskila: Maybe move these settings to always execute from RAM? I.e. within another function with section attribute macro 'sec_text_ram_core' defined.

benmccrea: @mkiiskila Took your suggestion. Also, added the ability to configure MCU_QSPIC_CTRLMODE_REG and re-added disabling interrupts. 

After more testing I'm confident this fix is safe.

benmccrea:
@kasjer @andrzej-kaczmarek Would one of you mind merging this? I was granted ASF commit access a while ago but for some reason I still don't have write access on mynewt-core.

PR #1993, opened 2019-09-07
OTP drivers and scripts

agross-korg:
Rebased this to the current master in use

### DIFF
--- a/hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
+++ b/hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
@@ -1,0 +1,595 @@
+#!/usr/bin/env python3
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+import serial
+import io
+import click
+import struct
+import binascii
+from typing import NamedTuple
+import os
+from cryptography.hazmat.primitives import serialization
+
+# Add in path to mcuboot.  This is required to pull in keys module from imgtool
+import sys
+sys.path.append(os.path.join(os.getcwd(), "repos", "mcuboot", "scripts",
+                "imgtool"))
+import keys as keys
+
+
+class Cmd(object):
+    OTP_READ_KEY = 0
+    OTP_WRITE_KEY = 1
+    FLASH_READ = 2
+    FLASH_WRITE = 3
+    OTP_READ_CONFIG = 4
+    OTP_APPEND_VALUE = 5
+    OTP_INIT = 6
+    FLASH_ERASE = 7
+
+
+class cmd_no_payload(NamedTuple):
+    som: int
+    cmd: int
+
+
+class cmd_read_key(NamedTuple):
+    som: int
+    cmd: int
+    segment: int
+    slot: int
+
+
+class cmd_write_key(NamedTuple):
+    som: int
+    cmd: int
+    segment: int
+    slot: int
+
+
+class cmd_append_value(NamedTuple):
+    som: int
+    cmd: int
+    length: int
+
+
+class cmd_response(NamedTuple):
+    som: int
+    cmd: int
+    status: int
+    length: int
+
+
+class cmd_flash_op(NamedTuple):
+    som: int
+    cmd: int
+    index: int
+    offset: int
+    length: int
+
+
+def crc16(data: bytearray, offset, length):
+    if data is None or offset < 0:
+        return
+
+    if offset > len(data) - 1 or offset + length > len(data):
+        return 0
+
+    crc = 0
+    for i in range(length):
+        crc ^= data[offset + i] << 8
+
+        for _ in range(8):
+            if (crc & 0x8000) > 0:
+                crc = (crc << 1) ^ 0x1021
+            else:
+                crc = crc << 1
+    return crc & 0xFFFF
+
+
+def validate_slot_index(ctx, param, value):
+    if value in range(8):
+        return value
+    else:
+        raise click.BadParameter("Slot value has to be between 0 and 7")
+
+
+def read_exact(device, length):
+    data = device.read(length)
+    if len(data) != length:
+        raise SystemExit("Failed to receive expected response, exiting")
+    return data
+
+
+@click.option('-i', '--index', required=True, type=int, help='key slot index',
+              callback=validate_slot_index)
+@click.option('-u', '--uart', required=True, help='uart port')
+@click.option('-s', '--segment', type=click.Choice(['signature', 'data',
+              'qspi']), help='OTP segment', required=True)
+@click.command(help='Read a single key from OTP key segment')
+def otp_read_key(index, segment, uart):
+    seg_map = {'signature': 0, 'data': 1, 'qspi': 2}
+
+    try:
+        ser = serial.Serial(port=uart, baudrate=1000000, timeout=1,
+                            bytesize=8, stopbits=serial.STOPBITS_ONE)
+    except serial.SerialException:
+        raise SystemExit("Failed to open serial port")
+
+    cmd = cmd_read_key(0xaa55aa55, Cmd.OTP_READ_KEY, seg_map[segment], index)
+    data = struct.pack('IIII', *cmd)
+
+    try:
+        ser.write(data)
+    except serial.SerialException:
+        raise SystemExit("Failed to write to %s" % uart)
+
+    data = read_exact(ser, 16)
+
+    response = cmd_response._make(struct.unpack_from('IIII', data))
+    if response.status == 0:
+        key = ser.read(32)
+        print("key:" + key.hex())
+    else:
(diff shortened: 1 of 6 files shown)


hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -0,0 +1,538 @@
...
+import io
+import click
+import struct
+import binascii
+from typing import NamedTuple
+from enum import Enum
+import os
+
+
+class Cmd(object):
```
utzig: I think `Cmd` should be a subclass of `Enum`, otherwise you might remove the `Enum` inclusion above.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -0,0 +1,538 @@
...
+        return
+
+    if offset > len(data) - 1 and offset + length > len(data):
+        return 0
+
+    crc = 0
+    for i in range(length):
+        crc ^= data[offset + i] << 8
+
+        for j in range(8):
```
utzig: This can be written `for _ in range(8)`

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -0,0 +1,538 @@
...
+    cmd: int
+    address: int
+    length: int
+
+
+def crc16(data: bytearray, offset, length):
+    if data is None or offset < 0:
+        return
+
+    if offset > len(data) - 1 and offset + length > len(data):
```
utzig: Should not this be an `or`?

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -0,0 +1,538 @@
...
+              'qspi']), help='OTP segment', required=True,)
+@click.command(help='Write a single key to OTP key segment')
+def otp_write_key(infile, index, segment, uart):
+
+    key = bytearray()
+    try:
+        with open(infile, "rb") as f:
+            buf = f.read(32)
+            key = struct.unpack('IIIIIIII', buf)
+    except IOError:
```
utzig: I believe something could go wrong here if you correctly open and read a file that is smaller than 32 bytes (`read` won't fail when there is less than 32), but `struct.unpack` is going to raise an exception that is not derived from `IOError`. One option would be to do a catch all (`except Exception`), or handle specific extra exceptions (`struct.error`).

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -0,0 +1,538 @@
...
+
+        response = cmd_response._make(struct.unpack_from('IIII', data))
+        if response.status == 0:
+            data = ser.read(response.length)
+            if len(data) != response.length:
+                raise SystemExit("Failed to receive response, exiting")
+            f.write(data)
+
+        else:
+            SystemExit("Error in read response, exiting")
```
utzig: Isn't this suppose to `raise` the exception? Also the `break` in the next line can be removed.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -0,0 +1,538 @@
...
+@click.option('-i', '--index', type=int, required=True,
+              help='Trim value id')
+@click.option('-u', '--uart', required=True, help='uart port')
+@click.command(help='Append trim value to OTP configuration script')
+def otp_append_trim(uart, trim, index):
+
+    data = []
+    data.append(0x90000000 + (len(trim) << 8) + index)
+    for entry in trim:
+        data.append(int(entry, 16))
```
utzig: Although this does not need to be efficient, one way to avoid the reallocation while expanding the list would be:

```
    data = [0 * (len(trim) + 1)]
    data[0] = 0x90000000 + (len(trim) << 8) + index
    for i, entry in enumerate(trim):
        data[i+1] = int(entry, 16)
```

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -0,0 +1,538 @@
...
+    msg = struct.pack('III', *cmd)
+
+    try:
+        ser.write(msg)
+    except serial.SerialException:
+        raise SystemExit("Failed to write to %s" % uart)
+
+    data = ser.read(16)
+    if len(data) != 16:
+        raise SystemExit("Failed to receive response, exiting")
```
utzig: Since this pattern is copied all over, one option would be to add a helper function which reads from the handle and raises the error, then you could just use something like `data = exact_read(handle, size)` and leave the size check + raise inside that helper function.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
agross-korg: good point, let me handle that case.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
agross-korg: oops, yes.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
agross-korg: fixed

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
agross-korg: fixed.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
agross-korg: at some point i also need to handle pem vs base64 encoded files.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
agross-korg: done

andrzej-kaczmarek:
@agross-korg could you please fix this PR to include patches related to OTP only? also if you agree with some/all of my comments, please update accordingly. I'd like to have this merged soon since https://github.com/apache/mynewt-core/pull/1996 depends on this code.

hw/mcu/dialog/da1469x/include/mcu/da1469x_otp.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define OTPM_BASE 0x10080000UL
+#define OTPM_SIZE 0x4096
```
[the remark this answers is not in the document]
agross-korg: This means splitting off this change into another repo at this point.  I was hoping to keep this self contained.

hw/mcu/dialog/da1469x/src/da1469x_otp.c
```diff
@@ -0,0 +1,148 @@
...
+    CRG_TOP->CLK_AMBA_REG |= CRG_TOP_CLK_AMBA_REG_OTP_ENABLE_Msk;
+
+    da1469x_otp_set_mode(OTPC_MODE_STBY);
+
+    /* set clk timing */
+    OTPC->OTPC_TIM1_REG = 0x0999101f;  /* 32 MHz default */
+    OTPC->OTPC_TIM2_REG = 0xa4040409;
+
+    /* Disable OTP clock */
+    CRG_TOP->CLK_AMBA_REG &= ~CRG_TOP_CLK_AMBA_REG_OTP_ENABLE_Msk;
```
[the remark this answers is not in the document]
agross-korg: done

hw/mcu/dialog/da1469x/include/mcu/da1469x_otp.h
```diff
@@ -0,0 +1,49 @@
...
+
+#define OTP_INVALID_SIZE_ALIGNMENT -1
+#define OTP_INVALID_ADDRESS -2
+#define OTP_PROGRAM_VERIFY_FAILED -3
```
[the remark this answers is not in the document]
agross-korg: done

hw/mcu/dialog/da1469x/src/da1469x_otp.c
```diff
@@ -0,0 +1,148 @@
...
+    uint32_t *src_addr = (uint32_t *)address;
+    uint32_t *dst_addr = dst;
+
+    if (address < OTPM_BASE || (address + num_bytes) > (OTPM_BASE + OTPM_SIZE)) {
+       return OTP_INVALID_ADDRESS;
+    }
+
+    if (num_bytes & 0x3) {
+       return OTP_INVALID_SIZE_ALIGNMENT;
+    }
```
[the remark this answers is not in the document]
agross-korg: i changed to offset, but lets leave the alignment check in place.

hw/mcu/dialog/da1469x/src/da1469x_otp.c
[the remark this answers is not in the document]
agross-korg: done

hw/mcu/dialog/da1469x/include/mcu/da1469x_otp.h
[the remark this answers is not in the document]
agross-korg: So I ended up moving these to the driver file itself.  The rationale is that the mcu.h currently resides in another repo and since only the driver now uses the definition, let's leave it there.  Ideally this would have been part of the CMSIS files provided by the manufacturer, but it is not.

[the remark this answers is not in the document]
andrzej-kaczmarek: this is common code for da1469x family, you just need to update `mcu.h` for `da14699` - 3rd party repos need to be adjusted to core, not the other way around...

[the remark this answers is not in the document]
agross-korg: or maybe it does.  not sure why this didnt work as I did try this yesterday but it worked today.   It might have been because I had the target set to mist vs dk.

PR #1996, opened 2019-09-12
hw/mcu/da1469x: Add support to use trim values from OTP

andrzej-kaczmarek:
We'll need to do something with these values later. This requires OTP driver from https://github.com/apache/mynewt-core/pull/1993

### DIFF
--- a/hw/mcu/dialog/da1469x/src/da1469x_trimv.c
+++ b/hw/mcu/dialog/da1469x/src/da1469x_trimv.c
@@ -1,0 +1,147 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


hw/mcu/dialog/da1469x/src/da1469x_trimv.c
```diff
@@ -0,0 +1,128 @@
...
+#define CS_WORD_TYPE_MASK       0xf0000000
+#define CS_WORD_TYPE_BOOTER     0x60000000
+#define CS_WORD_TYPE_TRIM       0x90000000
+
+struct da1469x_trimv_group {
+    uint8_t start;
+    uint8_t length;
+};
+
+static uint32_t g_mcu_trimv[TRIMV_COUNT_MAX];
```
mkiiskila: I don't know how often trim values are needed, but this seems like a lot of RAM to use for duplicating data which is readily readable from OTP.
I would rather just store pointers to OTP in g_mcu_trimv_groups, and call otp_read() when da1469x_trimv_get_group() is getting called.

andrzej-kaczmarek: that's a good idea, will fix. if code needs to add some trim values frequently, it can load these values to RAM - no need to load all values which we may not need at all.

vrahane:
There are conflicts, can you please resolve those and then we can merge this in. Looks like Marko's comments were addressed.

benmccrea:
LGTM. We tested with devices in the factory and the trim values are applied.

PR #2000, opened 2019-09-14
mcu/dialog: define adc pin selections in mcu.h

JustineKH:
Moved da14699 specific adc gpio pin selection out of da1469x_gpadc.c and da1469x_sdadc.c. Defining these in da14699 mcu.h for portability. @kasjer

### DIFF
--- a/hw/mcu/dialog/da14699/include/mcu/mcu.h
+++ b/hw/mcu/dialog/da14699/include/mcu/mcu.h
@@ -119,6 +119,28 @@ typedef enum {
 #define MCU_GPIO_PORT1(pin)		((1 * 32) + (pin))
 #define MCU_DMA_CHAN_MAX                    8
 
+#define MCU_PIN_GPADC_SEL0               MCU_GPIO_PORT1(9)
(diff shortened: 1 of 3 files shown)


hw/mcu/dialog/da14699/include/mcu/mcu.h
```diff
@@ -119,6 +119,28 @@ typedef enum {
 #define MCU_GPIO_PORT1(pin)		((1 * 32) + (pin))
 #define MCU_DMA_CHAN_MAX                    8
 
+#define MCU_GP_ADC_SEL0               MCU_GPIO_PORT1(9)
```
andrzej-kaczmarek: I'd use common prefix like `MCU_PIN` for all definitions, it will be easier to manage this in case we add more pins. also use `GPADC` and `SDADC` since this is how drivers are named. so this one would be `MCU_PIN_GPADC_SEL0`

benmccrea: @JustineKH I agree with andrzej about the naming. If you will make those changes and reflect the same changes to mynewt_runtime #69 then I think we should get this merged ASAP.

PR #2007, opened 2019-09-24
sensors/lis2dh12: Allow disabling one interface

kasjer:
In real life scenario accelerometer will be soldered to the board
and only one interface will be used to access it.
Current code always has built-in function for I2C and SPI due to
runtime checking of interface type.
This change adds syscfg values to enable interface in build
(default both are enabled for old time sake).

This allows to reduce image size.


vrahane:
I like the change but there was a reason for this. We had decided to keep both since all the MCU peripheral hals supported both simultaneously. Is that changing as well ? @wes3, @ccollins476ad and @mkiiskila, what do you think of the change ?

kasjer:
> I like the change but there was a reason for this. We had decided to keep both since all the MCU peripheral hals supported both simultaneously. Is that changing as well ? @wes3, @ccollins476ad and @mkiiskila, what do you think of the change ?

hifive1 board does not have i2c, just spi.

vrahane:
Ok, it totally makes sense then. I did not know that. Thanks.

PR #2008, opened 2019-09-24
drivers/sensors/lps33thw: Fixing packed struct pointer warning on gcc 9.

mlaz:


mlaz:
`Looks fine to me. More stack space but well, we got to stay up-to-date :-)`
Well, lps33hw is already doing this. We should probably review the other ones.

PR #2011, opened 2019-09-25
Change Reboot log from text to CBOR encoded.

manish-sudo:


manish-sudo:
> Generally, it looks fine but I think you should also add a dependency for tinycbor in `pkg.yml` for the `reboot` package

ACK

sys/reboot/src/log_reboot.c
```diff
@@ -175,41 +184,89 @@ log_reboot_write(const struct log_reboot_info *info)
...
+    if (!(flags & IMAGE_F_NON_BOOTABLE)) {
+    	off += snprintf(buf + off, sizeof buf - off, "%s ", "bootable");
+	}
+    if (state_flags & IMGMGR_STATE_F_CONFIRMED) {
+    	off += snprintf(buf + off, sizeof buf - off, "%s ", "confirmed");
+    }
+    if (state_flags & IMGMGR_STATE_F_PENDING) {
+    	off += snprintf(buf + off, sizeof buf - off, "%s ", "pending");
+    }
+    buf[off - 1] = '\0';
```
kasjer: in case of no flags (I don't know if it can happen) buf[off - 1] = 0; would write outside buf.
That will be caught by coverity tool.

sys/reboot/src/log_reboot.c
```diff
@@ -175,41 +184,89 @@ log_reboot_write(const struct log_reboot_info *info)
...
-                (unsigned long)info->pc);
+        cbor_encode_text_stringz(&map, "pc");
+        cbor_encode_int(&map, (unsigned long)info->pc);
+    }
+
+    state_flags = imgmgr_state_flags(boot_current_slot);
+    cbor_encode_text_stringz(&map, "flags");
+    off = 0;
+    if (state_flags & IMGMGR_STATE_F_ACTIVE) {
+        off += snprintf(buf + off, sizeof buf - off, "%s ", "active");
```
kasjer: inconsistent sizeof usages some have () and some don't

sys/reboot/src/log_reboot.c
```diff
@@ -220,8 +220,16 @@ log_reboot_write(const struct log_reboot_info *info)
 
     if (info->file != NULL) {
         cbor_encode_text_stringz(&map, "die");
+        off  = 0;
+
+        /* If die filename is longer than 1/3 of total allocated
+         * buffer, then trim the filename from left. */
+        if (strlen(info->file) > (sizeof(buf) / 3))
+        {
```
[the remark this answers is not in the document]
manish-sudo: ACK

sys/reboot/src/log_reboot.c
```diff
@@ -220,8 +220,16 @@ log_reboot_write(const struct log_reboot_info *info)
...
+            off = strlen(info->file) - (sizeof(buf)/3);
```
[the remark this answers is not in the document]
manish-sudo: ACK

sys/reboot/src/log_reboot.c
```diff
@@ -228,6 +230,24 @@ log_reboot_write(const struct log_reboot_info *info)
...
     }
 
+    state_flags = imgmgr_state_flags(boot_current_slot);
+    cbor_encode_text_stringz(&map, "flags");
+    off = 0;
+    if (state_flags & IMGMGR_STATE_F_ACTIVE) {
+        off += snprintf(buf + off, sizeof buf - off, "%s ", "active");
+    }
+    if (!(flags & IMAGE_F_NON_BOOTABLE)) {
+    	off += snprintf(buf + off, sizeof buf - off, "%s ", "bootable");
```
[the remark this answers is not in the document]
manish-sudo: ACK.
My IDE messed up with tabs.

sys/reboot/src/log_reboot.c
```diff
@@ -175,41 +184,89 @@ log_reboot_write(const struct log_reboot_info *info)
...
+    cbor_encode_text_stringz(&map, "rsn");
+    cbor_encode_text_stringz(&map, REBOOT_REASON_STR(info->reason));
+
+    cbor_encode_text_stringz(&map, "cnt");
+    cbor_encode_int(&map, reboot_cnt);
+
+    cbor_encode_text_stringz(&map, "img");
+    snprintf(buf, sizeof(buf), "%u.%u.%u.%u",
+                  ver.iv_major, ver.iv_minor, ver.iv_revision,
+                  (unsigned int) ver.iv_build_num);
```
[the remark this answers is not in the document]
manish-sudo: ACK

sys/reboot/src/log_reboot.c
```diff
@@ -175,41 +184,89 @@ log_reboot_write(const struct log_reboot_info *info)
...
+        snprintf(buf, sizeof(buf), "%s:%d",
+                &info->file[off], info->line);
+        cbor_encode_text_stringz(&map, buf);
     }
 
     if (info->pc != 0) {
-        off += snprintf(buf + off, sizeof buf - off, ", pc:0x%lx",
-                (unsigned long)info->pc);
+        cbor_encode_text_stringz(&map, "pc");
+        cbor_encode_int(&map, (unsigned long)info->pc);
```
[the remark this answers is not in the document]
manish-sudo: Didn't quite get it, care to elaborate little?

sys/reboot/src/log_reboot.c
manish-sudo: I thought of adding a protection earlier, but didn't thought it's absolutely required as technically there will be at-least one flag. Anyway, for the sanity I guarded this unlikely case.

sys/reboot/src/log_reboot.c
manish-sudo: So, the code I wrote had () and legacy code didn't. I was in dilemma again if I should parenthesis the existing sizeof.
Anyway, to address your point, I have removed parenthesis from sizeofs which I introduced to be consistent. pushing my changes in.

sys/reboot/src/log_reboot.c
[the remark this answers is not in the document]
kasjer: my bad, it should be **cast** not case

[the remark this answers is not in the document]
manish-sudo: Ah I see, I just carried it over. Do you have strong opinion about it? if so, I can make a new commit to address this.

sys/reboot/src/log_reboot.c
```diff
@@ -175,41 +184,90 @@ log_reboot_write(const struct log_reboot_info *info)
...
+    if (!(flags & IMAGE_F_NON_BOOTABLE)) {
+        off += snprintf(buf + off, sizeof buf - off, "%s ", "bootable");
     }
+    if (state_flags & IMGMGR_STATE_F_CONFIRMED) {
+        off += snprintf(buf + off, sizeof buf - off, "%s ", "confirmed");
+    }
+    if (state_flags & IMGMGR_STATE_F_PENDING) {
+        off += snprintf(buf + off, sizeof buf - off, "%s ", "pending");
+    }
+    if (off > 1) {
```
kasjer: if off == 0 would next cbor_encode_text_stringz encoded old values from hash, die or pc again?

sys/reboot/src/log_reboot.c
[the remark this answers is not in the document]
kasjer: No it's fine

sys/reboot/src/log_reboot.c
manish-sudo: good catch! cleared the buf here and other places before using, to avoid end up using stale data.

sys/reboot/src/log_reboot.c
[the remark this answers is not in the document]
manish-sudo: Addressed it anyway as I pushed a new commit.

sys/reboot/src/log_reboot.c
```diff
@@ -175,41 +184,93 @@ log_reboot_write(const struct log_reboot_info *info)
...
+    cbor_encode_int(&map, reboot_cnt);
+
+    cbor_encode_text_stringz(&map, "img");
+    snprintf(buf, sizeof buf, "%u.%u.%u.%u",
+                  ver.iv_major, ver.iv_minor, ver.iv_revision,
+                  (unsigned int)ver.iv_build_num);
+    cbor_encode_text_stringz(&map, buf);
+
+    cbor_encode_text_stringz(&map, "hash");
+    memset(buf, 0, sizeof buf);
```
kasjer: this is not really needed snprintf will work fine

sys/reboot/src/log_reboot.c
```diff
@@ -175,41 +184,93 @@ log_reboot_write(const struct log_reboot_info *info)
...
+        cbor_encode_int(&map, info->pc);
     }
 
-    /* Make sure we don't log beyond the end of the source buffer. */
-    if (off > sizeof buf) {
-        off = sizeof buf;
+    state_flags = imgmgr_state_flags(boot_current_slot);
+    cbor_encode_text_stringz(&map, "flags");
+    off = 0;
+    memset(buf, 0, sizeof buf);
```
kasjer: memset seems like overkill where` buf[0] = '\0';` would do

sys/reboot/src/log_reboot.c
manish-sudo: Since we are talking lot of hypothetical things, what if hash is empty?

sys/reboot/src/log_reboot.c
manish-sudo: ACK

sys/reboot/src/log_reboot.c
kasjer: what goes into string is ascii representation of hex encoded bytes, sizeof hash is not 0, so buf is always filled with something, and it should be null terminated after every round

manish-sudo: Ahh I see, I overlooked what we are using here is sizeof for looping, then it makes sense to assume buf will be null terminated.

vrahane:
@manish-sudo Thanks, one thing I forgot to mention is for future PRs please mention the modules in the commit message. I should have mentioned this before approving but yeah, it's nice to have to go back and look at the history and know which module the commit is addressed for.

PR #2023, opened 2019-10-01
mcu/dialog/da1469x: configurability for explicitly disabling dcdc

JustineKH:
Explicitly disable DCDC converter.


andrzej-kaczmarek:
discussed offline, we'll just keep dcdc disable API for now

PR #2024, opened 2019-10-01
hw/mcu/dialog: Add helper to enable/disable AMBA clocks

wes3:
Add a helper to enable/disable AMBA clocks in CRG_TOP->CLK_AMBA_REG.
The code was doing an un-protected read-modify-write and if the clock is modified inside ISR's this could cause unpredictable behavior.

The helper will enable/disable interrupts around the setting/clearing of bits in that register. The helpers added are:
da1469x_clock_amba_enable(mask)
da1469x_clock_amba_disable(mask)

### DIFF
--- a/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
+++ b/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
@@ -21,6 +21,7 @@
 #define __MCU_DA1469X_CLOCK_H_
 
 #include <stdint.h>
+#include "mcu/da1469x_hal.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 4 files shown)


hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
```diff
@@ -94,6 +95,36 @@ uint32_t da1469x_clock_lp_rcx_freq_get(void);
  */
 void da1469x_clock_lp_rcx_disable(void);
 
+/**
+ * Enable an AMBA clock
+ *
+ * @param idx
```
kasjer: idx or mask?

wes3: mask :-)

hw/mcu/dialog/da1469x/include/mcu/da1469x_hal.h
```diff
@@ -21,6 +21,7 @@
 #define __MCU_DA1469X_HAL_H_
 
 #include <assert.h>
+#include "mcu/cmsis_nvic.h"
```
kasjer: cmsis_nvic.h seems to be unrelated to change

wes3: I guess you can say it is unrelated. Looks like cmsis_nvic.h brings in mcu.h which is needed for files that include da1469x_hal.h. I will change it to mcu.h

PR #2026, opened 2019-10-04
[nrfx] wrap config defines with ifndef

saulpw:


kasjer:
Nice addition.
Could you please add empty lines between `#ifndef/#endif` pairs to make it more like other nrf52 headers. It's not strictly necessary by rules that are followed here, but it would look more consistent across similar headers.
Thanks.

saulpw:
@kasjer whitespace added.

PR #2027, opened 2019-10-07
fs_mount: fix conditional macro.

mlaz:


mlaz:
this was mistakenly open

PR #2029, opened 2019-10-07
Added different scalars for ildo and for vbus/vled.

pshah111:
Added scalars for different channels for ildo and for vbus/vled.

### DIFF
--- a/hw/drivers/adc/gpadc_da1469x/src/gpadc_da1469x.c
+++ b/hw/drivers/adc/gpadc_da1469x/src/gpadc_da1469x.c
@@ -30,6 +30,10 @@
 
 static struct da1469x_gpadc_dev *da1469x_gpadc_dev;
 
+// 1/0.180265 * 1024 = 5680, 1/0.06907 * 1024 = 14825, 1/0.065067 * 1024 = 15737
+// 1/0.068603 * 1024 = 14926, 1/0.066215 * 1024 = 15464, 1/0.170288 * 1024 = 6013
+static uint16_t ildo_scalar[] = {5680,14825,15737,14926,15464,6013};
+
 static void
 da1469x_gpadc_resolve_pins(uint32_t ctrl, int *pin0p, int *pin1p)
 {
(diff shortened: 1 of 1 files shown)


hw/drivers/adc/gpadc_da1469x/src/gpadc_da1469x.c
```diff
@@ -30,6 +30,10 @@
 
 static struct da1469x_gpadc_dev *da1469x_gpadc_dev;
 
+// 1/0.180265 * 1024 = 5680, 1/0.06907 * 1024 = 14825, 1/0.065067 * 1024 = 15737
+// 1/0.068603 * 1024 = 14926, 1/0.066215 * 1024 = 15464, 1/0.170288 * 1024 = 6013
```
vrahane: Can you please use C style comments instead of C++ style comments as per our coding style doc.

hw/drivers/adc/gpadc_da1469x/src/gpadc_da1469x.c
```diff
@@ -30,6 +30,10 @@
...
+static uint16_t ildo_scalar[] = {5680,14825,15737,14926,15464,6013};
```
vrahane: You can make it a const array since it never gets modified, also, we generally have spaces after `,`s inside the array. That would get the line beyond 79 chars, so you should try to wrap it, that is again part of our coding style.

hw/drivers/adc/gpadc_da1469x/src/gpadc_da1469x.c
```diff
@@ -128,22 +132,41 @@ da1469x_gpadc_configure_channel(struct adc_dev *adev, uint8_t cnum,
 {
     uint32_t ctrl;
     uint32_t ctrl2;
+    uint32_t mux;
     uint16_t refmv;
 
     ctrl = GPADC->GP_ADC_CTRL_REG;
     ctrl2 = GPADC->GP_ADC_CTRL2_REG;
+    mux = GPADC->GP_ADC_MUX_REG;
```
andrzej-kaczmarek: there is no such register in DA1469x...

hw/drivers/adc/gpadc_da1469x/src/gpadc_da1469x.c
pshah111: I didn't make it const since each scalar value will be trimmed on a per unit basis and stored in OTP. So, they will be pulled during initialization.

hw/drivers/adc/gpadc_da1469x/src/gpadc_da1469x.c
pshah111: The definition is from mynewt_runtime/hw/mcu/dialog/mist/include/DA10101AA-ALL.h

PR #2030, opened 2019-10-08
console improvements

kasjer:
This allows to have editable prompt line visible at the bottom of the terminal window.
This can be useful when logs are flooding console and it would be hard to type
complicated commands.

Additionally:
- completion for `select` fixed
- fixed auto completion for complete word (space added when word was already full)
- CTRL-C clears input line
- CTRL-L resets console if something goes wrong
- HOME/END putty work now


sjanc:
awesome! thanks for adding soft cursor

PR #2035, opened 2019-10-11
[WIP] DA1469x UART rx pullup

wes3:
This code will allow a user to configure an internal pullup on the UART rx pin. It is expected that a BSP would set UART_X_RX_ENABLE_PULLUP if there is no external pullup on the line.

<XXX: will add more here once we test this and see if it works>

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_uart.c
+++ b/hw/mcu/dialog/da1469x/src/hal_uart.c
@@ -40,6 +40,11 @@ struct da1469x_uart {
     volatile uint8_t tx_started : 1;
     volatile uint8_t rx_data;
 
+    /* Stores rx pin func for use during open/close and if there is a pullup */
+    mcu_gpio_func rx_pin_func;
+    int8_t        rx_pin;
+    uint8_t       rx_pullup;
+
     hal_uart_rx_char rx_func;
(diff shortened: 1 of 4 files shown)


hw/mcu/dialog/da1469x/src/hal_uart.c
```diff
@@ -463,6 +477,34 @@ hal_uart_config(int port, int32_t baudrate, uint8_t databits, uint8_t stopbits,
     if (!baudrate_cfg) {
         return SYS_ENOTSUP;
     }
+
+    /* Enable pullup if configured */
+    if (uart->cfg->rx_pullup) {
+        mcu_gpio_set_pin_function(uart->cfg->pin_rx, MCU_GPIO_MODE_INPUT_PULLUP,
+                                  uart->rx_pin_func);
+    }
```
mkiiskila: Rest of the pin config happens in hal_uart_init(). Is there a reason pullup configuration is here instead?

wes3: Debated putting it in hal_uart_init() as opposed to config and close but since hal_uart_init is only called once (typically) decided to put in hal_uart_config(). This would allow an app to close the uart, repurpose that pin, and not have to call hal_uart_init again.

PR #2046, opened 2019-10-16
base64,stats: add base64 fixedlen decode,fix stats

vrahane:
- persistent stats were failing on a downgrade if more stats were added
- Fix is to read whatever stats we can based on the size of the buffer
  and only fail if the stats fail
- Adding API to base64 to decode by fixed length

### DIFF
--- a/encoding/base64/src/base64.c
+++ b/encoding/base64/src/base64.c
@@ -167,6 +167,29 @@ base64_decode(const char *str, void *data)
     return q - (unsigned char *) data;
 }
 
+int
+base64_decode_maxlen(const char *str, void *data, int len)
+{
+    const char *p;
+    unsigned char *q;
+
+    q = data;
+    for (p = str; *p && (*p == '=' || strchr(base64_chars, *p)); p += 4) {
+        if (q - (unsigned char *)data >= len) {
+            break;
+        }
+        unsigned int val = token_decode(p);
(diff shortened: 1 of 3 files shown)


encoding/base64/src/base64.c
```diff
@@ -167,6 +167,29 @@ base64_decode(const char *str, void *data)
...
 
+int
+base64_decode_maxlen(const char *str, void *data, int len)
+{
+    const char *p;
+    unsigned char *q;
+
+    q = data;
+    for (p = str; *p && (*p == '=' || strchr(base64_chars, *p)); p += 4) {
+        if (q - (unsigned char *)data >= len) {
```
kasjer: It looks like function can write to memory beyond data + len which does not seem to be intentional.

encoding/base64/src/base64.c
```diff
@@ -167,6 +167,29 @@ base64_decode(const char *str, void *data)
     return q - (unsigned char *) data;
 }
 
+int
+base64_decode_maxlen(const char *str, void *data, int len)
```
kasjer: While it is visible that function body was copied from base64_decode() style could be refreshed since new lines do adhere to latest standards.

vrahane:
@kasjer opened a new PR for your comments, thanks: https://github.com/apache/mynewt-core/pull/2048

PR #2056, opened 2019-10-18
[DO NOT MERGE] : dialog_dk1469x-dk-pro: Fix otp_write_key

nkaje:
otp_write_key reads a base64 encoded key. When encoded, the key
is longer than 32 bytes. struct.unpack expects it to be decoded
and 32 bytes long. Hence decode it and then use.

### DIFF
--- a/hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
+++ b/hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
@@ -25,6 +25,7 @@ import binascii
 from typing import NamedTuple
 import os
 from cryptography.hazmat.primitives import serialization
+import base64
 
 # Add in path to mcuboot.  This is required to pull in keys module from imgtool
 import sys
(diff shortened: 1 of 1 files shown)


hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -179,9 +180,9 @@ def otp_write_key(infile, index, segment, uart):
             else:
                 # read key from base64 encoded AES key file
                 buf = f.read()
-                if isBase64(buf):
-                    if len(buf) != 32:
-                        raise SystemExit("AES key file has incorrect length")
+                buf = base64.b64decode(buf)
```
nkaje: Now that isBase64() not used, could we remove it?

agross-korg: yes

agross-korg: else:
                # read key from base64 encoded AES key file
                buf = f.read()

                try:
                    buf = base64.decodestring(buf)
                except ValueError:
                    raise SystemExit("Invalid base 64 file")

                if len(buf) != 32:
                    raise SystemExit("AES key file has incorrect length")

agross-korg: this would catch errors without spewing to stderr

nkaje: Done and new PS uploaded.

nkaje:
These changes are being made in https://github.com/apache/mynewt-core/pull/2061. Do not merge 2056

Issue #1968, opened 2019-08-27
stats: `static STATS_NAME_START()` keeps static qualifier as is when statistics names are disabled 


vrahane:
stats: `static STATS_NAME_START()` keeps static qualifier as is when statistics names are disabled.

There are couple of ways of solving this:
1. Remove static qualifier for all stat names
2. Add static qualifier to the macro and deal differently with it based on whether names are enabled or not

The above are just suggestions, there might be better ways of solving this.

PR #2066, opened 2019-10-23
stats: Get rid of static qualifier for STATS_NAME_START

manish-sudo:


kasjer:
Obviously change does something more beyond what is mention in the commit message.

It does not explain reasons behind the changed. Does it fix something? Is it some kind of improvement?

manish-sudo:
@kasjer, here is the context:
https://github.com/apache/mynewt-core/issues/1968

Also, while digging in the code I found that some of the code is still using the old "apache-mynewt-core/mgmt/imgmgr" which now has been replaced by mcumgr "apache-mynewt-mcumgr/cmd/img_mgmt" which has been addressed as well.

utzig:
@manish-sudo Pointing to an issue number here in the PR does not mean someone looking at git history will be able to understand why this change was made. You have to write it down in the commit, either copying from the body of the issue itself, or any other descriptive text that does not rely on external information.

kasjer:
@manish-sudo package name fixes were somehow obvious (separate commit would be much batter though).
The other part (getting rid of static) is not clear, what was done was explained, why it was done, not so much.

manish-sudo:
> @manish-sudo package name fixes were somehow obvious (separate commit would be much batter though).
> The other part (getting rid of static) is not clear, what was done was explained, why it was done, not so much.

Okay, let me close this PR then, I will break it down into two and also add some context in the commit. Thanks for your inputs!

kasjer:
> > @manish-sudo package name fixes were somehow obvious (separate commit would be much batter though).
> > The other part (getting rid of static) is not clear, what was done was explained, why it was done, not so much.
> 
> Okay, let me close this PR then, I will break it down into two and also add some context in the commit. Thanks for your inputs!

Thanks for understanding and you contribution.

PR #2075, opened 2019-10-29
kernel/os: Add os_mbuf_pack_chains

wes3:
This commit adds a new API for mbuf handling called os_mbuf_pack_chains.
This API, given two mbuf chains, will concatenate them and copy data
as needed so that each mbuf fully utilizes the data buffer in each mbuf.
Unused mbufs are discarded. This API sacrifices speed for memory and
may cause a lot of copying to occur within the mbuf chains.

Test code has been written for a number of reasonable cases.

### DIFF
--- a/kernel/os/selftest/src/testcases/os_mbuf_test_pack_chains.c
+++ b/kernel/os/selftest/src/testcases/os_mbuf_test_pack_chains.c
@@ -127,7 +127,7 @@ os_mbuf_test_pack_chains_create_chain(int num_mbufs, struct os_mbtpc_cd *mdata,
  *      with the bsd function rand() that returns 0..(2**31)-1.
  */
 unsigned int
-os_mbuf_test_pack_chans_rand(void)
+os_mbuf_test_pack_chains_rand(void)
 {
 	rand_next = rand_next * 1103515245 + 12345;
 	return ((unsigned int)(rand_next / 65536) % 32768);
@@ -189,7 +189,7 @@ TEST_CASE_SELF(os_mbuf_test_pack_chains)
     /* Fill the test data with random data */
     for (rc = 0; rc < 2048; ++rc) {
         os_mbuf_test_pack_chains_test_data[rc] =
-            (uint8_t)os_mbuf_test_pack_chans_rand();
+            (uint8_t)os_mbuf_test_pack_chains_rand();
     }
 
     /*
--- a/kernel/os/selftest/src/testcases/os_mbuf_test_pack_chains.c
+++ b/kernel/os/selftest/src/testcases/os_mbuf_test_pack_chains.c
@@ -1,0 +1,371 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


kernel/os/src/os_mbuf.c
```diff
@@ -1036,3 +1036,90 @@ os_mbuf_widen(struct os_mbuf *om, uint16_t off, uint16_t len)
 
     return 0;
 }
+
+struct os_mbuf *
+os_mbuf_pack_chains(struct os_mbuf *m1, struct os_mbuf *m2)
+{
+    uint16_t rem_len;
+    uint16_t copylen;
```
kasjer: slight in-consequence in naming length rem_len has _ while copylen does not

kernel/os/src/os_mbuf.c
```diff
@@ -1036,3 +1036,90 @@ os_mbuf_widen(struct os_mbuf *om, uint16_t off, uint16_t len)
...
+
+        /* Get a pointer to the next buf we want to absorb */
+        next = SLIST_NEXT(cur, om_next);
+
+        /*
+         * Is there trailing space in the mbuf? If so, copy data from
+         * following mbufs into the current mbuf
+         */
+        rem_len = OS_MBUF_TRAILINGSPACE(cur);
+        while (rem_len && next) {
```
kasjer: in other cases length is compared to 0 and next to NULL while here is shorter version is used. Not that it matters.

kernel/os/selftest/src/testcases/os_mbuf_test_pack_chains.c
```diff
@@ -0,0 +1,371 @@
...
+/*
+ * This is here cause I dont feel like calling rand :-)
+ *
+ *      Taken from the K&R C programming language book. Page 46.
+ *      returns a pseudo-random integer of 0..32767. Note that
+ *      this is compatible with the System V function rand(), not
+ *      with the bsd function rand() that returns 0..(2**31)-1.
+ */
+unsigned int
+os_mbuf_test_pack_chans_rand(void)
```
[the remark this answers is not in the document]
wes3: I have no good reason why I did not want to call rand() :-) I just wanted it all self-contained. I did have a reason why I used random data instead of, say, a pattern that increments by 1. I was worried that there might be some weird test patterns where the data would be the same but incorrect (if you see what I mean). I think with using random data we are pretty sure that wont be the case. Did not want to hash things or add a crc or checksum.

kernel/os/src/os_mbuf.c
wes3: Yes; I am not terribly consistent. I just thought remlen looked weird whereas copylen looks nice :-)

kernel/os/selftest/src/testcases/os_mbuf_test_pack_chains.c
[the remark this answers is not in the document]
wes3: As to the API with just one mbuf passed in, well, I am not going to change it now. I think this is fine. You can use it with just a single chain (pass NULL for m2) and it does make a sort of sense when "re-assembling" things (i.e. you have an existing chain and want to add an mbuf or mbuf chain to it).

Issue #2079, opened 2019-11-01
reboot log entry is in CBOR but is printed on console


mkiiskila:
compat> reset
000000 [ts=0us, mod=6 level=4 ]?crsndSOFTccnt.cimgg5.0.0.0dhashx@065b72e0c0d9af335d67cface6f7
d9d217aaefddffddde672eb8dd2059721c66eflagsx.active bootable confirmed?
Slinky 5.0.0.0

We should probably disable reboot logging with slinky at reset, if the data in it is going to be CBOR.

PR #2081, opened 2019-11-06
Reboot log entry should not be in CBOR when printed on console

vrahane:
This PR also adds a package for a generic logging cbor reader utility which would be used by shell instead of having it's own reader: `sys/log/util/cbor_reader`

This fixes https://github.com/apache/mynewt-core/issues/2079

**Before**
```
compat> reset
000000 [ts=0us, mod=6 level=4 ]?crsndSOFTccnt.cimgg5.0.0.0dhashx@065b72e0c0d9af335d67cface6f7
d9d217aaefddffddde672eb8dd2059721c66eflagsx.active bootable confirmed?
Slinky 5.0.0.0
```
**After**
```
compat> reset
000000 [ts=0us, mod=6 level=4 ]{_ "rsn": "SOFT", "cnt": 267, "img": "0.0.0.0", "hash": "9225ff83589e177fc4b3e483d9a6249866bd65b56e5d87dc7a5b4331f3
e85def", "flags": "active bootable confirmed"}
Slinky 5.0.0.0
```


sys/log/full/src/log_console.c
```diff
@@ -24,9 +24,28 @@
...
 #include "log/log.h"
+#if MYNEWT_VAL(LOG_VERSION) > 2
+#include "tinycbor/cbor.h"
+#include "tinycbor/cbor_buf_reader.h"
+#endif
 
 static struct log log_console;
 
+static int
+console_log_dump_cbor_entry(struct log *log, const void *dptr, uint16_t len)
```
ccollins476ad: You could remove the first parameter to this function (`struct log *log`).

sys/log/util/pkg.yml
```diff
@@ -0,0 +1,34 @@
+#
```
ccollins476ad: Does `sys/log/util` itself need to be a package?  It doesn't seem to contain any code of its own.

sys/log/util/cbor_reader/include/cbor_reader.h
```diff
@@ -0,0 +1,33 @@
+/*
```
ccollins476ad: I think `log_cbor_reader` would be a better name for this directory than `cbor_reader`.  If I see
```
#include "cbor_reader.h"
```

It wouldn't occur to me that the file includes a log header.

Also, I think there should be an extra subdirectory under `include` to maintain consistency, i.e.,
```
 sys/log/util/cbor_reader/include/cbor_reader/...
```
```
#include "cbor_reader/cbor_reader.h"
```

ccollins476ad:
```
000000 [ts=0us, mod=6 level=4 ]{_ "rsn": "SOFT", "cnt": 267, "img": "0.0.0.0", "hash": "9225ff83589e177fc4b3e483d9a6249866bd65b56e5d87dc7a5b4331f3
e85def", "flags": "active bootable confirmed"}
```
I wonder where that underscore before `rsn` comes from.

sys/log/full/src/log_console.c
vrahane: done

sys/log/util/pkg.yml
vrahane: done

sys/log/util/cbor_reader/include/cbor_reader.h
vrahane: done

vrahane:
@ccollins476ad Thanks for the review, fixed the issues.

vrahane:
I spoke to @ccollins476ad but just for the record I'll also mention this here. `cbor_value_to_pretty()` prints an underscore `_` if the length of the value is unknown.

`encoding/tinycbor/src/cborpretty.c`
```
278         if (!cbor_value_is_length_known(it)) {
279             if (fprintf(out, "_ ") < 0)
280                 return CborErrorIO;
281         }
```

PR #2084, opened 2019-11-06
[WIP - DO NOT MERGE] encoding/tinycbor: Break up cbor header

vrahane:
- Breaking up cbor header file to cbor_decoder_reader.h
  and cbor_encoder_writer.h

@sterlinghughes @mlaz @carlescufi @de-nordic

I have just made the change as a place holder and tested it out but I don't think this is necessary for Zephyr in anyway. Can we ask Dominik to take a look. If this is not required, I would like to not make this change at all in mynewt-core.

### DIFF
--- a/encoding/tinycbor/include/tinycbor/cbor.h
+++ b/encoding/tinycbor/include/tinycbor/cbor.h
@@ -32,6 +32,9 @@
 #include <string.h>
 #include <stdio.h>
 
+#include "tinycbor/cbor_buf_writer.h"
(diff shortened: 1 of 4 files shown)


vrahane:
@de-nordic Can I close this PR ? I think this is not needed.

de-nordic:
Yes, you may close it. Thanks.

ccollins476ad:
It looks like everyone agrees this should be closed.

PR #2086, opened 2019-11-07
bsp/sensorhub: Fix build

kasjer:
This fixes build for sensorhub bsp.
epos/apache-mynewt-core/hw/mcu/stm/stm32f4xx/src/clock_stm32f4xx.c:158:2: error: #error "PLLN value is invalid"
 #error "PLLN value is invalid"

Values provided in syscfg are most likely not valid and should be
updated.


utzig:
I guess that maybe we should just delete this BSP; not sure anyone has this HW, and it does not seem to exist anymore...

kasjer:
> I guess that maybe we should just delete this BSP; not sure anyone has this HW, and it does not seem to exist anymore...

I agree, I can't verify what should be put in syscfg.

kasjer:
BSP was removed, no need to fix build any more,

PR #2087, opened 2019-11-07
fix build error for LOG_VERSION < 3

kasjer:
Code was not building for LOG_VERSION == 2
- LOG_ETYPE_CBOR was not defined but was used
- ue_etype was not defined but used

for log_reboot code from before cbor conversion was restored (with some modifications that allow it to build).

### DIFF
--- a/sys/log/full/src/log_console.c
+++ b/sys/log/full/src/log_console.c
@@ -79,13 +79,15 @@ log_console_append_body(struct log *log, const struct log_entry_hdr *hdr,
         log_console_print_hdr(hdr);
     }
 
+#if MYNEWT_VAL(LOG_VERSION) < 3
+    console_write(body, body_len);
+#else
     if (hdr->ue_etype != LOG_ETYPE_CBOR) {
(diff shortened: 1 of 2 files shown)


vrahane:
@kasjer We wanted to get away from the non CBOR version is my understanding. I do agree there is no other way to implement reboot_log for LOG_VERSION 2, so we would have to revert the old code for LOG_VERSION 2. At some point we should get rid of LOG_VERSION 2 entirely since the MCUmgr mobile library does not understand log version 2 Vs 3.

vrahane:
@kasjer, I think you should close this PR. We would be removing support for `LOG_VERSION 2` based on what was discussed on the email chain. I created an issue for it as well: https://github.com/apache/mynewt-core/issues/2120.

vrahane:
Opened a PR for removing LOG_VERSION 2: https://github.com/apache/mynewt-core/pull/2157

kasjer:
LOG_VERSION 2 is no longer supported

PR #2089, opened 2019-11-08
kernel/os: count t_run_time in cpu time ticks conditionally

vrahane:
- task runtime - t_run_time variable can now be counted in cpu time
  ticks if OS_TASK_RUN_TIME_CPUTIME syscfg is set.


vrahane:
Thanks @wes3, changing as per review.

PR #2090, opened 2019-11-08
nrf52xxx/hal_spi.c, change to use MCU_TARGET instead of old MCU_NRF52832

ncasaril:
As per subject.


ncasaril:
Well, spotted. Comma is no longer crowded ;)

PR #2093, opened 2019-11-13
console history persistence

kasjer:
Console history now can be kept in log as well as in ram.

### DIFF
--- a/sys/console/full/history_log/src/history_log.c
+++ b/sys/console/full/history_log/src/history_log.c
@@ -1,0 +1,358 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <stdint.h>
+#include <ctype.h>
+#include <os/mynewt.h>
+#if MYNEWT_VAL(LOG_FCB)
+#include <fcb/fcb.h>
+#endif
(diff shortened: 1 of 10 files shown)


sys/console/full/history_log/src/history_log.c
```diff
@@ -0,0 +1,359 @@
...
+#endif
+#if MYNEWT_VAL(LOG_FCB2)
+#include <fcb/fcb2.h>
+#endif
+#include <log/log.h>
+#include <console/history.h>
+
+#define HISTORY_CACHE_SIZE MYNEWT_VAL(CONSOLE_HISTORY_LOG_CACHE_SIZE)
+
+#define VARIANT 2
```
andrzej-kaczmarek: some leftover?

sys/console/full/syscfg.yml
```diff
@@ -43,15 +43,31 @@ syscfg.defs:
...
         description: >
-            Number of lines to be stored in console history.
-            Set to "0" to disable console history.
+            This is no longer used.
+            Equivalent settings:
+                CONSOLE_HISTORY: ram
+                CONSOLE_HISTORY_RAM_HISTORY_SIZE: <n>   <- previous value of CONSOLE_HISTORY_SIZE
         value: 0
-    CONSOLE_MAX_PROMPT_LEN:
+        deprecated: 1
```
andrzej-kaczmarek: if this is not used (i.e. it does nothing as there is no new syscfg which "inherits" its value) then it should better be `defunct`. also I'd move it somewhere at the end of settings so it does not bother anyone

sys/console/full/syscfg.yml
```diff
@@ -43,15 +43,31 @@ syscfg.defs:
...
+    CONSOLE_HISTORY_AUTO_SEARCH:
+        description: >
+            When set to 1 history is automatically searched for lines starting with already typed
+            text.
+        value: 0
+    CONSOLE_PROMPT_MAX_LEN:
         description: 'Maximum number of characters for prompt'
         value: 16
-    CONSOLE_STICKY_PROMPT:
+    CONSOLE_PROMPT_PERSIST:
```
andrzej-kaczmarek: `STICKY` was nicer :)

sys/console/full/history_log/src/history_log.c
kasjer: indeed, removed

sys/console/full/syscfg.yml
kasjer: switched to CONSOLE_PROMPT_STICKY

sys/console/full/syscfg.yml
kasjer: moved to end and changed to defunct

PR #2094, opened 2019-11-14
oc_core: randomize uuid before use

nkaje:
Initialize the uuid buffer with random values before passing it to oc_gen_uuid().

This fixes the following warning generated from code coverage analysis.

```
CID 10028 (#1 of 1): Uninitialized scalar variable (UNINIT)10.
uninit_use_in_call: Using uninitialized element of array uuid.id when calling oc_gen_uuid.
```

### DIFF
--- a/net/oic/src/api/oc_uuid.c
+++ b/net/oic/src/api/oc_uuid.c
@@ -94,8 +94,21 @@ oc_uuid_to_str(const oc_uuid_t *uuid, char *buffer, int buflen)
 }
 
 void
+oc_gen_randomize_uuid(oc_uuid_t *uuid)
+{
+    int i;
+
+    for(i = 0; i < 16; i++) {
+            uuid->id[i] = (uint8_t)rand();
+    }
+}
+
+void
 oc_gen_uuid(oc_uuid_t *uuid)
 {
+  /* Set to Random values as per type 4 */
+  oc_gen_randomize_uuid(uuid);
+
   hal_bsp_hw_id(&uuid->id[0], sizeof(uuid->id));
 
   /*  From RFC 4122


net/oic/src/api/oc_core_res.c
```diff
@@ -236,6 +252,8 @@ oc_core_init_platform(const char *mfg_name, oc_core_init_platform_cb_t init_cb,
                                    core_resources[OCF_P].interfaces);
     oc_rep_set_uint(root, p, core_resources[OCF_P].properties & ~OC_PERIODIC);
 
+    /* Initialize with random values */
+    oc_core_randomize_uuid(&uuid);
     oc_gen_uuid(&uuid);
```
utzig: I honestly have no idea why you would initialize with random values. If this is something required by OIC, why does the library not do it already (in `oc_gen_uuid` maybe?)

nkaje: I looked into https://tools.ietf.org/html/rfc4122#page-14, as the oc_gen_uuid() is implementing type 4. This indicated that except from the 8th octet and 6th octet, others should be random.
The warning from code coverage tool is due to uninitialized variable. To fix this warning and satisfy the RFC requirement, the values need to be initialized with random values. Hence the randomization.

I can certainly move this call into oc_gen_uuid(). Let me know what you think.

net/oic/src/api/oc_core_res.c
```diff
@@ -207,6 +208,21 @@ oc_core_platform_handler(oc_request_t *req, oc_interface_mask_t interface)
...
 }
 
+static void oc_core_randomize_uuid(oc_uuid_t *pu)
+{
+    struct trng_dev *dev;
+    int i;
+
+    dev = (struct trng_dev *)os_dev_lookup("trng");
+    if (!dev) {
+        for(i=0; i<16; i++) {
```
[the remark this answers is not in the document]
nkaje: fixed

nkaje:
Target `nordic_pca10056_bleprph_oic` was failing on Travis CI. Hence added `pkg.deps` and `pkg.apis` entries to `net/oic/pkg.yml`. 
Alternatively, we could add these entries to @apache-mynewt-core\apps\bleprph_oic\pkg.yml, but then every app that uses these deps need to change. Since it's a dependency of `net/oic` on `trng`, it is added there.

net/oic/src/api/oc_uuid.c
```diff
@@ -93,11 +94,30 @@ oc_uuid_to_str(const oc_uuid_t *uuid, char *buffer, int buflen)
...
+    struct trng_dev *dev;
+    int i;
+
+    dev = (struct trng_dev *)os_dev_lookup("trng");
+    if (!dev) {
+        for(i = 0; i < 16; i++) {
+            uuid->id[i] = (uint8_t)rand();
+        }
+    } else {
+        trng_read(dev, (void *)&uuid->id, 16);
```
mkiiskila: I would recommend using rand() only, and not add dependency to TRNG for this.

net/oic/pkg.yml
```diff
@@ -23,11 +23,15 @@ pkg.author: "https://github.com/iotivity/iotivity-constrained"
 pkg.homepage: "https://www.iotivity.org/"
 pkg.keywords:
 
+pkg.apis:
+    - TRNG_HW_IMPL
```
ccollins476ad: This doesn't seem quite right.  The `net/oic` package does not implement the random number generator interface does it?

nkaje: No, net/oic will need to use `trng`. Not sure if I am setting the dependency right. If I don't, I hit
`Error: Unsatisfied APIs detected:
    * TRNG_HW_IMPL, required by: hw/drivers/trng`

ccollins476ad:
Barging in here, but in my opinion `net/oic` should not depend on a trng driver.  If we really need a random number generator, then it should require some sort of rng API (I don't think such an API exists at the moment).  I realize this comment isn't terribly helpful because it doesn't suggest a ready solution...

nkaje:
Thanks all for the feedback. I will upload a new PS using only `rand()`

PR #2095, opened 2019-11-15
sys/console: Fix NULL pointer dereference in log_find

kasjer:
When log_find() was called for non existing log after
checking all valid log pointers NULL pointer would be
used for checking name of the log.

Now strcmp() is called only if log pointer is not NULL.

### DIFF
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -297,12 +297,11 @@ log_find(const char *name)
     struct log *log;
 
     log = NULL;
-    do {
-        log = log_list_get_next(log);
+    while ((log = log_list_get_next(log)) != NULL) {
         if (strcmp(log->l_name, name) == 0) {
             break;
         }
-    } while (log != NULL);
+    }
 
     return log;
 }


sys/log/full/src/log.c
```diff
@@ -297,12 +297,11 @@ log_find(const char *name)
     struct log *log;
 
     log = NULL;
```
ccollins476ad: This initialization can be removed if you are so inclined.

kasjer: > This initialization can be removed if you are so inclined.
I'm afraid not.
Not that I'm not inclined, it would be wrong.

ccollins476ad: Doesn't `log` get initialized by the while loop?  `log_list_get_next()` has to get called at least once.

kasjer: Oh, I'm sorry I was not clear enough.
You are right that `log` is initialized but it is also used in the call `log_list_get_next(log)` so it's better that at that point it is NULL which means: start from the beginning.
It could easily be missed, and I should explain that in my first reply, I'm sorry for that.

ccollins476ad: Oops, you're right.

PR #2096, opened 2019-11-15
sys/log: Add persistent log index

kasjer:
It may be useful to keep log index even if after factory reset.

On startup:
A. (As usual) Firmware scans all persistent logs to determine the next available index.
B. Firmware reads the persistent_index value from flash, if available.
C. The next_index RAM variable is set to the greater of these two values.

On each log write:
1. Compare the next_index variable to persistent_index.
2. If the difference is less than 500, schedule an event to write a new persistent_index to flash.
3. The new persistent_index is equal to: next_index + 1000.

Note 1: the numbers 500 and 1000 were chosen arbitrarily.
Note 2: In step 2, the new persistent_index is written asynchronously because the logging task might not tolerate the time required to write to flash (e.g., ble_ll task).

### DIFF
--- a/sys/log/common/include/log_common/log_common.h
+++ b/sys/log/common/include/log_common/log_common.h
@@ -119,6 +119,10 @@ struct log;
 struct log_info {
 #if MYNEWT_VAL(LOG_GLOBAL_IDX)
     uint32_t li_next_index;
+#if MYNEWT_VAL(LOG_PERSIST_INDEX)
+    uint32_t li_persistent_index;
+    struct os_event li_persist_store_event;
+#endif
 #endif
     uint8_t li_version;
 };
--- a/sys/log/full/include/log/log.h
+++ b/sys/log/full/include/log/log.h
@@ -219,6 +219,10 @@ struct log {
     uint16_t l_max_entry_len;   /* Log body length; if 0 disables check. */
 #if !MYNEWT_VAL(LOG_GLOBAL_IDX)
     uint32_t l_idx;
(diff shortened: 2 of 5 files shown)


ccollins476ad:
I will take a closer look at the code, but for now I just have a high level comment.

Now that we support separate log indices per log (`!MYNEWT_VAL(LOG_GLOBAL_INDEX)`), I think much of the motivation for this feature is gone.  Prior to the separate index feature, there was a fear that the same index could be used by two different persistent logs.  For example:

1. Persistent log writes entry with index=1.
2. Console log writes entry with index=2.
3. Device reboots.
4. Persistent log writes entry with index=2.

If each log can have its own running index, this scenario can't happen *with persistent logs*.

Non-persistent logs (console, cbmem) continue to reuse indices, even if the separate index feature is enabled.  Do we care about this?  Ideally there would be no reuse by any logs whatsoever, but I wonder if reuse among transient logs is a big enough issue to warrant this feature.

If we don't care about reuse among transient logs, then all we care about is remembering the most-recent index when logs get erased (factory reset).  In this case, this feature can be simplified quite a bit:

1. No need to schedule an index-persist-event at write-time.  Just persist an index upon factory reset.
2. No need to persist a seperate index per log.  Just persist the greatest index+1 to a single config setting.  Or just write a dummy entry to each persistent log after the erase.

On the other hand, if we want to prevent reuse among transient logs, I do think we need the more complicated implementation.

I don't know the answer.  I'm just asking because it seems a shame to add such a messy feature for something that has never been an issue in practice (by messy I mean the feature itself, not the implementation :)).

PR #2098, opened 2019-11-18
[WIP] sensors/lsm303dlhc: Add bus driver support

kasjer:
Now lsm303dlhc can be configured with I2C bus drivers.


ccollins476ad:
@kasjer should this PR stay open or is it no longer needed?

kasjer:
Will be opened again when I have time to finish it

PR #2100, opened 2019-11-18
Sensors fixes

sjanc:


sjanc:
Yes, I was using 
[janc@ix ~]$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (Fedora 9.2.0-1.fc31) 9.2.0

vrahane:
oh wow, its a bummer though. Anyways, I think the changes look fine.

PR #2101, opened 2019-11-18
log/common: Fix misleading comment.

mlaz:
Currently `mcumgr log level_list` command isn't working because it expects LOG_LEVEL_IDX to return NULL instead of the string "UNKNOWN" when a non valid level index is provided.


vrahane:
Does the same thing happen with `newtmgr` ? I ask so because an `UNKNOWN` level has been around since a long time, maybe we should change the tool to handle `UNKNOWN` instead of changing the levels.

mlaz:
Ok, we can do that.

PR #2104, opened 2019-11-19
compiler/sim: Add --gc-sections option to the linker

rymanluk:
This patch make sure that image does not contain
not used functions and data as it is done for other targets


rymanluk:
> It would be nice to have it also in: sim-armv7, sim-mips, mips, xc32

I agree, but I need to setup environment for that to make sure it does not break anything. Will try to find time for that.

rymanluk:
Oops looks like 
GCC version
Apple LLVM version 9.0.0 (clang-900.0.38

does not support --gc-sections

rymanluk:
looks like for Apple there is other option - PR updated.

PR #2105, opened 2019-11-19
Remove internal CMAC (Dialog BLE Controller) driver

andrzej-kaczmarek:
We'll use external driver from mynewt-nimble instead.


andrzej-kaczmarek:
new CMAC driver is available here: https://github.com/apache/mynewt-nimble/pull/667

PR #2106, opened 2019-11-21
PR for 3 changes : NDEBUG build, fix SPI on STM32, bitbang uart tx only

brianwyld:
NDEBUG: currently mynewt-core fails to compile due to warnings when NDEBUG is defined (to create release version of code normally). This is because of:
 - assert() not being defined (leading to 'unused variable warnings')
 - assert.h not pulling in os/mynewt.h
 - non-definition of assert(0) meaning no return value for certain functions

SPI: in the STM32 code there is a 'weak' spi init function which should be overriden in another file; gcc does not take this into account. Fix by #ifdefing out the base spi function.

UART bitbang : 
 -allow use of this driver with only TX pin (remove syscfg test on the RX pin!=-1)
 - make its initialisation explicit by creation of the driver, rather than in pkg.yml which can't be overriden


hw/bus/src/bus.c
```diff
@@ -530,9 +530,12 @@ bus_node_unlock(struct os_dev *node)
      * structs are broken) and OS_BAD_MUTEX (unlock shall be only done by the
      * same task which locked it).
      */
-    assert(err == OS_OK || err == OS_NOT_STARTED);
-
-    return 0;
+    if (err == OS_OK || err == OS_NOT_STARTED) {
+        return 0;
+    }
+    // Anything else is assertable (or if no asserts, return error)
```
kasjer: C++ style comments are not used in mynewt

hw/bus/src/bus.c
```diff
@@ -530,9 +530,12 @@ bus_node_unlock(struct os_dev *node)
...
+    assert(0);
+    return -1;
```
kasjer: At this point rc has some error code (other then 0) why change something that may be meaningful to -1 that maybe less informative

hw/drivers/uart/uart_bitbang/pkg.yml
```diff
@@ -28,4 +28,4 @@ pkg.deps:
     - "@apache-mynewt-core/hw/drivers/uart"
 
 pkg.init:
-    uart_bitbang_pkg_init: 'MYNEWT_VAL(UARTBB_SYSINIT_STAGE)'
+#    uart_bitbang_pkg_init: 'MYNEWT_VAL(UARTBB_SYSINIT_STAGE)'
```
kasjer: if something is removed it should be gone and not commented out.

hw/hal/src/hal_flash.c
```diff
@@ -245,7 +245,7 @@ hal_flash_erase(uint8_t id, uint32_t address, uint32_t num_bytes)
     uint32_t end;
     uint32_t end_area;
     int i;
-    int rc;
+    int rc=0;
```
kasjer: missing spaces around =

hw/mcu/stm/stm32_common/src/hal_spi.c
```diff
@@ -172,6 +172,7 @@ stm32_resolve_spi_irq(SPI_HandleTypeDef *hspi)
 #endif
     default:
         assert(0);
+        return 0;
```
kasjer: returning 0 here, would lead to unpredictable behavior.

hw/mcu/stm/stm32_common/src/hal_uart.c
```diff
@@ -24,6 +24,7 @@
 #include "bsp/bsp.h"
 #include <assert.h>
 #include <stdlib.h>
+#include "os/mynewt.h"
```
kasjer: It seems strange that this file missed include file (please confirmed)

hw/mcu/stm/stm32l1xx/src/ext/Drivers/STM32L1xx_HAL_Driver/Src/stm32l1xx_hal_spi.c
```diff
@@ -170,7 +170,8 @@ static HAL_StatusTypeDef SPI_WaitOnFlagUntilTimeout(struct __SPI_HandleTypeDef *
 @endverbatim
   * @{
   */
-
+#if 0
+// BW : The weak direction is ignored by gcc, and so fails to pick up the proper implementation from stm32l1xx_hal_spi_ex.c
```
kasjer: C++ style comment should not be used.
Adding **BW :** is out of place

brianwyld:
Hi,

Thanks for your comments. I will update the code as per and attempt as 3
separate PRs.
To address specifc points:
 - compilation with  NDEBUG : even if it is mynewt 'policy' to have
assert() enabled generally, the code should at least compile when a
standard define is set and currently mynewt does not. The problem is that
currently it does not compile without warnings.
 - the hal_uart.c file in STM32 code is indeed missing the include file (I
agree this is strange but there you go...)

thanks

Brian

hw/mcu/stm/stm32_common/src/hal_uart.c
brianwyld: I confirm this file did not include os/mynewt.h (and hence does not compile if NDEBUG is defined)

hw/drivers/uart/uart_bitbang/pkg.yml
brianwyld: Can the pkg.init be made conditional on UARTBB_SYSINIT_STAGE being defined to a value? something like:
pkg.init.(UARTBB_SYSINIT_STAGE!=-1)
   uart_bitbang_pkg_init: 'MYNEWT_VAL(UARTBB_SYSINIT_STAGE)'

then if you don't want the init its just a case of defining its stage to be -1.

hw/bus/src/bus.c
brianwyld: Ok, wil return 'err'

hw/hal/src/hal_flash.c
brianwyld: ok

hw/mcu/stm/stm32_common/src/hal_spi.c
brianwyld: Will default to SPI1_IRQn

hw/mcu/stm/stm32l1xx/src/ext/Drivers/STM32L1xx_HAL_Driver/Src/stm32l1xx_hal_spi.c
brianwyld: Ok

kasjer:
Hi @brianwyld  you have better chances to merge you changes if you split you pr #2106. Removing restriction on bit bang uart (no need for RX) seems like reasonable and should not face with much resistance. Fix STM32 gpio is also separate thing. With assert related stuff you can still hit the wall and users who created mynewt in the first place may have something to say. I also refrained from pointing out every single c++ style comment that is still out there.

hw/drivers/uart/uart_bitbang/pkg.yml
brianwyld: Resolved by changing how I use bitbang, no more need to avoid its initialisation. Reverted to original.

brianwyld:
I don't understand why the CI build fails - it seems to be a problem with Travis not with the code?

brianwyld:
closing this PR to split it as requested

PR #2108, opened 2019-11-21
Hardware stack limiter for CM33

andrzej-kaczmarek:


kernel/os/src/arch/cortex_m33/m33/HAL_CM33.s
```diff
@@ -183,6 +183,11 @@ PendSV_Handler:
         STR     R12,[R1,#0]             /* Update stack pointer in current task */
         STR     R2,[R3]                 /* g_current_task = highest ready */
 
+        LDR     R12,[R2,#4]             /* get stack top of task we will start */
+        LDRH    R3,[R2,#8]              /* calculate stack size */
+        LSL     R3,R3,#2
```
kasjer: LSL R3, R3, #2
SUB R12, R12, R3
could be change to
SUB R12, R12, R3, LSL #2
saving 4 bytes

If on the other hand os_task_t stored stack bottom instead of stack top (what could be understood from @wes3 comment), code would be even shorter, and coverity tool would not issue false positive issues about out of bound access somewhere during task creation.

wes3: I would store it :-)

andrzej-kaczmarek:
@wes @kasjer 
I also opt for having it always enabled
and I'm fine with storing stack bottom in `os_task`, not sure if we need stack top, bottom and size stored at the same time but as far as I see there is code in e.g. mcumgr which unfortunately accesses `os_task` struct directly (which I don't like, this should be done via API) so perhaps it's better to just add stack bottom as some things may break if we remove stack top. do you have any strong opinion on this?
will also add comment to make sure no one moves these members in future

wes3:
@andrzej-kaczmarek I would simply add stack bottom if you ask me. I guess I was thinking making this "configurable" because stack bottom would really only be used by the M33 and as you said we store all those things. I cannot imagine folks getting upset about adding 4 bytes though. So my vote would be to always have it, no config, add stack bottom and put that comment about not moving stuff around.

mkiiskila:
I don't like adding another pointer to stack for the purpose of this feature only. To me this lgtm as is.

wes3:
@mkiiskila You would rather sacrifice that extra time every context switch for 4 bytes? I definitely would not.

andrzej-kaczmarek:
updated as per discussion on dev

andrzej-kaczmarek:
build will fail until https://github.com/apache/mynewt-mcumgr/pull/52 is merged

PR #2112, opened 2019-11-25
kernel/os: Fix sanity interval handling

andrzej-kaczmarek:
Previous check causes idle task to skip sanity check if wake up occured
at the same tick when sanity should happen. This should not be a problem
because iticks will be 0 and sanity check will be performed after wakeup
on next tick. The only side effect should be wake up on 2 consecutive
ticks. However...

It is (unlikely but) possible that (at least) one tick occurs between
check and entering critical section. If this happens on a tick where
sanity check is skipped, time to next sanity is calculated as a negative
value instead of 0 so depending on time to next sched/callout we may not
wake up for sanity and tickle watchdog on time.

### DIFF
--- a/kernel/os/src/os.c
+++ b/kernel/os/src/os.c
@@ -97,6 +97,7 @@ os_idle_task(void *arg)
     os_time_t iticks, sticks, cticks;
     os_time_t sanity_last;
     os_time_t sanity_itvl_ticks;
+    os_time_t sanity_to_next;
 
     sanity_itvl_ticks = (MYNEWT_VAL(SANITY_INTERVAL) * OS_TICKS_PER_SEC) / 1000;
     sanity_last = 0;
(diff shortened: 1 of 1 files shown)


kernel/os/src/os.c
```diff
@@ -111,7 +112,7 @@ os_idle_task(void *arg)
         ++g_os_idle_ctr;
 
         now = os_time_get();
-        if (OS_TIME_TICK_GT(now, sanity_last + sanity_itvl_ticks)) {
+        if (OS_TIME_TICK_GEQ(now, sanity_last + sanity_itvl_ticks)) {
```
kasjer: this changes sanity check frequency by one tick.
I think it could be left as it was except that line 139 could be change to
sanity_to_next = 0;

kasjer: now interval is accurate, disregard my previous comment

PR #2121, opened 2019-12-04
mcu: stm32: convert remaing families to syscfg clocking system

utzig:


utzig:
@kasjer This already configures the Olimex-P103 and Bluepill boards to run at 72MHz like you did in #2118.

PR #2123, opened 2019-12-05
Fix references to t_stacktop

andrzej-kaczmarek:


kernel/os/src/os_stacktrace.c
```diff
@@ -78,14 +78,16 @@ os_stacktrace(uintptr_t sp)
...
 
     sp &= ~(sizeof(uintptr_t) - 1);
     end = sp + OS_STACK_DEPTH_MAX;
 
     if (g_os_started && g_current_task) {
         t = g_current_task;
-        if (sp > (uintptr_t)t->t_stacktop && end > (uintptr_t)t->t_stacktop) {
-            end = (uintptr_t)t->t_stacktop;
+        stacktop = os_task_stacktop_get(t);
+        if (sp > (uintptr_t)stacktop && end > (uintptr_t)stacktop) {
```
kasjer: first condition `sp > stacktop` does not make sense in this check and will most likely result in extra values being printed in loop below.
In normal case **sp** will never be greater then **stacktop** making this condition to evaluate to false even if **end** is greater then **stacktop**.

PR #2126, opened 2019-12-10
hw/mcu/nordic: Fix hal_timer firing too early

andrzej-kaczmarek:
In hal_timer there is a workaround for an RTC compare issue which in
some cases fires timer callback too early. The problem with this code
is that timer value read in a callback will be *before* what was set
on timer and this is not what callback expects.

To fix this, in case we detect that new compare is too close to current
timer value we should not trigger timer too early but instead enable
TICK interrupt which occurs on every counter change and disable it once
condition for a workaround is no longer valid. This can cause an extra
interrupt on a tick with no timers being handled, but this is ok.

### DIFF
--- a/hw/mcu/nordic/nrf51xxx/src/hal_timer.c
+++ b/hw/mcu/nordic/nrf51xxx/src/hal_timer.c
@@ -200,14 +200,26 @@ nrf_timer_set_ocmp(struct nrf51_hal_timer *bsptimer, uint32_t expiry)
         delta_t = (int32_t)(expiry - temp);
 
         /*
-         * The nrf documentation states that you must set the output
-         * compare to 2 greater than the counter to guarantee an interrupt.
-         * Since the counter can tick once while we check, we make sure
-         * it is greater than 2.
+         * The nRF51xxx documentation states that COMPARE event is guaranteed
+         * only if value written to CC register is at least 2 greater than the
+         * current counter value. We also need to account for possible extra
+         * tick during calculations so effectively any delta less than 3 needs
+         * to be handled differently. TICK event is used to have interrupt on
+         * each subsequent tick so we won't miss any and in case we detected
+         * mentioned extra tick during calculations, interrupt is triggered
+         * immediately. Delta 0 or less means we should always fire immediately.
          */
-        if (delta_t < 3) {
+        if (delta_t < 1) {
+            rtctimer->INTENCLR = RTC_INTENCLR_TICK_Msk;
             NVIC_SetPendingIRQ(bsptimer->tmr_irq_num);
-        } else  {
+        } else if (delta_t < 3) {
+            rtctimer->INTENSET = RTC_INTENSET_TICK_Msk;
+            if (rtctimer->COUNTER != cntr) {
+                NVIC_SetPendingIRQ(bsptimer->tmr_irq_num);
+            }
+        } else {
+            rtctimer->INTENCLR = RTC_INTENCLR_TICK_Msk;
+
             if (delta_t < (1UL << 24)) {
                 rtctimer->CC[NRF_RTC_TIMER_CC_INT] = expiry & 0x00ffffff;
             } else {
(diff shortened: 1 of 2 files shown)


hw/mcu/nordic/nrf52xxx/src/hal_timer.c
```diff
@@ -168,11 +168,15 @@ nrf_timer_set_ocmp(struct nrf52_hal_timer *bsptimer, uint32_t expiry)
          * The nrf documentation states that you must set the output
          * compare to 2 greater than the counter to guarantee an interrupt.
          * Since the counter can tick once while we check, we make sure
-         * it is greater than 2.
+         * it is greater than 2. Otherwise enable TICK event so we can get an
+         * interrupt at proper tick (also we may get an extra interrupt, but
+         * we will just not handle any timer there).
          */
         if (delta_t < 3) {
```
wes3: so if I am understanding this correctly, this code will keep interrupting with the tick timer until we reach the expiration time. Given that is the case, I am wondering if we should fire off the timer if delta_t is zero. Otherwise we could end up being a tick late (at least that seems possible).

andrzej-kaczmarek: you're right. and now actually I think we should also set pending interrupt when delta==1 because the timer can tick once so TICK interrupt will happen on next tick which is also one tick too late. so something like this should work I think:
delta==2 enable tick and wait for proper tick (we may have 1 or 2 additional "empty" interrupts)
delta==1 enable tick and set irq pending (we may have 0 or 1 "empty" interrupt)
delta==0 set irq pending

does this make sense?

wes3: yes that seems to make sense to me. Too bad it cannot just work on the exact tick :-)

andrzej-kaczmarek:
@wes3 @kasjer fixed

andrzej-kaczmarek:
previous version (after 1st round of fixes) did not handle delta<0 properly, fixed now. also it was possible that if we have an "empty" interrupt before actual expiry value of 1st timer we would trigger another interrupt instantly so we could spin for a brief moment in an interrupt - also fixed. it seems to work fine now although I did not verify on nrf51 as I do not have one at the moment...

hw/mcu/nordic/nrf52xxx/src/hal_timer.c
```diff
@@ -165,14 +165,34 @@ nrf_timer_set_ocmp(struct nrf52_hal_timer *bsptimer, uint32_t expiry)
...
+         * less than 3 (we include also +2 since counter can tick once while we
+         * check).
          */
-        if (delta_t < 3) {
+        if (delta_t < 1) {
+            /* We are past or exactly at expiry value, trigger interrupt now */
+            rtctimer->INTENCLR = RTC_INTENCLR_TICK_Msk;
             NVIC_SetPendingIRQ(bsptimer->tmr_irq_num);
-        } else  {
+        } else if (delta_t == 1) {
```
kasjer: what about this:
```
        if (delta_t < 3) {
            rtctimer->INTENSET = RTC_INTENSET_TICK_Msk;
            /* Check if timer advanced, if so force interrupt in case line above was too late to enable tick */
            if (rtctimer->COUNTER != (temp & 0x00ffffff)) {
                NVIC_SetPendingIRQ(bsptimer->tmr_irq_num);
            }
```
Then in hal_timer_chk_queue() new logic with expired variable would not be needed I guess

hw/mcu/nordic/nrf51xxx/src/hal_timer.c
```diff
@@ -200,14 +200,34 @@ nrf_timer_set_ocmp(struct nrf51_hal_timer *bsptimer, uint32_t expiry)
         delta_t = (int32_t)(expiry - temp);
 
         /*
-         * The nrf documentation states that you must set the output
-         * compare to 2 greater than the counter to guarantee an interrupt.
-         * Since the counter can tick once while we check, we make sure
-         * it is greater than 2.
+         * The nRF52xxx documentation states that COMPARE event is guaranteed
```
kasjer: I wander if nRF52xxx should be a reference it this code?

PR #2129, opened 2019-12-11
hw/scripts: Fix another typo...

andrzej-kaczmarek:
The same kind of typo as 7bba744bf, I forgot to update it.


andrzej-kaczmarek:
> How come jlink script is fixed while commit message says that it's hw/mcu typo

typo ;)

PR #2133, opened 2019-12-16
sys/stats: Fix build with GCC 9.2.1

andrzej-kaczmarek:
This fixes following build error:

Compiling repos/apache-mynewt-core/sys/stats/full/src/stats.c
Error: In function 'stats_init',
    inlined from 'stats_module_init_internal' at repos/apache-mynewt-core/sys/stats/full/src/stats.c:148:10:
repos/apache-mynewt-core/sys/stats/full/src/stats.c:298:5: error: 'memset' offset [13, 16] from the object
at 'g_stats_stats' is out of the bounds of referenced subobject 's_hdr' with type 'struct stats_hdr'
at offset 0 [-Werror=array-bounds]
  298 |     memset((uint8_t *)shdr + offset, 0, size * cnt);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

I do not really like that we have separate macros for retrieving const
and non-const data pointer, but I do not see any easy way to workaround
this except for removing consts explicitly which basically original
private API did by casting to uint8_t*.

### DIFF
--- a/sys/stats/full/include/stats/stats.h
+++ b/sys/stats/full/include/stats/stats.h
@@ -48,6 +48,7 @@ struct stats_hdr {
     int s_map_cnt;
 #endif
     STAILQ_ENTRY(stats_hdr) s_next;
+    uint8_t data[0];
 };
 
 /**
@@ -59,6 +60,7 @@ struct stats_persisted_hdr {
     struct stats_hdr sp_hdr;
     struct os_callout sp_persist_timer;
     os_time_t sp_persist_delay;
+    uint8_t data[0];
 };
 
 #define STATS_SECT_DECL(__name)             \
--- a/sys/stats/full/src/stats.c
+++ b/sys/stats/full/src/stats.c
@@ -81,13 +81,9 @@ struct stats_registry_list g_stats_registry =
     STAILQ_HEAD_INITIALIZER(g_stats_registry);
 
 static size_t
-stats_offset(const struct stats_hdr *hdr)
+stats_data_offset(const struct stats_hdr *hdr)
 {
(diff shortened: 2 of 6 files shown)


andrzej-kaczmarek:
Perhaps someone has some better idea how to fix this... :)

sys/stats/full/include/stats/stats.h
```diff
@@ -48,6 +48,7 @@ struct stats_hdr {
     int s_map_cnt;
 #endif
     STAILQ_ENTRY(stats_hdr) s_next;
+    uint8_t data[0];
```
ccollins476ad, line 51: The C standard prohibits zero-length arrays (gcc allows them as an extension).  The pedantically correct way to do this is to use a "flexible array member", i.e.,
```
uint8_t data[]
```

sys/stats/full/src/stats.c
```diff
@@ -96,13 +92,24 @@ stats_size(const struct stats_hdr *hdr)
...
-    offset = stats_offset(hdr);
-    return (uint8_t *)hdr + offset;
+void *
+stats_data_rw(struct stats_hdr *hdr)
+{
+    if (hdr->s_flags & STATS_HDR_F_PERSIST) {
+        return &((struct stats_persisted_hdr *)hdr)->data[0];
+    } else {
+        return &hdr->data[0];
+    }
```
ccollins476ad, line 112: You might consider implementing this in terms of `stats_data_ro`.  If the stats object is not actually const (i.e., not actually located in read-only memory), then it is safe to cast away const from the result of `stats_data_ro` and return that.

I don't feel strongly about this at all.  Just throwing it out there.

PR #2134, opened 2019-12-16
hw/drivers/crypto: Add crypto_in_use flag.

wes3:
This commit adds a flag to the crypto driver called "in_use".
This flag, and associated API, is intended for use by ISR's to
perform a quick check to see if an underlying crypto hw peripheral
is being used so that the HW peripheral can be used in the ISR
or the use of an associated piece may impact an ongoing crypto
operation.

Currently, this is only implemented for the nrf52. Some changes
were made to the nrf52 crypto code. Some reasoning behind some
of the changes: 1) do not think a counter to bail out of the
loop waiting for the crypto to end or an error to occur is needed.
Either one of the two will occur. 2) Not quite sure if it is
possible but if both EVENTS_ENDECB and EVENTS_ERRORECB can be set
at the same time, the previous could assume success when there was
an error. 3) If EVENTS_ERRORECB is generated we simply try the
operation again.

Note that a somewhat un-related change was made to the encrypted
flash ef_crypto driver to add a restriction such that CRYPTO is
required to be defined when using this package.

### DIFF
--- a/hw/drivers/crypto/crypto_nrf52/src/crypto_nrf52.c
+++ b/hw/drivers/crypto/crypto_nrf52/src/crypto_nrf52.c
@@ -34,27 +34,34 @@ struct aes_128_data {
 static bool
 nrf52_crypto_ecb_start(const struct aes_128_data *crypto_data)
 {
-    int ctr = 0x100000;
-    bool rc = true;
+    bool rc;
+    uint32_t end;
+    uint32_t err;
 
+    /* Just to be sure, always stop ECB and clear END/ERROR registers */
+    NRF_ECB->TASKS_STOPECB = 1;
+    NRF_ECB->EVENTS_ENDECB = 0;
+    NRF_ECB->EVENTS_ERRORECB = 0;
+
+    /* Set the data pointer and start ECB */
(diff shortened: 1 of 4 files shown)


hw/drivers/crypto/crypto_nrf52/src/crypto_nrf52.c
```diff
@@ -34,27 +34,34 @@ struct aes_128_data {
...
     NRF_ECB->ECBDATAPTR = (uint32_t)crypto_data;
     NRF_ECB->TASKS_STARTECB = 1;
 
-    while (ctr-- > 0) {
-        if (NRF_ECB->EVENTS_ENDECB) {
+    while (1) {
+        end = NRF_ECB->EVENTS_ENDECB;
+        err = NRF_ECB->EVENTS_ERRORECB;
+        if (end || err) {
+            if (err) {
```
utzig: You could eliminate this `if` / `else` and `rc` variable by doing `return (err == 0);` at the end, although it would be less readable I suppose. I agree with @andrzej-kaczmarek on using `bool` for `in_use`, otherwise everything looks ok.

PR #2144, opened 2020-01-07
OTP Tool: Add information about supporting custom commands

nkaje:
Update comment about Cmds. For application/custom BSP specific
commands, define them in the custom BSP this module. Commands
can be assigned starting at an offset so that this command set
can be expanded if required.

### DIFF
--- a/hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
+++ b/hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
@@ -33,7 +33,10 @@ sys.path.append(os.path.join(os.getcwd(), "repos", "mcuboot", "scripts",
                 "imgtool"))
 import keys as keys
 
-
+# Cmds that apply for the dialog BSP are defined here.
+# Custom Commands are defined in the custom BSP starting
+# at a different offset so that the default command set
+# can be expanded if needed.
 class Cmd(object):
     OTP_READ_KEY = 0
     OTP_WRITE_KEY = 1
@@ -44,7 +47,6 @@ class Cmd(object):
     OTP_INIT = 6
     FLASH_ERASE = 7
     TEST_ALIVE = 8
-
 
 class cmd_no_payload(NamedTuple):
     som: int


hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -605,6 +615,141 @@ def test_alive_target(uart):
...
+    hwid = bytearray()
+    hwid_len = 10
+    try:
+        with open(infile, "rb") as f:
+                # read hwid file
+                buf = f.read()
+
+                # Expect 10 byte HWID
+                if len(buf) != 10:
+                    raise SystemExit("HWID has incorrect length")
```
utzig: You don't need to wrap the context manager inside a `try` block, just use "Failed to read key from file" here and remove the `except` below.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -605,6 +615,141 @@ def test_alive_target(uart):
...
+    msg = struct.pack('II', *cmd)
+
+    try:
+        ser.write(msg)
+    except serial.SerialException:
+        raise SystemExit("Failed to write %s" % uart)
+
+    data = read_exact(ser, 16)
+    response = cmd_response._make(struct.unpack_from('IIII', data))
+    if (response.status == 0 or response.status == 1):
```
utzig: Not "pythonic' to use parenthesis here.

utzig: A more pythonic way would be `if reponse.status in set([0, 1])`, but ok to use `or` as well.

utzig:
I am under the impression that every function starts with:

```
    try:
        ser = serial.Serial(port=uart, baudrate=1000000, timeout=1,
                            bytesize=8, stopbits=serial.STOPBITS_ONE)
    except serial.SerialException:
        raise SystemExit("Failed to open serial port")

    cmd = cmd_no_payload(0xaa55aa55, Cmd.READ_HWID)
    msg = struct.pack('II', *cmd)

    try:
        ser.write(msg)
    except serial.SerialException:
        raise SystemExit("Failed to write to %s" % uart)
```

where the only difference is the command used in `cmd_no_playload`, maybe a bit of refactoring for reuse?

Also default baudrate of 1_000_000 looks a bit weird, but I assume you know it works! ;-)

Apart from the comments, code looks fine.

sjanc:
what type of address is this?  Also, is BSP using this at all?

benmccrea:
@nkaje In light of the discussion we're having about using existing factory-written OTP data to obtain the HWID. I don't think we need or want a HWID Read/Write API in the script. 

For others reading this, we're proposing to derive the HWID using 6 words of OTP data that are written by Dialog during chip fab. For details see JIRA SM-217

andrzej-kaczmarek:
I do not really think this should be in default upstream BSP for DA1469x DK. This BSP does not use these information at all and also DA1469x MCU has unique id already written to OTP so default DK BSP should just use it. Any custom product based on DA1469x MCU should have its own BSP which can handle these settings in any way it wants.

nkaje:
Updated this PR to document how the custom commands can be implemented. No code change here.

PR #2146, opened 2020-01-08
bsp: b-l072z-lrwan1: LoRaWAN related updates

utzig:


utzig:
@wes3 This BSP has the SX1276 chip connected to SPI1. With c7e6261360861ac34274dd89df5cda35fbb8b964 I changed the SX127x code to allow any SPI index. I don't see any reason why this had to be fixed to SPI index 0, so let me know if there is some!

PR #2149, opened 2020-01-08
kernel/os/selftest: increase taskpool stack size

utzig:
The current taskpool stack size is being exhausted while running the mutex test case from kernel/os on travis-ci. Increase it by 512 bytes which should be enough.


utzig:
Ok, so this fixed the kernel/os/selftest issue, but there is still the util/rwlock/selftest issue, before running on a debug travis build, I will just try increasing the stack size of those test threads as well!

utzig:
> Ok, so this fixed the kernel/os/selftest issue, but there is still the util/rwlock/selftest issue, before running on a debug travis build, I will just try increasing the stack size of those test threads as well!

So it did not work and I dropped that commit. I will spawn a new debug build on travis and try to find the specific util/rwlock issue there.

PR #2153, opened 2020-01-10
sys/config: Allow spaces in value passed to CLI

ccollins476ad:
This is somewhat hacky but I think it solves the common case in a straightforward way.  Currently it is not possible to write a conf value containing spaces using the CLI.  If the user passes multiple values to the `config` write CLI command, e.g.,
```
config id/serial firstname lastname
```

the CLI command fails.

This PR changes the config command handler so that spaces are allowed.  All value tokens are concatenated into a space-delimited string, and the result is written to the config setting.

Issues with this approach:
* All whitespace characters get converted into regular spaces.  I don't think this is a huge issue since it is better than not allowing spaces at all.

I think anything better would be overkill (e.g., shell-style parsing).

### DIFF
--- a/sys/config/src/config_cli.c
+++ b/sys/config/src/config_cli.c
@@ -83,6 +83,48 @@ conf_dump_saved(void)
 }
 #endif
 
+#if (MYNEWT_VAL(CONFIG_CLI_RW) & 2) == 2
+static void
+shell_concat_vals(char *buf, int num_vals, char **vals)
+{
+    int remlen;
+    int len;
+    int off;
+    int i;
+
+    off = 0;
+    remlen = CONF_MAX_VAL_LEN;
+
+    for (i = 0; i < num_vals; i++) {
+        /* Add a space before all tokens except the first. */
+        if (i > 0) {
+            buf[off] = ' ';
+            off++;
+            remlen--;
+        }
+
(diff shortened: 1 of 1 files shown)


sjanc:
how about using "foo bar" syntax?

ccollins476ad:
> how about using "foo bar" syntax?

That is what I meant by "shell-style parsing".  That just seemed like a bit much ("overkill").

I don't think it's a deal breaker, but a quoting scheme also has backwards-compatibility implications.  Tokens get split by the shell package, then the token list gets passed to the appropriate command handler (as an `argc`, `argv` pair).  So any changes we make to token splitting will affect all shell commands.  If we want quoting to preserve literal text, then that changes the meaning of `"` characters.  If a user currently puts quotes in shell arguments, this solution would break their use case.  Of course we would add a way to escape quote characters (e.g., `\"`), but again - backwards compatibility.

My preference is to go with this simpler solution now.  This solution is compatible with any smarter feature that gets added to the shell package, so we could always add that later as needed.  If you (or anyone else) deem this as the wrong solution, I am totally fine with that - I can take a look at implementing proper argument parsing.

ccollins476ad:
I no longer need this feature, so I am closing this PR.  If we need this later, we can implement something better at that time.

PR #2155, opened 2020-01-10
Add travis checking of coding style

utzig:


sjanc:
I think we should also clarify our coding style docs with any rule that is enforced byt this script.

utzig:
@sjanc I agree, but my idea is to not enforce anything, this check should never fail the CI at least in the beginning, it'll be only for informational purposes.

utzig:
This should be functional now; the two main issues that remain:

1) It always checks style on whole files, even when the user made only small changes. It would be possible to isolate the specific changed sections by looking at the diff indexes but since those might not provide something that is valid C, I'm not sure how good it would be to try to style check.
2) There is some issue with alignment of parameters in functions calls. I have not found a way to ignore those, so I left it configured to align parameters that are on different lines below the `(` of the call. This seems to provide better results for our style, although it will be giving a few "false positive" style issues. Hope to fix this soon!

sjanc:
ad1) if using code diff is not an option, would it be possible to run check on master and then on PR and only show new issues (or differences)?

utzig:
> 1) if using code diff is not an option, would it be possible to run check on master and then on PR and only show new issues (or differences)?

Yes, possibly something doable.

PR #2161, opened 2020-01-21
hw/mcu/dialog: Apply trim and preferred values

andrzej-kaczmarek:
So far we assumed that 1st stage bootloader applied all trim values
stored in CS and they are valid. However, since power domains can be
disabled we need to make sure proper settings are reapplied once they
are enabled.

This patch automates this process:
- read trim values for each applicable PD on init
- apply trim values once PD is enabled (on 1st acquire)
- apply preferred values on top of this

This means that after waking up from extended sleep (where most of power
domains are disabled) all settings are reapplied correctly.


utzig:
@andrzej-kaczmarek Please rebase on latest `master`!

benmccrea:
@andrzej-kaczmarek LGTM

PR #2162, opened 2020-01-21
uncrustify: don't add braces to one-liner while

utzig:


utzig:
Fix the `while` issue seen in #2161

PR #2165, opened 2020-01-24
FIx: advance_internal should return error instead of using assert

de-nordic:
Assert may evaluate to empty expression which means that an error removed
from call to extract_number may get missed. It is better to return it,
and let it be processed further.

Signed-off-by: Dominik Ermel <dominik.ermel@nordicsemi.no>

### DIFF
--- a/encoding/tinycbor/src/cborparser.c
+++ b/encoding/tinycbor/src/cborparser.c
@@ -279,7 +279,9 @@ static CborError advance_internal(CborValue *it)
 {
     uint64_t length;
     CborError err = extract_number(it->parser, &it->offset,  &length);
-    assert(err == CborNoError);
+    if (err == CborNoError) {
+        return err;
+    }
 
     if (it->type == CborByteStringType || it->type == CborTextStringType) {
         assert(length == (size_t)length);


de-nordic:
For example, in zephyr this evaluates to empty expression which will make code proceed as if in case of success.
Additionally it cause compilation to warn about unused err variable.

encoding/tinycbor/src/cborparser.c
```diff
@@ -279,7 +279,9 @@ static CborError advance_internal(CborValue *it)
 {
     uint64_t length;
     CborError err = extract_number(it->parser, &it->offset,  &length);
-    assert(err == CborNoError);
+    if (err == CborNoError) {
+        return err;
+    }
```
utzig: I was trying to follow the unroll of this error up to check that it is handled somewhere up in the call chain, it seems to be OK. But the same assert is also being checked inside `cbor_value_enter_container`, should not that one be fixed as well, for consistency?

de-nordic: Yes. It should.

sjanc:
hi,  looks ok to me but you may have to rebase to fix CI issiues

encoding/tinycbor/src/cborparser.c
sjanc: hmm shouldn't if be !=  and not == ?

de-nordic:
Abandoned.

PR #2167, opened 2020-01-26
LIS2DH12 Sem Time

brolan-juul:
A LIS2DH12 semaphore on an interrupt has a max wait time of 4 seconds. This could cause unintended issues as 4 seconds may block the system from completing its other tasks. This also seems slightly excessive.

Reducing the wait time on a semaphore from 4 seconds to 1 second.


utzig:
hmmm, most of those style suggestions were actually good!

vrahane:
@utzig the style suggestions were good but the Pr did not change the lines pointed out by the bot :-)

PR #2174, opened 2020-01-31
Updates to crypto_test app; STM32 crypto HW driver; CBC/CTR speed

utzig:
* apps: crypto_test: add benchmarks for CBC/CTR
* stm32: crypto: make driver obey syscfg HW support
* crypto: use 32-bit XOR in CBC/CTR


utzig:
Could someone take a look?

hw/drivers/crypto/src/crypto.c
```diff
@@ -56,8 +61,20 @@ crypto_do_ctr(struct crypto_dev *crypto, const void *key, uint16_t keylen,
...
-            outbuf8[i] = inbuf8[i] ^ _out[i];
+        /*
+         * For full blocks increase speed by doing 32-bit XOR; maintain the
+         * stream semantics doing byte XORs for smaller sizes (end of buffer).
+         */
+        if (len == AES_BLOCK_LEN) {
+            inbuf32 = (uint32_t *)inbuf8;
+            outbuf32 = (uint32_t *)outbuf8;
+            for (i = 0; i < len / 4; i++) {
+                outbuf32[i] = inbuf32[i] ^ _out32[i];
```
kasjer: I'm not sure what is the usage of this function but unaligned access here could crash Cortex-M0

apps/crypto_test/src/main.c
```diff
@@ -282,6 +286,66 @@ run_benchmark(char *name, block_encrypt_func_t encfn, void *data, uint8_t iter)
...
+                printf("fail... blkidx=%u\n", blkidx);
+                for (j = 0; j < AES_BLOCK_LEN; j++) {
+                    printf("[%02x]<%02x> ", output[j],
+                            aes_128_cbc_expected[blkidx + j]);
+                }
+                return;
+            }
+        }
+    }
+    printf("done in %lu ticks\n", os_time_get() - t);
```
kasjer: I would not use ticks to display time, you could not compare results of two different boards not knowing what tick mean on each one.

hw/drivers/crypto/src/crypto.c
utzig: This should be fixed now, thanks!

apps/crypto_test/src/main.c
utzig: Comparing to different boards is not relevant IMO, what is being done here is to compare the HW acceleration speed to SW implementations using tinycrypt/mbedTLS.

PR #2175, opened 2020-02-01
da1469x: add initial crypto driver

utzig:
This is untested because I don't have HW! Best way to test is to run `apps/crypto_test` and check no errors are detected.

/cc @nkaje @agross-korg

### DIFF
--- a/hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
+++ b/hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
@@ -1,0 +1,246 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <string.h>
+#include <mcu/cmsis_nvic.h>
+#include <os/mynewt.h>
+#include <mcu/da1469x_clock.h>
+#include <crypto/crypto.h>
+#include <crypto_da1469x/crypto_da1469x.h>
+
+static struct os_mutex gmtx;
+
+#define VALID_AES_KEYLEN(x) (((x) == 128) || ((x) == 192) || ((x) == 256))
+
+static bool
+da1469x_has_support(struct crypto_dev *crypto, uint8_t op, uint16_t algo,
+                    uint16_t mode, uint16_t keylen)
+{
+    (void)op;
+
+    if (algo != CRYPTO_ALGO_AES || !VALID_AES_KEYLEN(keylen)) {
+        return false;
+    }
+
+    switch (mode) {
+    case CRYPTO_MODE_ECB:
+    case CRYPTO_MODE_CBC:
+    case CRYPTO_MODE_CTR:
+        return true;
+    }
+
+    return false;
+}
+
+static uint32_t
+da1469x_crypto_op(struct crypto_dev *crypto, uint8_t op, uint16_t algo,
+                  uint16_t mode, const uint8_t *key, uint16_t keylen,
+                  uint8_t *iv, const uint8_t *inbuf, uint8_t *outbuf,
+                  uint32_t len)
+{
+    uint8_t i;
+    uint8_t iv_save[AES_BLOCK_LEN];
+    unsigned int inc;
+    unsigned int carry;
+    unsigned int v;
+    uint32_t *keyreg;
+    uint32_t *keyp32;
+    uint32_t ctrl_reg;
+
+    if (!da1469x_has_support(crypto, op, algo, mode, keylen)) {
+        return 0;
+    }
+
+    if (mode == CRYPTO_MODE_CBC && op == CRYPTO_OP_DECRYPT) {
+        memcpy(iv_save, &inbuf[len-AES_BLOCK_LEN], AES_BLOCK_LEN);
+    }
+
+    os_mutex_pend(&gmtx, OS_TIMEOUT_NEVER);
+    da1469x_clock_amba_enable(CRG_TOP_CLK_AMBA_REG_AES_CLK_ENABLE_Msk);
+
+    /* enable AES / disable HASH */
+    ctrl_reg = ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_HASH_SEL_Msk;
+
+    /* Select mode
+     *
+     * Note: DS also mentions value 1 for ECB.
+     */
+    ctrl_reg &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Msk;
+    switch (mode) {
+    case CRYPTO_MODE_ECB:
+        ctrl_reg |= 0 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Pos;
+        break;
+    case CRYPTO_MODE_CTR:
+        ctrl_reg |= 2 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Pos;
+        break;
+    case CRYPTO_MODE_CBC:
+        ctrl_reg |= 3 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Pos;
+        break;
+    }
+
+    /* AES algorithm, there's no need to check algo here! */
+    ctrl_reg &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_Msk;
+
+    /*
+     * Note: DS also mentions value 3 for 256-bit key.
+     */
+    ctrl_reg &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_AES_KEY_SZ_Msk;
+    switch (keylen) {
+    case 128:
+        ctrl_reg |= 0 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_AES_KEY_SZ_Pos;
+        break;
+    case 192:
+        ctrl_reg |= 1 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_AES_KEY_SZ_Pos;
+        break;
+    case 256:
+        ctrl_reg |= 2 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_AES_KEY_SZ_Pos;
+        break;
+    }
+
+    /* activate key expansion */
+    ctrl_reg |= AES_HASH_CRYPTO_CTRL_REG_CRYPTO_AES_KEXP_Msk;
+
+    if (op == CRYPTO_OP_ENCRYPT) {
+        ctrl_reg |= AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ENCDEC_Msk;
+    } else {
+        ctrl_reg &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ENCDEC_Msk;
+    }
+
+    /*
+     * - no more data in
+     * - disable IRQ
+     * - write out all data
+     */
+    ctrl_reg &= ~(AES_HASH_CRYPTO_CTRL_REG_CRYPTO_MORE_IN_Msk |
+                  AES_HASH_CRYPTO_CTRL_REG_CRYPTO_IRQ_EN_Msk |
+                  AES_HASH_CRYPTO_CTRL_REG_CRYPTO_OUT_MD_Msk);
+
+    AES_HASH->CRYPTO_CTRL_REG = ctrl_reg;
+
+    switch (mode) {
+    case CRYPTO_MODE_CBC:
+    case CRYPTO_MODE_CTR:
+        AES_HASH->CRYPTO_MREG0_REG = os_bswap_32(((uint32_t *)iv)[3]);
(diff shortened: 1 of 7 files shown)


hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,236 @@
...
+da1469x_crypto_decrypt(struct crypto_dev *crypto, uint16_t algo, uint16_t mode,
+        const uint8_t *key, uint16_t keylen, uint8_t *iv, const uint8_t *inbuf,
+        uint8_t *outbuf, uint32_t len)
+{
+    return da1469x_crypto_op(crypto, CRYPTO_OP_DECRYPT, algo, mode, key,
+            keylen, iv, inbuf, outbuf, len);
+}
+
+static int
+da1469x_crypto_dev_open(struct os_dev *dev, uint32_t wait, void *arg)
```
kasjer: access to this function is possible ONLY in os_dev_open() that calls this function ONLY if dev is not NULL so function that checks only whether dev is not NULL seems redundant.

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,236 @@
...
+da1469x_crypto_dev_init(struct os_dev *dev, void *arg)
+{
+    struct crypto_dev *crypto;
+
+    crypto = (struct crypto_dev *)dev;
+    assert(crypto);
+
+    OS_DEV_SETHANDLERS(dev, da1469x_crypto_dev_open, NULL);
+
+    assert(os_mutex_init(&gmtx) == 0);
```
kasjer: I'm not sure if initializing mutex in assert is such a good idea

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,236 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <string.h>
+#include "mcu/cmsis_nvic.h"
```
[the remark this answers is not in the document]
utzig: Yeah, I don't know, I copy/pasted! :-P I have settled on <> but seems like every file around has both...

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,236 @@
...
+        uint16_t mode, uint16_t keylen)
+{
+    (void)op;
+
+    if (algo != CRYPTO_ALGO_AES || !VALID_AES_KEYLEN(keylen)) {
+        return false;
+    }
+
+    switch (mode) {
+    case CRYPTO_MODE_ECB: /* fallthrough */
```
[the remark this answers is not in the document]
utzig: Ok, thanks, fixed!

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,236 @@
...
+
+    /* Select mode */
+    AES_HASH->CRYPTO_CTRL_REG &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_Msk;
+    switch (mode) {
+    case CRYPTO_MODE_ECB:
+        AES_HASH->CRYPTO_CTRL_REG |= 0 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_Pos;
+        break;
+    case CRYPTO_MODE_CBC:
+        AES_HASH->CRYPTO_CTRL_REG |= 2 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_Pos;
+        memcpy(iv_save, &inbuf[len-AES_BLOCK_LEN], AES_BLOCK_LEN);
```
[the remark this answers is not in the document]
utzig: Done.

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,236 @@
...
+        for (i = AES_BLOCK_LEN; (inc != 0 || carry != 0) && i > 0; --i) {
+            v = carry + (uint8_t)inc + iv[i - 1];
+            iv[i - 1] = (uint8_t)v;
+            inc >>= 8;
+            carry = v >> 8;
+        }
+        break;
+    }
+
+    da1469x_clock_amba_disable(CRG_TOP_CLK_AMBA_REG_AES_CLK_ENABLE_Msk);
```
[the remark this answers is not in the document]
utzig: Done.

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
utzig: Ok, I removed the open function since it was doing nothing.

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
utzig: Ok, I separated both.

utzig:
Fixed review issues; added the driver initialization to MCU code which I had forgotten to do; fixed style issues.

utzig:
> We need to make sure there aren't any other possible conflicts where people are using the crypto driver somewhere else. I know we use the secure OTP keys in the bootloader, but that's self contained. The encrypted flash driver is not and could conflict with this.

Ok, I haven't seen there was an encrypted flash already, my bad.

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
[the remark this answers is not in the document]
ccollins476ad: Jumping in here...

This is a pedantic answer.  The original code looks correct to me.  From the C spec (C11 draft, 6.10.2, "Source file inclusion") (http://www.iso-9899.info/n1570.html):
> A preprocessing directive of the form
> ```# include <h-char-sequence> new-line```
> searches a sequence of implementation-defined places for a header identified uniquely by the specified sequence between the < and > delimiters, and causes the replacement of that directive by the entire contents of the header. How the places are specified or the header identified is implementation-defined.
>
> A preprocessing directive of the form
```# include "q-char-sequence" new-line```
> causes the replacement of that directive by the entire contents of the source file identified by the specified sequence between the " delimiters. The named source file is searched for in an implementation-defined manner. If this search is not supported, or if the search fails, the directive is reprocessed as if it read
```# include <h-char-sequence> new-line```
> with the identical contained sequence (including > characters, if any) from the original
directive.

In other words, `<>` is used to include headers from the *implementation* (compiler, linker, etc.).  `""` is used for everything else.  In practice there is not much difference.  I believe gcc searches implementation-specific directories first when `<>` is used, but last when `""` is used.

File: hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c [hunk 7]
```diff
@@ -0,0 +1,232 @@
...
+    case CRYPTO_MODE_CBC:
+    case CRYPTO_MODE_CTR:
+        AES_HASH->CRYPTO_MREG0_REG = ((uint32_t *)iv)[0];
+        AES_HASH->CRYPTO_MREG1_REG = ((uint32_t *)iv)[1];
+        AES_HASH->CRYPTO_MREG2_REG = ((uint32_t *)iv)[2];
+        AES_HASH->CRYPTO_MREG3_REG = ((uint32_t *)iv)[3];
+    }
+
+    /*
+     * Note: DS also mentions value 3 for 256-bit key.
```
nkaje: Based on my work during encrypted flash driver development, I made the following observations
        /*  
         * XXX: Datasheet for da1469x Rev 2.0 (CFR0011-120-00)
         * indicates that a value of 2 and 3 correspond to
         * 256 bit AES Key, However, only 2 works.
         */

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,232 @@
...
+          AES_HASH_CRYPTO_CTRL_REG_CRYPTO_OUT_MD_Msk);
+
+    /* AES algorithm, there's no need to check algo here! */
+    AES_HASH->CRYPTO_CTRL_REG &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Msk;
+
+    /*
+     * Note: DS also mentions value 1 for ECB.
+     */
+
+    /* Select mode */
```
nkaje: CFR0011-120-00: CRYPTO_ALG_MD: It defines the mode of operation of the AES
algorithm when the controller is configured for an
encryption/decryption processing
(CRYPTO_HASH_SEL = 0).
0x0 : ECB
0x1 : ECB
0x2 : CTR
0x3 : CBC

utzig:
@agross-korg @nkaje This would be incompatible with using the enc_flash driver for da1469x at the same time; to test this you would need to use `hw/drivers/flash/enc_flash/ef_crypto`.

utzig:
> @agross-korg @nkaje This would be incompatible with using the enc_flash driver for da1469x at the same time; to test this you would need to use `hw/drivers/flash/enc_flash/ef_crypto`.

That enc_flash for da1469x is better written as well, so I will pick some of that code...

agross-korg:
one mutex to rule them...

utzig:
> one mutex to rule them...

:rofl: yeah, that is one approach, but the `crypto` driver's idea is to add a lower layer that handles the HW crypto providing an abstraction that higher level drivers would use. There is still a mutex but higher levels don't see it and there are other advantages like mbedTLS AES calls being automatically routed to HW, etc.

utzig:
@agross-korg @nkaje What would be the best way to get this in? Does one of you want to take over, or should I merge and fix after testing is done? I am open to suggestions!

utzig:
This was updated with all findings by @nkaje (thanks a lot!); it now works correctly with `crypto_test` app but still have the limitation that the input buffer must be located in RAM (to be fixed!).

nkaje:
Results with apps/crypto_test

```
=== Test vectors ===
AES-128-ECB enc
        vector 0: ok, sz=16
        vector 1: ok, sz=16
        vector 2: ok, sz=16
        vector 3: ok, sz=16
AES-128-ECB dec
        vector 0: ok, sz=16
        vector 1: ok, sz=16
        vector 2: ok, sz=16
        vector 3: ok, sz=16
AES-128-CBC enc
        vector 0: ok, sz=16
        vector 1: ok, sz=16
        vector 2: ok, sz=16
        vector 3: ok, sz=16
AES-128-CBC dec
        vector 0: ok, sz=16
        vector 1: ok, sz=16
        vector 2: ok, sz=16
        vector 3: ok, sz=16
AES-128-CTR enc
        vector 0: ok, sz=16
        vector 1: ok, sz=16
        vector 2: ok, sz=8
        vector 3: ok, sz=16
AES-128-CTR dec
        vector 0: ok, sz=16
        vector 1: ok, sz=16
        vector 2: ok, sz=8
        vector 3: ok, sz=16
=== In-place encrypt/decrypt ===
AES-128-ECB enc: ok
AES-128-ECB dec: ok
AES-128-CBC enc: ok
AES-128-CBC dec: ok
AES-128-CTR enc: ok
AES-128-CTR dec: ok
=== iovec encrypt/decrypt ===
iov AES-128-ECB enc: ok
iov AES-128-ECB dec: ok
iov AES-128-CBC enc: ok
iov AES-128-CBC dec: ok
iov AES-128-CTR enc: ok
iov AES-128-CTR dec: ok
=== Benchmarks - iteration 1 ===
CRYPTO - running 30 iterations of 4096 block encrypt... done in 25 ticks
MBEDTLS - running 30 iterations of 4096 block encrypt... done in 55 ticks
TINYCRYPT - running 30 iterations of 4096 block encrypt... done in 330 ticks
=== Benchmarks - iteration 2 ===
CRYPTO - running 30 iterations of 4096 block encrypt... done in 25 ticks
MBEDTLS - running 30 iterations of 4096 block encrypt... done in 55 ticks
TINYCRYPT - running 30 iterations of 4096 block encrypt... done in 329 ticks
=== Benchmarks - iteration 3 ===
CRYPTO - running 30 iterations of 4096 block encrypt... done in 26 ticks
MBEDTLS - running 30 iterations of 4096 block encrypt... done in 55 ticks
TINYCRYPT - running 30 iterations of 4096 block encrypt... done in 330 ticks
=== CRYPTO benchmarks ===
AES-128-CBC - running 50 iterations of 4096 block encrypt... done in 46 ticks
AES-128-CTR - running 50 iterations of 4096 block encrypt... done in 47 ticks
=== Concurrency [8 tasks] ===
task0 [0 fails / 256 ok] done
task1 [0 fails / t2t5t6t totk] done
ask2 [0 fails / 256 ok] donaea
asaka3 [0 fails / 256 ok] done
sk4 [0 fsasisls / 256 ok] done
k5 [0 fails / 256 ok] dkokne
6 [0 fails / 256 ok] done
7 [0 fails / 256 ok] done
```

hw/drivers/crypto/crypto_da1469x/src/crypto_da1469x.c
```diff
@@ -0,0 +1,246 @@
...
+    case CRYPTO_MODE_CBC:
+        ctrl_reg |= 3 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Pos;
+        break;
+    case CRYPTO_MODE_CTR:
+        ctrl_reg |= 2 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Pos;
+        break;
+    }
+
+    /* AES algorithm, there's no need to check algo here! */
+    ctrl_reg &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Msk;
```
nkaje: This should be `ctrl_reg &= ~AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_Msk`

utzig: Done.

utzig:
> Minor nit: CBC and CTR cases can be rearranged for better readability - mask is 0, 2 and 3

Done!

PR #2176, opened 2020-02-04
da1469x: support blocks that are not ENC_FLASH_BLK multiples.

nkaje:
Support encryption, decryption operation for block lengths != ENC_FLASH_BLK (16)

### DIFF
--- a/hw/drivers/flash/enc_flash/ef_da1469x/src/hw_enc_flash.c
+++ b/hw/drivers/flash/enc_flash/ef_da1469x/src/hw_enc_flash.c
@@ -88,7 +88,7 @@ get_key_size_mask(int keysize)
 }
 
 void
-do_encrypt(const struct hal_flash *h_dev, uint32_t *ctr, const uint8_t *src, uint8_t *tgt, int off, int cnt)
+do_encrypt(const struct hal_flash *h_dev, uint32_t *ctr, const uint8_t *src, uint8_t *tgt)
 {
     /* Select AES CTR, set CRYPTO_ALG_MD bits to 10 */
     uint32_t algo_sel = (2 << AES_HASH_CRYPTO_CTRL_REG_CRYPTO_ALG_MD_Pos);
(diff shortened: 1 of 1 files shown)


hw/drivers/flash/enc_flash/ef_da1469x/src/hw_enc_flash.c
```diff
@@ -132,18 +133,34 @@ do_encrypt(const struct hal_flash *h_dev, uint32_t *ctr, const uint8_t *src, uin
...
 
     ctr[0] = (uint32_t) ((blk_addr - h_dev->hf_base_addr) / ENC_FLASH_BLK);
 
     os_sem_pend(&dev->ef_sem, OS_TIMEOUT_NEVER);
-    do_encrypt(h_dev, ctr, src, tgt, off, cnt);
+    ef_da_get_block(h_dev, ctr, blk_addr, blk);
+    b = &blk[off];
+    for (i = 0; i < cnt; i++) {
+        *tgt++ = *b++ ^ *src++;
+    }
```
utzig: The code is doing CTR-mode in SW, and the HW seems to be configured to do CTR-mode in HW as well, so the results won't be what you would expect; it still garbles data but not in a standard way. I think this can either be changed to use the HW in ECB mode and leave the code above as is, or adapt the code to send the same `src` received by `enc_flash_crypt_arch` to `do_encrypt` and remove this block of XOR operations above, changing for a simple `memcpy` of `cnt` amount...

nkaje: Modified by removing the` XOR` operations and using `memcpy` instead.

hw/drivers/flash/enc_flash/ef_da1469x/src/hw_enc_flash.c
```diff
@@ -139,12 +139,15 @@ enc_flash_crypt_arch(struct enc_flash_dev *edev, uint32_t blk_addr,
     struct eflash_da1469x_dev *dev = EDEV_TO_DA1469X(edev);
     const struct hal_flash *h_dev = edev->efd_hwdev;
     uint32_t ctr[4] = {0};
+    uint8_t blk[ENC_FLASH_BLK] = {0};
 
     ctr[0] = (uint32_t) ((blk_addr - h_dev->hf_base_addr) / ENC_FLASH_BLK);
```
utzig: Btw, it would be interesting to find out if the endianness of the `uint32_t` written to `CRYPTO_MREGx_REG` is correct by just writing the value or if we need to do `os_bswap_32` here (or in the register writing above!). The security is still OK, but the values would be different from the expected ones and the counter would increment in a weird way (instead of +1). A way to test it would be to write one block of a vector to address 0 of flash and read using JTAG/SWD and check that it matches the encrypted vector.

PR #2180, opened 2020-02-07
sys/config: Support for unsigned integer types

ccollins476ad:
Prior to this commit, only signed integer types were supported for persisted settings.  This PR adds support for unsigned integer types.

### DIFF
--- a/sys/config/src/config.c
+++ b/sys/config/src/config.c
@@ -152,21 +152,20 @@ conf_parse_and_lookup(char *name, int *name_argc, char *name_argv[])
 int
 conf_value_from_str(char *val_str, enum conf_type type, void *vp, int maxlen)
 {
-    int32_t val;
-    int64_t val64;
-    uint32_t uval;
-    uint64_t uval64;
+    int64_t val;
+    uint64_t uval;
     char *eptr;
 
     if (!val_str) {
         goto err;
     }
     switch (type) {
+    case CONF_BOOL:
     case CONF_INT8:
     case CONF_INT16:
     case CONF_INT32:
-    case CONF_BOOL:
-        val = strtol(val_str, &eptr, 0);
+    case CONF_INT64:
+        val = strtoll(val_str, &eptr, 0);
         if (*eptr != '\0') {
             goto err;
         }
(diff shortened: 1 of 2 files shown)


sys/config/src/config.c
```diff
@@ -194,6 +196,34 @@ conf_value_from_str(char *val_str, enum conf_type type, void *vp, int maxlen)
...
+            if (uval > UINT8_MAX) {
+                goto err;
+            }
+            *(uint8_t *)vp = uval;
+        } else if (type == CONF_UINT16) {
+            if (uval > UINT16_MAX) {
+                goto err;
+            }
+            *(uint16_t *)vp = uval;
+        } else if (type == CONF_UINT32) {
```
utzig: Since this is inside a `case` I don't think `type` could anything but `CONF_UINT32`here!

sys/config/src/config.c
```diff
@@ -194,6 +196,34 @@ conf_value_from_str(char *val_str, enum conf_type type, void *vp, int maxlen)
...
+            *(uint32_t *)vp = uval;
+        }
+        break;
+    case CONF_UINT64:
+        uval64 = strtoull(val_str, &eptr, 0);
```
utzig: I see there's special handling for `UINT64` here and also for `INT64` above, but I am not sure why you can't just use `strtoll` for all of them.

sys/config/src/config.c
```diff
@@ -252,6 +283,21 @@ conf_str_from_value(enum conf_type type, void *vp, char *buf, int buf_len)
...
+            uval = *(uint8_t *)vp;
+        } else if (type == CONF_UINT16) {
+            uval = *(uint16_t *)vp;
+        } else {
+            uval = *(uint32_t *)vp;
+        }
+        snprintf(buf, buf_len, "%lu", (unsigned long)uval);
+        return buf;
+    case CONF_UINT64:
+        snprintf(buf, buf_len, "%llu", *(unsigned long long *)vp);
```
utzig: My comment regarding special handling for 64-bit int also applies here...

sys/config/src/config.c
ccollins476ad: I agree.  My guess is that the "ll" versions of these functions felt too heavy weight to use for the <=32-bit types.  If I were implementing it from scratch, I think I would just use the "ll" functions across the board.  I am reluctant to change this existing code though... this feels like an easy place to introduce a bug.

sys/config/src/config.c
ccollins476ad: Good point.  I will change it.

sys/config/src/config.c
ccollins476ad: Thanks, I agree here too.  I would rather not change the old code and risk breaking something.  And I don't change the old code, I think it is best to keep the new code consistent with the old.

sys/config/src/config.c
utzig: It looks like both `strtoul` and `strtoull` are mapped to `strntoumax` inside baselibc, just using a different cast!

ccollins476ad: OK @utzig, you convinced me to change everything :).  I also noticed `conf_value_from_str()` had a bug in its range checks for signed integers, so I have fixed that.

Would you mind taking a look at the latest commits?

utzig: > Would you mind taking a look at the latest commits?

Just did, looks good!

PR #2181, opened 2020-02-07
sys/config: Extended callbacks

ccollins476ad:
This PR is needed for the "scfg" feature: https://github.com/apache/mynewt-core/pull/2182.

The four conf_handler callbacks (get, set, commit, and export) do not accept a `void *arg` parameter.  This means every config group must define a dedicated set of callback functions, making it impossible wrap the config API in a generic way.

This PR allows config groups to be defined as "extended".  An extended config group uses slightly different callbacks.  These callbacks do take a `void *arg` parameter.

To define an extended config group, set its `ch_ext` member to 1.

### DIFF
--- a/sys/config/include/config/config.h
+++ b/sys/config/include/config/config.h
@@ -26,6 +26,7 @@
 
 #include <os/queue.h>
 #include <stdint.h>
+#include <stdbool.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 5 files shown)


sys/config/include/config/config.h
```diff
@@ -157,14 +162,40 @@ struct conf_handler {
      * The name of the conifguration item/subtree
      */
     char *ch_name;
+
+    /**
+     * Whether to use the extended callbacks.
+     * 0=standard
+     * 1=extended
+     */
+    uint8_t ch_ext : 1;
```
utzig: I think this should rather be `bool`, although it would not be a problem to use a bitfield in practice, unless someone wrote a configuration to flash and moved that flash to a CPU with different endiannes! :-P

ccollins476ad: I agree, it should be a bool.  I think I was bothered by the size increase of this struct so I went overboard.  But since there is only one field like this, we're not saving any space anyway.  I'll change it to bool.

PR #2186, opened 2020-02-10
stm32: unify peripheral initialization

utzig:
This moves all STM32 BSPs to use a unified peripheral initialization, as is done in nrf5x and da1469x. Lots of testing and few fixes remain before it's done for review...


utzig:
@kasjer So far I have not tested SPI, I2C and ADC on any boards but plan to do it later today. Thanks a lot for taking a look already!

utzig:
@kasjer Ok, so this won't be an easy review! But I did test every HW driver on at least one BSP; also I did a fairly thorough review of all PIN configurations, etc, so I hope there will be nothing obviously broken.

PR #2192, opened 2020-02-12
Add generic API torretrieve provisioned data from BSP

andrzej-kaczmarek:


hw/hal/include/hal/hal_bsp.h
```diff
@@ -79,8 +80,73 @@ int hal_bsp_hw_id_len(void);
...
+ * may be larger than input value.
+ *
+ * Caller can set \p data to NULL to query for required \p length to read
+ * complete data. In such case \p length input value does not matter.
+ *
+ * If provided buffer is too small to store all provisioned data, returned data
+ * may be either truncated (check \p length to see complete data length) or
+ * error is returned. The former can usually happen for variable-length data
+ * where application should first query data length, the latter can usually
+ * happen for fixed-length data where data length is well known.
```
ccollins476ad: This is just a personal thing, but I dislike it when a library makes me do extra work to check for truncation.  I prefer it when it is always an error if the supplied buffer is too small.  I could be mistaken, but looking at the list of data types, I don't see any that can be safely truncated.

One alternative would be to have a second function that allows truncation.

Just my two cents.

andrzej-kaczmarek:
@ccollins476ad @wes3 you're right, this is too confusing. my intention was to allow to read some data even if app uses some buffer of hardcoded length and it's too small, but this does not really make sense as incomplete data are not really useful... I changed this API to return data length as previously, but if buffer is too small it will always return an error. so caller does not need to check for truncation since this won't happen.

also I included some code in common hal_bsp.c so each BSP does not need to implement it (i.e. c&p...)

I keep this as a draft for now since I am still not really sure if this does what I would expect, so any comments regarding usability of this API are welcome :)

PR #2195, opened 2020-02-18
Add full console example to console documentation

dwld:
Added example of event queue handling for full console.


sjanc:
looks good, thanks for improving our docs!

PR #2196, opened 2020-02-18
Full console: suppress new line when echo is disabled

dwld:
When echo is switched off via console_echo(0), also new line should be suppressed.

### DIFF
--- a/sys/console/full/include/console/console.h
+++ b/sys/console/full/include/console/console.h
@@ -61,6 +61,12 @@ int console_read(char *str, int cnt, int *newline);
 #endif
 void console_blocking_mode(void);
 void console_non_blocking_mode(void);
+/**
+ * Switch console echo state for input data on (on != 0) or off ( on= 0) 
+ *
+ * @param on Let console know if it should echo the input (!= 0),
+ *        or disable the echo (= 0).
+ */
 void console_echo(int on);
 
 int console_vprintf(const char *fmt, va_list ap);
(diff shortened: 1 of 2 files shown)


kasjer:
Hi @dwld, I tell you why this takes so much time with such simple reasonable change.
For a long time console code that was modified by a few people did not really handled few control characters resulting in some echo while console was configured to be silent except for NLIP protocol.
In late 2019 I changed it so not echo for CR LF (and other control characters showed) but one company that uses mynewt was affected since they relayed on this not so obvious behavior.
So 153dfb8b7d13f0cf16298fd6e3b5235909fe2f30 restored echo for CR/LF just as not to break `backward compatibility`.
I probably should have add conditional code for allowing this echo if someone really needs it, instead of always, which is not something that would be expected from the code.
So that is why no one is rushing to accept this.

dwld:
Hi @kasjer, thanks for the explanation. I understand. For our use cases, this behavior breaks the asynchronous communication. Would the introduction of another switch to suppress the echo for CR/LF be an option?

kasjer:
Hi @dwld quick question, are you using nlip protocol or some other custom protocol and you are not really interested in nlip handling?
If you are not going to use nlip (that seem like reasonable option) that I will add switch to just disable nlip handling and this will reduce code size and runtime cpu usage.
But what I've noticed when looking at you proposal is that console_echo() function that is public api for disabling console echo and it should not be used inside console code to turn on/off echo for other reason as it is done currently. And this also cries for fix.
btw you proposal for call console_echo(echo); would not make any difference and could be just totally dropped to achieve same goal.
Fixes are on the way...

PR #2200, opened 2020-02-21
cborencoder: Add functions calculating size of encoded data

de-nordic:
The functions will allow to check how much space in encoded stream would
data take, given the value or input size, in case of encoding strings.

Signed-off-by: Dominik Ermel <dominik.ermel@nordicsemi.no>


de-nordic:
@mlaz please notify whoever there is needed to review this.

vrahane:
@de-nordic We do have a have a count encoder, which does the same thing.

de-nordic:
> 
> 
> @de-nordic We do have a have a count encoder, which does the same thing.

Thx. Didn't know about it. I will use the count encoder.

de-nordic:
@vrahane, thanks I have been able to solve problem with use of count encoder. This PR is no longer needed.

PR #2201, opened 2020-02-24
Add base62 encoder/decoder

kasjer:
Base62 encoding code and unit tests.


utzig:
Why would someone use `base62` instead of `base64`?

kasjer:
it does not have / and it seems that it could be more convenient when used in URLs that use / to separate stuff.

PR #2204, opened 2020-02-24
Dialog 1V8 Power Source

brolan-juul:
Moving from DCDC to LDO source for 1V8. This will reduce noise on the rail.


andrzej-kaczmarek:
code looks good, although it does not really solve any problem since someone may want different DCDC configuration and there's no way to do this currently. since there are quite extensive settings for DCDC in Dialog, I think it would be the best if BSP can configure registers directly and then MCU code should retain them properly (to restore after wakeup) and enable DCDC. here's PR which allows this: https://github.com/apache/mynewt-core/pull/2205

PR #2206, opened 2020-02-25
Update nrfx to 2.1.0

sjanc:


utzig:
@sjanc The license checks which were failing are now passing because they were updated to use `.rat-excludes` with https://github.com/JuulLabs-OSS/mynewt-travis-ci/pull/78

PR #2207, opened 2020-02-25
Update RTT to latest available revision

andrzej-kaczmarek:


utzig:
@andrzej-kaczmarek If this code comes from some outside project, please add `hw/drivers/rtt` to `.style_ignored_dirs`.

andrzej-kaczmarek:
@utzig style check should pass on PR which updates `.style_ignored_dirs` or it will only pass on `master` after this is merged?

utzig:
> @utzig style check should pass on PR which updates `.style_ignored_dirs` or it will only pass on `master` after this is merged?

It should but I think it is buggy; unfortunately right now I can' t even open a PR on this repo to test changes; feel free to merge the update is correct.

PR #2211, opened 2020-02-26
NRF52, Re-added TASKS_RESUME in hal_i2c_master_write

ncasaril:
In mynewt_1_5 setting regs->TASKS_STARTTX was followed by setting regs->TASKS_RESUME in hal_i2c_master_write. This was removed in mynewt_1_6 resulting in rare but persistent hangs in the i2c master. I.e. once the hang was triggered no future i2c transactions could be performed. Not even disabling and reenabling the driver helped.

This change adds setting regs->TASKS_RESUME back. No hangs have been observed after this change.


sjanc:
This was removed in bb1938c0d5f448cb6a1d4152c03995652959a562 (hw/mcu/nordic: Remove unnecessary RESUME task trigger).

@andrzej-kaczmarek thoughts?

ncasaril:
Normally the RESUME trigger is not needed but if I suspect that if the previous transaction was interrupted (isr during transfer for instance) then perhaps it can shake things lose again. 
The hang only happens for me with plenty of interrupts, or debugger interrupt.

PR #2219, opened 2020-02-28
Fix json decoding of empty array

dwld:
Without the fix, decoding an empty array leads to JSON_ERR_BADTRAIL.
When peek is "positive", the read cursor needs to be forwarded to discard the closing "]".

### DIFF
--- a/encoding/json/src/json_decode.c
+++ b/encoding/json/src/json_decode.c
@@ -526,6 +526,7 @@ json_read_array(struct json_buffer *jb, const struct json_array_t *arr)
     json_skip_ws(jb);
 
     if (json_peek(jb) == ']') {
+        jb->jb_read_next(jb);
         goto breakout;
     }
 


dwld:
I have two questions:
* Does it make sense to keep both version of the json test suite (selftest and hosttest)?
* How can a recheck of the CI system be triggered? To me, it seems that the check did not fail due to actual errors caused by the commit.

ccollins476ad:
> Does it make sense to keep both version of the json test suite (selftest and hosttest)?

You're right - the host test is not really necessary.  The only reason it is there is to serve as an example host test!  I see mynewt-core has gotten a few other host tests in the meantime, so maybe this one should be removed.

> How can a recheck of the CI system be triggered? To me, it seems that the check did not fail due to actual errors caused by the commit.

I see a "Restart build" button on the Travis page, but maybe you need special permissions to see it.  I don't see any test failures though.  Was it failing earlier?

PR #2227, opened 2020-03-04
fix incorrect check in lis2ds12_set_full_scale

gonccalo:
Checking if `fs` has a value bigger than `LIS2DS12_FS_16G` is incorrect because `LIS2DS12_FS_4G` and `LIS2DS12_FS_8G` have a bigger numeric value and are valid configurations for this device.


hw/drivers/sensors/lis2ds12/src/lis2ds12.c
```diff
@@ -520,7 +520,7 @@ lis2ds12_set_full_scale(struct sensor_itf *itf, uint8_t fs)
     int rc;
     uint8_t reg;
 
-    if (fs > LIS2DS12_FS_16G) {
+    if (fs > LIS2DS12_FS_8G) {
```
kasjer: Code was wrong, thanks @gonccalo, because it rejected correct value.
This check implies that value could be wrong, but it could be wrong in more ways that this condition checks.
So condition that validates correctness of full scale argument should rather be:
if (0 != (fs & ~LIS2DS12_CTRL_REG1_FS))

Condition fs > LIS2DS12_FS_8G on the first look suggest that 8G is limit for this sensor.
So @gonccalo would you mind amending this.

gonccalo:
I also found some problems in the `lis2ds12_set_rate` function.
The `LIS2DS12_DATA_RATE_LP_10BIT_800HZ` is defined as 0x70 when it should be 0xF0 according to the documentation. The condition check for the rate argument is also not the best one (and is incorrect right now).
Screenshot from 2020-03-05 17-14-04

vrahane:
@gonccalo, appreciate your changes and bug fix. Please keep up the good work, we need all the help we can get 🙂

rymanluk:
@gonccalo , we have noticed that something went wrong when squashing and merging your PR and  your patch is in the tree but with incorrect author information. Sorry about that, we need to look closer into github to figure out what happen. Promise it will not happen again. Thanks!

utzig:
> @gonccalo , we have noticed that something went wrong when squashing and merging your PR and your patch is in the tree but with incorrect author information. Sorry about that, we need to look closer into github to figure out what happen. Promise it will not happen again. Thanks!

https://github.com/isaacs/github/issues/1368

utzig:
I'd rather disable both "Squash and merge" and "Create a merge commit" options, one is a Github rather then a git concept, the other changes how the author organized his/her commits. Authors should be responsible for squashing, rebasing, and organizing commits.

PR #2229, opened 2020-03-05
da1469x/hal_flash: Fix write boundaries

vrahane:
- Write boundaries in the hal flash write callback were partially
  correct, 0->0x80000 address remap also needs to be considered as per the datasheet.


andrzej-kaczmarek:
@ccollins476ad we need to check native QSPIF address space as well because app is linked there while bootloader is linked at remapped address space. This is because QSPI is remapped at 0 by ROM bootloader with an offset depending on product headers so our bootloader always starts at the same address, but at the same time this makes app base address change in remapped address space so it's easier to use native address space there.

PR #2233, opened 2020-03-09
Fix console echo handling

kasjer:
1. Echo could be turned on after some nlip bytes handling, now if user code disabled echo with console_echo(0) it will not be enabled in nlip code.

2 Additionally nlip handling can be disabled to reduce code size.


kasjer:
Hi @dwld could you please checked if this solution fixes problem that you detected?

utzig:
```
         default:
             insert_char(&input->line[cur], byte, end);
-            /* Falls through. */
+        /* Falls through. */
         case '\r':
-            /* Falls through. */
+        /* Falls through. */
         case '\n':
```

Funny one, the style checker does not know what "fall through" means and thinks that your comment applies to the next `case`!

dwld:
> Hi @dwld could you please checked if this solution fixes problem that you detected?

Hi @kasjer, thanks for your work. In my current project, I do not require NLIP. I just tested this PR with CONSOLE_ECHO: 0 and CONSOLE_NLIP: 0. Unfortunately, the new line is still there, since it is printed by lines 1168 and 1169 in console.c (`console_filter_out('\r');` and `console_filter_out('\n');`).

Maybe we could add an option like your introduced CONSOLE_NLIP_ECHO_LF for NLIP to enable/disable (whatever is more reasonable) the new line echo in non-NLIP mode.

PR #2239, opened 2020-03-12
dialog_da1469x: enhance otp and serial loader tools.

nkaje:
This PR brings in enhancement updates to the OTP and Serial Loading Tools.

### DIFF
--- a/hw/bsp/dialog_da1469x-dk-pro/da1469x_serial.py
+++ b/hw/bsp/dialog_da1469x-dk-pro/da1469x_serial.py
@@ -20,6 +20,8 @@ import serial
 import io
 import click
 import os
+import os.path as path
+from datetime import datetime
 
 
 @click.argument('infile')
@@ -27,7 +29,7 @@ import os
 @click.command(help='Load the provided file using serial load protocol')
 def load(infile, uart):
     try:
-        ser = serial.Serial(port=uart, baudrate=115200, timeout=10,
+        ser = serial.Serial(port=uart, baudrate=115200, timeout=0.010,
                             bytesize=8, stopbits=serial.STOPBITS_ONE)
     except serial.SerialException:
         raise SystemExit("Failed to open serial port")
@@ -51,21 +53,32 @@ def load(infile, uart):
     # - If XOR matches, host sends 0x6 as final ack
     # - board boots image after receiving ACK from host
 
-#    if len(msg) == 0:
-#        raise SystemExit("Read timed out, exiting")
-
-    print("Please reset board to enter ROM uart recovery")
+    som_detected = False
+    reset_triggered = False
+    prev = datetime.now()
+    reset_delay_us = 250000
 
     while True:
+        elapsed = datetime.now() - prev
+        if elapsed.seconds >= 15:
+            raise SystemExit("Failed to receive SOM, aborting")
+        if not som_detected and not reset_triggered:
+            if elapsed.microseconds >= reset_delay_us:
+                print("Triggering SWD reset...")
+                reset = path.join(path.dirname(__file__), 'reset.sh &')
+                os.system(reset)
+                reset_triggered = True
+
         msg = ser.read(1)
-        print(msg)
-        if msg[0] == 0x2:
+        if len(msg) > 0 and msg[0] == 0x2:
+            print(msg)
+            som_detected = True
             print("Detected serial boot protocol")
             msg = bytes([0x1])
             msg += bytes([len(data) & 0xff])
             msg += bytes([len(data) >> 8])
             ser.write(msg)
-
+            ser.timeout = 10
             msg = ser.read()
             if len(msg) == 0:
                 raise SystemExit("Failed to receive SOH ACK, aborting")
--- a/hw/bsp/dialog_da1469x-dk-pro/da1469x_serial.py
+++ b/hw/bsp/dialog_da1469x-dk-pro/da1469x_serial.py
@@ -26,8 +26,9 @@ from datetime import datetime
 
 @click.argument('infile')
 @click.option('-u', '--uart', required=True, help='uart port')
+@click.option('-r', '--reset_script', required=True, help='Custom reset script to switch to serial load')
 @click.command(help='Load the provided file using serial load protocol')
-def load(infile, uart):
+def load(infile, uart, reset_script):
     try:
         ser = serial.Serial(port=uart, baudrate=115200, timeout=0.010,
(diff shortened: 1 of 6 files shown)


hw/bsp/dialog_da1469x-dk-pro/da1469x_serial.py
```diff
@@ -51,21 +52,32 @@ def load(infile, uart):
...
+    reset_delay_us = 250000
 
     while True:
+        elapsed = datetime.datetime.now() - now
+        if elapsed.seconds >= 15:
+            raise SystemExit("Failed to receive SOM, aborting")
+        if not som_detected and not reset_triggered:
+            if elapsed.microseconds >= reset_delay_us:
+                print("Triggering SWD reset...")
+                os.system("./reset.sh &")
```
utzig: Doing this the script can only run `reset.sh` in the local directory, if I run it from repo root, eg, `./hw/bsp/dialog_da1469x-dk-pro/da1469x_serial.py`, it would not find the reset script. Not strictly required though.

hw/bsp/dialog_da1469x-dk-pro/otp_tool.py
```diff
@@ -593,16 +616,18 @@ def test_alive_target(uart):
         raise SystemExit("Failed to write to %s" % uart)
 
     data = ser.read(rlen)
-    print(data)
     if len(data) != rlen:
         raise SystemExit("Failed to receive response, exiting")
 
     response = cmd_response._make(struct.unpack_from('IIII', data))
+    if response.som != 0x55aa55aa:
+        raise SystemExit("SOM not valid, invalid response")
```
utzig: Some copy/paste here. Although the checks must be done there could be a bit or reuse so in case `0x55aa55aa` or the error message needs to change no find/replace is required.

hw/bsp/dialog_da1469x-dk-pro/reset.sh
```diff
@@ -0,0 +1,66 @@
...
+openocd -f ${CFG_FILE_DIR}/ft4232h.cfg -f ${CFG_FILE_DIR}/steam.cfg -f ${CFG_FILE_DIR}/bindto.cfg -d > .openocd_log 2>&1 &"
+EOF
+fi
+cat >> ${GDB_CMD_FILE} << EOF
+target remote ${ip}:3333
+mon adapter_khz ${freq}
+mon reset halt
+mon resume
+
+set *(int *)0x100C0050 = 1
```
utzig: What does this do?

hw/bsp/dialog_da1469x-dk-pro/reset.sh
```diff
@@ -0,0 +1,66 @@
...
+  GDB=arm-none-eabi-gdb
+fi
+
+GDB_CMD_FILE=${GDB_CMD_FILE:-".gdb_load"}
+OPENOCD_LOG_FILE=${OPENOCD_LOG_FILE:-".openocd_log"}
+CFG_FILE_DIR=${CFG_FILE_DIR:-"/srv/juul-test/jtest/openocd"}
+
+cat > ${GDB_CMD_FILE} << EOF
+set pagination off
+EOF
```
utzig: Adding an extra empty line here should make it easier to read.

hw/bsp/dialog_da1469x-dk-pro/reset.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ "$ip" == "localhost" ]; then
+cat >> ${GDB_CMD_FILE} << EOF
```
utzig: This and the `cat` in the `else` could be indented, no?

hw/bsp/dialog_da1469x-dk-pro/reset.sh
```diff
@@ -0,0 +1,66 @@
...
+shell sh -c "trap '' 2; openocd -f ${CFG_FILE_DIR}/ft4232h.cfg -f ${CFG_FILE_DIR}/steam.cfg -d > ${OPENOCD_LOG_FILE} 2>&1 &"
+EOF
+else
+cat >> ${GDB_CMD_FILE} << EOF
+shell sh -c "trap '' 2; ssh -tt osf@$ip docker-compose exec jtest \
+openocd -f ${CFG_FILE_DIR}/ft4232h.cfg -f ${CFG_FILE_DIR}/steam.cfg -f ${CFG_FILE_DIR}/bindto.cfg -d > .openocd_log 2>&1 &"
+EOF
+fi
```
utzig: Adding an empty line should make it easier to read.

hw/bsp/dialog_da1469x-dk-pro/da1469x_serial.py
agross-korg: ah good point.  this really needs to be "reset.sh &" and the pathing must include the path to it.

utzig: I am not a big run of `os.system` either, so I would use subprocess module. You could fix this with:

```
import os.path as path
import subprocess

reset = path.join(path.dirname(__file__), 'reset.sh')
subprocess.call(reset.split())
```

nkaje: For some reason, the serial read from the device after the reset doesn't receive expected data. I addressed all other comments except for switching os.system -> subprocess.call()

hw/bsp/dialog_da1469x-dk-pro/reset.sh
nkaje: HW RESET REQ. added a comment.

utzig: > HW RESET REQ. added a comment.

Yeah, I checked the DS this morning, I think it would be possible to add a comment but it's OK anyway.

nkaje:
Logs from OpenOCD based system
```
./da1469x_serial_new.py load -u /dev/ttymxc4 -r ./reset_ocd.sh 
Triggering SWD reset...
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x16011b16 in ?? ()
adapter speed: 1000 kHz
target halted due to debug-request, current mode: Thread 
xPSR: 0xf9000000 pc: 0x2000028c msp: 0x20040000
b'\x02'
Detected serial boot protocol
Loading application to RAM
```

vrahane:
Changes look fine. I merged it since it was approved by both @utzig and @agross-korg

PR #2242, opened 2020-03-18
Prepare for Apache Mynewt 1.8.0 release

sjanc:


repository.yml
```diff
@@ -36,6 +36,7 @@ repo.versions:
     "1.5.0": "mynewt_1_5_0_tag"
     "1.6.0": "mynewt_1_6_0_tag"
     "1.7.0": "mynewt_1_7_0_tag"
+    "1.8.0": "mynewt_1_8_0_tag"
```
utzig: Shouldn't it be `mynewt_1_8_0_rc1_tag`?

sjanc: no, it was never needed
we used to have both rc and non-rc tags for rc release

for 1.8 we don't even need non-rc tag since newt should be able to handle those now :) (fingers crossed)

PR #2245, opened 2020-03-20
STM32 Allow reboot to bootloader

kasjer:
STM32 MCUs have embedded bootloader that can be activated on reset when BOOT pin(s) are in certain state.
This change allows to start bootloader from application that is running from flash, so bootloader can be activated even if BOOT pins are soldered to boot system from flash.

### DIFF
--- a/hw/mcu/stm/stm32_common/include/stm32_common/mcu.h
+++ b/hw/mcu/stm/stm32_common/include/stm32_common/mcu.h
@@ -106,6 +106,8 @@ extern "C" {
  */
 #define MCU_AFIO_PIN_NONE       (0xFFFF)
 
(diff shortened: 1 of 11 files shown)


utzig:
@kasjer This looks nice, I'm gonna try to test it this afternoon.

brianwyld:
Hi,
Does this mean it may have an impact on using HSE with STM32?
thanks

Brian

kasjer:
> Hi, Does this mean it may have an impact on using HSE with STM32? thanks Brian

Hi @brianwyld, this is adds new function that can be used to start ST bootloader.
If you don't use this function in your code, there is no impact.

brianwyld:
Ok, thanks!

PR #2252, opened 2020-03-27
[RFC] I2S API and driver for STM32 and NRF52 MCUs

kasjer:
I2S interface is present on several supported boards.
This introduces API for I2S input and output devices.
Drivers for following MCUs are included
- STM32F4xx
- STM32F1xx (not tested)
- NRF52


kasjer:
PR split into few smaller ones

PR #2253, opened 2020-03-27
bsp/pinetime: Add bsp for PineTime smartwatch

caspermeijn:
This adds the bsp boilerplate including the pin definitions for the peripherals. A description of the device can be found here: https://wiki.pine64.org/index.php/PineTime

I plan to add drivers for the peripherals later.

### DIFF
--- a/hw/bsp/pinetime/bsp.yml
+++ b/hw/bsp/pinetime/bsp.yml
@@ -1,0 +1,62 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+bsp.arch: cortex_m4
+bsp.compiler: compiler/arm-none-eabi-m4
+bsp.linkerscript:
+    - "hw/bsp/pinetime/pinetime.ld"
+    - "@apache-mynewt-core/hw/mcu/nordic/nrf52xxx/nrf52.ld"
+bsp.linkerscript.BOOT_LOADER.OVERWRITE:
+    - "hw/bsp/pinetime/boot-pinetime.ld"
+    - "@apache-mynewt-core/hw/mcu/nordic/nrf52xxx/nrf52.ld"
+bsp.part2linkerscript: "hw/bsp/pinetime/split-pinetime.ld"
+bsp.downloadscript: "hw/bsp/pinetime/pinetime_download.sh"
+bsp.debugscript: "hw/bsp/pinetime/pinetime_debug.sh"
+
+bsp.flash_map:
+    areas:
+        # System areas.
+        FLASH_AREA_BOOTLOADER:
+            device: 0
+            offset: 0x00000000
+            size: 16kB
+        FLASH_AREA_IMAGE_0:
+            device: 0
+            offset: 0x00008000
+            size: 232kB
+        FLASH_AREA_IMAGE_1:
+            device: 0
+            offset: 0x00042000
+            size: 232kB
+        FLASH_AREA_IMAGE_SCRATCH:
+            device: 0
+            offset: 0x0007c000
+            size: 4kB
+
+        # User areas.
+        FLASH_AREA_REBOOT_LOG:
+            user_id: 0
+            device: 0
+            offset: 0x00004000
+            size: 16kB
+        FLASH_AREA_NFFS:
+            user_id: 1
+            device: 0
+            offset: 0x0007d000
+            size: 12kB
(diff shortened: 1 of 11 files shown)


hw/bsp/pinetime/pkg.yml
```diff
@@ -0,0 +1,37 @@
...
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+    - nrf52
+    - Pine64
+    - PineTime
+
+pkg.cflags:
+    - '-DNRF52'
+    # TODO: Move to more specific variable, however this currently causes a build failure '-DNRF52832_XXAA'
```
mlaz: I'm wondering which would be the build failure here. It looks like NRF52832_XXAA would be the right value.

hw/bsp/pinetime/syscfg.yml
```diff
@@ -0,0 +1,53 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# Settings this BSP defines.
+syscfg.defs:
+
+# Settings this BSP overrides.
+syscfg.vals:
```
mlaz: Any reasons for not having `CONFIG_FCB_FLASH_AREA`, `REBOOT_LOG_FLASH_AREA` and `COREDUMP_FLASH_AREA` set ?
Also there are no timer configurations for BLE controller, is there a reason for that?

hw/bsp/pinetime/README.md
```diff
@@ -0,0 +1,56 @@
...
+newt build boot-pinetime
+newt load boot-pinetime
+
+newt target create blinky-pinetime
+newt target set blinky-pinetime \
+    app=apps/blinky \
+    bsp=@apache-mynewt-core/hw/bsp/pinetime \
+    build_profile=debug
+newt run blinky-pinetime 0
+```
```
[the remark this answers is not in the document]
caspermeijn: I am not a natural writer, but I like the idea of a blinky tutorial. The [nRF52](https://mynewt.apache.org/latest/tutorials/blinky/nRF52.html) tutorial helped me a lot, so I will try to create a page like that.

However I would like to document the status of the BSP, as not drivers are available yet. Do you have ideas on how to do that?

hw/bsp/pinetime/pkg.yml
caspermeijn: I believe some library didn't build because it expected `NRF52`. I will look at it again and fix the actual problem ;-)

hw/bsp/pinetime/syscfg.yml
caspermeijn: No, I don't have a reason for that, other than that is was not in the [BSP porting tutorial](https://mynewt.apache.org/latest/os/core_os/porting/port_bsp.html). Do you have a good example of that?

hw/bsp/pinetime/pkg.yml
caspermeijn: So the first build problem I get is in `hw/mcu/nordic/nrf52xxx/include/nrfx_config.h`. This seems easily solvable, however there are more assumptions in the code for `NRF52` to be defined. Most of the bsp use `-DNRF52`, so I suggest to leave the TODO as-is and fix this later.

hw/bsp/pinetime/syscfg.yml
caspermeijn: I created a PR for the blinky tutorial: https://github.com/apache/mynewt-documentation/pull/91

And I added the flash areas as with most other BSP.

mlaz: Yes, take look at [bsp/nordic_pca10040](https://github.com/apache/mynewt-core/blob/master/hw/bsp/nordic_pca10040) . You will probably need to upbate your `syscfg.yml` and `hal_bsp_get_nvic_priority` to be able to build and run [bleprph](https://github.com/apache/mynewt-nimble/tree/master/apps/bleprph).

hw/bsp/pinetime/pkg.yml
mlaz: Ok, nevermind, this is dealt with on nrfx, `NRF52` is ok.

hw/bsp/pinetime/README.md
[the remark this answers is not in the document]
caspermeijn: I created a blinky tutorial for PineTime. I also creates a new section in OS User Guide for Board support. The first page is the status of PineTime BSP.

caspermeijn:
I split the documentation into a seperate PR in the hope to make progress on the BSP while the docs are discussed https://github.com/apache/mynewt-core/pull/2254

hw/bsp/pinetime/bsp.yml
```diff
@@ -0,0 +1,62 @@
...
+
+bsp.arch: cortex_m4
+bsp.compiler: compiler/arm-none-eabi-m4
+bsp.linkerscript:
+    - "hw/bsp/pinetime/pinetime.ld"
+    - "@apache-mynewt-core/hw/mcu/nordic/nrf52xxx/nrf52.ld"
+bsp.linkerscript.BOOT_LOADER.OVERWRITE:
+    - "hw/bsp/pinetime/boot-pinetime.ld"
+    - "@apache-mynewt-core/hw/mcu/nordic/nrf52xxx/nrf52.ld"
+bsp.part2linkerscript: "hw/bsp/pinetime/split-pinetime.ld"
```
danielkucera, line 28: where is the `split-pinetime.ld` file???

caspermeijn: I think this line should not be there. I think this is supposed to be part of [split images](http://mynewt.apache.org/latest/os/modules/split/split.html). But this is currently not implemented for PineTime.

PR #2254, opened 2020-03-29
docs/os/bsp: Add board support section

caspermeijn:
Add a new section to the documentation which lists the status of BSPs.
The first page is the status of the PineTime BSP.

This is part of an effort to add PineTime support to mynewt. This documents the progress of that effort.


mlaz:
Ok, I changed the title of this PR to [RFC] since this is a structural change to doc. Anyway I agree with it and I think we should encourage new BSP PRs to also include an entry under this new section.

PR #2256, opened 2020-04-01
sys/id: Add optional id/serial_mfg field for storing manufacturer serial numbers

benmccrea:
This PR adds a new optional field, `id/serial_mfg` to be used for alternate serial numbers on devices during the manufacturing process.

The syscfg `ID_SERIAL_MFG_PRESENT` is disabled by default.


benmccrea:
Closing in favor of 2257

PR #2259, opened 2020-04-02
ARM semihosting console on pinetime

caspermeijn:
This adds ARM semihosting console support to mynewt. It is only the tx side. And it is enabled by default for PineTime BSP.

This first commits adds the actual semihosting driver. It originates from mbed-os, with some modification to make it compatible with mynewt.

The second commit adds semihosting to the full console implementation.

The third commit adds support for the CFG_POST_INIT variable in the openocd script. This was needed to enable semihosting during debug sessions.

The last commit edits the PineTime BSP to enable the semihosting console in syscfg and debug.sh.

### DIFF
--- a/.style_ignored_dirs
+++ b/.style_ignored_dirs
@@ -5,6 +5,7 @@ apps/coremark
 crypto/mbedtls
 crypto/tinycrypt
 hw/drivers/rtt
+hw/drivers/semihosting
 libc/baselibc
 net/ip/lwip_base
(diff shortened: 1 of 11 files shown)


caspermeijn:
I changed to using GDB_CMD_FILE instead of CFG_POST_INIT, which is a bit more elegant

caspermeijn:
Added raw semihosting driver to the .style_ignored_dirs

PR #2267, opened 2020-04-14
I2S API for mynewt

kasjer:
This introduces API for accessing I2S devices.
API handles input, output and duplex mode.
Further PRs with drivers for different MCU will follow.


kasjer:
Code was present in #2268

PR #2269, opened 2020-04-15
hw/drivers/i2s: Add I2S driver for NRF52 family

kasjer:
I2S driver implementation for NRF52 MCUs.

### DIFF
--- a/hw/drivers/i2s/i2s_nrf52/src/i2s_nrf52.c
+++ b/hw/drivers/i2s/i2s_nrf52/src/i2s_nrf52.c
@@ -1,0 +1,235 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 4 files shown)


hw/drivers/i2s/i2s_nrf52/src/i2s_nrf52.c
```diff
@@ -0,0 +1,235 @@
...
+        i2s->sample_size_in_bytes = 4;
+        break;
+    }
+
+    i2s->direction = I2S_INVALID;
+    if (cfg->nrfx_i2s_cfg.sdin_pin != NRFX_I2S_PIN_NOT_USED) {
+        i2s->direction = I2S_IN;
+    }
+    if (cfg->nrfx_i2s_cfg.sdout_pin != NRFX_I2S_PIN_NOT_USED) {
+        i2s->direction |= I2S_OUT;
```
sjanc, line 130: while the resulting value seems fine it is a bit unclear to do that with enum values

kasjer: Values were designed that way so I2S_IN and I2S_OUT have separate bits and it's easier to check if it's input, output or both for duplex mode.

sjanc: ok, fair enough :)

PR #2271, opened 2020-04-17
hw/drivers/i2s: Add I2S driver for STM32F1 family

kasjer:
Code implements I2S driver for STM32F1xx MCUs that have it.

### DIFF
--- a/hw/drivers/i2s/i2s_stm32f1/src/i2s_stm32f1.c
+++ b/hw/drivers/i2s/i2s_stm32f1/src/i2s_stm32f1.c
@@ -1,0 +1,475 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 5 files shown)


hw/drivers/i2s/i2s_stm32f1/src/i2s_stm32f1.c
```diff
@@ -0,0 +1,475 @@
...
+
+    if (cfg->mode == I2S_MODE_MASTER_TX || cfg->mode == I2S_MODE_SLAVE_TX) {
+        __HAL_LINKDMA(&stm32_i2s->hi2s, hdmatx, stm32_i2s->hdma_spi);
+    } else {
+        __HAL_LINKDMA(&stm32_i2s->hi2s, hdmarx, stm32_i2s->hdma_spi);
+    }
+
+    i2s_init_interrupts(cfg);
+end:
+    return rc;
```
utzig, line 236: With a `return` as the only thing here, you could also just `return` in the failure tests; it is no clear if it was done on purpose.

PR #2275, opened 2020-04-18
Battery for PineTime BSP

caspermeijn:
This initializes the ADC for battery voltage measurement for PineTime BSP.
This also places a ifdef around the  battery shell code, as it doesn't compile when BATTERY_SHELL is disabled.

### DIFF
--- a/hw/bsp/pinetime/src/hal_bsp.c
+++ b/hw/bsp/pinetime/src/hal_bsp.c
@@ -30,6 +30,9 @@
 #include "mcu/nrf52_hal.h"
 #include "mcu/nrf52_periph.h"
 #include "sgm4056/sgm4056.h"
+#include "battery/battery_adc.h"
+#include "adc_nrf52/adc_nrf52.h"
+#include <nrf_saadc.h>
 
 /** What memory to include in coredump. */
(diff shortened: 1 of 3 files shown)


hw/bsp/pinetime/src/hal_bsp.c
```diff
@@ -89,6 +92,51 @@ hal_bsp_get_nvic_priority(int irq_num, uint32_t pri)
...
+static struct adc_chan_cfg hal_bsp_adc_channel_config = {
+    .gain = ADC_GAIN1_6,
+    .reference = ADC_REFERENCE_INTERNAL,
+    .acq_time = ADC_ACQTIME_10US,
+    .pin = NRF_SAADC_INPUT_AIN7,
+    .differential = false,
+    .pin_negative = -1,
+};
+
+static struct battery hal_bsp_battery_dev = {};
```
kasjer, line 110: It would be the first time in mynewt code to put = {}; for non stack variable. Not that there is something wrong with this it's just not happened before.

caspermeijn: You are right. This is only valid in C++. And if I read the C spec right it will be initialized to zero because of the static declaration.

PR #2279, opened 2020-04-30
TinyUSB integration

kasjer:
## TinyUSB integration ##
TinyUSB is know for a while to have support for mynewt.
This work makes it easier to use it with mynewt base projects.
This work includes:
- USB task creation
- mynewt glue to CDC
- mynewt glue to HID
- hardware initialization code that could not be used as is from TinyUSB repository (BSP code duplicates in both repositories)
- Console over USB that can be used instead of UART or RTT

## ~~Before this can be merge~~ ##
~~**newt** tool needs to support **pkg.include_dirs** keyword that is introduced in by this PR https://github.com/apache/mynewt-newt/pull/397~~
~~Once this is approved pull request to TinyUSB repository will be made https://github.com/hathach/tinyusb/pull/392 to change to **pkg.yml** to use this new keyword.
Without this it is possible to use TinyUSB but a lot of packages would need~~
```yml
pkg.cflags:
    - "-I@tinyusb/src"
```


hw/usb/tinyusb/README.md
```diff
@@ -0,0 +1,48 @@
...
+To easy add support for CDC or HID class device simply add
+**hw/usb/tinyusb/std_descriptors** package to project
+```yaml
+pkg.deps:
+    - "@apache-mynewt-core/hw/usb/tinyusb/std_descriptors"
+```
+
+To use console over USB, set syscfg value
+```yaml
+    CONSOLE_USB: 1
```
rymanluk, line 47: We need also USBD_CDC to be set as it is not done automatically.
Need to figure out how to do it automatically.

mlaz:
I just tested with stm32f411 and slinky works perfectly. With nrf52840 I am getting the CDC tty device but no console.

hw/usb/tinyusb/cdc_console/src/cdc_console.c
```diff
@@ -0,0 +1,147 @@
...
+    return c;
+}
+
+void
+console_rx_restart(void)
+{
+    os_eventq_put(os_eventq_dflt_get(), &rx_receive_event);
+}
+
+bool connected;
```
sjanc, line 70: nitpick: could be static and usually those are declared on top

sys/console/full/syscfg.yml
```diff
@@ -26,6 +26,9 @@ syscfg.defs:
     CONSOLE_SEMIHOSTING:
         description: 'Set console output to ARM semihosting'
         value: 0
+    CONSOLE_USB:
```
mlaz, line 29: This can't be enabled if `CONSOLE_UART` is, it would be nice to have it described on `syscfg.restrictions`

PR #2291, opened 2020-05-18
libc/baselibc: Fix definitions for printing 32bit integers

sjanc:


ncasaril:
> I think this header should be gone or reduced to minimum.
> PRI... macros are provided by **inttypes.h** from **gcc**.

Perhaps the solution then is to remove the inttypes.h and use the one from the compiler instead? 
From what I can see the declarations for **strmtoimax** and **strntoumax** will need to be moved somewhere else in that case. Other than that, I think this would be the cleanest solution.

sjanc:
agreed, I've updated PR, inttypes.h is now gone and extensions were moved to separate header file

kasjer:
There is another problem I'm afraid.
Old, removed version of **inttypes.h** defined **PRId8 PRIi8 PRIo8 PRIu8 PRIx8 PRIX8 PRId16 PRIi16 PRIo16 PRIu16 PRIx16 PRIX16** without **hh** or **h** just d,u,i,x and such.
GCC version of those have hh or h prefix so for printing uint8_t we have `"%hhu"` format specifier uint16_t `"%hu"`.
This h is not handled in printf-like functions and it will lead runtime break in nimble **ble_uind.c** and in cbor **cbor_pretty.c**.
h and hh is handled in scanf-like so it make sense to support it in printf as well.

caspermeijn:
It seems like this change breaks my build in a surprising way. I use Rust for my application and that uses the clang backend. When clang parses the mynewt header files it needs  `#include <inttypes.h>` to understand the uint8_t. However after merging this pull request I get the error `/usr/lib64/clang/10.0.0/include/inttypes.h:21:15: fatal error: 'inttypes.h' file not found`. 

It seems like clang provided version of inttypes.h is just a placeholder as it does `#include_next <inttypes.h>`, which it apparently can't find.

If I revert commit c1f0300413fc981741d37d655654de1dbca5d4df, then the build succeeds again.

PR #2294, opened 2020-05-26
mbedtls_aes_alt: add setkey for decyption, validate input, fix data type and return type

nkaje:
Let https://github.com/apache/mynewt-core/pull/2175 merge and then merge the top change.

### DIFF
--- a/hw/drivers/crypto/src/mbedtls_aes_alt.c
+++ b/hw/drivers/crypto/src/mbedtls_aes_alt.c
@@ -23,6 +23,7 @@
 
 #include "crypto/crypto.h"
 #include "crypto/aes_alt.h"
+#include "mbedtls/aes.h"
 
 void
 mbedtls_aes_init(mbedtls_aes_context *ctx)
@@ -40,9 +41,9 @@ mbedtls_aes_free(mbedtls_aes_context *ctx)
     memset(ctx, 0, sizeof(*ctx));
 }
 
-int
-mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
-        unsigned int keybits)
+static int
+mbedtls_aes_setkey(mbedtls_aes_context *ctx, const unsigned char *key,
+                   unsigned int keybits)
 {
     switch (keybits) {
     case AES_128_KEY_LEN * 8:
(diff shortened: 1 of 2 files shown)


utzig:
> Let #2175 merge and then merge the top change.

Just merged it!

hw/drivers/crypto/src/mbedtls_aes_alt.c
```diff
@@ -53,15 +54,59 @@ mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
...
 int
 mbedtls_aes_crypt_ecb(mbedtls_aes_context *ctx, int mode,
-        const unsigned char input[16], unsigned char output[16])
+                      const unsigned char input[16], unsigned char output[16])
 {
-    return crypto_encrypt_aes_ecb(ctx->crypto, ctx->key, ctx->keylen,
-            (const uint8_t *)input, (uint8_t *)output, AES_BLOCK_LEN);
+    int ret = crypto_encrypt_aes_ecb(ctx->crypto, ctx->key, ctx->keylen,
+                                     (const uint8_t *)input, (uint8_t *)output, AES_BLOCK_LEN);
+    if (ret == AES_BLOCK_LEN) {
```
utzig, line 105: I think `return (ret == AES_BLOCK_LEN) ? 0 : -1;` would look a bit better in this case.

hw/drivers/crypto/src/mbedtls_aes_alt.c
```diff
@@ -53,15 +54,59 @@ mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
         return 0;
     }
 
-    return -1;
+    return MBEDTLS_ERR_AES_INVALID_KEY_LENGTH;
+}
+
+int
```
utzig, line 60: Could this be `static`?

utzig:
I am under the impression that `mbedtls_aes_setkey_enc` and `mbedtls_aes_setkey_dec` do the same thing, so maybe the code could be moved into a single function? Nice catch on the error return, the reason for only having `_enc` is that we only use CTR-mode for flash and image encryption and the CTR code in mbedTLS does not check return values!

hw/drivers/crypto/src/mbedtls_aes_alt.c
```diff
@@ -41,8 +42,8 @@ mbedtls_aes_free(mbedtls_aes_context *ctx)
 }
 
 int
```
utzig, line 44: `static` maybe?

hw/drivers/crypto/src/mbedtls_aes_alt.c
nkaje: done

hw/drivers/crypto/src/mbedtls_aes_alt.c
nkaje: done

nkaje:
> I am under the impression that `mbedtls_aes_setkey_enc` and `mbedtls_aes_setkey_dec` do the same thing, so maybe the code could be moved into a single function? Nice catch on the error return, the reason for only having `_enc` is that we only use CTR-mode for flash and image encryption and the CTR code in mbedTLS does not check return values!

made a common function.

nkaje:
> LGTM
> 
> The `return (ret == AES_BLOCK_LEN ? 0 : -1);` looks a bit weird to me, because I am very used to `return (ret == AES_BLOCK_LEN) ? 0 : -1;`, but given the low precedence of the ternary conditional it should be fine!

done.

PR #2296, opened 2020-05-27
crypto_test: add syscfg option to use RAM based input

nkaje:
Some targets (such as Dialog DA1469x), Crypto block can not
readily access the input buffer in flash. In most use-cases,
the buffers reside in RAM. Add a syscfg option to move the input
to RAM before invoking crypto functions. By default this
config is disabled.

### DIFF
--- a/apps/crypto_test/src/main.c
+++ b/apps/crypto_test/src/main.c
@@ -171,6 +171,9 @@ run_test_vectors(struct crypto_dev *crypto, struct test_vectors *test_mode)
     int i;
     uint32_t sz;
     uint32_t asksz;
+#if MYNEWT_VAL(CRYPTOTEST_LOCAL_BUF)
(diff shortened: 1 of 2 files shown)


kasjer:
I'm not convinced that this is correct solution to problem.
If accessing flash is problematic on some platform driver for this platform should have this switch inside.
Considering that additional buffer is taken from stack it would not matter if this stack was allocated from stack in driver code.
If test application can suffer from bad memory access so can any other user app, so why apply cure to this test app only.

utzig:
> I'm not convinced that this is correct solution to problem.

This is not a driver, it's a test app, and I believe this is a reasonable trade-off to have as optional while developing a driver further. The da1469x can use flash as input for the crypto unit as long as the proper memory mappings are done, which is not available just yet.

nkaje:
https://github.com/apache/mynewt-core/pull/2327 is added to check the address and mapping in the crypto driver.

PR #2297, opened 2020-05-28
hw/mcu/dialog: Add callbacks on enter/exit sleep

andrzej-kaczmarek:
This patch adds callbacks to be called before and after deep sleep.
Those are intended to be registered by BSP which can then decide to
save/restore state of certain peripherals and release/acquire power
domains in order to save some power while M33 goes into deep sleep.

Note that similar power saving can usually be achieved if application
uses os_dev_open/close consistently, i.e. opens device only when
needed, which should properly acquire and release power domains
(although the actual behavior depends on a driver implementation).
However, there may still be some cases where quick save/restore is
possible and this can be handled in new callbacks.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/da1469x_sleep.c
+++ b/hw/mcu/dialog/da1469x/src/da1469x_sleep.c
@@ -22,6 +22,7 @@
 #include "mcu/da1469x_pd.h"
 #include "mcu/da1469x_pdc.h"
 #include "mcu/da1469x_prail.h"
+#include "mcu/da1469x_sleep.h"
 #include "mcu/mcu.h"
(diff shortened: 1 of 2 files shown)


hw/mcu/dialog/da1469x/src/da1469x_sleep.c
```diff
@@ -110,4 +128,9 @@ da1469x_sleep(os_time_t ticks)
     __WFI();
 }
 
+void
```
kasjer, line 131: This looks like function is duplicated. Does it compile?

kasjer: I did not noticed #ifdef no problem

PR #2300, opened 2020-05-29
TCP/IPv4 adaptor for CoAP

ccollins476ad:
This transport uses mn_sockets.

Unlike other transports, this one does not listen for new connections on its own.  Instead, the application needs to populate the transport with connections to be managed using oc_tcp4_add_conn().  There are a few too many parameters that need to be specified when listening for TCP connections.

### DIFF
--- a/net/oic/include/oic/port/mynewt/stream.h
+++ b/net/oic/include/oic/port/mynewt/stream.h
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * stream.h: Utility code for adaptors that implement the CoAP-over-TCP
+ * protocol (Bluetooth, TCP/IP, etc.).
+ */
+
+#ifndef H_MYNEWT_TCP_
+#define H_MYNEWT_TCP_
+
+#include "os/mynewt.h"
+#include "oic/oc_ri.h"
+#include "oic/oc_ri_const.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Indicates whether a transport-specific endpoint matches the provided
+ * description.
+ */
+typedef bool oc_stream_ep_match(const void *ep, const void *ep_desc);
+
+/**
+ * Fills the given endpoint data structure according to the provided
+ * description.
+ */
+typedef void oc_stream_ep_fill(void *ep, const void *ep_desc);
+
+/**
+ * Used for reassembling CoAP-over-TCP packets.  A transport should only have
+ * one reassembler.
+ *
+ * The user must initialize ALL members.  After initialization, the user should
+ * not directly access any members.
+ */
+struct oc_stream_reassembler {
+    STAILQ_HEAD(, os_mbuf_pkthdr) pkt_q; /* Use STAILQ_HEAD_INITIALIZER() */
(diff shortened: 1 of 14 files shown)


vrahane:
I guess we can call datagram transport as UDP but well it won't be a correct term for BLE :-)

vrahane:
@ccollins476ad I think there is a style check issue most probably, hence the CI failure, rest looks good.

net/oic/src/messaging/coap/engine.c
```diff
@@ -231,17 +231,19 @@ coap_receive(struct os_mbuf **mp)
...
+            /* handle responses */
+            if (message->type == COAP_TYPE_CON) {
+                erbium_status_code = EMPTY_ACK_RESPONSE;
+            } else if (message->type == COAP_TYPE_ACK) {
+                /* transactions are closed through lookup below */
+            } else if (message->type == COAP_TYPE_RST) {
 #ifdef OC_SERVER
-            /* cancel possible subscriptions */
-            coap_remove_observer_by_mid(OC_MBUF_ENDPOINT(m), message->mid);
+                /* cancel possible subscriptions */
```
vrahane, line 243: more spaces than required

ccollins476ad: I think it is correct.  The `#ifdef` just makes it look confusing.

```
        if (!oc_endpoint_use_tcp(&endpoint)) {
            /* handle responses */
            if (message->type == COAP_TYPE_CON) {
                erbium_status_code = EMPTY_ACK_RESPONSE;
            } else if (message->type == COAP_TYPE_ACK) {
                /* transactions are closed through lookup below */
            } else if (message->type == COAP_TYPE_RST) {
#ifdef OC_SERVER
                /* cancel possible subscriptions */
                coap_remove_observer_by_mid(OC_MBUF_ENDPOINT(m), message->mid);
#endif
            }
```

net/oic/include/oic/port/mynewt/stream.h
```diff
@@ -0,0 +1,95 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Indicates whether a transport-specific endpoint matches the provided
+ * description.
+ */
+typedef bool oc_stream_ep_match(const void *ep, void *ep_desc);
```
kasjer, line 40: Any reason why only ep is const?

net/oic/src/port/mynewt/stream.c
```diff
@@ -0,0 +1,112 @@
...
+#include "oic/messaging/coap/coap.h"
+
+int
+oc_stream_reass(struct oc_stream_reassembler *r, struct os_mbuf *om1,
+                void *ep_desc, struct os_mbuf **out_pkt)
+{
+    struct os_mbuf_pkthdr *pkt1;
+    struct os_mbuf_pkthdr *pkt2;
+    struct os_mbuf *om2;
+    uint8_t hdr[sizeof (struct coap_tcp_hdr32)];
```
kasjer, line 36: This is the only place where space is after sizeof

net/oic/src/port/mynewt/tcp4_adaptor.c
```diff
@@ -0,0 +1,425 @@
...
+ * under the License.
+ */
+
+#include <assert.h>
+#include <string.h>
+#include <stdint.h>
+
+#include "os/mynewt.h"
+
+uint8_t oc_tcp4_transport_id = -1;
```
kasjer, line 26: Slightly unorthodox way to initialize unsigned value.
Other place in this commit use UINT8_MAX

net/oic/src/port/mynewt/tcp4_adaptor.c
```diff
@@ -0,0 +1,425 @@
...
+    struct mn_socket *sock;
+    oc_tcp4_err_fn *err_cb;
+    void *err_arg;
+    SLIST_ENTRY(oc_tcp4_conn) next;
+};
+
+struct os_mempool oc_tcp4_conn_pool;
+static os_membuf_t oc_tcp4_conn_buf[
+    OS_MEMPOOL_SIZE(MYNEWT_VAL(OC_TCP4_MAX_CONNS),
+                    sizeof (struct oc_tcp4_conn))
```
kasjer, line 120: Second place with space after sizeof few lines above there was no space

net/oic/include/oic/port/mynewt/stream.h
ccollins476ad: Thanks.  I agree `ep_desc` should also be const here.  I also think it should be const in the "fill" callback.  I have fixed both.

net/oic/src/port/mynewt/stream.c
ccollins476ad: Thanks, I have fixed this (and the other style issues you identified).

PR #2301, opened 2020-05-29
oic tcp4_adaptor: Don't loop if nothing to tx

ccollins476ad:
This is a fix that I meant to include in the previous PR (https://github.com/apache/mynewt-core/pull/2300).  Sorry for the confusion!

Prior to commit:

When a managed TCP connection had no pending transmits, the code would
enqueue another "tx" event.  As a result, the "tx" event would always be
executing or enqueued!

Fix:

If a connection has no pending transmits, move on to the next
connection.  Do not enqueue a "tx" event.


ccollins476ad:
Thanks @vrahane.

PR #2302, opened 2020-06-03
stm32: fix invalid uart access

utzig:
This fixes and invalid access when only UART_1, or UART_2 are enabled.

The uarts structure in memory is allocated according to the number of uarts enabled. When only UART_2 is enabled for example, only 1 element of uarts is enabled, but when trying to access it the index 2 was being used. This commit adds a helper function that fixes the indexing.


utzig:
This was first reported by @wes3. Unfortunately I am short on time to test it today, will try tomorrow (anyway I believe the solution is correct! :-P)

utzig:
> anyway I believe the solution is correct! :-P

Hmmm, maybe not, `if (port < 0 || port >= UART_CNT)`, this code will not allow it to work. Will fix and test tomorrow.

wes3:
@utzig Hey Fabio. I will wait for the changes to approve the PR. It looks good though but I do agree you should add the check that was deleted.

hw/mcu/stm/stm32_common/src/hal_uart.c
```diff
@@ -38,6 +38,31 @@ struct hal_uart {
     const struct stm32_uart_cfg *u_cfg;
 };
 static struct hal_uart uarts[UART_CNT];
+static struct hal_uart *
+uart_index_by_port(int port)
```
kasjer, line 42: Function name is misleading, It does not return index.

hw/mcu/stm/stm32_common/src/hal_uart.c
```diff
@@ -194,12 +216,9 @@ hal_uart_blocking_tx(int port, uint8_t data)
...
 
-    if (port < 0 || port >= UART_CNT) {
-        return;
-    }
-    u = &uarts[port];
-    if (!u->u_open) {
-        return;
+    u = uart_index_by_port(port);
+    if (!u || u->u_open) {
+        return -1;
```
kasjer, line 221: Original return did not have -1

utzig:
@kasjer @wes3 The reported issues were fixed, plus a few places where invalid accesses could happen. I added a new commit for the nucleo-f439zi which allows to use another UART and did some proper testing.

PR #2309, opened 2020-06-09
hw/mcu/dialog: Add system clock choice

kasjer:
System clock was always set to XTAL32M.

Now it can be also RC32M and PLL96
PLL96 can be turned on even if it is not used for system clock,
this is needed if USB is to be used.

### DIFF
--- a/hw/mcu/dialog/da1469x/syscfg.yml
+++ b/hw/mcu/dialog/da1469x/syscfg.yml
@@ -66,6 +66,17 @@ syscfg.defs:
         description: >
             Specifies settling time [ms] of 32K crystal oscillator.
         value: 8000
+    MCU_SYSCLK_SOURCE:
+        description: >
+            Specifies system clock source
+        choices:
+            - XTAL32M
+            - PLL96
+        value: XTAL32M
+    MCU_PLL_ENABLE:
+        description: >
+            Enable 96MHz PLL
+        value: 0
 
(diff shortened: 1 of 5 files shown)


hw/mcu/dialog/da1469x/syscfg.yml
```diff
@@ -392,3 +403,4 @@ syscfg.restrictions:
     - "!SPI_1_MASTER || (SPI_1_MASTER_PIN_SCK && SPI_1_MASTER_PIN_MOSI && SPI_1_MASTER_PIN_MISO)"
     - "!SPI_0_SLAVE || (SPI_0_SLAVE_PIN_SCK && SPI_0_SLAVE_PIN_MOSI && SPI_0_SLAVE_PIN_MISO && SPI_0_SLAVE_PIN_SS)"
     - "!SPI_1_SLAVE || (SPI_1_SLAVE_PIN_SCK && SPI_1_SLAVE_PIN_MOSI && SPI_1_SLAVE_PIN_MISO && SPI_1_SLAVE_PIN_SS)"
+    - "PLL_ON || MCU_SYSCLK_SOURCE != PLL96"
```
andrzej-kaczmarek, line 406: should be `MCU_PLL_ON` although I'd change that name to `MCU_PLL_ENABLE` (like `MCU_DCDC_ENABLE`)

PR #2311, opened 2020-06-10
kernel/os: added package dependency

KKopyscinski:
If RTT was enabled, rtt/SEGGER_RTT.h was being  included
in os.c, but that package was lacking, causing build error.

### DIFF
--- a/kernel/os/pkg.yml
+++ b/kernel/os/pkg.yml
@@ -29,6 +29,9 @@ pkg.deps:
     - "@apache-mynewt-core/sys/sysinit"
     - "@apache-mynewt-core/util/mem"
 
+pkg.deps.CONSOLE_RTT:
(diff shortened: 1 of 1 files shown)


kernel/os/pkg.yml
```diff
@@ -29,6 +29,9 @@ pkg.deps:
     - "@apache-mynewt-core/sys/sysinit"
     - "@apache-mynewt-core/util/mem"
 
+pkg.deps.CONSOLE_RTT:
+    - "@apache-mynewt-core/hw/drivers/rtt"
```
andrzej-kaczmarek, line 33: `RTT` setting is defined by this package so this dependency does not make any sense

KKopyscinski: Indeed error was on my side, and this is not needed

PR #2316, opened 2020-06-13
Fix NLIP receives

ccollins476ad:
Incoming NLIP frames were not being properly decoded.  This bug has two causes:

1. The trailing '\n' character was not being stripped before passing input to `nlip_process()`.  The base64 decode functions were recently made more strict such that they reject invalid input.  The old behavior was to stop at the first invalid character and return the partially decoded string.

The fix here is to trim the trailing '\n'.

2. The base64 code contained a buffer overrun bug.  An internal function expected a string but was passed an unterminated character array.  When the internal function called `strlen()` on its input, the result was undefined behavior.  Consequently, base64 decoding would fail with some optimization levels.

The fix is to change the intermediate function (`process_token()`) to accept a length argument.  The first argument is now treated as a regular character array (not a string).

The base64 changes that broke this are from this PR: https://github.com/apache/mynewt-core/pull/2287

### DIFF
--- a/encoding/base64/src/base64.c
+++ b/encoding/base64/src/base64.c
@@ -131,19 +131,26 @@ token_decode(const char *token, int len)
     int i;
     unsigned int val = 0;
     int marker = 0;
-    if (len < 4)
+
+    if (len < 4) {
         return DECODE_ERROR;
+    }
+
     for (i = 0; i < 4; i++) {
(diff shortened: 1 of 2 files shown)


ccollins476ad:
The base64 unit tests are failing.  I will reopen after I have fixed that.

PR #2317, opened 2020-06-15
net/oic: Fail init if any interface fails

ccollins476ad:
oc_connectivity_init() would succeed if at least one interface was successfully initialized.  This allowed interfaces to silently fail.

This commit changes oc_connectivity_init() so that it fails unless *all* interfaces successfully initialize.

### DIFF
--- a/net/oic/src/port/mynewt/adaptor.c
+++ b/net/oic/src/port/mynewt/adaptor.c
@@ -172,21 +172,25 @@ oc_connectivity_shutdown(void)
 int
 oc_connectivity_init(void)
 {
-    int rc = -1;
+    int rc;
     int i;
     const struct oc_transport *ot;
 
     oc_conn_init();
+
     for (i = 0; i < OC_TRANSPORT_MAX; i++) {
         if (!oc_transports[i]) {
             continue;
         }
+
         ot = oc_transports[i];
-        if (ot->ot_init() == 0) {
-            rc = 0;
+        rc = ot->ot_init();
+        if (rc != 0) {
+            return -1;
         }
     }
-    return rc;
+
+    return 0;
 }
 
 void


kasjer:
Code is fine, could you explain however why this is needed. If this is obvious for oic then forget I asked. I was just curious why code that work OK in one way now has to be changed.

ccollins476ad:
Thanks @vrahane and @kasjer.

> Code is fine, could you explain however why this is needed. If this is obvious for oic then forget I asked. I was just curious why code that work OK in one way now has to be changed.

I briefly rationalized this change in my commit message, but I probably could have gone into more detail.  In the current code (without this PR), interfaces can fail to initialize and the application gets no feedback about the failure.  For example, say a device acts as a CoAP server with two interfaces: 1) serial, and 2) Bluetooth.  The call to `oc_connectivity_init()` can have four possible outcomes:

| Serial  | Bluetooth | Return code |
| ------- | --------- | ----------- |
| Success | Success   | Success     |
| Success | Fail      | Success     |
| Fail    | Success   | Success     |
| Fail    | Fail      | Fail        |

To me this just seems wrong.  If an application expects two interfaces to be available, it should be notified if one of them fails.  Perhaps the ideal solution would be to implement a more complicated feedback mechanism such that the application knows exactly which interfaces succeeded and which ones failed, but that seemed like overkill to me.  Absent a more sophisticated feedback mechanism, I think it is best to err on the side of failure.  So my PR changes `oc_connectivity_init()` to use the below table instead:

| Serial  | Bluetooth | Return code |
| ------- | --------- | ----------- |
| Success | Success   | Success     |
| Success | Fail      | Fail        |
| Fail    | Success   | Fail        |
| Fail    | Fail      | Fail        |

As for why the current code now needs to be changed: I think no one noticed this until now!  I think I may be the first person to have an interface fail to initialize.  The reason I made this change was to prevent others from dealing with the frustrating debugging experience that I just had :).  One interface failed to initialize for a weird reason (a mempool buffer wasn't properly aligned which caused os_mempool_init to fail, causing an interface's init function to fail) but there was no indication of the failure.  Much later, CoAP commands sent through the failed interface weren't getting processed.

All that said, I am definitely OK with keeping things as-is if you feel this change is not an improvement :).

kasjer:
I'm fine with change, this area is so distant from what I touch on daily basis that it was just curiosity. If one interface fails and that is enough to fail, why try others though?

ccollins476ad:
> I'm fine with change, this area is so distant from what I touch on daily basis that it was just curiosity. If one interface fails and that is enough to fail, why try others though?

Good point!  I have modified the PR so that it returns on the first failure.

kasjer:
Thanks for explanation.
There was no objection on my part, just curiosity.

PR #2319, opened 2020-06-21
Add tickles sleep for STM32F1 CPU

jipanienko:
Tickles sleep can be used when LSE oscillator is used and OS_TICKS_PER_SEC is set to power of 2.

### DIFF
--- a/hw/mcu/stm/stm32_common/src/hal_os_tick.c
+++ b/hw/mcu/stm/stm32_common/src/hal_os_tick.c
@@ -22,7 +22,7 @@
 #include <hal/hal_os_tick.h>
 
 /*
- * XXX implement tickless mode.
+ * XXX implement tickless mode for MCU other then STM32F1.
  */
 
 /*
@@ -43,6 +43,8 @@ __WFI(void)
                     "bx lr");
 }
 #endif
+
+#if MYNEWT_VAL(STM32_CLOCK_LSE) == 0 || (((32768 / OS_TICKS_PER_SEC) * OS_TICKS_PER_SEC) != 32768) || !defined(STM32F1)
 
 void
 os_tick_idle(os_time_t ticks)
@@ -76,3 +78,5 @@ os_tick_init(uint32_t os_ticks_per_sec, int prio)
     DBGMCU->CR |= (DBGMCU_CR_DBG_STOP | DBGMCU_CR_DBG_STANDBY);
 #endif
 }
+
+#endif
--- a/hw/mcu/stm/stm32_common/syscfg.yml
+++ b/hw/mcu/stm/stm32_common/syscfg.yml
@@ -262,3 +262,8 @@ syscfg.defs:
     ADC_2:
         description: "ADC_2"
         value:  0
+
+    OS_TICKS_PER_SEC:
+        description: >
+            Desired ticks frequency in Hz.
+        value: 1000
--- a/hw/mcu/stm/stm32f1xx/src/rtc_tick_stm32f1xx.c
+++ b/hw/mcu/stm/stm32f1xx/src/rtc_tick_stm32f1xx.c
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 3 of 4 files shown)


hw/mcu/stm/stm32_common/syscfg.yml
```diff
@@ -262,3 +262,7 @@ syscfg.defs:
     ADC_2:
         description: "ADC_2"
         value:  0
+
+    OS_TICKS_PER_SEC:
+        description:
```
kasjer, line 267: Description not provided.

hw/mcu/stm/stm32f1xx/src/system_stm32f1xx.c
```diff
@@ -69,6 +69,12 @@ void SystemClock_Config(void);
 void
 SystemInit(void)
 {
+    /* Reset all peripherals. */
+    RCC->APB1RSTR = 0xFFFFFFFF;
+    RCC->APB2RSTR = 0xFFFFFFFF;
+    RCC->APB1RSTR = 0x00000000;
+    RCC->APB2RSTR = 0x00000000;
+
```
kasjer, line 77: This code does not seem to be relevant to the change, isn't it?

hw/mcu/stm/stm32_common/src/hal_os_tick.c
```diff
@@ -20,6 +20,7 @@
 #include <assert.h>
 #include "os/mynewt.h"
 #include <hal/hal_os_tick.h>
+#include <stm32_common/stm32_hal.h>
 
 /*
  * XXX implement tickless mode.
```
kasjer, line 26: PR seems to make this comment partially obsolete :)

hw/mcu/stm/stm32_common/src/hal_os_tick.c
```diff
@@ -76,3 +79,111 @@ os_tick_init(uint32_t os_ticks_per_sec, int prio)
...
+    }
+}
+
+void
+os_tick_init(uint32_t os_ticks_per_sec, int prio)
+{
+    uint32_t sr;
+
+    __HAL_RCC_RTC_CONFIG(RCC_RTCCLKSOURCE_LSE);
+    __HAL_RCC_RTC_ENABLE();
```
kasjer, line 163: This may not be enough. ST HAL code for changing clock source does a lot of other things since changing it may not be possible after it was first set (for example if bootloader was not configured to use LSE and application is change may not be applied).
Have you checked that it will work in this scenario?
Maybe ST HAL could be used for this?

hw/mcu/stm/stm32_common/src/hal_os_tick.c
```diff
@@ -76,3 +79,111 @@ os_tick_init(uint32_t os_ticks_per_sec, int prio)
     DBGMCU->CR |= (DBGMCU_CR_DBG_STOP | DBGMCU_CR_DBG_STANDBY);
 #endif
 }
+
```
brianwyld, line 82: IMHO such a large block of STM32F1 specific code should not be the stm32_common files. Should be in the specific MCU directory instead... especially if we then have tickless code for the STM32L1 (as per my PR), STM32L0/L4, etc etc as each one requires a different approach.

kasjer:
Hi @jipanienko, I checked code and it looks ok. I thought I understand how it should work more or less however when I tried this patch to see if it works it did not work at all.
I tried blinky example with following syscfg values that seems to be needed to trigger your code:
```yml
    STM32_CLOCK_LSE: 1
    OS_TICKS_PER_SEC: 128
```
But with just those changes, blinky no longer works.
Am I missing something?

jipanienko:
@kasjer settings are good.
crash is because mcuboot leaves timer running and LSE take time to established in that time interrupt crash because mcuboot memory is no valid and interrupt vectors are not changed yet to blinky or whatever

for workaround start LSE in bootloader STM32_CLOCK_LSE: 1
issue for mcuboot https://github.com/JuulLabs-OSS/mcuboot/issues/759 and https://github.com/JuulLabs-OSS/mcuboot/issues/529

hw/mcu/stm/stm32f1xx/src/rtc_tick_stm32f1xx.c
```diff
@@ -0,0 +1,135 @@
...
+        if (LL_RTC_IsActiveFlag_SEC(RTC)) {
+            rtc_cnt = (RTC->CNTH << 16u) + RTC->CNTL;
+            ticks--;
+        }
+        rtc_cnt += ticks - 1;
+        /* All flags cleared, disable write protection */
+        RTC->CRL = RTC_CRL_CNF_Msk;
+
+        LL_RTC_ALARM_Set(RTC, rtc_cnt);
+        LL_RTC_EnableIT_ALR(RTC);
```
kasjer, line 83: I think that SEC interrupt should be disabled otherwise even if alarm is set SEC interrupt will fire with OS_TICKS_PER_SEC frequency.

PR #2320, opened 2020-06-21
Add syscfg option to select HFCLK source for NRF52 devices

alvarop:
This is needed to be able to run NRF52 devices without an external 32MHz crystal.
BLE will not work when set to HFINT, but everything else should.
The only restriction is that the LFCLK source cannot be LFSYNTH
In the rare case where MCU_HFCLK_SOURCE is set to HFINT and BLE is enabled, an assert(0) will be hit in nrf52_clock_hfxo_request/release


alvarop:
I have a couple of boards without a 32MHz crystal (No BLE) and this was the simplest way I could think of of making them work without changing too much. 

Open to suggestions if y'all have a better syscfg name/description.

alvarop:
Updated the description/title to match the latest changes.

PR #2321, opened 2020-06-21
Add support for all UARTS on STM32

jeremywood:
Adds support for configuring UART3-UART7, increasing the maximum number of configurable UARTs on STM32 from 3 to 8.

### DIFF
--- a/hw/mcu/stm/stm32_common/src/hal_uart.c
+++ b/hw/mcu/stm/stm32_common/src/hal_uart.c
@@ -63,6 +63,37 @@ uart_by_port(int port)
     if (port == 2) {
         return &uarts[index];
     }
+    index++;
+#endif
+#if MYNEWT_VAL(UART_3)
+    if (port == 3) {
+        return &uarts[index];
+    }
+    index++;
+#endif
+#if MYNEWT_VAL(UART_4)
+    if (port == 4) {
+        return &uarts[index];
+    }
+    index++;
+#endif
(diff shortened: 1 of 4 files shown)


hw/mcu/stm/stm32_common/src/hal_uart.c
```diff
@@ -63,6 +63,37 @@ uart_by_port(int port)
...
+    if (port == 6) {
+        return &uarts[index];
+    }
+    index++;
+#endif
+#if MYNEWT_VAL(UART_7)
+    if (port == 7) {
+        return &uarts[index];
+    }
+    index++;
```
utzig, line 96: No need to increment here.

kasjer:
Quick check of number of UARTs/USARTs on ST page reveals that STMF423 and STMF413 (that has BSP in Mynewt) have 10 of them. While I don't have problem with this code PR, title is misleading suggesting that all UARTs are covered.
Since there are two more UARTS out there, way not cover it all here and we could skip next PR with additional 2 in the future.

hw/mcu/stm/stm32_common/src/hal_uart.c
jeremywood: Fixed.

jeremywood:
@kasjer Good catch on the STMF4XX support for 10 UARTS. I've searched across all of the STM headers, and that does appear to be the highest port count.

PR #2323, opened 2020-06-24
net/ip: Re-arrange lwip

wes3:
This commit changes how the mynewt lwip code is arranged in
the repository. Hopefully, this new arrangement makes it easier
to include lwip_base so other lwip implementations can use it
without having to bring the mynewt lwip code. Packages requiring
the base lwip code (and not the mynewt lwip code) can depend
on net/ip/lwip_base. Applications wanting to include the mynewt
lwip code should depend on net/ip/lwip_mn. None of the previous
functionality should change; the code was simply moved from
net/ip/src to net/ip/lwip_mn/src and a new package created.


ccollins476ad:
The change looks good to me, but an automated build is failing:
```
newt build -q -l info stm32f7discovery_iptest

Error: In file included from net/ip/lwip_base/include/lwip/etharp.h:45,

                 from net/ip/lwip_base/include/netif/etharp.h:2,

                 from hw/drivers/lwip/stm32_eth/src/stm32_eth.c:35:

net/ip/lwip_base/include/lwip/opt.h:51:10: fatal error: lwipopts.h: No such file or directory

 #include "lwipopts.h"

          ^~~~~~~~~~~~

compilation terminated.```

utzig:
Probably have to change this dependency as well: https://github.com/apache/mynewt-core/blob/master/net/ip/lwip_base/pkg.yml#L31

wes3:
@utzig No I do not think that is the case. Well, this was part of my issue with how things were arranged. The file lwipopts.h was placed in net/ip/include. However, you may not want this particular version of lwipopts.h. That particular version is now in /net/ip/lwip_mn/include/lwipopts.h.

One issue I was having was that when I was trying to include a different package with its own lwipopts.h, the one from net/ip/include was being included as opposed to the one I wanted.

One way this does work is to include net/ip/lwip_mn to the stm32f7discovery target but that seems a bit odd.

utzig:
> @utzig No I do not think that is the case. Well, this was part of my issue with how things were arranged. The file lwipopts.h was placed in net/ip/include. However, you may not want this particular version of lwipopts.h. That particular version is now in /net/ip/lwip_mn/include/lwipopts.h.

Well, then you still have to remove the `net/ip` dependency inside `net/ip/lwip_base`, no?

> 
> One issue I was having was that when I was trying to include a different package with its own lwipopts.h, the one from net/ip/include was being included as opposed to the one I wanted.
> 
> One way this does work is to include net/ip/lwip_mn to the stm32f7discovery target but that seems a bit odd.

I don't think you should be changing BSPs. I believe you can add an extra dependency to `net/ip/lwip_mn` here: https://github.com/apache/mynewt-core/blob/master/hw/drivers/lwip/stm32_eth/pkg.yml#L30

wes3:
@utzig 
1) If that dependency were removed, how would lwip_base get access to the include directories in net/ip? Maybe I misunderstand something here.
2) Indeed, a better place for the dependency is where you suggest. BTW: my original placement had it in the target and not the BSP but that is no matter.

utzig:
> 1. If that dependency were removed, how would lwip_base get access to the include directories in net/ip? Maybe I misunderstand something here.

I misunderstood something yesterday, my bad.

PR #2329, opened 2020-07-12
remove timer 4 initialisation from loraping

jipanienko:
Timer is inited and never used and some BSP dont have timer 4.


kasjer (approved):
@wes3 and @ccollins476ad agreed in #793 that it was not supposed to be there.
Somehow it was forgotten, thanks @jipanienko.

PR #2330, opened 2020-07-12
remove STATS_GET macro from stats/stub

jipanienko:
Macro STATS_GET don't make sense in stub (it do in full)


kasjer (approved):
Indeed there is not point to have this macro, thanks.

PR #2331, opened 2020-07-14
Fix 2 bytes I2C read on stm32f1

jipanienko:
2 byte read receive wrong data for seconde byte.
First byte read is NAKed and seconde is trash.
Register access sequecnce is wrong for 2 byte read.
Better sequence stolen from HAL_I2C_Master_Receive from hw/mcu/stm/stm32f1xx/src/ext/Drivers/STM32F1xx_HAL_Driver/Src/stm32f1xx_hal_i2c.c
Now NAK is send on seconde byte not first.

### DIFF
--- a/hw/mcu/stm/stm32_common/src/stm32_driver_mod_i2c_v1.c
+++ b/hw/mcu/stm/stm32_common/src/stm32_driver_mod_i2c_v1.c
@@ -695,14 +695,14 @@ HAL_StatusTypeDef HAL_I2C_Master_Receive_Custom(I2C_HandleTypeDef *hi2c,
     }
     else if(hi2c->XferSize == 2U)
     {
-      /* Disable Acknowledge */
-      hi2c->Instance->CR1 &= ~I2C_CR1_ACK;
-
       /* Enable Pos */
       hi2c->Instance->CR1 |= I2C_CR1_POS;
 
       /* Clear ADDR flag */
       __HAL_I2C_CLEAR_ADDRFLAG(hi2c);
+
+      /* Disable Acknowledge */
+      hi2c->Instance->CR1 &= ~I2C_CR1_ACK;
     }
     else
     {


hw/mcu/stm/stm32_common/src/stm32_driver_mod_i2c_v1.c
```diff
@@ -695,14 +695,14 @@ HAL_StatusTypeDef HAL_I2C_Master_Receive_Custom(I2C_HandleTypeDef *hi2c,
...
-      hi2c->Instance->CR1 &= ~I2C_CR1_ACK;
-
       /* Enable Pos */
       hi2c->Instance->CR1 |= I2C_CR1_POS;
 
       /* Clear ADDR flag */
       __HAL_I2C_CLEAR_ADDRFLAG(hi2c);
+
+      /* Disable Acknowledge */
+      hi2c->Instance->CR1 &= ~I2C_CR1_ACK;
```
utzig, line 705: This driver is shared between F1/L1/F4, and for L1/F4 the order of operations would best remain unchanged, so maybe add some `#if MYNEWT_VAL(MCU_STM32F1)` here?

jipanienko:
It turns that there is more problem with I2C on F1 (ST hal driver have more workaround for problem in errata sheet). I may do PR with more F1 specified code.
@utzig what you think about stm32_drive_mod_i2c_v3.c with F1 code?

utzig:
> @utzig what you think about stm32_drive_mod_i2c_v3.c with F1 code?

If it's justifiable by the amount of changes, it is totally fine. Btw, you don't need to close PRs, you can update this one with new changes.

jipanienko:
> > @utzig what you think about stm32_drive_mod_i2c_v3.c with F1 code?
> 
> If it's justifiable by the amount of changes, it is totally fine. Btw, you don't need to close PRs, you can update this one with new changes.

Changes needed in only in two function, I add modified _Custom functions for F1 in same v1 file, thank you @utzig.

PR #2335, opened 2020-07-17
Add INA226 current sensor driver

jipanienko:


jipanienko:
> Looks good, the STATS indentation needs to be fixed as per the indentation check. Rest looks fine. I do not have the part to test with me but codewise it looks ok.

Every STATS_NAME in mynewt have this indentation, style checker rule are bad

PR #2337, opened 2020-07-21
Add busy I2C state recovery on STMF1/F4/L1

jipanienko:
Sometime glitche on SDA SCL lines cause I2c to enter
busy state that never end.
All calls to read or write will fail later because of BUSY flag.
Datasheet recommens software reset in that case.
Workaround applied to transmit only, it could be on receive
but that should do.

### DIFF
--- a/hw/mcu/stm/stm32_common/src/stm32_driver_mod_i2c_v1.c
+++ b/hw/mcu/stm/stm32_common/src/stm32_driver_mod_i2c_v1.c
@@ -373,7 +373,16 @@ HAL_StatusTypeDef HAL_I2C_Master_Transmit_Custom(I2C_HandleTypeDef *hi2c,
       /* Wait until BUSY flag is reset */
       if(I2C_WaitOnFlagUntilTimeout(hi2c, I2C_FLAG_BUSY, SET, I2C_TIMEOUT_BUSY_FLAG, tickstart) != HAL_OK)
       {
-        return HAL_BUSY;
+        /* Stuck in busy state, datasheet recommends software reset */
+        HAL_I2C_DeInit(hi2c);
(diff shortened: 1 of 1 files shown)


hw/mcu/stm/stm32_common/src/stm32_driver_mod_i2c_v1.c
```diff
@@ -373,7 +373,18 @@ HAL_StatusTypeDef HAL_I2C_Master_Transmit_Custom(I2C_HandleTypeDef *hi2c,
       /* Wait until BUSY flag is reset */
       if(I2C_WaitOnFlagUntilTimeout(hi2c, I2C_FLAG_BUSY, SET, I2C_TIMEOUT_BUSY_FLAG, tickstart) != HAL_OK)
       {
-        return HAL_BUSY;
```
utzig, line 376: I wonder if this needs to be applied to other families as well or just F1?

jipanienko: Same suggestion in all datasheet for F1/F4/L1, commit message changed.

kasjer:
@jipanienko is there a way to reproduce wrong behavior to see that this fixes problem?

jipanienko:
> @jipanienko is there a way to reproduce wrong behavior to see that this fixes problem?

While transmission is going, short SCL line to ground for a while.
F1 stacks busy frequently.

kasjer (approved):
Verified that grounding SCL line during transmission leaves controller in permanent busy state (tested on STM32F407), and with this fix transmission is resumed.
@jipanienko did you tried any other device that uses stm32_driver_mod_i2c_v2.c?
If not, I will check if similar solution can be applied there.
Thanks for the fix.

PR #2344, opened 2020-08-04
bsp/pinetime - Change loading tool to J-Link

KKopyscinski:
Default tool to flash pinetime with Mynewt should be J-Link, not STLink.
This is consistent with other bsps using Nordic chips, and is easier to use.

### DIFF
--- a/hw/bsp/pinetime/pinetime_debug.sh
+++ b/hw/bsp/pinetime/pinetime_debug.sh
@@ -27,13 +27,18 @@
 #  - RESET set if target should be reset when attaching
 #  - NO_GDB set if we should not start gdb to debug
 #
-. $CORE_PATH/hw/scripts/openocd.sh
+. $CORE_PATH/hw/scripts/jlink.sh
 
 FILE_NAME=$BIN_BASENAME.elf
-CFG="-f interface/stlink.cfg -f target/nrf52.cfg"
-EXTRA_GDB_CMDS='monitor arm semihosting enable'
-# Exit openocd when gdb detaches.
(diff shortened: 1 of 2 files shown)


utzig:
Does the Pinetime even include a bundled J-Link?

KKopyscinski:
It does not have included any debugger, only 4 wires to connect to what user has. So I think it's good idea to use J-Link, as it works good with newt load and avoids hassle with openOCD

utzig:
> It does not have included any debugger, only 4 wires to connect to what user has. So I think it's good idea to use J-Link, as it works good with newt load and avoids hassle with openOCD

So, you think that because using a JLink suits YOU it's OK to break the configuration of other people including the original contributors? Why don't  you use a ST-Link? It costs US$ 1!

KKopyscinski:
Well, I didn't saw that that way; I just thought as it is using nrf52xxx chip AND can be flashed using J-Link it improves consistency with other bsps, namely nordic_pca100xx. Now, I see that it might be an issue, so I think it might be useful to add second bsp - something like pinetime-jlink, to allow both solutions.

utzig:
> Well, I didn't saw that that way; I just thought as it is using nrf52xxx chip AND can be flashed using J-Link it improves consistency with other bsps, namely nordic_pca100xx. Now, I see that it might be an issue, so I think it might be useful to add second bsp - something like pinetime-jlink, to allow both solutions.

J-Link has nothing to do with nRF5x, it just happens to be on every Nordic devkit. You could instead add an environment variable and use it to decide if you prefer to use ST-Link as is now, or J-Link as you want to.

KKopyscinski:
Ok, so I think this PR may be just closed now. Maybe I'll add this feature in the future, but I agree that current state is fine.

PR #2345, opened 2020-08-05
da1469x_sleep: Adding deep sleep pre/post callbacks, and QSPI configurability in deep sleep

JustineKH:
Adding configurability to da1469x_m33_sleep.S QSPI in deep sleep. MCU_DEEP_SLEEP_QSPIC_QUAD_DISABLE.
Adding retention for more IPRs.
Adding pre and post callbacks for dep_sleep

 @andrzej-kaczmarek

### DIFF
--- a/hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
+++ b/hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
@@ -208,8 +208,8 @@ da1469x_m33_wakeup:
     add     r1, r0, NVIC_IPR_OFFSET
     ldmia   r3!, {r4-r5}
     stmia   r0!, {r4-r5}
-    ldmia   r3, {r0, r2, r4-r12}
-    stmia   r1!, {r0, r2, r4-r12}
+    ldmia   r3!, {r0, r2, r4-r12}
+    stmia   r1, {r0, r2, r4-r12}
 
  /* Restore SCB state */
     ldmia   r3!, {r4-r10}
--- a/hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
+++ b/hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
@@ -39,7 +39,7 @@ da1469x_m33_sleep_state:
     .space 40   /* R4-R12, LR */
 .saved_nvic:
     .space 8    /* ISER[0..1] */
-    .space 40   /* IPR[0..39] */
+    .space 44   /* IPR[0..43] */
 .saved_scb:
     .space 28   /* SCR, CCR, SHPR[0..11], CPACR */
 #if MYNEWT_VAL(HARDFLOAT)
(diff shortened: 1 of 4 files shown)


hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
```diff
@@ -149,6 +166,32 @@ da1469x_m33_sleep:
...
+    orr     r1, r1, #0x3c   /* Set IO2/IO3 DAT and OEN */
+    str     r1, [r0, #(QSPIC_CTRLMOD_OFFSET)]
+    mov     r1, #1          /* Set single */
+    str     r1, [r0, #(QSPIC_CTRLBUS_OFFSET)]
+    mov     r1, #8          /* Enable CS */
+    str     r1, [r0, #(QSPIC_CTRLBUS_OFFSET)]
+    mov     r1, #0xff       /* Exit QPI mode */
+    strb    r1, [r0, #(QSPIC_WRITEDATA_OFFSET)]
+    mov     r1, #16         /* Disable CS */
+    str     r1, [r0, #(QSPIC_CTRLBUS_OFFSET)]
```
andrzej-kaczmarek, line 187: code above this line is not necessary since we are already in single mode - it was done before wfi

hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
```diff
@@ -39,7 +39,7 @@ da1469x_m33_sleep_state:
     .space 40   /* R4-R12, LR */
 .saved_nvic:
     .space 8    /* ISER[0..1] */
-    .space 41   /* IPR[0..40] */
+    .space 44   /* IPR[0..40] */
```
kasjer, line 42: comment does not match new value it looks like it should be IPR[0..43]

hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
```diff
@@ -94,8 +94,8 @@ da1469x_m33_sleep:
     add     r1, r0, NVIC_IPR_OFFSET
     ldmia   r0!, {r4-r5}
     stmia   r3!, {r4-r5}
-    ldmia   r1!, {r2, r4-r12}
-    stmia   r3!, {r2, r4-r12}
+    ldmia   r1!, {r2}
```
kasjer, line 97: There is something wrong here.
Only **IPR[0..3]** is saved here (and restored later).
If it is intentional `space 44` could be `space 4`.
But most likely those line should be changed to copy one more word, R0 could be used for this as it's current value will not be used later.
```
ldmia   r1, {r0, r2, r4-r12}
stmia   r3!, {r0, r2, r4-r12}
```

hw/mcu/dialog/da1469x/src/arch/cortex_m33/da1469x_m33_sleep.S
```diff
@@ -174,8 +208,8 @@ da1469x_m33_wakeup:
     add     r1, r0, NVIC_IPR_OFFSET
     ldmia   r3!, {r4-r5}
     stmia   r0!, {r4-r5}
-    ldmia   r3!, {r2, r4-r12}
-    stmia   r1!, {r2, r4-r12}
+    ldmia   r3!, {r2}
+    stmia   r1!, {r2}
```
kasjer, line 212: I think it should be coping 11 words not just one.
```
 ldmia   r3!, {r0, r2, r4-r12}
 stmia   r1, {r0, r2, r4-r12}
```

PR #2347, opened 2020-08-06
docs: Add section about conditional settings

ccollins476ad:


docs/os/modules/sysinitconfig/sysinitconfig.rst
```diff
@@ -352,6 +352,47 @@ use the flash area named ``FLASH_AREA_NFFS`` in the ``syscfg.yml`` file.
...
+be the value of a single syscfg setting or an arbitrary expression
+involving many settings.
+
+Examples of Conditional Settings
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Example 6
+`````````
+
+In this example, a package depends on `lib/pkg2` only if `MY_SETTING` has a
```
utzig, line 370: Unlike Markdown where inline literals use a single backquote ("`"), reStructuredText uses double backquotes ("``") around the text.

ccollins476ad: Thanks @utzig!  I will need to render this text locally to make sure I am not messing up more stuff.

By the way, I just noticed that there already was a "conditional settings" section at the bottom of the page!  I think the one I added is a bit better because it is more succinct and it talks about using arbitrary expressions.  So, I removed the old one.

PR #2351, opened 2020-08-08
PineTime SPI flash

caspermeijn:
Use spi flash for the second image. 
This makes it possible to double the size of the image.
To be compatible with other community firmwares the bootloader size is
increased. This causes the reboot log to shrink.
An extra SPI flash area is assigned for bootloader assets (for example
a boot logo).
The image could be increased by another 12kB, but I want to wait for
other firmwares to do that as well.

Also use the bus driver for the SPI flash, so that the SPI display can be
operated in parallel.


hw/bsp/pinetime/src/hal_bsp.c
```diff
@@ -162,15 +181,51 @@ hal_bsp_charger_init(void)
...
 #endif
 
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+static struct bus_spi_node hal_bsp_display_spi;
+struct bus_spi_node_cfg hal_bsp_display_spi_cfg = {
+    .node_cfg.bus_name = MYNEWT_VAL(BSP_FLASH_SPI_BUS),
+    .pin_cs = LCD_CHIP_SELECT_PIN,
+    .mode = BUS_SPI_MODE_3,
+    .data_order = HAL_SPI_MSB_FIRST,
+    .freq = 8000,
```
utzig, line 191: Could have used `MYNEWT_VAL(SPIFLASH_BAUDRATE)`

caspermeijn:
Thanks

PR #2355, opened 2020-08-17
sensors: Add current and voltage support

jipanienko:
current.h and voltage.h are used in INA219 and INA226 and there were not added yet.
Sensors shell also has support for current and voltage.

### DIFF
--- a/hw/sensor/include/sensor/current.h
+++ b/hw/sensor/include/sensor/current.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2020 Jesus Ipanienko
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


hw/sensor/include/sensor/current.h
```diff
@@ -0,0 +1,25 @@
...
+#define __SENSOR_CURRENT_H__
+
+#include "os/mynewt.h"
+#include "sensor/sensor.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Data representing a singular read from a pressure sensor
```
kasjer, line 11: It should not be **pressure** but current I suppose.

hw/sensor/include/sensor/voltage.h
```diff
@@ -0,0 +1,25 @@
...
+#define __SENSOR_VOLTAGE_H__
+
+#include "os/mynewt.h"
+#include "sensor/sensor.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Data representing a singular read from a pressure sensor
```
kasjer, line 11: Change pressure to voltage please.

hw/sensor/include/sensor/voltage.h
```diff
@@ -0,0 +1,25 @@
...
+
+    /* Validity */
+    uint8_t svd_voltage_is_valid:1;
+};
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __VOLTAGE_HUMIDITY_H__ */
```
kasjer, line 25: It should be __SENSOR_VOLTAGE_H__

PR #2358, opened 2020-08-20
hw/mcu/dialog: Make 1V8 and 1V8P rails configurable

andrzej-kaczmarek:
This adds some configuration options to control 1V8x rails. For now,
only LDOs are selectable with possibly more options to come, if needed.


andrzej-kaczmarek:
@JustineKH ^^

PR #2359, opened 2020-08-20
hw/mcu/dialog: Add all reset reason checks

JustineKH:
Adding all checks for reboot reasons per da1469x ds.

### DIFF
--- a/hw/hal/include/hal/hal_system.h
+++ b/hw/hal/include/hal/hal_system.h
@@ -72,6 +72,8 @@ enum hal_reset_reason {
     HAL_RESET_SYS_OFF_INT = 7,
     /** Restart due to DFU */
     HAL_RESET_DFU = 8,
+    /** Restart reason other */
+    HAL_RESET_OTHER = 256,
 };
 
 /**
--- a/sys/reboot/src/log_reboot.c
+++ b/sys/reboot/src/log_reboot.c
@@ -207,7 +207,7 @@ log_reboot_write(const struct log_reboot_info *info)
     }
 
     cbor_encode_text_stringz(&map, "rsn");
-    cbor_encode_text_stringz(&map, REBOOT_REASON_STR(info->reason));
+    cbor_encode_text_stringz(&map,log_reboot_reason_str(info->reason));
 
     cbor_encode_text_stringz(&map, "cnt");
     cbor_encode_int(&map, reboot_cnt);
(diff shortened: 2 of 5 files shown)


hw/hal/include/hal/hal_system.h
```diff
@@ -72,6 +72,10 @@ enum hal_reset_reason {
     HAL_RESET_SYS_OFF_INT = 7,
     /** Restart due to DFU */
     HAL_RESET_DFU = 8,
+    /** Reserved 1 */
+    HAL_RESET_WATCHDOG2 = 9,
```
andrzej-kaczmarek, line 76: this file is common for all MCUs and imo it does not really need to have unique mapping for every single reset reason for each MCU. in this case, CMAC watchdog could be just mapped to `HAL_RESET_WATCHDOG`. in generic case, it would be just better to add `HAL_RESET_OTHER` and map any reset reason that does not map to defined value - MCU can provide `mcu_reset_cause` to read actual reset reason and appropriate `#define`s in `mcu.h`.

andrzej-kaczmarek:
my proposal to fix this would be as follows:
- add smth like `HAL_RESET_OTHER = 256` for any reset reason that is specific for MCU, that means it neither maps to common list nor it makes sense to add it to that list (e.g. CMAC watchdog)
- `hal_reset_cause` can either just return `HAL_RESET_OTHER` for such reason or if more mappings are requires, it can return `HAL_RESET_OTHER + 1`, `HAL_RESET_OTHER + 2` and so on; the behavior whether to return just `HAL_RESET_OTHER` for all reasons or more detailed values could be potentially configurable via syscfg, although I don't think this is really necessary
- `sys/reboot` code should use some proper function instead of ugly `REBOOT_REASON_STR` to stringify reset reason - for known mappings it will return proper string, for others it can return hex value or smth so it can be logged properly

this way we do not bloat reset reason line with too many symbols, but also allow to return some more info if needed. would like to hear what other thinks about above idea, perhaps there's something smarter to do here

JustineKH:
@andrzej-kaczmarek 
I agree that this list of reboot reasons could get out of control if we keep adding non-standard reasons. Could we just add a callback function for the reason and SYSCFG to enable it? Seems that some users could be directly calling `REBOOT_REASON_STR`, and expect it to exist.

andrzej-kaczmarek:
you do not need to remove `REBOOT_REASON_STR`

sys/reboot/src/log_reboot.c
```diff
@@ -383,6 +384,50 @@ reboot_conf_export(void (*func)(char *name, char *val),
     return 0;
 }
 
+void
```
andrzej-kaczmarek, line 387: `const char *`

sys/reboot/src/log_reboot.c
```diff
@@ -383,6 +384,50 @@ reboot_conf_export(void (*func)(char *name, char *val),
...
+log_reboot_reason_str(enum hal_reset_reason reason, char *buf)
+{
+    char str_reason[MYNEWT_VAL(REBOOT_LOG_REBOOT_REASON_SIZE)];
```
andrzej-kaczmarek, line 390: use `static` buffer with fixed size

sys/reboot/src/log_reboot.c
```diff
@@ -383,6 +384,50 @@ reboot_conf_export(void (*func)(char *name, char *val),
...
+
+    if (reason >= HAL_RESET_OTHER) {
+        snprintf(str_reason,MYNEWT_VAL(REBOOT_LOG_REBOOT_REASON_SIZE),"OTHER: 0x%X",reason - HAL_RESET_OTHER);
+        strcpy(buf, str_reason);
+        return;
+    }
+
+    switch (reason) {
+    case HAL_RESET_POR:
+        snprintf(str_reason,MYNEWT_VAL(REBOOT_LOG_REBOOT_REASON_SIZE),"HARD");
```
andrzej-kaczmarek, line 400: you can just return e.g. `"HARD"` for plain strings
for values that have to be converted to string just do `snprintf` into static buffer of fixed size which is large enough to hold that string and return pointer to that buffer - it's safe as long as it is 'static`

sys/reboot/src/log_reboot.c
```diff
@@ -383,6 +384,50 @@ reboot_conf_export(void (*func)(char *name, char *val),
...
+        snprintf(str_reason,MYNEWT_VAL(REBOOT_LOG_REBOOT_REASON_SIZE),"REQUESTED");
+        break;
+    case HAL_RESET_SYS_OFF_INT:
+        snprintf(str_reason,MYNEWT_VAL(REBOOT_LOG_REBOOT_REASON_SIZE),"SYSTEM_OFF_INT");
+        break;
+    case HAL_RESET_DFU:
+        snprintf(str_reason,MYNEWT_VAL(REBOOT_LOG_REBOOT_REASON_SIZE),"DFU");
+        break;
+    default:
+        snprintf(str_reason,MYNEWT_VAL(REBOOT_LOG_REBOOT_REASON_SIZE),"UNKNOWN");
```
andrzej-kaczmarek, line 424: would be better to stringify this like `"UNKNOWN (%d)"` so actual value is also logged

andrzej-kaczmarek (changes requested):

andrzej-kaczmarek (approved):

PR #2362, opened 2020-08-23
apps/rust_blinky: Add blinky app using Rust language

caspermeijn:
This added the Rust blinky application as explained in https://mynewt.apache.org/latest/tutorials/other/rust.html

Then it extends it by using bindgen to generate Rust code from the actual C-header as used for the C code. Then use the generated code in the Rust application to make it compatible with all BSP.

### DIFF
--- a/apps/rust_blinky/build.rs
+++ b/apps/rust_blinky/build.rs
@@ -1,0 +1,69 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2020 Casper Meijn <casper@meijn.net>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+extern crate bindgen;
+
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+use std::str;
+
+fn main() {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("bindings.rs");
+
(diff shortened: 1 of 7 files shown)


apps/rust_blinky/build.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+use std::str;
+
+fn main() {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("bindings.rs");
+
+    let mut builder = bindgen::Builder::default()
+        .clang_arg("--target=thumbv7m-none-eabi")
```
utzig, line 29: Is this needed?

apps/rust_blinky/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    /* Initialize all packages. */
+    unsafe {
+        bindings::sysinit_start();
+        bindings::sysinit_app();
+        bindings::sysinit_end();
+    }
+
+    /* Turn on the LED */
+    unsafe {
+        bindings::hal_gpio_init_out(bindings::LED_BLINK_PIN as i32, 1);
```
utzig, line 42: `LED_BLINK_PIN` is usually a #define in `bsp/bsp.h`, does bindgen generate a binding for it?

utzig:
I tested this now on `frdm-k64f`  and `b-l072z-lrwan1` and works fine as long as I change the `bindings::LED_BLINK_PIN` to the appropriate pin number for both boards; not sure why `LED_BLINK_PIN` does not work for me. I also removed the line I commented above, `.clang_arg("--target=thumbv7m-none-eabi")`, and it still worked OK. Also would be nice to silence all those  `warning: <something> is never used` messages generated by `bindgen`. Otherwise looks good and works well!

apps/rust_blinky/src/lib.rs
caspermeijn: For the PineTime BSP it does, but that is a simple number. I see that your bsp.h do some calculation in the macro, maybe that is different. I am not sure how to solve that.

apps/rust_blinky/build.rs
caspermeijn: I needed it at a earlier stage, but it seems like it is not needed anymore ;-)

caspermeijn:
Updated the commit. I fixed the license, removed the clang target and silenced the "is never used" warnings.

PR #2363, opened 2020-08-28
Allow oc gatt client as BLE central

bgiori:
This PR enables an oc client to work with ble gatt transport adaptor in as a ble central (rather than a peripheral). This had never been done and therefore did not work. As a client and a central, we need to write to a the request characteristic and listen for notifications on the response characteristic. 

These changes add a new syscfg flag `OC_BLE_CENTRAL` to `net/oic/syscfg.yml` with a default value 0. Setting this value to 1 defines a function for receiving response characteristic notifications and enables characteristic write requests.


vrahane:
@bgiori Can you please squash the merges into one ?

bgiori:
squash my commits into one?

you should be able to do so with the "squash and merge" feature on github.

PR #2364, opened 2020-09-01
Update NRFX to release 2.3.0

sjanc:


sjanc:
@vrahane Could you check if it is actually enough to change include for nrf91 as stated in build error and things work?

Error: In file included from hw/mcu/nordic/nrf91xx/src/hal_flash.c:25:
hw/mcu/nordic/src/ext/nrfx/templates/nrfx_config_nrf9160.h:36:2: error: #error "This file should not be included directly. Include nrfx_config.h instead."
 #error "This file should not be included directly. Include nrfx_config.h instead."
  ^~~~~

kasjer:
It would be nicer (IMHO) if code provided by Nordic was in separate commit as it should not be checked for styl and such.

kasjer (approved):
LGTM

PR #2366, opened 2020-09-02
hw/mcu/dialog: Add syscfg to allow setting a custom DMA_IRQn priority.

benmccrea:
For hw/mcu/dialog/da1469x, adds a syscfg `MCU_DMA_IRQ_PRIO` to allow setting the interrupt priority of DMA_IRQn at system initialization time.


kasjer (approved):
LGTM

PR #2369, opened 2020-09-03
stm32: Disable WFI when OS_SYSVIEW is defined

kasjer:
Reading RAM or FLASH from Debugger on STM32 MCUs in SLEEP (WFI)
returns wrong values.
If makes it almost impossible to use with SystemView when application
runs at high clock speed and spend a lot of time in SLEEP mode.
SystemView is often not able to correctly read header for RTT
virtually preventing tool to detect instrumentation.

This change replaces WFI with busy loop waiting for interrupts when
OS_SYSVIEW is defined.

### DIFF
--- a/hw/mcu/stm/stm32_common/src/hal_os_tick.c
+++ b/hw/mcu/stm/stm32_common/src/hal_os_tick.c
@@ -26,6 +26,15 @@
  */
 
 /*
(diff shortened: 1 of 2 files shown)


andrzej-kaczmarek (approved):

utzig (approved):
LGTM

PR #2370, opened 2020-09-03
hw/drivers/i2s: Add sysview interrupt tracing to STM drivers

kasjer:
This adds interrupt tracing code so SystemView can be
used for load analyzes.


utzig (approved):
LGTM

PR #2371, opened 2020-09-03
hw/drivers/i2s: Add buffer overwrite check

kasjer:
It is easy to accidentally write past the sample buffer.
Such writes could lead to difficult to trace memory corruptions.

With this change most common errors will be detected early.
To have checking code working syscfg value I2S_BUFFER_GUARD
should specify how many bytes before and after actual data
is reserved and filled with pattern.

### DIFF
--- a/hw/drivers/i2s/src/i2s.c
+++ b/hw/drivers/i2s/src/i2s.c
@@ -103,6 +103,7 @@ static void
 i2s_buffers_from_pool(struct i2s *i2s, struct i2s_buffer_pool *pool)
 {
     int i;
+    int j;
     int sr;
     struct i2s_sample_buffer *buffers;
     uintptr_t sample_data;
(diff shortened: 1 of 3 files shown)


hw/drivers/i2s/src/i2s.c
```diff
@@ -316,7 +323,16 @@ i2s_buffer_put(struct i2s *i2s, struct i2s_sample_buffer *buffer)
 {
     int sr;
     int rc = I2S_OK;
+    int i;
 
+    if (I2S_BUFFER_GUARD) {
+        const uint8_t *pre = (uint8_t *)(buffer->sample_data) - 1;
```
sjanc, line 329: nitpick: should be declared on function top

sjanc (approved):

hw/drivers/i2s/src/i2s.c
```diff
@@ -125,10 +126,16 @@ i2s_buffers_from_pool(struct i2s *i2s, struct i2s_buffer_pool *pool)
     sample_data = (uintptr_t)&buffers[pool->buffer_count];
 
     for (i = 0; i < pool->buffer_count; ++i) {
+        for (j = 0; j < I2S_BUFFER_GUARD; ++j, ++sample_data) {
+            *(uint8_t *)sample_data = 0xFD;
```
wes3, line 130: Another nit-pick: Maybe define the buffer guard value and use the definition instead of the hard-coded value.

PR #2377, opened 2020-09-17
64KB block and chip erase for Nordic nrf52 QSPI

mikevoyt:
## Description
Without a hff_erase implementation for HAL flash functions, the HAL flash layer will default to sector erase for all
erase functions, regardless of length.

This PR implements hff_erase, and will issue a chip erase or 64KB block erase where possible based on the address and length of area to be erased.

## How was this tested?

On the nRF52840 DK (PCA10056) using the provided MX25R6435F QSPI flash, I measured a 4K sector erase of about 84 mS.  With this patch, a 64KB block erase averages about 1164 mS.  This saves about 176 mS per 64K erase.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
@@ -66,25 +66,29 @@
 
 static int
 nrf52k_qspi_read(const struct hal_flash *dev, uint32_t address,
-        void *dst, uint32_t num_bytes);
+                 void *dst, uint32_t num_bytes);
 static int
 nrf52k_qspi_write(const struct hal_flash *dev, uint32_t address,
-        const void *src, uint32_t num_bytes);
+                  const void *src, uint32_t num_bytes);
 static int
 nrf52k_qspi_erase_sector(const struct hal_flash *dev,
-        uint32_t sector_address);
+                         uint32_t sector_address);
 static int
 nrf52k_qspi_sector_info(const struct hal_flash *dev, int idx,
-        uint32_t *address, uint32_t *sz);
+                        uint32_t *address, uint32_t *sz);
 static int
 nrf52k_qspi_init(const struct hal_flash *dev);
+static int
+nrf52k_qspi_erase(const struct hal_flash *dev, uint32_t address,
+                  uint32_t size);
 
 static const struct hal_flash_funcs nrf52k_qspi_funcs = {
     .hff_read = nrf52k_qspi_read,
     .hff_write = nrf52k_qspi_write,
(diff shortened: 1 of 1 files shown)


hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
```diff
@@ -226,16 +246,53 @@ nrf52k_qspi_erase_sector(const struct hal_flash *dev,
...
+    address &= ~0xFFFU;
+    end = address + size;
+
+    if (end == MYNEWT_VAL(QSPI_FLASH_SECTOR_COUNT) *
+                   MYNEWT_VAL(QSPI_FLASH_SECTOR_SIZE)) {
+        return erase_block(0, NRF_QSPI_ERASE_LEN_ALL);
+    }
+
+    while (size) {
+#if MYNEWT_VAL(SPIFLASH_BLOCK_ERASE_64BK)
```
andrzej-kaczmarek, line 272: this is syscfg from different package, do not use it here. either make it always enabled or alternatively add new syscfg to MCU package

mikevoyt: @andrzej-kaczmarek I've removed the `SPIFLASH_BLOCK_ERASE_64BK` conditional.

hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
```diff
@@ -218,30 +223,77 @@ nrf52k_qspi_write(const struct hal_flash *dev, uint32_t address,
...
+    return 0;
+}
+
+static int
+nrf52k_qspi_erase(const struct hal_flash *dev, uint32_t address,
+                  uint32_t size)
+{
+
+    uint32_t end;
+    int rc;
```
kasjer, line 262: rc is not used any more, it should be removed

kasjer (changes requested):
@mikevoyt I tested this PR and and works (once unused variable is remove).
There is something that could be done though.
First commit message was clean and it was obvious what the change is.
Next few make sense only in context of PR.
I can squash them in github but then your authorship will be messaged up.
I suggest to fix rc, then squash all commits, force push to the same branch, so we can see one clean commit that adds 64KB erase, and that will go to the git history.

hw/mcu/nordic/nrf52xxx/src/hal_qspi.c
mikevoyt: @kasjer no problem, I've removed `int rc`, and forced pushed the branch so now we have one clean commit.

kasjer (approved):
Works and look great, thank you @mikevoyt

PR #2378, opened 2020-09-21
hw/ipc_nrf5340: Add support for passing GPIO control

sjanc:
This allows to configure GPIOs that should be controlled by Network
Core. For race-free execution GPIOs are configured from IPC subsystem
when Networking Core is force-off.


kasjer (approved):
LGTM

ncasaril (approved):
Looks good 👍

PR #2382, opened 2020-09-30
allow more then one battery

### LINKED ISSUE #2380: multiple instances of `struct battery`
It seems multiple instances of `struct battery` is half supported. It is not configurable and changing
https://github.com/apache/mynewt-core/blob/a25a08db9b884be617cb4c96b2056076d873c530/hw/battery/src/battery.c#L28
to >1 will raise other issue(s). Example the code below assumes early loop exit otherwise the assert is raised:
https://github.com/apache/mynewt-core/blob/a25a08db9b884be617cb4c96b2056076d873c530/hw/battery/src/battery.c#L813-L821

5frank:
Fixes #2380

### DIFF
--- a/hw/battery/src/battery.c
+++ b/hw/battery/src/battery.c
@@ -25,7 +25,7 @@
 #include <battery/battery_prop.h>
 #include <battery/battery_drv.h>
 
-#define BATTERY_MAX_COUNT 1
+#define BATTERY_MAX_COUNT MYNEWT_VAL(BATTERY_MAX_COUNT)
 
 #define GET_BIT(arr, bit)   ((arr)[(bit) / 32] & (1 << ((bit) % 32)))
 
--- a/hw/battery/syscfg.yml
+++ b/hw/battery/syscfg.yml
@@ -26,6 +26,11 @@ syscfg.defs:
             Specify the eventq to be used by the battery manager
         value:
 
+    BATTERY_MAX_COUNT:
+        description: >
+            Maximum number of batteries.
+        value: 1
+
     BATTERY_MAX_PROPERTY_COUNT:
         description: >
             Maximum number of supported battery properties. Number should


hw/battery/src/battery.c
```diff
@@ -25,7 +25,12 @@
 #include <battery/battery_prop.h>
 #include <battery/battery_drv.h>
 
+
+#if MYNEWT_VAL(BATTERY_MAX_COUNT)
```
kasjer, line 29: The way to go in that case is to create syscfg definition in **hw/battery/syscfg.yml**.
Without it package uses value that is never defined.
something like that:
```yml
    BATTERY_MAX_COUNT:
        description: Maximum number of batteries.
        value: 1
```
Then it would be also possible to omit
```c
#if MYNEWT_VAL(BATTERY_MAX_COUNT)
```
... and just leave it like this:
```c
#define BATTERY_MAX_COUNT MYNEWT_VAL(BATTERY_MAX_COUNT)
```

hw/battery/src/battery.c
```diff
@@ -818,7 +823,7 @@ battery_init(struct os_dev *dev, void *arg)
     }
     os_mutex_release(&battery_manager.bm_lock);
 
-    assert(i < BATTERY_MAX_COUNT);
+    assert(i <= BATTERY_MAX_COUNT);
```
kasjer, line 826: This change is not needed. Condition was correct.
On the other hand mutex release should be done later and that could be fixed.

kasjer (changes requested):

hw/battery/src/battery.c
5frank: Agree that is the right way! 
I was following the style you used here:
https://github.com/apache/mynewt-core/blob/f5832e09b0679ce14032850f2d797dc088dbb9fd/hw/battery/include/battery/battery.h#L38-L42

hw/battery/src/battery.c
5frank: Agree! Thanks for the feedback! I didn't think that one through

hw/battery/src/battery.c
kasjer: You are quite right, BATTERY_DRIVERS_MAX is also missing from syscfg.yml.
Here is pull request #2384 that fixes that.

If you still want to contribute your change with syscfg definition please don't hesitate to use force push to your branch so we can limit number of commits with names like `fixes review comments`, and we can merge you commits without squashing them first (that can drop authorship).

kasjer (approved):
Looks good.
I squashed it manually to have just nice commit instead of two.
No worries though @5frank, you are still the author of the change.

PR #2387, opened 2020-10-06
sys/log/cbmem: Fix return code for early break in log_cbmem_walk

caspermeijn:
The return code of succesfull log_cbmem_walk was 0, even when the loop
ended early. In case of an early loop break, we want to return the rc of
the walked function.
This is needed to make `newtmgr log show --all` work, as the rc is
passed to the client to indicate more data available.

The return codes of the log_walk functions seem a bit of a mess, as they both indicate an error state as rc of the walk_func. To bad C doesn't support multiple return values.


ccollins476ad (approved):
Thanks @caspermeijn.  Good catch!

PR #2391, opened 2020-10-13
apps: Fix cbmem sizes

caspermeijn:
The size passed to cbmem_init is in bytes. Therefore the datatype of
cbmem_buf needs to be one byte long.


ccollins476ad (approved):
Thanks, @caspermeijn!

PR #2394, opened 2020-10-20
hw/mcu/dialog: Acquire PD_SYS for DMA

andrzej-kaczmarek:
DMA controller resides in PD_SYS domain so we cannot turn off that
domain if any DMA channes is acquired. That basically means on 1st
acquired DMA channel we need to also acquire PD which we can release
after last DMA channel is released. Note that acquiring PD_SYS means
M33 cannot be powered down in deep sleep.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/da1469x_dma.c
+++ b/hw/mcu/dialog/da1469x/src/da1469x_dma.c
@@ -21,6 +21,7 @@
 #include <stdint.h>
 #include "os/os_trace_api.h"
 #include "mcu/da1469x_dma.h"
+#include "mcu/da1469x_pd.h"
 #include "mcu/cmsis_nvic.h"
 #include "defs/error.h"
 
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
@JustineKH ^^

agross-korg:
I really don't like the overall design of the power management.  This patch fixes the issue of instability when we go to deep sleep while the DMA is active.  This patch also introduces increased power usage if we happen to go to deep sleep while the DMA is active.

agross-korg:
I'm saying that once we fix this we have more work to do to make this power efficient.

kasjer (approved):

agross-korg (approved):

nkaje (approved):

wes3 (approved):
This is odd. I thought I added a review but it appears not. I took the code and did some rudimentary testing.

PR #2395, opened 2020-10-20
da1469x_pd: add function to get power domain reference counts

JustineKH:
da1469x_pd.c: Add function to get power domain reference counts.
hal_gpio.c: Restore Px_DATA_REG from retained values. Seeing issues/hangs when attempting to restore it from itself, when it is possibly not retained when PD_COM is down.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_gpio.c
+++ b/hw/mcu/dialog/da1469x/src/hal_gpio.c
@@ -89,6 +89,7 @@ static struct hal_gpio_irq hal_gpio_irqs[HAL_GPIO_MAX_IRQ];
 
 #if MYNEWT_VAL(MCU_GPIO_RETAINABLE_NUM) >= 0
 static uint32_t g_mcu_gpio_latch_state[2];
+static uint32_t g_mcu_gpio_data_latch_state[2];
 static uint8_t g_mcu_gpio_retained_num;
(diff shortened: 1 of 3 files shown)


wes3 (approved):
I approve of the changes but my preference would be to return the native type of an element in a structure if you are returning that element. Not sure everyone would agree with that btw. Returning and int will work but my preference would be that this return uint8_t. You could also have done the following and saved some typing but no need to change :-)

assert(pd < PD_COUNT);
return g_da1469x_pd_data[pd].refcnt;

Anyway, all good and no need to change anything.

andrzej-kaczmarek:
adding new API is fine, although I only see it being used to make workarounds like in the 2nd patch which should not be merged upstream unless this fix can be explained.

hw/mcu/dialog/da1469x/src/hal_gpio.c
```diff
@@ -473,6 +474,10 @@ mcu_gpio_enter_sleep(void)
         return;
     }
 
+    /* Save pins states and their latched values */
```
benmccrea, line 477: @JustineKH Are interrupts already disabled when `mcu_gpio_enter_sleep` gets called? If not,  an interrupt could change pin levels before entering sleep.

andrzej-kaczmarek: the whole sleep code is run with interrupts disabled since it's called from 'os_tick_idle`

PR #2398, opened 2020-10-21
hw/mcu/dialog: Support updated TLV values for booting

agross-korg:
This patch adds handling of both legacy values and updated values for the
IMAGE_TLV_AES_NONCE and IMAGE_TLV_SECRET_ID TLVs.  The reason for the change
is to avoid overlapping TLV usage with upstream mcuboot.


nkaje (approved):
LGTM

PR #2399, opened 2020-10-23
flash_loader: Fix awk script

kasjer:
flash_loader.gdb generated script that was later sourced back
into gdb with file_sz variable computed in shell.

This script had extra backslash that was not really needed.
awk complained about this extra \


kasjer:
Thanks to @JustineKH and @brolan-juul for helping with this issue.

andrzej-kaczmarek (approved):

PR #2400, opened 2020-10-23
stm32/hal_spi: Fix alternate function spec for SPI_4

kasjer:
SPI_4 maps to STM SPI5.
Alternate function for this block is not always GPIO_AF5_SPI5.
For some MCUs (STM32F411) it should be GPIO_AF6_SPI5.
So now one that is defined by STM HAL will be used.


vrahane (approved):
Looks good.

PR #2401, opened 2020-10-23
hw/mcu/dialog: Keep GPIO mux powered if non-GPIO pins enabled

andrzej-kaczmarek:
If any pin has non-GPIO function we should not allow PD_COM to be
disabled since this will also reset GPIO mux settings and this can
lead to unpredictable behavior.

However, if there are no pins configured or all of them are used
as standard GPIO, we can just latch their state and disable PD_COM
safely.


hw/mcu/dialog/da1469x/src/hal_gpio.c
```diff
@@ -90,6 +90,7 @@ static struct hal_gpio_irq hal_gpio_irqs[HAL_GPIO_MAX_IRQ];
 #if MYNEWT_VAL(MCU_GPIO_RETAINABLE_NUM) >= 0
 static uint32_t g_mcu_gpio_latch_state[2];
 static uint8_t g_mcu_gpio_retained_num;
+static uint32_t g_mcu_gpio_can_sleep;
```
kasjer, line 93: for now this variable is used only with values 0/1 maybe uint32_t is too much

PR #2402, opened 2020-10-24
hw/battery: Fix strncpy warning

caspermeijn:
GCC complains about a strncpy call when building using build_profile speed.
This warning indicates a possible string truncation. The only proper
solution I could find was to decrease the buf_size and explicitly set a
zero byte.

Error: In function 'battery_prop_get_name',
    inlined from 'battery_find_property_by_name' at repos/apache-mynewt-core/hw/battery/src/battery.c:392:9:
repos/apache-mynewt-core/hw/battery/src/battery.c:379:5: error: 'strncpy' specified bound 20 equals destination size [-Werror=stringop-truncation]
  379 |     strncpy(buf, driver_prop->bdp_name, buf_size);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/hw/battery/src/battery.c
+++ b/hw/battery/src/battery.c
@@ -376,7 +376,8 @@ battery_prop_get_name(const struct battery_property *prop, char *buf,
     const struct battery_driver_property *driver_prop =
             &driver->bd_driver_properties[prop->bp_prop_num - driver->bd_first_property];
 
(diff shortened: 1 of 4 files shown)


kasjer (approved):
LGTM

PR #2403, opened 2020-10-25
Detect connected programmer to download and debug firmware

rafacouto:
Here is a new bash funtion in _hw/scripts/common.sh_ to detect connected programmers. I've added detection by USB vendor:product IDs, but other methods could be implemented. At the moment I've tested with Segger's **jlink** nrf5x on-board, **cmsis-dap** on STM32 and **stlink-v2** and **stlink-v2.1**.

Detected programmer is put on **DETECTED_PROGRAMMER variable** so download or debug script for some board will be able to use the bash function. For example, scripts for pinetime BSP are using this feature.

### DIFF
--- a/hw/scripts/common.sh
+++ b/hw/scripts/common.sh
@@ -89,6 +89,9 @@ detect_programmer() {
     echo "$USB_DEV" | grep -q -i '0483:3748'
     [ $? -eq 0 ] && DETECTED_PROGRAMMER='stlink-v2'
 
+    echo "$USB_DEV" | grep -q -i '0483:374b'
+    [ $? -eq 0 ] && DETECTED_PROGRAMMER='stlink-v2-1'
+
     echo "$USB_DEV" | grep -q -i '1366:1015'
     [ $? -eq 0 ] && DETECTED_PROGRAMMER='jlink'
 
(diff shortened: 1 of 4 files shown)


caspermeijn:
I like this idea. I have a ST-link V2 myself, so I will not benefit directly. This patch work on my machine.

Would it make sense for some of the code to move to `openocd.sh`? This way more BSPs can benefit from this change.

rafacouto:
> Would it make sense for some of the code to move to `openocd.sh`? This way more BSPs can benefit from this change.

You are right: moved openocd code to openocd.sh script

hw/scripts/common.sh
```diff
@@ -76,3 +76,27 @@ parse_extra_jtag_cmd() {
     echo $NEW_EXTRA_JTAG_CMD
     EXTRA_JTAG_CMD=$NEW_EXTRA_JTAG_CMD
 }
+
+# Try to detect connnected programmers
+detect_programmer() {
+
+    # scan USB for well-known VID:PID
+    USB_DEV=$(ls /sys/bus/hid/devices)
```
kasjer, line 84: Are you sure that _stlink_ and _jlink_ are **HID** devices.
It does not show up on my Linux (for Windows there is no such folder but we can leave it for now).
`lsusb` on the other hand does show instances of stlink, jlink and dap.
Maybe command is valid for Mac?

rafacouto: I've trusted on sysfs too much :angel: _lsusb_ command is the right way. Reviewing now... Thanks for your comments :+1:

kasjer (approved):
LGTM

PR #2405, opened 2020-10-28
hw/drivers/i2s: Add I2S driver for DA1469x family

kasjer:
I2S driver implementation for DA1469X MCUs.
Only TX part for now (speaker).

Three data formats supported 16/16 16/32 and 32/32.

### DIFF
--- a/hw/drivers/i2s/i2s_da1469x/include/i2s_da1469x/i2s_da1469x.h
+++ b/hw/drivers/i2s/i2s_da1469x/include/i2s_da1469x/i2s_da1469x.h
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef _I2S_DA1469X_H
+#define _I2S_DA1469X_H
+
+#include <stdint.h>
+
+struct da1469x_dma_buffer {
+    uint16_t size;
+    uint8_t *buffer;
+};
+
+#define I2S_DA1469X_DMA_BUFFER_DEF(_name, _size) \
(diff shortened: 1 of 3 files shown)


hw/drivers/i2s/i2s_da1469x/src/i2s_da1469x.c
```diff
@@ -0,0 +1,433 @@
...
+        .priority = 0,
+        .burst_mode = 0,
+        .bus_width = data_format == I2S_DATA_FRAME_16_32 ? MCU_DMA_BUS_WIDTH_2B : MCU_DMA_BUS_WIDTH_4B,
+    };
+
+#if MYNEWT_VAL_CHOICE(MCU_SYSCLK_SOURCE, PLL96)
+    system_clock = 96000000;
+#else
+    system_clock = 32000000;
+#endif
```
andrzej-kaczmarek, line 334: There's `SystemCoreClock` global defined by CMSIS which holds system core clock. In DA1469x MCU we do not have it defined properly, but perhaps it would be better to do so instead of hardcoding clocks here.

hw/drivers/i2s/i2s_da1469x/syscfg.yml
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
```
andrzej-kaczmarek, line 19: no need for empty syscfg file

hw/drivers/i2s/i2s_da1469x/include/i2s_da1469x/i2s_da1469x.h
```diff
@@ -0,0 +1,71 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef _I2S_DA1469X_H
+#define _I2S_DA1469X_H
+
+#include <stdint.h>
+
+struct i2s;
```
andrzej-kaczmarek, line 25: this struct is not used in .h file so forward declaration is redundant

andrzej-kaczmarek (changes requested):

hw/drivers/i2s/i2s_da1469x/src/i2s_da1469x.c
```diff
@@ -0,0 +1,430 @@
...
+        rregs->DMA_LEN_REG = number_of_transfers;
+        rregs->DMA_CTRL_REG |= DMA_DMA0_CTRL_REG_DMA_ON_Msk;
+    }
+    lregs->DMA_CTRL_REG |= DMA_DMA0_CTRL_REG_DMA_ON_Msk;
+}
+
+struct pcm_div {
+    uint16_t div;
+    uint16_t fdiv;
+};
```
andrzej-kaczmarek, line 245: now we have two identical definitions of the same struct ;)

kasjer: fixed

andrzej-kaczmarek (approved):

PR #2406, opened 2020-11-02
hw/bsp/pinetime: Setup I2C

rafacouto:
Enable the I2C port 1 for the Pinetime. Maximum speed of 400 KHz is set since connected I2C slave devices are >= 400 KHz.


vrahane (approved):
Looks good.

PR #2407, opened 2020-11-03
This change adds protection against rollover at high altitudes.

supervillain101:
Fix was provided by Bosch.

At high altitudes there is a potential for rollover in the compensate_pressure() function.  To prevent that, Bosch provided a fix with a divide by 10 in the affected code and then a multiply by 10 after to bring the final value back to the expected value.

### DIFF
--- a/hw/drivers/sensors/bmp388/src/bmp388.c
+++ b/hw/drivers/sensors/bmp388/src/bmp388.c
@@ -1343,7 +1343,8 @@ compensate_pressure(const struct bmp3_uncomp_data *uncomp_data,
     BMP388_LOG_ERROR("*****partial_data4 high32bit = 0x%x low32bit = 0x%x\n", (uint32_t)((partial_data4)>>32),(uint32_t)(partial_data4&0xffffffff));
 #endif
 
-    partial_data5 = (partial_data4 * uncomp_data->pressure) / 512;
+    partial_data5 = (partial_data4 * (uncomp_data->pressure / 10)) / 512;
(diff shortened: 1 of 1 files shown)


vrahane (changes requested):
Please write better commit messages. Rest looks fine.

vrahane:
@supervillain101 What I meant by that is, we generally write commit messages like this:

`sensors/bmp388: Adding protection against rollover at high altitudes.`

It gives a small description of the fix along with the module that you are changing.

wes3 (commented):
Some unsolicited comments here. I prefer using parentheses so you do not need to think about associativity and precedence of operators. I think there should be parentheses like this: partial_data5 = (partial_data4 * (uncomp_data->pressure / 10)) / 512;

I thought that * and / have same precedence and are evaluated left to right but I could be wrong; I do not remember this stuff as I always use parentheses so I do not have to worry about it but I do not think you get the same answer.

Another thing: why was the value 10 chosen and why was uncomp_data->pressure chosen to divide? I would divide the larger number as opposed to the smaller number but maybe the pressure is generally larger than the other. One final note here: it is always better to divide by powers of 2 so if 8 works I would have chosen 8 :-)

vrahane:
I agree with what @wes3 is saying, it makes sense to use parenthesis above.

supervillain101:
@vrahane Changes made to parens and commit message.

vrahane (approved):
Looks good.

PR #2408, opened 2020-11-04
gdbmacros: Add script to extract cmac core

kasjer:
Core file from DA1469x platform contains embedded memory
dump from Cortex-M0 core.

cmac_core_dump.gdb adds command to extract cmac core to separate core
file that can be analyzed.

Crashes inside interrupt handler were not tested.

### DIFF
--- a/compiler/gdbmacros/cmac_core_dump.gdb
+++ b/compiler/gdbmacros/cmac_core_dump.gdb
@@ -1,0 +1,97 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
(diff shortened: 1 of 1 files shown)


wes3 (commented):
I generated a core using the script and am looking at it along with the application elf file. I can see variables no problem (at least they seem reasonable). What I do not see are M0 core registers using info reg. Not sure if these were expected to be part of the generated core.

kasjer:
Registers should be there, if they can be found.
I tested it crashes where core crash originated on CMAC side (not sure what would happen if M33 crashes).
Only crashes in process mode were tested so if crash happened during interrupt that may be not handled correctly.

compiler/gdbmacros/cmac_core_dump.gdb
```diff
@@ -0,0 +1,97 @@
...
+    find /w /1 $_cmac_ram_start, +$_ram_size, $_lr, $_pc
+    # exception pointer from address of LR
+    set $_ep = (uint32_t *)($_ - 20)
+    # SP in cmac address space
+    set $_cmac_sp = (uint32_t *)((int)$_ep - (int)$_cmac_ram_start + 0x20000000)
+
+    # Exception in system mode, Trap Frame just before exception frame
+    if $_ep[-1] == 0xFFFFFFF1 || $_ep[-1] == 0xFFFFFFF9
+        set $_tf = $_ep - 10
+    else
```
wes3, line 51: btw: what is _tf used for?

andrzej-kaczmarek: I think it should be either `$_tp` here or `$_tf` below :)

wes3: Yes, I figured that was a typo :-)

kasjer: fixed

wes3 (approved):
LGTM

PR #2410, opened 2020-11-05
hw/mcu/dialog: Use 256kHz clock during XTAL32M settling

kasjer:
RC32K was used to wait for XTAL32M to settle.
Unfortunately it's declared frequency range is 25kHz to 39kHz.
XTAL32M settling time was taken into account for CMAC wakeup
calculation but it can be very inaccurate when RC32K is used.

This can result in CMAC crashes due to assert when Cortex-M0
wakes up to late.

This switches to 256kHz clock (derived from RC32M) that is much
mode precise.


benmccrea:
@kasjer I've been testing on this for a couple of days and have not seen a CMAC crash.  
@andrzej-kaczmarek If this looks good to you, would you approve?

andrzej-kaczmarek (approved):

PR #2411, opened 2020-11-09
log: yield periodically while dumping logs

nkaje:
The logs are processed in the default task.
In cases where there are a lot of logs to be
processed, the idle task may get blocked from
running, resulting in a watchdog bite. Address
this by periodically putting ourselves on the
sleep list.


kasjer:
Maybe instead of yielding code could feed watchdog.
If we have two tasks that are busy even if they go to sleep periodically it may happen that they will switch between them and idle will never feed watchdog.

But it could be that watchdog is not initialized at all then feeding it could not be the right thing to do, but yielding without watchdog running will just make walk slower.

benmccrea:
@nkaje When it yields to another task, will a problem occur if the new task creates a new log entry while the dump is still in progress?

PR #2413, opened 2020-11-19
STM32 RTC based os tick

kasjer:
STMF1 already had support for tickles sleep.
Now other MCU with quite distinct RTC also can be configured to LSE/RTC as system tick source.

RTC is added to F0/F4/L0/L1/L4/WB55 which share similar hardware design.
Small changes are present for L0/L1/F0 MCUs.

Some families have low power clocks other then RTC that could be used as tick source but some don't hence common code.

RTC in ST MCUs can be powered by battery and then time can be kept even when MCU is shut down.
Wall clock changes made in current API are stored as UTC time in RTC and are restored at boot time if RTC is powered by battery.

This is first step to use low power STOP mode when tick can wake up system while high speed clocks are stopped.


hw/mcu/stm/stm32_common/src/hal_os_tick.c
```diff
@@ -54,6 +55,315 @@ __WFI(void)
...
+    },
+    .AlarmMask = RTC_ALARMMASK_ALL,
+    .AlarmSubSecondMask = 0,
+    .Alarm = RTC_ALARM_A,
+};
+
+static uint32_t
+rtc_time_to_sub_seconds(const RTC_TimeTypeDef *time)
+{
+    int32_t sub_seconds;
```
sjanc, line 122: is this signed type on purpose?

kasjer: it was, but now I see that it does not need to be, I will change it to unsigned.

hw/mcu/stm/stm32_common/src/hal_os_tick.c
```diff
@@ -54,6 +55,315 @@ __WFI(void)
...
+{
+    struct os_timeval utc;
+    struct clocktime ct;
+    ct.year = 2000 + date->Year;
+    ct.mon = date->Month;
+    ct.day = date->Date;
+    ct.hour = time->Hours;
+    ct.min = time->Minutes;
+    ct.sec = time->Seconds;
+    ct.usec = ((time->SecondFraction - time->SubSeconds) * 15625) >> SUB_SECONDS_BITS;
```
sjanc, line 298: ct.dow is left uninitialized here

kasjer: ct.dow is not used for conversions from clocktime, it is filled when conversion is done other way round.
day of week can be incorrect in hardware, MCU will not enforce correct value so there is no guarantee that value extracted from registers is correct.
I saw no point in computing correct value since conversion function will not use it any way.

vrahane:
@kasjer Looks good to me but needs a rebase.

kasjer:
> @kasjer Looks good to me but needs a rebase.

@vrahane I rebased with necessary corrections, thanks for checking out.

PR #2415, opened 2020-11-21
BSP de-initialization to be executed before boot loader start application

kasjer:
This introduces empty hal_bsp_deinit() function
that is intended to be called by bootloader
before it calls application reset handler.

mcuboot calls hal_bsp_init() at start it can leave
various resources initialized before application
starts.
In can leave to situation when such resource (timer/uart)
is configured with interrupt enabled.
Then in bad case interrupt can be activated before resource
is reconfigured, or interrupt are remapped but often RAM (bss and data)
are already wiped out and interrupt (with handler still from
bootloader will try to execute with its data structures corrupted.

matching mcuboot pr: https://github.com/mcu-tools/mcuboot/pull/886

### DIFF
--- a/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
+++ b/hw/bsp/ada_feather_nrf52/src/hal_bsp.c
@@ -101,3 +101,8 @@ hal_bsp_init(void)
     /* Create all available nRF52840 peripherals */
     nrf52_periph_create();
 }
+
+void
+hal_bsp_deinit(void)
+{
+}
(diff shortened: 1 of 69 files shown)


utzig (approved):

wes3 (commented):
I think the idea of hal_bsp_deinit() is a good one. I have not looked at all the registers for the stm32 code you have added. Are these just the reset values of these registers?

kasjer:
> I think the idea of hal_bsp_deinit() is a good one. I have not looked at all the registers for the stm32 code you have added. Are these just the reset values of these registers?

Pretty much so.
For enable registers we have reset value (so for most of them 0 with some exceptions like RAM and FLASH).
For reset registers, all bits that are available are first set then reset.

PR #2419, opened 2020-11-25
hw/mcu/dialog: Add SystemCoreClock variable

kasjer:
There was extarn for SystemCoreClock but variable was not defined
like on others MCUs.

This adds variable and sets it according to current system clock
selection.


andrzej-kaczmarek (approved):
lgtm

PR #2423, opened 2020-11-30
hw/mcu/dialog: Disconnect UART RX pin when grounded

kasjer:
When UART RX pin is shorted to ground and UART is enabled it enters
busy state.
In this state busy interrupt fires constantly it prevents non-interrupt code
to be executed and can lead to watchdog reboot.
In function hal_uart_config() pullup was added to RX pin to enable configuration
(which was not successful in busy state).
This counter measure worked fined when RX pin was floating and just happen to be
low for a while but internal pull-up resistor was enough to make is stable.
However when RX pin is shorted to ground internal pull-up will not solve the problem.

- now RX pin is not routed to UART during configuration to prevent busy state at that time.
- RX can be disconnected from UART once it enters busy state, pin is reconfigured for
GPIO with interrupt to detect end of short condition. (This functionality is optionall
and can be enabled with syscfg value MCU_UART_DICONNECT_RX_ON_BUSY).

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_uart.c
+++ b/hw/mcu/dialog/da1469x/src/hal_uart.c
@@ -25,6 +25,7 @@
 #include "mcu/da1469x_hal.h"
 #include "mcu/mcu.h"
 #include "hal/hal_uart.h"
+#include "hal/hal_gpio.h"
 #include "defs/error.h"
 #include "os/os_trace_api.h"
 #include "os/util.h"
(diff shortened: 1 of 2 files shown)


agross-korg:
This works pretty well.  I tested using both an mcuboot and an application image.  The only thing I did notice is that most of the early output from the application (device TX) is lost.  However, after the image boots, output is working.  The patches only affect RX so this was kind of interesting.  Have you seen anything similar?

kasjer:
> This works pretty well. I tested using both an mcuboot and an application image. The only thing I did notice is that most of the early output from the application (device TX) is lost. However, after the image boots, output is working. The patches only affect RX so this was kind of interesting. Have you seen anything similar?

I did not noticed, I will check some more.

agross-korg:
works fine for me.  Scratch the comments on the app issues.  We use a separate driver for uart on that, so that is a separate issue.

agross-korg (approved):

PR #2425, opened 2020-12-02
log_fcb: optimize finding records towards the end of circular buffer

nkaje:
In scenarios where the interested record lies towards the end of a
large circular buffer, walking it from the beginning can take up
significant amount of time.

It is possible that the thread that walks the log is shared among
other user interfaces and until the bookmarks are built up, flash
based walks and lookups can make the system quite slow to respond.

Optimize this by beginning to walk from the flash area that is closer
to the latest one in use.

In the tests performed, the walk time was reduced by 2 orders of
magnitude (e.g 9000 ms to 50 ms).

### DIFF
--- a/sys/log/full/src/log_fcb.c
+++ b/sys/log/full/src/log_fcb.c
@@ -32,6 +32,71 @@
 
 static int log_fcb_rtr_erase(struct log *log);
 
+static int
+fcb_get_fa_hdr(struct fcb *fcb, struct log *log, struct fcb_entry *fcb_entry, struct log_entry_hdr *hdr)
+{
+    int rc;
+
+    rc = fcb_getnext(fcb, fcb_entry);
+    if (rc == 0) {
+        return log_read_hdr(log, fcb_entry, hdr);
+    } else {
+        return rc;
+    }
+}
+
+/**
+ * Helper function to find start point for walking, given an offset.
+ * for a non-zero offset, instead of walking from the beginning,
+ * walk from the last applicable area, check first entry and compare
+ * against the given offset. Repeat this until an offset less than the
+ * given offset is found, start walking from there.
+ */
+static int
+fcb_walk_back_find_start(struct fcb *fcb, struct log *log, struct log_offset *log_offset, struct fcb_entry *fcb_entry)
+{
+    struct flash_area *fap;
+    struct log_entry_hdr hdr;
+    struct fcb_entry iter_entry = {0};
+    int rc;
+
+    /**
+     * If the provided lo_index is less than the oldest entry,
+     * simply return the oldest. Else, start from the active area and search back.
+     */
+    iter_entry.fe_area = fcb->f_oldest;
+    rc = fcb_get_fa_hdr(fcb, log, &iter_entry, &hdr);
+    if (rc != 0) {
+        return rc;
+    }
+
+    if (hdr.ue_index >= log_offset->lo_index) {
+        goto found_ent;
+    }
+
+    fap = fcb->f_active.fe_area;
+
+    for (hdr.ue_index = log_offset->lo_index; hdr.ue_index >= log_offset->lo_index;) {
+        memset(&iter_entry, 0, sizeof(iter_entry));
+        iter_entry.fe_area = fap;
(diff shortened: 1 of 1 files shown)


sys/log/full/src/log_fcb.c
```diff
@@ -102,17 +132,24 @@ log_fcb_find_gte(struct log *log, struct log_offset *log_offset,
...
+    lwga.log_offset     = log_offset;
+    lwga.fcb_entry      = out_entry;
+    rc = fcb_walk(fcb, fap, fcb_walk_gte, (void *)&lwga);
+    /* Not found, try the active area */
+    if (rc == 0) {
+        fap = fcb->f_active.fe_area;
+        rc  = fcb_walk(fcb, fap, fcb_walk_gte, (void *)&lwga);
+    }
+
+    assert(rc == FCB_FOUND);
```
nkaje, line 152: expect to find every time, not sure assert is the right thing to do here though.

nkaje:
To test `sys/log: optimize header read based on image hash`, use this block immediately after the optimized read.
```
+    /* verify */
+    bytes_read = log_read(log, dptr, &ue_imghash[0],
                               LOG_BASE_ENTRY_HDR_SIZE, LOG_IMG_HASHLEN);
-        if (bytes_read != LOG_IMG_HASHLEN) {
-            return SYS_EIO;
-        }
+    if (bytes_read != LOG_IMG_HASHLEN) {
+        return SYS_EIO;
     }
+    if (memcmp(hdr->ue_imghash, &ue_imghash[0], LOG_IMG_HASHLEN) != 0) {
+        console_printf("imghash read optimization check failed");
+        return SYS_EIO;
+    }
```

sys/log/full/src/log_fcb.c
agross-korg: or return an error

sys/log/full/src/log.c
```diff
@@ -955,18 +955,17 @@ log_read_hdr(struct log *log, const void *dptr, struct log_entry_hdr *hdr)
...
+    if (bytes_read != (LOG_BASE_ENTRY_HDR_SIZE + LOG_IMG_HASHLEN)) {
+        return SYS_EIO;
+    }
+#else
     bytes_read = log_read(log, dptr, hdr, 0, LOG_BASE_ENTRY_HDR_SIZE);
     if (bytes_read != LOG_BASE_ENTRY_HDR_SIZE) {
         return SYS_EIO;
     }
-
-    if (hdr->ue_flags & LOG_FLAGS_IMG_HASH) {
```
vrahane, line 963: I think this is going to cause issues, the image hash was part of the header for backwards compatibility. Removing it might cause issues for other people.

nkaje: Thanks Vipul. The gains from this is less compared to the other change, so updated the PR and dropped this change. The unit tests should be probably expanded to ensure the backward compatibility use-cases and this change could be revisited.

sys/log/full/src/log_fcb.c
```diff
@@ -32,6 +32,41 @@
...
+fcb_walk_back_find_start(struct fcb *fcb, struct log *log, struct log_offset *log_offset, struct fcb_entry *fcb_entry)
+{
+    struct flash_area *fap;
+    struct log_entry_hdr hdr;
+    struct fcb_entry iter_entry;
+    int rc;
+
+    fap = fcb->f_active.fe_area;
+
+    for (hdr.ue_index = log_offset->lo_index; hdr.ue_index >= log_offset->lo_index; fap--) {
```
kasjer, line 52: **fap** can't be just decremented, it is circular buffer, sometimes it wraps around so fap should be checked for this condition.

kasjer (changes requested):

kasjer:
Just one more, if lo_index is less then anything that is in fcb, loop will continue forever or maybe it will exit with rc != 0, but later this error will prevent walking altogether.

nkaje:
> Just one more, if lo_index is less then anything that is in fcb, loop will continue forever or maybe it will exit with rc != 0, but later this error will prevent walking altogether.

Thanks for the review! Added a check to address this case.

kasjer (approved):
LGTM

PR #2426, opened 2020-12-02
syscfg: OS_TICKS_PER_SEC definition workaround

kasjer:
OS_TICK_PER_SEC is configurable via syscfg in fe310/nrf52/stmf1.
nrf52 and stmf1 had this changed not so long ago.
Unfortunately the way it was defined:
#define OS_TICKS_PER_SEC MYNEWT_VAL(OS_TICKS_PER_SEC)
collides with existing usage where some other
syscfg values refer to definition OS_TICKS_PER_SEC
(not MYNEWT_VAL(..) which is not valid for all MCUs)
and that makes compilation fail with undefined MYNEWT_VAL() function.

If OS_TICKS_PER_SEC is to be configurable (and that seems reasonable
for many cases) it would be good to define it for all MCUs,
then change few yml files that refer to plain OS_TICKS_PER_SEC
to use MYNEWT_VAL() instead.

For now as workaround MYNEWT_VAL(OS_TICKS_PER_SEC) is expanded
in 3 files to avoid preprocessor problem.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/include/mcu/cortex_m4.h
+++ b/hw/mcu/nordic/nrf52xxx/include/mcu/cortex_m4.h
@@ -28,7 +28,7 @@
 extern "C" {
 #endif
 
-#define OS_TICKS_PER_SEC    MYNEWT_VAL(OS_TICKS_PER_SEC)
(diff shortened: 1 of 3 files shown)


kasjer:
Alternative solution applied

PR #2427, opened 2020-12-07
hw/drivers/bmp388: Fix compilation warning

kasjer:
Compiler detects this:
bmp388.c:3596:12: error: 'rc' may be used uninitialized in this function

Fixed by initializing variable to 0.


vrahane (approved):
LGTM

PR #2428, opened 2020-12-07
apps/ble: Use store/config instead of store/ram

kasjer:
store/ram was deprecated
Now application are configured to use store/config and
BLE_STORE_CONFIG_PERSIST set to 0.


andrzej-kaczmarek (approved):

kasjer:
This https://github.com/apache/mynewt-nimble/pull/889 needs to be merged first to make CI work

PR #2429, opened 2020-12-07
hw/drivers/i2s: Fix build for MCUs that have fewer I2S

kasjer:
I2S number can very from MCU to MCU.
I2S2 and I2S3 are common but not many MCUs have 5 I2S blocks.

This adds proper preprocessor guards against non existing configurations.


utzig (approved):
LGTM

PR #2430, opened 2020-12-07
standardize OS_TICKS_PER_SEC definition

kasjer:
In several instances OS_TICKS_PER_SEC was configurable via syscfg.

This PR proposes to have same behavior across all MCUs.

Now only one definition in os_time.h exists and it gets it form syscfg.
All MCUs have definition moved to syscfg.
~~In few cases other syscfg values that referred to OS_TICKS_PER_SEC
are using MYNEWT_VAL(OS_TICKS_PER_SEC) to avoid circular
dependency problem.~~

This requires **newt** with https://github.com/apache/mynewt-newt/pull/423 included.

### DIFF
--- a/hw/mcu/ambiq/apollo2/include/mcu/cortex_m4.h
+++ b/hw/mcu/ambiq/apollo2/include/mcu/cortex_m4.h
@@ -26,8 +26,6 @@
 extern "C" {
 #endif
 
-#define OS_TICKS_PER_SEC    (128)
-
 static inline void
 hal_debug_break(void)
 {
--- a/hw/mcu/ambiq/apollo2/syscfg.yml
+++ b/hw/mcu/ambiq/apollo2/syscfg.yml
@@ -123,3 +123,6 @@ syscfg.defs:
                     APOLLO2_TIMER_SOURCE_TMRPINA
                     APOLLO2_TIMER_SOURCE_BUCKA
         value:
+
+syscfg.vals:
+    OS_TICKS_PER_SEC: 128
--- a/hw/mcu/arc/snps/include/mcu/mcu_arc.h
+++ b/hw/mcu/arc/snps/include/mcu/mcu_arc.h
@@ -24,8 +24,6 @@
 extern "C" {
 #endif
 
-#define OS_TICKS_PER_SEC    (100)
-
 static inline void
 hal_debug_break(void)
 {
--- a/hw/mcu/dialog/cmac/include/mcu/cortex_m0.h
+++ b/hw/mcu/dialog/cmac/include/mcu/cortex_m0.h
@@ -24,8 +24,6 @@
 extern "C" {
 #endif
 
-#define OS_TICKS_PER_SEC    (128)
-
 static inline void
 hal_debug_break(void)
 {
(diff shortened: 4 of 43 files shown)


net/ip/native_sockets/syscfg.yml
```diff
@@ -27,7 +27,7 @@ syscfg.defs:
         description: >
             The frequency at which to poll for received data.  Units
             are OS ticks.
-        value: 'OS_TICKS_PER_SEC / 5'
+        value: 'MYNEWT_VAL(OS_TICKS_PER_SEC) / 5'
```
kasjer, line 30: It turns out that such expressions are not handled sensible by newt tool.

caspermeijn (commented):
I think it makes sense to standardize. I found this problem recently as well, when I wanted to use `hw/util/button`.

Some questions out of curiosity:
- Why would someone want to change this value?
- Why do all mcu define their own value? 
- Would it make sense to standardize the actual value?

caspermeijn:
I can't compile with this version of mynewt-core:
```
$ newt build klok-pinetime
Building target targets/klok-pinetime
Error: Priority violations detected (Packages can only override settings defined by packages of lower priority):
    Package: @apache-mynewt-core/hw/mcu/nordic/nrf52xxx overriding setting: OS_TICKS_PER_SEC defined by @apache-mynewt-core/kernel/os

Setting history (newest -> oldest):
    OS_TICKS_PER_SEC: [@apache-mynewt-core/hw/mcu/nordic/nrf52xxx:128, @apache-mynewt-core/kernel/os:]

```

andrzej-kaczmarek:
@caspermeijn you need to use latest newt (master) to build

wes3:
@caspermeijn I can answer some of the questions you posed.

1) Why would someone want to change this value?
Some may want to change this value so that the time resolution for the os tick to be greater or less than the default value set. Some folks may want a 1 msec OS tick resolution whereas some may want longer. If you use 128 ticks per second the rate is, I believe 7.8125 msecs per tick.
2) Why do all mcu define their own value?
The main reason for this, I believe, is the default timer used for generating the os tick. We wanted to be able to generate the tick period exactly. Consider a 1 MHz timer and a 32768 Hz timer. For the 32768 timer, generating a 10msec os tick is not easy given integer number of ticks. That is why 128 is chosen for those MCUs as the 7.8125msec period is an integer number of ticks for a 32768 timer. For 1MHz, a 10msec resolution was chosen since that is an integer number of ticks for that timer.
3) Would it make sense to standardize the actual value?
I think, given the reason all mcus define their own values, no. There are really only two values I think used that are "standard": 128 and 100. So this seems ok to me.

caspermeijn:
@andrzej-kaczmarek Thanks, with latest newt it builds.

@wes3 So basically both question are for to get more accurate time handling. Got it.

@kasjer `hw/util/button` build correctly for me with these changes.

andrzej-kaczmarek (approved):

caspermeijn (approved):

kasjer:
re-base was required over already approved version due to conflicts in pic32 code

PR #2433, opened 2020-12-10
remove OS_TICKS_PER_SEC from syscfg when not needed

kasjer:
Button and native sockets had some time values that measured in ticks
but value used indicated that normal time values could be applied.
syscfg values were updated to be more natural for human the machine.

### DIFF
--- a/hw/util/button/src/button.c
+++ b/hw/util/button/src/button.c
@@ -370,14 +370,14 @@ button_exec_fsm(button_t *button, int action)
  do_pressed:
 #if MYNEWT_VAL(BUTTON_USE_LONG)
     if (button->mode & BUTTON_FLG_LONG) {
-        os_callout_reset(callout, MYNEWT_VAL(BUTTON_LONGHOLD_TICKS));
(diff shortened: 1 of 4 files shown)


hw/util/button/src/button.c
```diff
@@ -370,14 +370,14 @@ button_exec_fsm(button_t *button, int action)
  do_pressed:
 #if MYNEWT_VAL(BUTTON_USE_LONG)
     if (button->mode & BUTTON_FLG_LONG) {
-        os_callout_reset(callout, MYNEWT_VAL(BUTTON_LONGHOLD_TICKS));
+        os_callout_reset(callout, MYNEWT_VAL(BUTTON_LONGHOLD_TIME_MS) * OS_TICKS_PER_SEC / 1000);
```
andrzej-kaczmarek, line 373: use `os_time_ms_to_ticks32`

andrzej-kaczmarek (changes requested):

andrzej-kaczmarek (approved):

caspermeijn:
LGTM
I like that is improves usability in the process of fixing a build failure. The `defunct` option produces a clear message when upgrading.

PR #2435, opened 2020-12-11
hw/drivers/uart: Add uart_open function

kasjer:
This adds thin wrapper to open UART device.
Generic os_dev_open has opaque pointer that in case of UART has
configuration but it's not clear what should be passed as argument
without looking to UART code.

With this change parameter check is enhanced as well as readability.


mlaz (approved):
LGTM

andrzej-kaczmarek (approved):

PR #2436, opened 2020-12-14
Add da1469x uart driver

kasjer:
Driver that shuts down UART when RX line goes down.
Uses de-bouncing to wait for RX line to be stable high again before
re-enabling UART.

### DIFF
--- a/hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
+++ b/hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
@@ -1,0 +1,563 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <assert.h>
+#include <string.h>
+#include <bsp.h>
+#include "uart/uart.h"
+#include "mcu/da1469x_pd.h"
+#include "mcu/da1469x_hal.h"
+#include "hal/hal_gpio.h"
+#include "uart_da1469x/uart_da1469x.h"
+
+static void da1469x_uart_uart_isr(void);
+static void da1469x_uart_uart2_isr(void);
+static void da1469x_uart_uart3_isr(void);
+
+struct da1469x_uart_hw_data {
+    UART2_Type *regs;
+    IRQn_Type irqn;
+    mcu_gpio_func rx_pin_func;
+    mcu_gpio_func tx_pin_func;
+    mcu_gpio_func rts_pin_func;
+    mcu_gpio_func cts_pin_func;
+    /* Mask for (RE)SET_CLK_COM_REG */
+    uint8_t clk_com_mask;
+    void (*isr)(void);
+};
+
+static const struct da1469x_uart_hw_data da1469x_uart0_hw = {
+    .regs = (UART2_Type *)UART,
+    .irqn = UART_IRQn,
+    .rx_pin_func = MCU_GPIO_FUNC_UART_RX,
+    .tx_pin_func = MCU_GPIO_FUNC_UART_TX,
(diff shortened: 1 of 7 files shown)


hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
```diff
@@ -0,0 +1,563 @@
...
+    .rts_pin_func = MCU_GPIO_FUNC_UART3_RTSN,
+    .cts_pin_func = MCU_GPIO_FUNC_UART3_CTSN,
+    .clk_com_mask = CRG_COM_RESET_CLK_COM_REG_UART3_ENABLE_Msk,
+    .isr = da1469x_uart_uart3_isr,
+};
+
+static struct da1469x_uart_dev *da1469x_dev0;
+static struct da1469x_uart_dev *da1469x_dev1;
+static struct da1469x_uart_dev *da1469x_dev2;
+
```
wes3, line 81: I am wondering if it is a good idea to wra the definitions of da1469x_uartX_hw structures and the *da1469x_devX pointers with MYNEWT_VAL(UART_X) if you see what I mean

kasjer: da1469x_uart_dev_create() has series of **if** statements:
```
    if (MYNEWT_VAL(UART_0) && uart_num == 0) {
```
references to **da1469x_devX** and **da1469x_uartX_hw** are only used directly there and compiler and linker get rid of cases when **UART_x** is not set. Interrupt handlers are also left out.
This way we still have benefit of compiler checking code that otherwise could be dropped out by pre-processor without syntax check.

wes3: OK. I did not check all the references to those so all good.

hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
```diff
@@ -0,0 +1,563 @@
...
+    /* Reset UART */
+    CRG_COM->RESET_CLK_COM_REG = dev->hw->clk_com_mask;
+    CRG_COM->SET_CLK_COM_REG = dev->hw->clk_com_mask;
+    uart->UART2_SRR_REG = UART2_UART2_SRR_REG_UART_UR_Msk |
+                          UART2_UART2_SRR_REG_UART_RFR_Msk |
+                          UART2_UART2_SRR_REG_UART_XFR_Msk;
+
+    /* Configure baudrate */
+    baudrate_cfg = (32000000 - 1 + (uc->uc_speed / 2)) / uc->uc_speed;
```
andrzej-kaczmarek, line 280: should we better use systemclock value here?

hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
```diff
@@ -0,0 +1,563 @@
...
+    } else if (MYNEWT_VAL(UART_1) && uart_num == 1) {
+        dev->hw = &da1469x_uart1_hw;
+        assert(da1469x_dev1 == NULL);
+        da1469x_dev1 = dev;
+    } else if (MYNEWT_VAL(UART_2) && uart_num == 2) {
+        dev->hw = &da1469x_uart2_hw;
+        assert(da1469x_dev2 == NULL);
+        da1469x_dev2 = dev;
+    } else {
+        return OS_EINVAL;
```
andrzej-kaczmarek, line 558: nitpick: `OS_ENOENT` makes more sense

hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    if (odev->od_flags & OS_DEV_F_STATUS_OPEN) {
+        return OS_EBUSY;
+    }
+
+    dev->uc = *uc;
+
+    if (uc->uc_speed < 1200 || uc->uc_speed > 1000000) {
+        return SYS_ENOTSUP;
```
andrzej-kaczmarek, line 381: use OS error codes, as above, since this is os_* call

andrzej-kaczmarek (changes requested):

hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
kasjer: There is DivN block after system clock so for 96MHz PLL 32MHz is source for UART clock, same for no PPL and XTAL32M.
Maybe if system clock is set to LP_CLK, value should be different.
I can apply 32MHz limit to expression if you think UART is likely to be used with LP clock.

hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
kasjer: fixed

hw/drivers/uart/uart_da1469x/src/uart_da1469x.c
kasjer: fixed

andrzej-kaczmarek (approved):

PR #2439, opened 2020-12-15
Fixed MMC starting frequency

pengyun961012:
Apparently the lowest clock rate acceptable by hal_spi_init is 125KHz, so if 100KHz is used hal_spi_init will always fail.


kasjer:
> Apparently the lowest clock rate acceptable by hal_spi_init is 125KHz, so if 100KHz is used hal_spi_init will always fail.

@pengyun961012 hal_spi_init is MCU specific code. Are you sure all implementations have this limitation? (I have not checked it, so it's not a tricky question).
If not, what MCU/BSP are you referring to?

kasjer (approved):
@pengyun961012 I checked that Nordic MCUs has this limitation. All others seems to handle frequency in more relaxed way, meaning if frequency is not exactly as requested, closest one will be used instead.
Thanks for spotting this.

PR #2440, opened 2020-12-16
allow extra time for corefile to be created in NMI handler

supervillain101:

### DIFF
--- a/kernel/os/src/arch/cortex_m33/os_fault.c
+++ b/kernel/os/src/arch/cortex_m33/os_fault.c
@@ -24,6 +24,7 @@
 #include "console/console.h"
 #include "hal/hal_system.h"
 #include "os_priv.h"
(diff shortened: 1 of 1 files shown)


kasjer (commented):
@supervillain101 your change looks solid and I would not hesitate to approve it, but...
it would be more consistent with commit style in mynewt to remove 2 extra lines that got into commit and are not really relevant to change.

Additionally commit message title could be more precise like:
```
os/cortex_m4: tickle watchdog before core dump

allow extra time for corefile to be created in NMI handler
```

I suggest to remove those two extra empty lines.
Then ```git commit --amend``` (update commit message during this commit)
then
```git push -f <your_fork>```

supervillain101:
Done - thanks

kasjer (approved):
LGTM

Issue #2393, opened 2020-10-18
Low power with nRF52


danielkucera:
Hi,

I'm tuning the low power consumption with nRF52832. I was able to get to 170uA in System On mode with no modifications to mynewt-core.

But to get to 1.9uA I had to add following patch:
```
diff --git a/hw/mcu/nordic/nrf52xxx/src/hal_os_tick.c b/hw/mcu/nordic/nrf52xxx/src/hal_os_tick.c
index bca24c440..10bc94967 100644
--- a/hw/mcu/nordic/nrf52xxx/src/hal_os_tick.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_os_tick.c
@@ -154,6 +154,7 @@ os_tick_idle(os_time_t ticks)
         nrf52_os_tick_set_ocmp(ocmp);
     }
 
+    NRF_CLOCK->TASKS_HFCLKSTOP = 1;
     __DSB();
     __WFI();
 ```

AFAIK, HFCLK is managed by softdevice but since mynewt is not using it, it has to manage HFCLK by itself.

Would you consider adding this to the codebase or is there something I am missing?

**Comment by andrzej-kaczmarek** [comment]
there's `nrf52_clock_hfxo_request` function which is used to start HFXO and refcount its users, so it may be just that some code requested HFXO but did not release it. for sure you should not stop it as above.

**Comment by danielkucera** [comment]
In whole mynewt-core there is not a single call to `nrf52_clock_hfxo_request` or `nrf52_clock_hfxo_release`.
Am I missing something?

**Comment by danielkucera** [comment]
Okay, I've put 
```
NRF_CLOCK->TASKS_HFCLKSTOP = 1;
```
into my app `main()` and it has the same effect - power usage drops.
Is it possible that it's on by default or enabled during mcuboot? 
Should it be handled in `SystemInit()` ?

**Comment by andrzej-kaczmarek** [comment]
I recall two other places where HFXO can be started:
- on init, if LFCLK source is set to synth clock (this would have to be set explicitly)
- in hal_timer, when timer that uses HFCLK is configured

So in case mcuboot uses some timer, then it is possible that such timer started HFXO and it is not stopped. In such case I guess it would be reasonable to explicitly stop HFXO in `hal_system_clock_start` - so you can check if this helps and if so, patches are welcome :)

**Comment by danielkucera** [comment]
following seems to fix the issue with latest master
```
[...]

PR #2441, opened 2020-12-16
hw/mcu/nordic/nrf52xxx: Using hfxo_request and release. Addresses #2393.

ncasaril:
hal_timer and hal_system_clock_start both enabled the HFXO directly. 
Now they use the nrf52_clock_hfxo_request, and where suitable also nrf52_clock_hfxo_release.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_timer.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_timer.c
@@ -26,6 +26,7 @@
 #include "hal/hal_timer.h"
 #include "nrf.h"
 #include "mcu/nrf52_hal.h"
+#include "mcu/nrf52_clock.h"
 
 /* IRQ prototype */
 typedef void (*hal_timer_irq_handler_t)(void);
@@ -652,17 +653,7 @@ hal_timer_config(int timer_num, uint32_t freq_hz)
 
 #if MYNEWT_VAL_CHOICE(MCU_HFCLK_SOURCE, HFXO)
     /* Make sure HFXO is started */
-    if ((NRF_CLOCK->HFCLKSTAT &
-         (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) !=
-        (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) {
-        NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
-        NRF_CLOCK->TASKS_HFCLKSTART = 1;
-        while (1) {
-            if ((NRF_CLOCK->EVENTS_HFCLKSTARTED) != 0) {
-                break;
-            }
-        }
-    }
+    nrf52_clock_hfxo_request();
 #endif
     hwtimer = bsptimer->tmr_reg;
 
@@ -723,6 +714,12 @@ hal_timer_deinit(int timer_num)
     }
     bsptimer->tmr_enabled = 0;
     bsptimer->tmr_reg = NULL;
+
+#if MYNEWT_VAL_CHOICE(MCU_HFCLK_SOURCE, HFXO)
+    if (timer_num != 5) {
+        nrf52_clock_hfxo_release();
+    }
+#endif
     __HAL_ENABLE_INTERRUPTS(ctx);
 
 err:
--- a/hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
+++ b/hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
@@ -44,7 +44,16 @@ nrf52_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf52_clock_hfxo_refcnt < 0xff);
     if (nrf52_clock_hfxo_refcnt == 0) {
-        NRF_CLOCK->TASKS_HFCLKSTART = 1;
+        /* Check the current STATE and SRC of HFCLK */
+        if ((NRF_CLOCK->HFCLKSTAT &
+             (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) !=
+            (CLOCK_HFCLKSTAT_SRC_Xtal << CLOCK_HFCLKSTAT_SRC_Pos |
+             CLOCK_HFCLKSTAT_STATE_Running << CLOCK_HFCLKSTAT_STATE_Pos)) {
+            NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
+            NRF_CLOCK->TASKS_HFCLKSTART = 1;
+            while (!NRF_CLOCK->EVENTS_HFCLKSTARTED) {
+            }
+        }
         started = 1;
     }
     ++nrf52_clock_hfxo_refcnt;
--- a/hw/mcu/nordic/nrf52xxx/src/hal_system.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_system.c
@@ -22,6 +22,7 @@
 #include "hal/hal_system.h"
 #include "hal/hal_debug.h"
 #include "nrf.h"
+#include "mcu/nrf52_clock.h"
 
 /**
  * Function called at startup. Called after BSS and .data initialized but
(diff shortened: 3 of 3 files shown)


hw/mcu/nordic/nrf52xxx/src/hal_timer.c
```diff
@@ -652,18 +653,7 @@ hal_timer_config(int timer_num, uint32_t freq_hz)
...
-        (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) {
-        NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
-        NRF_CLOCK->TASKS_HFCLKSTART = 1;
-        while (1) {
-            if ((NRF_CLOCK->EVENTS_HFCLKSTARTED) != 0) {
-                break;
-            }
-        }
-    }
-#endif
```
kasjer, line 666: **#endif** was lost in the process

ncasaril: Thanks @kasjer. 

In the hal_timer_deinit I've added the same ifdef around the release and also checking that it's not timer 5 to match hal_timer_config. 

In the deinit, I realise that the if-statement checks for tmr_rtc which is only enabled for timer_num=5 and so the release could go into the else section there. However using the timer_num as in hal_timer_config may be easier to understand?

kasjer:
@ncasaril this patch slightly changes how code used to work.
Before it waited for XTAL to settle now it just starts it without waiting.
I don't know the implication of this. I guess that when XTAL will start settle time is short enough and it does not matter
when change form HFINT to HFXO will take place.
On the other hand if XTAL fails to start (I'm not sure what are the odds for this provided that hardware is correct) code will ignore this and will continue to run on HFINT, while before the change, it would stuck in loop and probably watchdog would cause the reboot.
That's why I added @wes3 as the original author who may have some insight.

For me it looks OK

wes3:
@ncasaril and @kasjer I was looking over the changes and I was thinking the same thing regarding settling time. The code to request the hfxo just turns it on and leaves without waiting for it to settle as you say and I was thinking of the implications of that. I think I modeled this code after some code I saw in an SDK from nordic. The idea was to check to see if the clock was started; if so do nothing. If not, clear the started event, turn on the clock, and wait for it to settle. I figured it would be pretty catastrophic if we never see EVENTS_HFCLKSTARTED so that is why it just loops forever and lets the watchdog fire. Yes, that assumes there is a watchdog running :-) There is an errata for the chip that states you could see that event erroneously if the clock takes longer than 400 usecs to settle btw so that would have to be handled differently.

So I guess we just need to decide if we care about using a timer or the radio before the clock settles. I would suspect most settling times for most boards to be in the hundreds of usecs so probably not an issue but who knows? I guess if we were being very cautious and "more correct" (bad term) we would move the code that waits for settling into hfxo_request if the clock is not on. So some thoughts on what we could do if we wanted:
1) Make sure on startup that the HFXO is off. This is to deal with any bootloaders that may have turned it on.
2) In HFXO request, wait to settle if the reference count is zero.
3) Related to 2, if all things are working properly, the HFXO should be off if the reference count is zero. Not sure if we care about that case and want to add an assert there for debug (a debug assert that can get compiled out) but this is not critical.

If both of you feel that there really is no need to wait for it to settle I am fine with the code the way it is.

ncasaril:
@wes3 and @kasjer Thanks for your input. 
I've tried to accomodate your suggestions with moving the wait code into hfxo_request. 
It feels like a good solution and that's now the only place that I can find that enables the HFCLK. 

That said, I think the NRF_CLOCK->HFCLKSTAT isn't used correctly to detect if the HFXO is used. It only checks if the task has been started, not if we're actually running off the HFXO. 

Leave me with it for a bit and I'll see if I can get clarity into how it's meant to work. 

So, not ready for merge just yet.

ncasaril:
@wes3  In the hfxo_request I'm waiting for confirmation that the HFCLK has started. I removed the check on the HFCLKSTAT as it was only checking if the task had been triggered (bit 16) and not if the HFXO was used (bit 0) thus not actually starting the HFXO. 

If you're happy with how it looks now I'll clean up the commits.

wes3 (commented):
This all seems good to me but I am not in a position currently to test this on a board.  BTW: I never really checked to see if checking only bit 16 in HFCLKSTAT register is sufficient to see if the HFXO is currently running. I do not know why the code got removed to check that the source was the HFXO (bit 0 being 1) but not a big deal if you tested this and it works.

ncasaril:
Sorry guys for all the back and forth on this one and thanks for the input so far.

I'm now checking both the STATE and the SRC of the HFCLK before starting it in hfxo_request. 

Note that I've not covered the case of using LFCLK_SYNT + HFINT as that combination is restricted by syscfg. 
If you feel this is needed I can add the ifdefs needed for this is hal_system_clock_start.

@andrzej-kaczmarek  I think we need your input to see that we're not breaking anything here.

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
```diff
@@ -44,7 +44,18 @@ nrf52_clock_hfxo_request(void)
     __HAL_DISABLE_INTERRUPTS(ctx);
     assert(nrf52_clock_hfxo_refcnt < 0xff);
     if (nrf52_clock_hfxo_refcnt == 0) {
-        NRF_CLOCK->TASKS_HFCLKSTART = 1;
+        /* Check the current STATE and SRC of HFCLK */
+        if ((NRF_CLOCK->HFCLKSTAT &
+             (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) !=
+            (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) {
```
danielkucera, line 50: Just out of  curiosity... What is this condition supposed to test?

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
```diff
@@ -44,7 +44,18 @@ nrf52_clock_hfxo_request(void)
...
+            NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
+            NRF_CLOCK->TASKS_HFCLKSTART = 1;
+            while (1) {
+                if ((NRF_CLOCK->EVENTS_HFCLKSTARTED) != 0) {
+                    break;
+                }
```
danielkucera, line 56: what about using this here?
```
while (!NRF_CLOCK->EVENTS_HFCLKSTARTED) {
}
```

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
wes3: I was speaking to @ncasaril about this. The specification shows two bits in that register. SRC is bit 0 and is 0 for HFINT and 1 for HFXO. The STATE bit is bit 16 and is 0 if HFCLK not running and 1 if running. So it seemed best to check both bits being 1. It might be possible to just check SRC and if 1 all is good. Looking at one of the versions of the SDK from nordic it does appear the check both bits.

danielkucera: Okay, then this is not very obvious... I would suggest comparing to `CLOCK_HFCLKSTAT_SRC_Xtal` and `CLOCK_HFCLKSTAT_STATE_Running` defines, something like:
```
        if ((NRF_CLOCK->HFCLKSTAT &
             (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) !=
            (CLOCK_HFCLKSTAT_SRC_Xtal << CLOCK_HFCLKSTAT_SRC_Pos | 
             CLOCK_HFCLKSTAT_STATE_Running << CLOCK_HFCLKSTAT_STATE_Pos)) {
```

ncasaril: I agree that this is clearer. The reason we used the other if-statment was that this was the original statement used and I wanted to keep the changes to a minimum. I'll change so it's clearer.

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
ncasaril: Again, same reason - minimum changes left this one in there. Very happy to simplify this.

kasjer (approved):
LGTM, if no one has anything to add soon I will merge it very soon.

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
```diff
@@ -44,7 +44,16 @@ nrf52_clock_hfxo_request(void)
...
     if (nrf52_clock_hfxo_refcnt == 0) {
-        NRF_CLOCK->TASKS_HFCLKSTART = 1;
+        /* Check the current STATE and SRC of HFCLK */
+        if ((NRF_CLOCK->HFCLKSTAT &
+             (CLOCK_HFCLKSTAT_SRC_Msk | CLOCK_HFCLKSTAT_STATE_Msk)) !=
+            (CLOCK_HFCLKSTAT_SRC_Xtal << CLOCK_HFCLKSTAT_SRC_Pos |
+             CLOCK_HFCLKSTAT_STATE_Running << CLOCK_HFCLKSTAT_STATE_Pos)) {
+            NRF_CLOCK->EVENTS_HFCLKSTARTED = 0;
+            NRF_CLOCK->TASKS_HFCLKSTART = 1;
+            while ((NRF_CLOCK->EVENTS_HFCLKSTARTED) != 0) {
```
danielkucera, line 54: this logic is reversed - `NRF_CLOCK->EVENTS_HFCLKSTARTED` has to be non-zero to escape the loop

kasjer: thanks @danielkucera

kasjer (changes requested):

hw/mcu/nordic/nrf52xxx/src/nrf52_clock.c
ncasaril: Dang, too quick. Thanks guys.

kasjer (approved):
LGTM

PR #2442, opened 2020-12-16
net/lora/node: RxDone event and rx_win1 timeout

ncasaril:
LoRaNode thread can access RxDone data before it's filled in if the submitting thread has a lower priority, typically not a good idea. 

Second commit addresses the rx_win1 and rx_win2 issues where we can be stuck waiting for a timeout on rx_win1 because
rx_win2 has already started (and prevented a timeout of rx_win1).


kasjer (approved):
LGTM

PR #2445, opened 2020-12-18
sensor creator init split

kasjer:
This solves the problem of sensor_creator usage along with SPI flash that is present in flash map.

If BSP uses SPI flash and sensor creator with SPI devices that are connected to the same
SPI bus as flash, not initializing chip select pin for those other devices may lead to wrong
reads from flash.
It happens when uninitialized yet GPIO is still perceived by sensor as active and sensor responds
to requests that should be answered by flash.

This adds sensor_dev_init() that has initialization code taken from sensor_dev_create().
Thai also changes flash_map initialization stage to accommodate for other initialization that
could be needed before flash_map can be initialized.


vrahane (approved):
Looks good. That was a really good catch. The config for the sensors should have taken care of this situation as it would configure the spi accordingly but this seems like a cleaner solution.

PR #2452, opened 2021-01-10
Fix wrong desc_hid_report

beeender:
HID_REPORT_ID is needed when using TUD_HID_REPORT_DESC_KEYBOARD. By
passing `(int_id, )` to the macro, the hid device can be recognized
under Linux with an error message "unknown main item tag 0x0". The
normal keys works fine surprisingly in this case but not the modifiers.
Defining the report in this way would cause a Windows device failure
with code 10.

See "HID Report Descriptor Template" comments in `hid_device.h` for more
information.


kasjer (approved):
LGTM

kasjer:
Thanks @beeender

PR #2453, opened 2021-01-11
Add function to check usb console init

beeender:
Add usb_cdc_console_is_init() to return the usb cdc console
initialization status. Without this, usb console log won't work.


kasjer (approved):
Great

PR #2457, opened 2021-01-20
[RFC] coap: negation in tcp check in coap_receive()

KKopyscinski:
When using tcp protocol responses are not handled. Removed negation from
protocol check in handling responses.

### DIFF
--- a/net/oic/src/messaging/coap/engine.c
+++ b/net/oic/src/messaging/coap/engine.c
@@ -244,6 +244,10 @@ coap_receive(struct os_mbuf **mp)
                 coap_remove_observer_by_mid(OC_MBUF_ENDPOINT(m), message->mid);
 #endif
             }
+        } else {
+            /* CoAP over TCP doesn't support Confirmable, Non-confirmable,
+               Acknowledgment and Reset messages are simply not supported.  */
+            erbium_status_code = SERVICE_UNAVAILABLE_5_03;
         }
 
         /* Open transaction now cleared for ACK since mid matches */


KKopyscinski:
When using application bleprph_oic I noticed that coap messages with response codes ( method code = 225, 226 and 228) were not handled. When using build from Nimble 1.8 this problem hasn't occured. I see that the fragment of code impacting handling of these messages was altered in commit 9064dbafc0fab24833c1bae77c52755e02108ea6. Previously there was no check here; although there was `Fix this` in comment. I don't fully understand what the issue was - the responses should be handled only for tcp, all protocols, or all beside tcp?

mkiiskila:
See https://tools.ietf.org/html/rfc8323#section-3. Coap over TCP does not do confirmed messages, acknowledgments etc.
I'll have to say no to this PR, I suspect the issue you're after is somewhere else.

KKopyscinski:
@mkiiskila Thank you for your response, it was very helpful! If I understand correctly, message types matter only for non-reliable transports such as UDP. As a result, only message codes between 1 through 4 should be handled properly (COAP_GET - COAP_DELETE). For TCP message types don't matter. What I can see is that for TCP message codes outside 1-4 range are not handled at all. Shouldn't they met with SERVICE_UNAVAILABLE_5_03 response? I pushed the code that redirects this case here: https://github.com/apache/mynewt-core/blob/433b35e380296d375d5f250f11ad92f90c3e2057/net/oic/src/messaging/coap/engine.c#L295
without it handling ends up here:
https://github.com/apache/mynewt-core/blob/433b35e380296d375d5f250f11ad92f90c3e2057/net/oic/src/messaging/coap/engine.c#L273
but there is no transaction to be sent.

KKopyscinski:
Sample logs illustrating the issue:
- Without patch
`003019 [ts=23585900us, mod=4 level=0 ]ble_hs_hci_evt_acl_process(): conn_handle=1 pb=2 len=17 data=0x0d 0x00 0x04 0x00 0x12 0x0c 0x00 0x71 0xe1 0x01 0xff 0x73 0x61 0x6d 0x70 0x6c 0x65 
003019 [ts=23585900us, mod=7 level=0 ]oc_gatt rx seg 1-200014bc-10
003019 [ts=23585900us, mod=4 level=0 ]host tx hci data; handle=1 length=5
003019 [ts=23585900us, mod=4 level=0 ]ble_hs_hci_acl_tx(): 0x01 0x00 0x05 0x00 0x01 0x00 0x04 0x00 0x13 
003019 [ts=23585900us, mod=7 level=0 ]oc_buffer_rx: CoAP: received datalen=10
003020 [ts=23593712us, mod=7 level=0 ]Token (len 1)   Parsed: CoAP version: 0, token: 0x0100, mid: 0
003020 [ts=23593712us, mod=7 level=0 ]  type: CON`
- With patch (send service not supported)
`001646 [ts=12859320us, mod=4 level=0 ]ble_hs_hci_evt_acl_process(): conn_handle=1 pb=2 len=17 data=0x0d 0x00 0x04 0x00 0x12 0x0c 0x00 0x71 0xe1 0x01 0xff 0x73 0x61 0x6d 0x70 0x6c 0x65 
001646 [ts=12859320us, mod=7 level=0 ]oc_gatt rx seg 1-200014bc-10
001646 [ts=12859320us, mod=4 level=0 ]host tx hci data; handle=1 length=5
001646 [ts=12859320us, mod=4 level=0 ]ble_hs_hci_acl_tx(): 0x01 0x00 0x05 0x00 0x01 0x00 0x04 0x00 0x13 
001646 [ts=12859320us, mod=7 level=0 ]oc_buffer_rx: CoAP: received datalen=10
001646 [ts=12859320us, mod=7 level=0 ]Token (len 1)   Parsed: CoAP version: 0, token: 0x0100, mid: 0
001647 [ts=12867132us, mod=7 level=0 ]  type: CON
001647 [ts=12867132us, mod=7 level=0 ]coap_tx: 0x20001544
001647 [ts=12867132us, mod=7 level=0 ]coap_tx: serialized 2 B (header len 2, payload len 0)
001647 [ts=12867132us, mod=7 level=1 ]coap_send_message(): (2) 
001647 [ts=12867132us, mod=7 level=0 ]oc_buffer_tx: GATT procedure initiated: notify; att_handle=14
001647 [ts=12867132us, mod=4 level=0 ]host tx hci data; handle=1 length=9
001647 [ts=12867132us, mod=4 level=0 ]ble_hs_hci_acl_tx(): 0x01 0x00 0x09 0x00 0x05 0x00 0x04 0x00 0x1b 0x0e 0x00 0x00 0xa3 `

KKopyscinski:
Closed because the issue was not in the code. We have CoAP ported from ContikiOS, which follows RFC 7959 and received messages came from system following RFC 8323. Between these versions were introduced new message type - CSM, which is sent at the beginning of every connection. If device following newer spec doesn't receive response for CSM message (and it doesn't, because we do not handle these), it disconnects.

PR #2460, opened 2021-01-25
stm32 spi related issues

kasjer:
- black_vet6 had wrong sector size in syscfg value 4094 was obvious mistake
- nucleo-f103 was configured to use internal HSI clock for PLL source resulting in sub-optimal system clock
- STMF0xx had incorrect alternate function setting for SPI2 (5 instead of 0)


utzig (approved):
LGTM

PR #2463, opened 2021-02-02
hw/mcu/da1469x: Fix HAL UART creation

kasjer:
Introduction of dedicated UART driver broke build when user
decided to use HAL driver instead.
When HAL_UART was set in syscfg.yml da1469x_periph.c would not
build due to conditional compilation inconsistency.

This also changes HAL_UART syscfg to UART_HAL which is more
consistent with driver name.

### DIFF
--- a/hw/mcu/dialog/da1469x/pkg.yml
+++ b/hw/mcu/dialog/da1469x/pkg.yml
@@ -39,10 +39,10 @@ pkg.deps.TRNG:
 pkg.deps.CRYPTO:
     - "@apache-mynewt-core/hw/drivers/crypto/crypto_da1469x"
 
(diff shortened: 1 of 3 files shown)


benmccrea (approved):

benmccrea:
@kasjer LGTM. I tested with `HAL_UART: 1` and all works.

kasjer:
> @kasjer LGTM. I tested with `HAL_UART: 1` and all works. There is a CI failure (style?). I approved but I'll wait for someone else to review the CI failure before merging.

@benmccrea Right now CI fails due to missing linker script **stm32l073.ld** which has nothing to do with this PR. This happened while CI was not really working for a while. Pending #2468 addresses this problem.
So I think we can merge this one.
Tanks for testing it out.

utzig (approved):
LGTM

PR #2470, opened 2021-02-04
[KINETIS] Turn pyocd into a reusable script

utzig:
Move custom pyocd flashing/debugging scripts to a common hw/script that can be used by other KINETIS BSPs and update it to recent pyocd APIs.


mlaz (approved):
LGTM!

PR #2472, opened 2021-02-05
MTB in core dump

kasjer:
- MTB registers are stored in RAM when crash happens so MTB can be used check what lead to crash
- MTB is also stopped in __assert_function() not only in default interrupt handler
- disassemble of non existing memory region does not stop script for continuing
- instruction preceding exception are displayed

### DIFF
--- a/hw/scripts/micro-trace-buffer.py
+++ b/hw/scripts/micro-trace-buffer.py
@@ -37,6 +37,12 @@ class MicroTraceBuffer(gdb.Command):
 
     def __init__(self):
         super(MicroTraceBuffer, self).__init__("mtb", gdb.COMMAND_STATUS, gdb.COMPLETE_NONE)
+        try:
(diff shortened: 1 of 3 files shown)


hw/scripts/micro-trace-buffer.py
```diff
@@ -37,6 +37,12 @@ class MicroTraceBuffer(gdb.Command):
 
     def __init__(self):
         super(MicroTraceBuffer, self).__init__("mtb", gdb.COMMAND_STATUS, gdb.COMPLETE_NONE)
+        try:
+            mtb_state_at_crash = gdb.lookup_symbol("mtb_state_at_crash")[0]
```
kasjer, line 41: this can also detect mtb_state_at_crash variable on live system and it this case real addresses should be used

kasjer: fixed by checking content of mtb_state_at_crash not only variable existence

andrzej-kaczmarek (approved):

PR #2477, opened 2021-02-11
Enable Mbed TLS self-tests which were not running

utzig:
Enable extra self-tests from Mbed TLS. Added ARIA, CMAC, MD2, MD4 and J-PAKE tests. The test order was sorted for easier reading.


mlaz (approved):
LGTM

PR #2478, opened 2021-02-11
Add more Mbed TLS alt implementations support

utzig:
* crypto: mbedtls: allow EC customizations - Add options to implement EC point operations, ECDH and ECDSA with external libraries. This can be used to accelerate math in custom HW based implementations.

* crypto: mbedtls: allow MPI size customization - This syscfg can be used to lower the amount of RAM used by MPIs, when it is known that less than 1024 bytes are required, eg, EC operations (or small RSA).

* crypto: mbedtls: allow alt bignum implementation - Add options that allows overriding the bignum (aka MPI) implementation so that HW accelerators for math in modular fields can be used as alternative to SW based MPI calculation. PS: This is not a standard Mbed TLS option, so care must be taken when upgrading.

* crypto: mbedtls: default secp256r1 to enabled - secp224r1, which is not a commonly used curve, was enabled by default so
disable it and instead enable secp256r1 (aka EC256), which is used by MCUboot for example.


mlaz (approved):
LGTM

nkaje (approved):
LGTM

PR #2479, opened 2021-02-15
hw/mcu/dialog: Retain CM_ERROR_DIS_REG

andrzej-kaczmarek:
Latest CMAC PHY uses some hackery with CM_ERROR_DIS_REG to handle
tx_late errors so we need to retain that register.


benmccrea:
LGTM

benmccrea (approved):

PR #2480, opened 2021-02-15
Kinetis hash driver updates

utzig:
Rename the k64f hash driver to kinetis to avoid making it look like it is specific to k64f, because some kinetis families happen to have other HW units that can also do the HASH acceleration, eg `LTC` (instead of `mmCAU`). But no extra support is being added here. Also enable the driver to run on the frdm-k82f.


kasjer (approved):
LGTM

PR #2482, opened 2021-02-16
Updates to k64f aka kinetis trng driver

utzig:
* trng: k64f: rename to trng_kinetis - This makes it less specific, since the driver should also work on other Kinetis families.

* trng: kinetis: update driver for newer families - Older Kinetis families used the RNGA HW module for random number generation, newer models have a slightly different HW module called TRNG; this commit adds TRNG API usage compatibility, and adds syscfg to allow BSPs to choose from one of the HW modules.

* hw: bsp: frdm-k82f: enable TRNG usage

* trng: kinetis: remove delay in poller - This delay was causing a race condition on trng_test during console_printf, and can be safely removed.

* trng: kinetis: seed RNGA with SIM->UIDL - This follows the recommendation from NXP SDK.


hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
```diff
@@ -18,7 +18,23 @@
...
 #include <string.h>
+#include "os/mynewt.h"
+
+#if MYNEWT_VAL(KINETIS_TRNG_USE_RNGA)
+#define USE_RNGA 1
 #include "fsl_rnga.h"
+#elif MYNEWT_VAL(KINETIS_TRNG_USE_TRNG)
+#define USE_TRNG 1
+#include "fsl_trng.h"
+#define TRNG_START(base) \
```
kasjer, line 29: maybe use do {} while (0) to prevent accidents with if (cond) TRNG_START();
though I know mynewt style calls for {} in each if

hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
```diff
@@ -154,9 +186,16 @@ kinetis_trng_dev_open(struct os_dev *dev, uint32_t wait, void *arg)
...
 
+#if USE_RNGA
         RNGA_Init(RNG);
-
-        kinetis_rnga_start();
+#elif USE_TRNG
+        (void)TRNG_GetDefaultConfig(&default_config);
+        TRNG_Init(TRNG0, &default_config);
+#else
+        assert(0);
```
kasjer, line 195: this case was take care of in line 35
```c
#error "Unsupported TRNG interface"
```

hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
utzig: Done.

hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
utzig: I removed it.

utzig:
Fix for CI was submitted here: https://github.com/apache/mynewt-mcumgr/pull/113

kasjer (approved):
LGTM

PR #2483, opened 2021-02-17
Updates to k64f aka kinetis crypto driver

utzig:
* crypto: k64f: rename to kinetis - Make it clear this driver should work on other Kinetis family members as well; the k64f uses a mmCAU which is one of the possible, and oldest, of the acceleration units in those parts.
* kinetis: crypto: add initial LTC support
* bsp: frdm-k82f: enable CRYPTO acceleration


kasjer (approved):
LGTM

utzig:
Rebased because of conflicts with #2482, no changes were made...

PR #2484, opened 2021-02-17
Update crypto_test for AES-256

utzig:
* apps: crypto_test: add AES-256 vectors - Add AES-256 vectors from NIST and expand inplace and iovect to test AES-256.
* apps: crypto_test: fix warnings from syscfg - Enabling/disabling some of the crypto_test syscfgs resulted in warnings from non-used variables, etc. This commit fixes the issues so each individual syscfg can be isolatedly used.

Tested on frdm-k82f (fully functional), nucleo-f439 (fully functional), frdm-k64f (no AES-256 support, not sure if it's a driver or HW issue), nrf52840/pca10056 (no AES-256? no decryption!)

### DIFF
--- a/apps/crypto_test/src/main.c
+++ b/apps/crypto_test/src/main.c
@@ -46,6 +46,12 @@ struct test_vectors {
 
 static struct os_mutex mtx;
 
+#if (MYNEWT_VAL(CRYPTOTEST_VECTORS_ECB) || \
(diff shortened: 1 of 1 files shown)


kasjer (approved):
LGTM

mlaz (approved):
LGTM

PR #2487, opened 2021-02-18
[KINETIS] Multiple fixes: boot, systick and debug.

mlaz:
- Kinetis hal_os_tick now ran by SysTick instead of PIT.
- frdm-k82f clock_config files updated.
- frdm-k82f FOPT now boots straight from flash and doesn't wait for NMI.
- frdm-k64f and frdm-k64f BSP debug scripts updated so Core and peripheral registers can be read.


hw/bsp/frdm-k82f/src/arch/cortex_m4/startup_MK82F25615.S
```diff
@@ -272,7 +272,7 @@ __isr_vector:
     .long 0xFFFFFFFF
     .long 0xFFFFFFFF
     .long 0xFFFFFFFF
-    .long 0xFFFFFFFE
+    .long 0xFFFF3DFE
```
utzig, line 275: If you are setting `BOOTSRC_SEL` to internal flash shouldn't `BOOTPIN_OPT` be 1 to select `BOOTSRC_SEL`, in other words `0xFFFF3FFE`?

utzig (changes requested):
Please fix hw/mcu/nxp/kinetis/src/hal_os_tick.c as suggested. Left one doubt, but it is running pretty well here, thanks!

hw/bsp/frdm-k82f/src/arch/cortex_m4/startup_MK82F25615.S
mlaz: Yes, you are right, I misread the doc. That is what I intended to do.

mlaz: Fixed.

mlaz:
All fixed.

utzig (approved):
LGTM

PR #2491, opened 2021-02-18
Fix Kinetis HASH driver with small length buffers

utzig:
* apps: hash_test: add small varlen inputs test - Add a new test that hashes multiple small variable length strings into a digest, where each string is smaller than SHA-256 block length.
* hw: drivers: hash_kinetis: fix small buffer hashing - Hashing a message with multiple blocks of length<=64 (SHA-256 block length) was broken due to proper usage of the mmCAU padding. This commit fixes the handling of the pad data.

### DIFF
--- a/apps/hash_test/src/main.c
+++ b/apps/hash_test/src/main.c
@@ -254,6 +254,71 @@ run_stream_test(struct hash_dev *hash)
     }
 }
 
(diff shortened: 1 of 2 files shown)


mlaz (approved):
LGTM

PR #2494, opened 2021-02-20
Fix STM32 HASH stream support

utzig:
apps: hash_test: add single byte multi-write test - Expand string stream test to also run on SHA-224 (when available). Add new single byte multi-write test, where a single "a" is written 257 (because odd numbers are good!) times, running on both SHA-224 and SHA-256 (where available).

hw: drivers: hash_stm32: fix stream support - The STM32 HASH HW is configured to work on a word (32-bit) write, and the only write that can be less than 4-bytes is the last write (_finish()). To be compatible with streams that are non word-boundary sized, a new pad (aka state) was added that caches writes that are less than 4 byte, so only words are ever written in an _update(), and _finish() flushes the remaining bytes that were less than 4-bytes.

### DIFF
--- a/apps/hash_test/src/main.c
+++ b/apps/hash_test/src/main.c
@@ -210,19 +210,21 @@ run_stream_test(struct hash_dev *hash)
     uint8_t outbuf[HASH_MAX_DIGEST_LEN];
     struct hash_generic_context ctx;
     struct stream_data *stream;
+    uint16_t algo;
     int i, d;
     int rc;
 
-    for (d = 0; d < sizeof(streams)/sizeof(streams[0]); d++) {
+    for (d = 0; d < ARRAY_SIZE(streams); d++) {
         stream = &streams[d];
         printf("%s: ", stream->name);
+        algo = stream->algo;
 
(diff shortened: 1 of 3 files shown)


apps/hash_test/src/main.c
```diff
@@ -275,50 +299,133 @@ run_varlength_sha256_test(struct hash_dev *hash)
...
+    struct hash_generic_context ctx;
+    int i, j;
+    int rc;
+    uint8_t outbuf[HASH_MAX_DIGEST_LEN];
+    struct stream_data *bytewrite;
+    uint16_t algo;
+
+    char byte = 'a';
+
+    for (j = 0; j < sizeof bytewrites /sizeof bytewrites[0]; j++) {
```
kasjer, line 384: missing space after /, which should be detected by style check
os/kernel/util.h has ARRAY_SIZE which could be used

utzig: Fixed.

kasjer (approved):
Not tested, but looks good to me

PR #2495, opened 2021-02-22
hw: bsp: frdm-k82f: enable HW based CBC/CTR

utzig:
Avoid linking in CBC/CTR in SW, because this device can do it in HW.


mlaz (approved):
Looks good

PR #2496, opened 2021-02-22
sys/console: Add syscfg to allow restoring echo state after NLIP commands

benmccrea:
Currently when an NLIP command is finished processing, console echo is always turned back on.

This PR adds an optional syscfg, `CONSOLE_NLIP_RESTORE_ECHO`.

`CONSOLE_NLIP_RESTORE_ECHO: 0` (default setting) retains the existing console behavior, in which completion of an NLIP command will turn on console echo.

`CONSOLE_NLIP_RESTORE_ECHO: 1` will cause the echo on/off state to be restored to the state it was in before an NLIP command was processed.


kasjer (approved):

vrahane (approved):
Looks fine to me.

PR #2498, opened 2021-02-23
hw/mcu: Fix nRF5340 netcore build with synthesize LF clock

sjanc:
Network core doesn't have secure peripherals.


vrahane (approved):
LGTM

PR #2500, opened 2021-02-24
hw/sensor: fix possible null dereference in sensor_set_thresh()

KKopyscinski:
Added assert to check if stt is not NULL before dereferencing it

### DIFF
--- a/hw/sensor/src/sensor.c
+++ b/hw/sensor/src/sensor.c
@@ -1970,6 +1970,11 @@ sensor_set_thresh(const char *devname, struct sensor_type_traits *stt)
     struct sensor *sensor;
     int rc;
 
+    if (stt == NULL) {
+        rc = SYS_EINVAL;
+        goto err;
+    }
+
     sensor = sensor_get_type_traits_byname(devname, &stt_tmp,
                                            stt->stt_sensor_type);
     if (!sensor) {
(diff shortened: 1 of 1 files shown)


hw/sensor/src/sensor.c
```diff
@@ -1970,14 +1970,15 @@ sensor_set_thresh(const char *devname, struct sensor_type_traits *stt)
     struct sensor *sensor;
     int rc;
 
+    assert(stt != NULL);
```
andrzej-kaczmarek, line 1973: you should better return error here

KKopyscinski: done

hw/sensor/src/sensor.c
```diff
@@ -1970,14 +1970,19 @@ sensor_set_thresh(const char *devname, struct sensor_type_traits *stt)
     struct sensor *sensor;
     int rc;
 
+    if (stt != NULL) {
```
andrzej-kaczmarek, line 1973: so now I am confused... we want to avoid null dereference or force it? :D

also update commit message

KKopyscinski: avoid null, of course... :)
updated

andrzej-kaczmarek (approved):

PR #2501, opened 2021-02-24
hw: handle hal_spi_enable/disable errors

KKopyscinski:
hal_spi_enable/disable errors were handled, but not everywhere. This
unifies handling manner.

### DIFF
--- a/hw/drivers/flash/at45db/src/at45db.c
+++ b/hw/drivers/flash/at45db/src/at45db.c
@@ -446,7 +446,6 @@ at45db_init(const struct hal_flash *hal_flash_dev)
     }
 
     hal_spi_set_txrx_cb(dev->spi_num, NULL, NULL);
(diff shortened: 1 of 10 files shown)


hw/drivers/flash/at45db/src/at45db.c
```diff
@@ -446,7 +446,10 @@ at45db_init(const struct hal_flash *hal_flash_dev)
     }
 
     hal_spi_set_txrx_cb(dev->spi_num, NULL, NULL);
-    hal_spi_enable(dev->spi_num);
+    rc = hal_spi_enable(dev->spi_num);
```
kasjer, line 449: i would just return rc all the time since there is nothing to do after if (rc) statement

KKopyscinski: ok, done

kasjer (approved):
LGTM

PR #2502, opened 2021-02-24
bsp/nordic_pca10095_net: Set empty IPC region for bootloader build

sjanc:
IPC is not used in bootloader but linker were complaining about IPC
region being not present.


kasjer (approved):

vrahane (approved):
Looks good.

PR #2503, opened 2021-02-25
sensor_creator: initialize sensor configs with zeros

KKopyscinski:
Some default sensor configurations have uninitialized member fields,
which later were used as value to set in [sensor_name]_config().
Now all values are 0 by default, and non-zero values are set.


kasjer (approved):
LGTM

PR #2505, opened 2021-02-25
mmc: Decrease mmc/sd initialization time

kasjer:
SD specification section "6.4.1 Power Up"
requires 74 clocks with high CS.
Code was sending 592 clocks (it was sending 74 bytes).

Now code sends 10 bytes (80 cycles).


utzig (approved):
LGTM

PR #2508, opened 2021-02-27
net/osdp: Use internal data structures and remove dynamic memory allocations

vikrant-proxy:
- Cleanup commented code
- Make secure mode on by default


vrahane:
I have already reviewed this quite in detail, can others please take a look. With this we have tried to do the following:
- Reduce duplication of code
- Removed its inherent data structures 
- Used the mynewt BSD data structures
- Removed dynamic memory allocation here and used mempools instead
- Remove the ringbuffer and use the internal one since there was too much overhead involved

vrahane (approved):
Looks good to me.

PR #2511, opened 2021-03-01
Replace strtok with strtok_r

### LINKED ISSUE #2510: Unsafe function strtok used in several places
Function is not thread safe and there is safe replacement strtok_r.

mkasenberg:
strtok is not thread safe.

Fixes #2510

### DIFF
--- a/hw/drivers/sensors/bno055/src/bno055_shell.c
+++ b/hw/drivers/sensors/bno055/src/bno055_shell.c
@@ -115,6 +115,7 @@ bno055_shell_cmd_sensor_offsets(int argc, char **argv)
     uint16_t val;
     uint16_t offsetdata[11] = {0};
     char *tok;
+    char *tok_ptr;
 
     rc = 0;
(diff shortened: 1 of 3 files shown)


hw/drivers/sensors/bno055/src/bno055_shell.c
```diff
@@ -141,15 +142,15 @@ bno055_shell_cmd_sensor_offsets(int argc, char **argv)
...
+        tok = strtok_r(argv[2], ":", &tok_ptr);
         i = 0;
         do {
             val = parse_ll_bounds(tok, 0, UINT16_MAX, &rc);
             if (rc) {
                 return bno055_shell_err_invalid_arg(argv[2]);
             }
             offsetdata[i] = val;
-            tok = strtok(0, ":");
+            tok = strtok_r(0, ":", &tok_ptr);
```
sjanc, line 153: make it NULL instead of 0

sjanc (approved):

PR #2512, opened 2021-03-01
Fixes to Kinetis/K8x port

utzig:
hw: mcu: kinetis: remove custom systick - Remove custom systick implementation and allow Kinetis MCUs to use the existing `SysTick_Handler` implementation already available for Cortex-M4 MCUs.

hw: mcu: kinetis: k8x: fix flash write alignment - Flash in Kinetis devices has either 4 or 8 byte write alignment, depending on the specific model. For K8x this is 4 bytes so fix the current syscfg.

hw: mcu: kinetis: uninterrupted flash programming - Disable interrupts while executing the flash commands; this avoids hardfaults due to interrupts executing code when FCCOBx registers are being written. This is only protecting `FTFx_CMD_Program` which is the function that is called from the Kinetis HAL's flash_write; other flash programming routines might also need similar protection. I've not seen any issues with erase operations.

### DIFF
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -299,6 +299,7 @@ MK82F25615_features.h
 system_MK82F25615.h
 system_MK82F25615.c
 fsl_clock.h
+fsl_ftfx_controller.c
 
 # LWIP - BSD License.
 lwip_base
--- a/hw/mcu/nxp/kinetis/MK8xF/src/fsl_ftfx_controller.c
+++ b/hw/mcu/nxp/kinetis/MK8xF/src/fsl_ftfx_controller.c
@@ -1,0 +1,1513 @@
+/*
+ * Copyright 2013-2016 Freescale Semiconductor, Inc.
+ * Copyright 2016-2020 NXP
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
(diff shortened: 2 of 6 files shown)


hw/mcu/nxp/src/ext/nxp-kinetis-sdk/drivers/fsl_ftfx_controller.c
```diff
@@ -516,7 +517,9 @@ status_t FTFx_CMD_Program(ftfx_config_t *config, uint32_t start, const uint8_t *
         }
 
         /* calling flash command sequence function to execute the command */
+        OS_ENTER_CRITICAL(sr);
         returnCode = ftfx_command_sequence(config);
+        OS_EXIT_CRITICAL(sr);
```
mlaz, line 523: I would suggest to move this file somewhere else so we don't have to maintain changes to this when we update the SDK.

utzig: Done.

utzig:
@mlaz I force pushed again adding RAT excludes, style checks can be ignored...

mlaz (approved):
LGTM

PR #2517, opened 2021-03-04
net/osdp: update pkg.yml with a better description

vikrant-proxy:
Update pkg.yml with a better description.


vrahane (approved):
Looks good.

PR #2518, opened 2021-03-05
sys log shell improvements

kasjer:
- Added function to get last log index
- Dumping log from shell with partial log name match
```shell
log my*
```
- Limit number of logs shown using index
```shell
log 1000 mylog
```
- List streaming logs
```shell
log -l
218506 console (stream)
218506 mylog
```


utzig:
Need to update `sys/log/full` to depend on `util/parse`

utzig (approved):
LGTM

PR #2519, opened 2021-03-05
Remove unistd.h from files that don't need it

kasjer:
unistd.h is included in several files that seems not to need
it for anything.


utzig (approved):
LGTM

PR #2520, opened 2021-03-09
Kinetis QSPI driver and frdm-k82f BSP updates

utzig:
QSPI driver updates:

* Clean up LUT definitions and add a few extra MX25U3235F commands
* Remove page size limitations on reads/writes, now reads can be any size and writes can be hf_align sized.
* Add proper synchronization so multiple consumers can use the flash driver concurrently.
* Clean up syscfg naming and break QSPIA and QSPIB apart.

Kinetis K8x and frdm-k82f updates:

* The flash on the K8x family has 4 byte alignment, but due to a bug in MCUboot, it is unable to upgrade when different alignments are used between different flash devices; change the internal flash alignment temporarily to match the one used in the QSPI flash.
* Update flash areas to only store the bootloader and image0 in the internal flash, leaving other partitions in one of the QSPI flashes. Move MCUboot mode to "move" instead of scratch, but leave a scratch partition because FCB needs it. Use different partitions for FCB and the FS.
* Update syscfg QSPI definitions to match QSPI changes in the MCU.
* Use correct syscfg to access QSPI flash.


hw/mcu/nxp/kinetis/src/hal_qspi.c
```diff
@@ -37,62 +37,69 @@
...
-
-    /* Seq7: Erase Sector */
-    /* CMD:  0x20 - Sector Erase, single pad */
-    /* ADDR: 0x18 - 24 bit address, single pad */
-    [28] = QSPI_LUT_SEQ(QSPI_CMD, QSPI_PAD_1, 0x20, QSPI_ADDR, QSPI_PAD_1, 0x18),
-
-    /* Match MISRA rule */
-    [63] = 0
+#define AHB_BUFFER_SIZE MYNEWT_VAL(QSPI_FLASH_PAGE_SIZE)
+#define QSPI_HAS_CLR (defined(FSL_FEATURE_QSPI_SOCCR_HAS_CLR_LPCAC) && \
```
kasjer, line 41: same effect could be
```
#define QSPI_HAS_CLR FSL_FEATURE_QSPI_SOCCR_HAS_CLR_LPCAC
```

kasjer (approved):
LGTM, not tested

hw/mcu/nxp/kinetis/src/hal_qspi.c
utzig: Fixed, thanks!

PR #2523, opened 2021-03-11
hw: kinetis: Only allow aligned block erases

utzig:
This fixes an issue where an erase command with sizes larger or equal than a block erase (32K or 64K) could delete the whole block even when it was not aligned. A block erase command is accepted for any address inside the block, but if a flash area is not aligned to this block size, it will result in erasing data which are not part of the expected requested range. With this change the alignment must be valid when doing block erases.


mlaz (approved):
LGTM

PR #2525, opened 2021-03-11
[DNM] travis: update Linux builds to use gcc 8

utzig:
The currently gcc-multilib 7 is causing failures on Travis due to unmet dependencies. Try bumping it to gcc 8.


utzig:
Nope, didn't work...

PR #2531, opened 2021-03-15
bsp/black_vet6: Fix ETH configuration

kasjer:
Some of the ETH pins were set up to port G.
There are not Port G pins on MCU on this board.
But there are pins from port B that can be used.

IRQ was hard coded to PA3 while ETH code allows
for this pin to be undefined, so now default
value is -1.

PHY address were hardcoded now they can be changed
in syscfg.

### DIFF
--- a/hw/bsp/black_vet6/src/hal_bsp.c
+++ b/hw/bsp/black_vet6/src/hal_bsp.c
@@ -261,14 +261,15 @@ const struct stm32_eth_cfg os_bsp_eth0_cfg = {
     .sec_port_mask[2] = (1 << 1) | (1 << 4) | (1 << 5),
 
     /*
-     * PORTG
-     *   PG11 - ETH_RMII_TXEN
-     *   PG13 - ETH_RMII_TXD0
-     *   PG14 - ETH_RMII_TXD1
+     * PORTB
+     *   PB11 - ETH_RMII_TXEN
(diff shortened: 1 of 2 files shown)


utzig (approved):
LGTM

hw/bsp/black_vet6/src/hal_bsp.c
```diff
@@ -261,14 +261,15 @@ const struct stm32_eth_cfg os_bsp_eth0_cfg = {
...
+     *   PB11 - ETH_RMII_TXEN
+     *   PB12 - ETH_RMII_TXD0
+     *   PB13 - ETH_RMII_TXD1
      */
-    .sec_port_mask[6] = (1 << 11) | (1 << 13) | (1 << 14),
-    .sec_phy_type = SMSC_8710_RMII,
-    .sec_phy_irq = MCU_GPIO_PORTA(3)
+    .sec_port_mask[1] = (1 << 11) | (1 << 12) | (1 << 13),
+    .sec_phy_type = MYNEWT_VAL(STM32_ETH_PHY_TYPE),
+    .sec_phy_irq = MYNEWT_VAL(STM32_ETH_PHY_IRQ)
```
utzig, line 271: oh, there is a missing `,` here, nicely caught by the style checker (without knowing it!).

PR #2533, opened 2021-03-16
mcu/nrf5340: HAL gpio irq should use GPIOTE0

vrahane:
- hal gpio irq should use GPIOTE0 but GPIOTE1 was getting used


utzig (approved):
LGTM

PR #2538, opened 2021-03-18
hw/util/buzzer: Tone generator with the PWM peripheral

rafacouto:
This package implements methods to use a PWM peripheral as a tone generator ready to connect a passive buzzer to a MCU pin. It generates a square wave of the specified frequency and 50% duty cycle.

It depends on the PWM abstract driver provided by Apache Mynewt. Review if the PWM driver interface is implemented for your platform.


andrzej-kaczmarek:
imo this should be `hw/util/buzzer` since it's not a driver for a specific device

apps/buzzer/beep/pkg.yml
```diff
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-pkg.name: apps/beep
+pkg.name: apps/buzzer/beep
```
vrahane, line 20: I think you should just call it apps/buzzer and place the app under it as well.

vrahane (approved):
Looks good to me.

PR #2539, opened 2021-03-18
Add QSPI for NRF5340

kasjer:
This PR adds:
- hal for QSPI
- QSPI flasd device in pca10095 bsp

bsp.yml is not yet modified to have this flash used for any purpose.
It is possible to move slot 1 to this external flash.

Encryption is not implemented yet.


vrahane (approved):
LGTM

vrahane:
@kasjer There are some coding style fixes, not sure if we want to do those considering some of this stuff is from ARM.

sjanc (approved):

PR #2542, opened 2021-03-22
kinetis/hal_gpio.c: Fix hal_gpio_init_out

mlaz:
This PR fixes `hal_gpio_init_out`, which is not setting the initial value to the configured pin. 
There has also been removed a `port_pin_config_t` instance, since the only port configuration needed in this case is the multiplexer.


utzig (approved):

utzig:
For some reason the CI also run on the branch (seems like a branch was pushed upstream). I just went forward and merged because I had to actually fix the same issue this morning.

PR #2543, opened 2021-03-22
net/ip: Update receive windows in lwip_recvfrom

kasjer:
Socket like api function lwip_recvfrom did not called
lwip_recved which is needed to update TCP window.
Without this change socket is able to receive only
TCP_WND bytes which is by default 4380.

Now tcp_recved() is called inside lwip_recvfrom() much
like in other implementations.
It would be slightly better to call it when mbuf is freed
but that is not possible easily right now.

### DIFF
--- a/net/ip/lwip_mn/src/lwip_socket.c
+++ b/net/ip/lwip_mn/src/lwip_socket.c
@@ -593,6 +593,11 @@ lwip_recvfrom(struct mn_socket *ms, struct os_mbuf **mp,
 #endif
             }
         }
+#if LWIP_TCP
+        if (s->ls_type == MN_SOCK_STREAM) {
+            tcp_recved(s->ls_pcb.tcp, m->omp_len);
+        }
+#endif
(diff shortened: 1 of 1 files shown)


net/ip/lwip_mn/src/lwip_socket.c
```diff
@@ -593,6 +593,9 @@ lwip_recvfrom(struct mn_socket *ms, struct os_mbuf **mp,
 #endif
             }
         }
+        if (s->ls_type == MN_SOCK_STREAM) {
```
mkiiskila, line 596: Should make this code conditional on #if LWIP_TCP.

mkiiskila (approved):
Good catch; approved, but needs the #ifdef I mentioned.

net/ip/lwip_mn/src/lwip_socket.c
kasjer: done

PR #2548, opened 2021-03-29
hw: mcu: kinetis: Fix dual flash QSPI support

utzig:
Update the QSPI flash driver so it can work correctly with two flashes connected to the two QSPI channels. This is currenctly limited to using two QSPI flashes of the same type.


kasjer (approved):
LGTM, not tested nor compiled though just common sense

PR #2550, opened 2021-03-31
drivers/da1469x_charger: Enable custom value for NTC

kasjer:
Battery temperature values from CHARGER_TEMPSET_PARAM_REG
could be incorrectly decoded in shell command `charger dump decode`
when NTC value was different then datasheet recommended 15k ohm.

Now real pull-up resistor value is taken into account when values
are displayed.
Value of the resistor can be set in syscfg file.

### DIFF
--- a/hw/drivers/chg_ctrl/da1469x_charger/src/da1469x_charger_shell.c
+++ b/hw/drivers/chg_ctrl/da1469x_charger/src/da1469x_charger_shell.c
@@ -373,11 +373,53 @@ static const char *const tdie_max[] = {
     NULL,
 };
 
+/*
+ * When pull-up resistor for NTC is different then 15k code needs to
+ * recalculate temperature settings. It uses table with voltages measured
+ * at NTC for values from -10 to 53 deg C and converts voltage to
+ * to values that would be measured IF recommended 15K resistor was used.
+ */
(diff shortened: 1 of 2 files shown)


hw/drivers/chg_ctrl/da1469x_charger/src/da1469x_charger_shell.c
```diff
@@ -373,11 +373,53 @@ static const char *const tdie_max[] = {
...
 };
 
+/*
+ * When pull-up resistor for NTC is different then 15k code needs to
+ * recalculate temperature settings. It uses table with voltages measured
+ * at NTC for values from -10 to 53 deg C and converts voltage to
+ * to values that would be measured IF recommended 15K resistor was used.
+ */
+#if MYNEWT_VAL(DA1469X_CHARGER_NTC_VALUE) != 15000
+static uint16_t Vntc_tab[] = {
```
sjanc, line 383: nitpick: could be static const

sjanc (approved):

hw/drivers/chg_ctrl/da1469x_charger/src/da1469x_charger_shell.c
kasjer: fixed

PR #2552, opened 2021-04-01
hw/drivers/flash/spiflash: add JEDEC_MFC_XTX and XT25F32B chip

ildar:
for PineTime smart watch


mlaz:
@caspermeijn can you have a look at this? and try it out, maybe?

kasjer (approved):

PR #2554, opened 2021-04-06
mcu/nrf5340: Add MCU_APP_CORE/MCU_NET_CORE syscfg values

kasjer:
ipc driver was depending on `BSP_NRF5340` (which is specific to
nordic_pca100095) to determine which core was code built for.

Custom bsp would need to introduce same name to use IPC driver.
Now MCUs define own constants that could be used.


sjanc (approved):

vrahane (approved):

vrahane:
Cc: @vikrant-proxy

PR #2558, opened 2021-04-09
nrf5340 combine app/net build

kasjer:
This adds a way to have single image used to upgrade code on both cores.
Combine build advantages:
- easy upgrade deployment
- possible upgrade off application side only
- synchronization of code between net and app side
- **newtmgr** just works
- **mcuboot** works (but could be changed to embrace this approach batter)
- similar pattern of combine build is already present for Dialog network core (made possible by @andrzej-kaczmarek, scripting know-how part borrowed from his cmac)

For this to work **mcuboot** on **network** core have to be build with
```yml
    BOOTUTIL_OVERWRITE_ONLY: 1
```

In the **application** core target
```yml
    NRF5340_EMBED_NET_CORE: 1
```

### DIFF
--- a/hw/bsp/nordic_pca10095_net/bsp.yml
+++ b/hw/bsp/nordic_pca10095_net/bsp.yml
@@ -45,10 +45,10 @@ bsp.flash_map:
             device: 0
             offset: 0x01008000
             size: 100kB
-        # TODO this should map to app flash
+        # This maps to app flash and uses vflash
         FLASH_AREA_IMAGE_1:
-            device: 0
-            offset: 0x01021000
+            device: 1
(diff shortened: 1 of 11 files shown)


vikrant-proxy:
@kasjer can you mark `build_net_core.sh` as executable?

vrahane (approved):
@hardik-proxy and @vikrant-proxy have both verified the OTA DFU works as expected. Apart from the small comment above about making the script executable, rest of it looks good.

kasjer:
> @kasjer can you mark `build_net_core.sh` as executable?

done

vrahane (approved):

PR #2563, opened 2021-04-12
mcu/nrf5340: Add trng to netcore. Use existing trng_nrf52 driver

vikrant-proxy:
- Add nrfx header to trng_nrf52.c


vrahane (approved):
I think this looks fine.

PR #2564, opened 2021-04-12
nrf5340/nrf52xxx: Add support for enabling instruction/data cache in hal_system_init() and add dependencies for BUS DRIVER

vrahane:
- nrf5340: Add support for enabling instruction/data cache in hal_system_init() based on syscfg and add dependencies for BUS_DRIVER

- nrf52xxx: Enable instruction code cache in hal_system_init() based on syscfg

### DIFF
--- a/hw/mcu/nordic/nrf5340/src/hal_system.c
+++ b/hw/mcu/nordic/nrf5340/src/hal_system.c
@@ -36,7 +36,6 @@
 void
 hal_system_init(void)
 {
-
 #if MYNEWT_VAL(MCU_ICACHE_ENABLED)
     NRF_NVMC_S->ICACHECNF = 1;
 #endif
@@ -81,7 +80,6 @@ hal_debugger_connected(void)
 void
 hal_system_clock_start(void)
 {
-
 #if MYNEWT_VAL(MCU_LFCLK_SOURCE)
     uint32_t regmsk;
     uint32_t regval;
--- a/hw/mcu/nordic/nrf5340/src/hal_system.c
+++ b/hw/mcu/nordic/nrf5340/src/hal_system.c
@@ -36,6 +36,15 @@
 void
 hal_system_init(void)
 {
+
+#if MYNEWT_VAL(MCU_ICACHE_ENABLED)
+    NRF_NVMC_S->ICACHECNF = 1;
+#endif
+
+#if MYNEWT_VAL(MCU_CACHE_ENABLED)
+    NRF_CACHE_S->ENABLE = 1;
+#endif
+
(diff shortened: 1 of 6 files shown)


andrzej-kaczmarek:
split this into two separate PRs or at least do not mix unrelated bus driver and cache changes in single commit

sjanc:
why not just always enable icache than?

vrahane:
We can enable cache always but the problem with that is that sometimes the performance actually degrades instead of becoming better. It should be left to the developer to decide if he/she wants to enable it or not. Another thing is, if someone wants to monitor the performance of the cache, example: hit/miss counts, they would enable the cache and look at those registers, so, keeping it conditional is better.

vrahane:
@andrzej-kaczmarek yes agreed, will add a separate commit for it.

vrahane:
@andrzej-kaczmarek added a separate commit for the BUS driver changes.

hw/mcu/nordic/nrf5340/src/hal_system.c
```diff
@@ -72,6 +81,7 @@ hal_debugger_connected(void)
 void
 hal_system_clock_start(void)
 {
+
```
sjanc, line 84: nitpick

hw/mcu/nordic/nrf5340/src/hal_system.c
```diff
@@ -36,6 +36,15 @@
 void
 hal_system_init(void)
 {
+
```
sjanc, line 39: why empty line?

utzig:
@vrahane The Travis fail on the branch is a bit weird, did you first push the branch upstream instead of to your own fork?

sjanc:
fair enough, two nitpicks though

hw/mcu/nordic/nrf5340/src/hal_system.c
vrahane: done

hw/mcu/nordic/nrf5340/src/hal_system.c
vrahane: removed

vrahane:
@utzig your are right, I did push to origin by mistake and then deleted it right after. The recent commits should fix that I guess.

utzig:
> @utzig your are right, I did push to origin by mistake and then deleted it right after. The recent commits should fix that I guess.

Looks like it did, seems to require a rebase to sync with Travis again, thanks!

vrahane:
@utzig done!

sjanc (approved):

PR #2566, opened 2021-04-14
flash_map: make mfg support optional

utzig:
A flash map configuration may be stored in the manufacturing meta region, and be used instead of the one build time generated from bsp.yml. This is currently hard-coded in the flash map code. For users who don't use a mmr, this adds somewhere between 500 and 1000 bytes of code who serves no purpose. This PR adds a syscfg to allow enabling/disabling the use of the MMR in the flash map thus making the mfg dependency optional.

PS: I am adding `FLASH_MAP_SUPPORT_MFG` unset, which breaks backwards compatibility, but on the assumption that most users **don't use** an MMR.

### DIFF
--- a/sys/flash_map/pkg.yml
+++ b/sys/flash_map/pkg.yml
@@ -27,8 +27,10 @@ pkg.keywords:
 
 pkg.deps:
     - "@apache-mynewt-core/sys/defs"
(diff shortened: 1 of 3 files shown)


kasjer (commented):
+1

kasjer (approved):

PR #2567, opened 2021-04-14
crypto: mbedtls: optional montgomery curve support

utzig:
Add a new syscfg to enable/disable Curve448 support. Both Curve25519 and Curve448 are montgomery form elliptic curves, and bring in a lot of supporting code. Curve25519 was already disabled by default, but there was no option to disable Curve448, so the montgomery support code was built whenever MbedTLS based EC code was being used. Leaving this option disabled reduces MbedTLS' size by about 2KB (for EC users).


kasjer (approved):
LGTM

PR #2569, opened 2021-04-16
hw/mcu/cmac: Fix os_cputime and LLT wrap around handling

andrzej-kaczmarek:
os_cputime runs at 32768Hz but it uses LLT internally which runs at
1MHz. This means both timers will wrap around at different time. This
means after one of timers wraps around (os_cputime wraps earlier),
calculations that use values converted between both timers will fail.

To fix this we "manually" extend LLT value to 64-bit internally so it
takes zillion of years before it wraps around. Also conversions from
os_cputime (a.k.a. HAL) value to LLT value take os_cputime value wrap
around into account and return proper 64-bit LLT value.

### DIFF
--- a/hw/mcu/dialog/cmac/include/mcu/cmac_timer.h
+++ b/hw/mcu/dialog/cmac/include/mcu/cmac_timer.h
@@ -27,9 +27,27 @@
 extern "C" {
 #endif
 
+/*
+ * Equals to LLT value after HAL timer wrapped around,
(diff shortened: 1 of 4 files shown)


andrzej-kaczmarek:
this is different approach to https://github.com/apache/mynewt-core/pull/2561

wes3 (approved):
LGTM

PR #2570, opened 2021-04-16
stm32/hal_uart: Fix blocking write

kasjer:
Function hal_uart_blocking_tx() had incorrect condition
checking if UART is open, this prevented non-blocking writes.

Condition:
    if (!u || u->u_open)
was obviously wrong and is now changed to
    if (!u || !u->u_open)


utzig (approved):
LGTM

PR #2571, opened 2021-04-19
nrf53/hal: icache config enable reg is in netcore

vikrant-proxy:
https://infocenter.nordicsemi.com/index.jsp?topic=%2Fps_nrf5340%2Fnvmc.html


vrahane (approved):
Looks good, good catch.

PR #2577, opened 2021-04-21
mcu/stm32: Move .init_array sections to flash

kasjer:
content of .init_array* sections is meant to be executed only
once during system start.
For no apparent reason those arrays were put in data section.
This results in wast of RAM when C++ code is built.

Sections are now moved to .text.


utzig (approved):
LGTM

PR #2580, opened 2021-04-22
mcu/nrf5340: Allow changing HFCLK divider

kasjer:
MCU starts with HFCLK divder set to 2, meaning APP core
runs at 64MHz.
Now divider can be set to 1 in syscfg allowing to run
MCU at 128MHz.


vrahane (approved):
Apart from the syntax issues, code looks good.

kasjer:
I think that file is formatted by Nordic, it was left in this form so we can easily spot what was changed in mynewt version .

PR #2587, opened 2021-05-05
fs: littlefs: add migrate/readonly build options

utzig:
Add syscfg options that enable building the v1 metadata format migration to v2, and read-only support (for smaller build sizes).


mlaz (approved):
LGTM!

PR #2595, opened 2021-05-14
stm32/hal_spi: Enable SPI in hal_spi_enable

kasjer:
Functions hal_spi_enable() and hal_spi_disable() did not do
anything useful.

Now like many other implementations (NRF, DA1469x) those functions
actually enable and disable SPI.

Without this change SPI peripheral could be enabled by
ST HAL functions like HAL_SPI_Transmit_Receive() but
if that was the case chip select would be asserted before
SPI peripheral was enabled.
In some cases it could lead to clock PIN going to inactive state
after chip select was active.
With this change SPI pins are in correct state before CS is activated during transmission.


utzig (approved):
LGTM

PR #2596, opened 2021-05-14
hw/drivers/spiram: Add SPI RAM driver

kasjer:
This adds driver for SPI RAMs.

### DIFF
--- a/hw/drivers/ram/spiram/src/spiram.c
+++ b/hw/drivers/ram/spiram/src/spiram.c
@@ -1,0 +1,302 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <string.h>
+#include <assert.h>
+
+#include "os/mynewt.h"
+#include <bsp/bsp.h>
+#include <hal/hal_spi.h>
+#include <hal/hal_gpio.h>
+#include <spiram/spiram.h>
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+#include <bus/drivers/spi_common.h>
+#endif
+
+static inline int
+spiram_lock(struct spiram_dev *dev)
+{
+#if MYNEWT_VAL(OS_SCHEDULING)
+    return os_mutex_pend(&dev->lock,
+                         os_time_ms_to_ticks32(MYNEWT_VAL(SPIRAM_LOCK_TIMEOUT)));
+#else
+    return 0;
+#endif
+}
+
+static inline void
+spiram_unlock(struct spiram_dev *dev)
+{
+#if MYNEWT_VAL(OS_SCHEDULING)
+    int rc = os_mutex_release(&dev->lock);
+    assert(rc == 0);
+#endif
+}
+
+void
+spiram_cs_activate(struct spiram_dev *dev)
+{
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    struct bus_spi_node *node = (struct bus_spi_node *)&dev->dev;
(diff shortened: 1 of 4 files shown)


hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -0,0 +1,273 @@
...
+    struct bus_node_callbacks cbs = {
+        .init = init_node_cb,
+    };
+    int rc;
+
+    bus_node_set_callbacks((struct os_dev *)dev, &cbs);
+
+    dev->characteristics = spi_cfg->characteristics;
+
+    rc = bus_spi_node_create(name, &dev->dev, &spi_cfg->node_cfg, NULL);
```
utzig, line 188: Could just use a `return` here and `rc` is not needed.

hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -0,0 +1,273 @@
...
+#else
+        spiram_cs_activate(dev);
+        hal_spi_txrx(dev->spi_num, cmd, NULL, dev->characteristics->address_bytes + 1);
+        hal_spi_txrx(dev->spi_num, buf, NULL, size);
+        spiram_cs_deactivate(dev);
+#endif
+    }
+    spiram_unlock(dev);
+
+    return rc;
```
utzig, line 166: `rc` will always be `0`, so this function could as well return `void`.

hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -0,0 +1,286 @@
...
+#include <spiram/spiram.h>
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+#include <bus/drivers/spi_common.h>
+#endif
+
+static inline void
+spiram_lock(struct spiram_dev *dev)
+{
+#if MYNEWT_VAL(OS_SCHEDULING)
+    os_mutex_pend(&dev->lock, OS_TIMEOUT_NEVER);
```
vrahane, line 36: I think error handling needs to be there for this along with a syscfg for the timeout.

hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -0,0 +1,286 @@
...
+#endif
+}
+
+static inline void
+spiram_unlock(struct spiram_dev *dev)
+{
+#if MYNEWT_VAL(OS_SCHEDULING)
+    os_mutex_release(&dev->lock);
```
vrahane, line 44: Error handling needs to be done here as well.

hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -0,0 +1,286 @@
...
+                      const struct spiram_cfg *spi_cfg)
+{
+    int rc;
+
+    dev->spi_num = spi_cfg->spi_num;
+    dev->characteristics = spi_cfg->characteristics;
+    dev->ss_pin = spi_cfg->ss_pin;
+    dev->spi_settings = spi_cfg->spi_settings;
+
+    rc = os_dev_create(&dev->dev, name, OS_DEV_INIT_SECONDARY, 0, spiram_init, NULL);
```
vrahane, line 236: I think an assert would be best here since device creation is an init thing.

hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -0,0 +1,286 @@
...
+    spiram_unlock(dev);
+
+    return rc;
+}
+
+int
+spiram_read(struct spiram_dev *dev, uint32_t addr, void *buf, uint32_t size)
+{
+    uint8_t cmd[5] = { SPIRAM_READ };
+    int i;
```
vrahane, line 98: Why is this an int ? A uint8_t is enough.

hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -0,0 +1,286 @@
...
+
+    return rc;
+}
+
+int
+spiram_write(struct spiram_dev *dev, uint32_t addr, void *buf, uint32_t size)
+{
+    uint8_t cmd[4] = { SPIRAM_WRITE };
+    int rc = 0;
+    int i;
```
vrahane, line 140: Why is this an int ? A uint8_t is enough.

hw/drivers/ram/spiram/src/spiram.c
kasjer: done

hw/drivers/ram/spiram/src/spiram.c
kasjer: done

hw/drivers/ram/spiram/src/spiram.c
kasjer: int is a native type and it's usage usually produces more optimal code.
In this case for ARM build both debug and optimized profiles result in shorter binary when int is used instead of uint8_t.

hw/drivers/ram/spiram/src/spiram.c
kasjer: see above

vrahane (approved):

PR #2597, opened 2021-05-18
sensor_creator: Add LIS2DW12 configuration for bus drivers

kasjer:
LIS2DW12 has support for bus driver.
Now it can be used with sensor_creator.

### DIFF
--- a/hw/sensor/creator/src/sensor_creator.c
+++ b/hw/sensor/creator/src/sensor_creator.c
@@ -595,15 +595,47 @@ static struct sensor_itf i2c_0_itf_lpst = {
 };
 #endif
 
-#if MYNEWT_VAL(I2C_0) && MYNEWT_VAL(LIS2DW12_OFB)
+#if MYNEWT_VAL(LIS2DW12_OFB)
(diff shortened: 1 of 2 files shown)


vrahane (approved):
LGTM

hw/sensor/creator/src/sensor_creator.c
```diff
@@ -595,16 +595,48 @@ static struct sensor_itf i2c_0_itf_lpst = {
...
+        { -1, 0, 0 },
+    }
+};
+#else
+#if MYNEWT_VAL(I2C_0)
 static struct sensor_itf i2c_0_itf_lis2dw12 = {
     .si_type = SENSOR_ITF_I2C,
-    .si_num  = 0,
+    .si_num = 0,
     .si_addr = 0x18,
```
vikrant-proxy, line 631: I think this should use the syscfg, as it is variable

vrahane: +1, it should be `LIS2DW12_OFB_ITF_ADDR`

vrahane (approved):

PR #2599, opened 2021-05-19
spiram fix build issues

kasjer:
There were two problems:
- there was link error SPIRAM_0 was not defined
- there was build warning for non bus driver build

Apart from those this PR does not change any logic.


vrahane (approved):
LGTM

PR #2600, opened 2021-05-21
net/osdp: add a polling timeout for non-secure session

vikrant-proxy:
- Raise a timeout flag if the peripheral device does not hear from the control panel for said time interval


vrahane (approved):
Looks good.

PR #2603, opened 2021-05-27
kernel/os/src: replace SYS_ by OS_ errorcodes

fjmolinas:
This PR uses the correct error code macros for the file. This is useful to avoid redefining MACROS when only mynewt-core/kernel/os is being imported in an external project.


fjmolinas:
Ah I didn't realize its actually matching what is described in the API because `OS_E%` are non-negative while `SYS_E%` are. Since this would mean an API change this is not a small change anymore, and not worth it.

PR #2605, opened 2021-05-27
hw/drivers/uart: Add MAX3107 driver

kasjer:
This adds driver for max3107 (SPI to UART device).

### DIFF
--- a/hw/drivers/uart/max3107/include/max3107/max3107.h
+++ b/hw/drivers/uart/max3107/include/max3107/max3107.h
@@ -121,6 +121,18 @@ int max3107_write(struct max3107_dev *dev, const void *buf, size_t size);
 int max3107_read(struct max3107_dev *dev, void *buf, size_t size);
 
 /**
+ * Get real baudrate of the UART
+ *
+ * Baudrate requested during opening of the device may not be possible with
+ * clock that is supplied to max3107.
+ * User can check what is the actual value baudrate calling this function.
+ *
+ * @param dev - device to check baudrate
+ * @return baudrate of the device
+ */
+uint32_t max3107_get_real_baudrate(struct max3107_dev *dev);
+
+/**
  * Returns number of bytes in RX FIFO that can be read.
  *
  * @param dev - MAX3107 device
--- a/hw/drivers/uart/max3107/include/max3107/max3107.h
+++ b/hw/drivers/uart/max3107/include/max3107/max3107.h
@@ -1,0 +1,189 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef __DRIVERS_MAX3107_H_
+#define __DRIVERS_MAX3107_H_
+
+#include <stdint.h>
+#include <uart/uart.h>
+#include <hal/hal_spi.h>
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+#include <bus/drivers/spi_common.h>
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct max3107_dev;
+
+struct max3107_cfg {
+#if MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    struct bus_spi_node_cfg node_cfg;
+#else
+    struct hal_spi_settings spi_settings;
+    int spi_num;
+    int ss_pin;
+#endif
+    /**< External clock/oscillator frequency in Hz */
+    uint32_t osc_freq;
+    /**< IRQ Pin */
+    int8_t irq_pin;
+    /** RX FIFO trigger level */
+    uint8_t rx_trigger_level;
+    /** TX FIFO trigger level */
+    uint8_t tx_trigger_level;
+    int8_t ldoen_pin;
+    /**< 1 - External crystal oscillator, 0 - external clock */
+    int8_t crystal_en: 1;
+    /**< 1 - To disable PLL */
+    uint8_t no_pll: 1;
+    /**< 1 - To enable x 4 mode */
+    uint8_t allow_mul_4: 1;
+    /**< 1 - To enable x 4 mode */
+    uint8_t allow_mul_2: 1;
+};
+
+typedef enum max3107_error {
+    /** Error during SPI transaction device may be in unpredictable state, close device */
+    MAX3107_ERROR_IO_FAILURE = 0x01,
+    /** RX overrun, some data was lost */
+    MAX3107_ERROR_UART_OVERRUN = 0x02,
+    /** Parity error on incoming data */
+    MAX3107_ERROR_UART_PARITY = 0x04,
+    /** Framing error on incoming data */
+    MAX3107_ERROR_UART_FRAMING = 0x08,
+} max3107_error_t;
+
+struct max3107_client {
+    /**< Function that will be called from process context when data can be read. */
+    void (*readable)(struct max3107_dev *dev);
+    /**< Function that will be called from process context when data can be written. */
+    void (*writable)(struct max3107_dev *dev);
+    /**< Function that will be called from process context when break is detected. */
+    void (*break_detected)(struct max3107_dev *dev);
+    /**< Function that will be called from process context when receive errors are detected. */
+    void (*error)(struct max3107_dev *dev, max3107_error_t errcode);
+};
+
+/**
+ * Configure UART parameters.
+ *
+ * This function sets baud rate, data size, stop bits, parity, flow control.
+ *
+ * @param dev  - device to configure.
+ * @param conf - UART parameters.
+ *
+ * @return 0 on success, error code otherwise.
+ */
+int max3107_config_uart(struct max3107_dev *dev, const struct uart_conf_port *conf);
+
+/**
+ * Write to UART TX FIFO.
+ *
+ * @param dev  - device to write to
+ * @param buf  - data to write
+ * @param size - number of bytes to write
+ *
+ * @return negative value on failure,
+ *         0 output FIFO full or size is 0
+ *         number of bytes written (can be less then size)
+ */
+int max3107_write(struct max3107_dev *dev, const void *buf, size_t size);
+
+/**
+ * Read UART RX FIFO bytes
+ *
+ * This function does not wait for data it just reads whatever is in RX FIFO.
+ * Can return 0 if RX FIFO is empty.
+ *
+ * @param dev  - device to read from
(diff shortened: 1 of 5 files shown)


hw/drivers/uart/max3107/include/max3107/max3107.h
```diff
@@ -0,0 +1,172 @@
...
+    uint8_t no_pll: 1;
+    /**< 1 - To enable x 4 mode */
+    uint8_t allow_mul_4: 1;
+    /**< 1 - To enable x 4 mode */
+    uint8_t allow_mul_2: 1;
+};
+
+struct max3107_client {
+    /**< Function that will be called from process context when data can be read. */
+    void (*readable)(struct max3107_dev *dev);
```
vrahane, line 61: I think we do need to return `int` here and for all the function pointers below.

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1059 @@
...
+        dev->regs.clock.clk_source &= ~CLCSOURCE_PLLBYPASS;
+    }
+    if (dev->cfg.allow_mul_4) {
+        dev->regs.clock.brg_config = BRGCONFIG_4XMODE;
+    } else if (dev->cfg.allow_mul_2) {
+        dev->regs.clock.brg_config = BRGCONFIG_2XMODE;
+    } else {
+        dev->regs.clock.brg_config = 0;
+    }
+    max3107_compute_clock_config(dev->cfg.osc_freq, conf->uc_speed, &dev->regs.clock);
```
vrahane, line 287: You are returning `best_br` here, should we be using it instead for our config or something ? Do we need to verify it with our specified config.

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1059 @@
...
+    dev->regs.modes.irda = 0;
+
+    rc = max3107_write_regs(dev, MAX3107_REG_MODE1, &dev->regs.modes.mode1, 6);
+    if (rc) {
+        goto end;
+    }
+
+    /* RTS activated at FIFO level 120. */
+    dev->regs.fifo.flow_lvl = 0xFF;
+    dev->regs.fifo.fifo_trg_lvl = 0x38;
```
vrahane, line 311: Does this `0x38` come from the datasheet ?

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1059 @@
...
+    rc = max3107_write_regs(dev, MAX3107_REG_FLOWLVL, &dev->regs.fifo.flow_lvl, 2);
+    if (rc) {
+        goto end;
+    }
+
+    dev->regs.flow.flow_ctrl = conf->uc_flow_ctl ? (FLOWCTRL_AUTOCTS | FLOWCTRL_AUTORTS) : 0;
+    max3107_write_reg(dev, MAX3107_REG_FLOWCTRL, dev->regs.flow.flow_ctrl);
```
vrahane, line 318: We need to handle error here.

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1059 @@
...
+static void
+max3107_disable_rx_int(struct max3107_dev *dev)
+{
+    max3107_irqen_clear(dev, IRQEN_RXTRGIEN | IRQEN_LSRERRIEN);
+}
+
+static void
+max3107_enable_rx_int(struct max3107_dev *dev)
+{
+    max3107_irqen_set(dev, IRQEN_RXTRGIEN | IRQEN_LSRERRIEN);
```
vrahane, line 556: Handle error here.

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1059 @@
...
+}
+
+static void
+max3107_disable_tx_int(struct max3107_dev *dev)
+{
+    max3107_irqen_clear(dev, IRQEN_TXEMTYIEN | IRQEN_TXTRGIEN);
```
vrahane, line 562: Handle error here.

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1059 @@
...
+}
+
+static void
+max3107_enable_tx_int(struct max3107_dev *dev)
+{
+    max3107_irqen_set(dev, IRQEN_TXEMTYIEN | IRQEN_TXTRGIEN);
```
vrahane, line 568: Handle error here.

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1059 @@
...
+int
+max3107_set_client(struct max3107_dev *dev, const struct max3107_client *client)
+{
+    hal_gpio_irq_disable(dev->cfg.irq_pin);
+
+    dev->client = client;
+
+    if (client) {
+        hal_gpio_irq_enable(dev->cfg.irq_pin);
+        max3107_isr_cb(dev);
```
vrahane, line 617: Handle error here.

vrahane (changes requested):
So, the code looks really clean, we do need to handle errors in various places. I think adding STATS and LOG messages for figuring out what is going wrong would be helpful here.

hw/drivers/uart/max3107/syscfg.yml
```diff
@@ -0,0 +1,134 @@
...
+
+    MAX3107_0_SPI_NUM:
+        description: >
+            SPI interface used for SPI RAM.
+        value: 0
+
+    MAX3107_0_OSC_FREQ:
+        description: >
+            External clock/oscillator frequency.
+        value:
```
vrahane, line 87: This value needs to be set to a default cause `clock_config` uses it, or you might need to handle the case where this is not set.

vrahane:
Also, a basic CLI to query registers of the chip would be useful for debugging.

hw/drivers/uart/max3107/include/max3107/max3107.h
kasjer: Driver does not care if client reacts to notifications. So return value is not needed.

hw/drivers/uart/max3107/src/max3107.c
kasjer: Other drivers silently set register values from requested baud rate. NRF5x for example outlines in product specification what is the actual baud rate for requested one. For MCUs, code calculates it in much simpler way since there is only one clock involved (or takes values from table) and you don't get information what is the actual speed at all.
In MAX3107 case which external clock/oscillator, code finds most accurate settings and returns it so user can verify it during the development phase by just reading return value of this functions instead of checking registers and computing it manually.
So this value is informative only. For example on my dev board that has 36868400 clock I could get as high as 19,7 MHz and not 24MHz that seems to be possible from MAX3107 datasheet.

hw/drivers/uart/max3107/syscfg.yml
kasjer: There is not default value, it must be set according to actual external component value.

hw/drivers/uart/max3107/include/max3107/max3107.h
vrahane: How is the driver going to detect read/write errors without a return value ?

hw/drivers/uart/max3107/src/max3107.c
vrahane: I see, makes sense. some comment or a log statement will be helpful in understanding what the real Vs specified config is.

hw/drivers/uart/max3107/syscfg.yml
vrahane: Yeah, you are right since this is an external oscillator but there needs to be a range or something associated with this.

hw/drivers/uart/max3107/include/max3107/max3107.h
kasjer: Driver does not care if client reacts to notifications. So return value is not needed.

hw/drivers/uart/max3107/src/max3107.c
kasjer: Separate function to get actual baudrate added

hw/drivers/uart/max3107/src/max3107.c
kasjer: Fifo trigger level now can be set with syscfg, so user can experiment with other values.

hw/drivers/uart/max3107/src/max3107.c
kasjer: All cases are now handled.
Client will be notified about catastrophic failures in which case closing and opening device should be done.

hw/drivers/uart/max3107/syscfg.yml
kasjer: Default value is not provided.
Range check is added to allow for valid frequencies 0.5-35 MHz.

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1090 @@
...
+max3107_read_regs(struct max3107_dev *dev, uint8_t addr, void *buf, uint32_t size)
+{
+    int rc;
+    bool locked;
+#if !MYNEWT_VAL(BUS_DRIVER_PRESENT)
+    uint8_t fast_buf[8];
+#endif
+
+    rc = max3107_lock(dev);
+    locked = rc == 0;
```
sjanc, line 82: shouldn't this return if failed to lock device?

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1090 @@
...
+}
+
+int
+max3107_write_regs(struct max3107_dev *dev, uint8_t addr, const uint8_t *buf, uint32_t size)
+{
+    int rc;
+    bool locked;
+    uint8_t fast_buf[17];
+
+    rc = max3107_lock(dev);
```
sjanc, line 128: ditto

hw/drivers/uart/max3107/src/max3107.c
```diff
@@ -0,0 +1,1090 @@
...
+    rc = max3107_write_reg(dev, MAX3107_REG_FLOWCTRL, dev->regs.flow.flow_ctrl);
+    if (rc) {
+        goto end;
+    }
+
+    dev->regs.ints.irq_en = IRQEN_LSRERRIEN | IRQEN_RXTRGIEN;
+    dev->regs.ints.lsr_int_en = LSRINTEN_FRAMEERRIEN | LSRINTEN_PARITYIEN |
+                                LSRINTEN_RBREAKIEN | LSRINTEN_ROVERRIEN | LSRINTEN_RTIMEOUTIEN;
+    rc = max3107_write_regs(dev, MAX3107_REG_IRQEN, &dev->regs.ints.irq_en, 3);
+end:
```
sjanc, line 333: this label doesn't do any cleanup etc so maybe avoid all those gotos?

hw/drivers/uart/max3107/src/max3107.c
kasjer: yes

hw/drivers/uart/max3107/src/max3107.c
kasjer: if code was to be modified with some kind of tracing single place with return is easier to manage.
My local version (with changes that are for debug purposes only) does just that.

sjanc: alright, fair enough

kasjer:
> So, the code looks really clean, we do need to handle errors in various places. I think adding STATS and LOG messages for figuring out what is going wrong would be helpful here.

I will add separate PR for that if needed.

kasjer:
> Also, a basic CLI to query registers of the chip would be useful for debugging.

I will add separate PR for that that.

vrahane (approved):
Looks good.

PR #2606, opened 2021-05-27
Improve bus write read transactions

kasjer:
Change extends bus driver operations to allow faster transfer.

New driver function **write_read** allow for faster transaction implementation.

Existing function **bus_node_write_read_transact()** was using **write** that **read**.
It turns out that for short transfers (1 or 2 bytes written) several bytes read it is worth
to implement custom function that writes and read.

This code also adds first implementation of such function for spi_hal driver that can
benefit around 50% time reduction on single transfer.

### DIFF
--- a/hw/bus/drivers/spi_hal/src/spi_hal.c
+++ b/hw/bus/drivers/spi_hal/src/spi_hal.c
@@ -191,6 +191,66 @@ bus_spi_write(struct bus_dev *bdev, struct bus_node *bnode, const uint8_t *buf,
     return rc;
 }
 
+static int
+bus_spi_write_read(struct bus_dev *bdev, struct bus_node *bnode,
+                   const uint8_t *wbuf, uint16_t wlength,
+                   uint8_t *rbuf, uint16_t rlength,
+                   os_time_t timeout, uint16_t flags)
+{
+    struct bus_spi_hal_dev *dev = (struct bus_spi_hal_dev *)bdev;
+    struct bus_spi_node *node = (struct bus_spi_node *)bnode;
+    int rc;
+    uint8_t buf[16];
+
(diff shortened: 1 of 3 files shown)


hw/bus/drivers/spi_hal/src/spi_hal.c
```diff
@@ -191,6 +191,66 @@ bus_spi_write(struct bus_dev *bdev, struct bus_node *bnode, const uint8_t *buf,
...
+
+    /* XXX update HAL to accept const instead */
+
+#if MYNEWT_VAL(SPI_HAL_USE_NOBLOCK)
+    if (wlength + rlength <= sizeof(buf)) {
+        memcpy(buf, wbuf, wlength);
+        memset(buf + wlength, 0, rlength);
+        rc = hal_spi_txrx_noblock(dev->spi_dev.cfg.spi_num, buf, buf, wlength + rlength);
+        if (rc == 0) {
+            os_sem_pend(&dev->sem, OS_TIMEOUT_NEVER);
```
vrahane, line 218: This does need a specific timeout value. same with the one below.

vrahane:
I do like the idea and I think this is needed for sure. As per above dummy reads/writes are handled by the underlying api which is great.

hw/bus/drivers/spi_hal/src/spi_hal.c
kasjer: > This does need a specific timeout value. Same with the ones below.
Structure of the block was put together from write and read that use timeout like that.
While I'm not against changing this I will keep it like that in this PR.
If anyone feels that it should be changed, all places could be applied in separate PR(s).

vrahane (approved):
LGTM.

PR #2607, opened 2021-05-28
sensor_creator: Add non-bus driver SPI config for LIS2DW12

kasjer:
LIS2DW12 could be used with SPI and I2C when bus driver was used.
SPI configuration was missing for non-bus driver usage.

This also restores default I2C interface number 0 for LIS2DW12.

### DIFF
--- a/hw/sensor/creator/src/sensor_creator.c
+++ b/hw/sensor/creator/src/sensor_creator.c
@@ -624,17 +624,21 @@ static struct sensor_itf lis2dw12_itf = {
     }
 };
 #else
+static struct sensor_itf lis2dw12_itf = {
 #if MYNEWT_VAL(LIS2DW12_OFB_I2C_NUM) >= 0
-static struct sensor_itf i2c_0_itf_lis2dw12 = {
     .si_type = SENSOR_ITF_I2C,
     .si_num = MYNEWT_VAL(LIS2DW12_OFB_I2C_NUM),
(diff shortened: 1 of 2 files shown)


sjanc (approved):

hw/sensor/creator/src/sensor_creator.c
```diff
@@ -624,19 +624,23 @@ static struct sensor_itf lis2dw12_itf = {
...
+static struct sensor_itf lis2dw12_itf = {
 #if MYNEWT_VAL(LIS2DW12_OFB_I2C_NUM) >= 0
-static struct sensor_itf i2c_0_itf_lis2dw12 = {
     .si_type = SENSOR_ITF_I2C,
     .si_num = MYNEWT_VAL(LIS2DW12_OFB_I2C_NUM),
     .si_addr = MYNEWT_VAL(LIS2DW12_OFB_ITF_ADDR),
+#elif MYNEWT_VAL(LIS2DW12_OFB_SPI_NUM) >= 0
+    .si_type = SENSOR_ITF_SPI,
+    .si_num = MYNEWT_VAL(LIS2DW12_OFB_SPI_NUM),
+    .si_cs_pin = MYNEWT_VAL(BME280_OFB_CS),
```
vikrant-proxy, line 635: shouldn't this be `LIS2DW12_OFB_CS` ?

kasjer: Yeap, it was a rush job to fix coverity build, thanks for spotting.
Fix is coming #2609

PR #2604, opened 2021-05-27
kernel/os/include/os/os_trace_api: avoid unused parameter warning

fjmolinas:
This PR voids case the unused parameters in `os_trace_api`.

PR #2608, opened 2021-05-28
sys/log/stub: avoid unused parameter warning

fjmolinas:
Similar to #2604 this PR voids casts unused parameters, useful when only including `os/*` and not wanting to disable the warning for all compilation units (I'm using Werror in my build).

### DIFF
--- a/sys/log/stub/include/log/log.h
+++ b/sys/log/stub/include/log/log.h
@@ -43,17 +43,25 @@ static inline int
 log_register(const char *name, struct log *log, const struct log_handler *h,
              void *arg, uint8_t level)
 {
+    (void)name;
+    (void)log;
+    (void)h;
+    (void)arg;
+    (void)level;
     return 0;
 }
 
 static inline void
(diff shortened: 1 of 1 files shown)


sys/log/stub/include/log/log.h
```diff
@@ -43,38 +43,62 @@ static inline int
 log_register(const char *name, struct log *log, const struct log_handler *h,
              void *arg, uint8_t level)
 {
+    (void)name;
+    (void)log;
+    (void)log_handler;
```
kasjer, line 48: @fjmolinas maybe you don't have access to travis output to see why build failed.

```c
(void)log_handler;
```
should be
```c
(void)h;
```

kasjer (changes requested):

sys/log/stub/include/log/log.h
fjmolinas: Face palm... copy pasta, it seems I did a poor compile test, sorry about that.

kasjer (approved):
LGTM

PR #2610, opened 2021-05-31
nxp: kinetis: decrease LPUART RAM usage

utzig:
Refactor LPUart driver to only allocated RAM proportional to the number of enabled Uarts; each UART uses about 230 bytes for its state, so having a single UART enabled should provide a significant reduction in RAM usage.

The driver's flash usage was also reduced, by about 20% when a single UART is being used.

### DIFF
--- a/hw/mcu/nxp/kinetis/src/hal_lpuart.c
+++ b/hw/mcu/nxp/kinetis/src/hal_lpuart.c
@@ -70,25 +70,112 @@ struct hal_uart {
 };
 
 /* UART configurations */
-static struct hal_uart uarts[FSL_FEATURE_SOC_LPUART_COUNT];
+#define UART_CNT (((uint8_t)(MYNEWT_VAL(UART_0) != 0)) + \
+                  ((uint8_t)(MYNEWT_VAL(UART_1) != 0)) + \
+                  ((uint8_t)(MYNEWT_VAL(UART_2) != 0)) + \
+                  ((uint8_t)(MYNEWT_VAL(UART_3) != 0)) + \
+                  ((uint8_t)(MYNEWT_VAL(UART_4) != 0)))
 
-static uint8_t const s_uartExists[] = NXP_UART_EXISTS;
-static uint8_t const s_uartEnabled[] = NXP_UART_ENABLED;
+static struct hal_uart uarts[UART_CNT];
+static struct hal_uart *
+uart_by_port(int port)
+{
+    int index;
+
+    (void)index;
+    (void)uarts;
+
+    index = 0;
+#if MYNEWT_VAL(UART_0)
+    if (port == 0) {
+        return &uarts[index];
+    }
+    index++;
+#endif
+#if MYNEWT_VAL(UART_1)
+    if (port == 1) {
+        return &uarts[index];
+    }
+    index++;
+#endif
+#if MYNEWT_VAL(UART_2)
+#   if FSL_FEATURE_SOC_LPUART_COUNT < 3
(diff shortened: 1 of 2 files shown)


hw/mcu/nxp/kinetis/src/hal_lpuart.c
```diff
@@ -410,7 +501,8 @@ hal_uart_config(int port,
     uconfig.enableTx = true;
     uconfig.enableRx = true;
 
-    NVIC_SetVector(u->u_irq, (uint32_t)s_uartirqs[port]);
+
```
kasjer, line 504: extra empty line

hw/mcu/nxp/kinetis/src/hal_lpuart.c
```diff
@@ -345,12 +433,15 @@ hal_uart_config(int port,
...
-    if (port >= FSL_FEATURE_SOC_LPUART_COUNT) {
+    u = uart_by_port(port);
+    if (!u || !u->u_configured || u->u_open) {
         return -1;
     }
-    u = &uarts[port];
-    if (!u->u_configured || u->u_open) {
+
+    uart_irq_fn = s_uartirqs[port];
+    if (!uart_irq_fn) {
```
kasjer, line 444: this check will never fail, it seems

hw/mcu/nxp/kinetis/src/hal_lpuart.c
```diff
@@ -70,25 +70,112 @@ struct hal_uart {
...
 static void uart_irq2(void);
+#endif
+#if MYNEWT_VAL(UART_3)
 static void uart_irq3(void);
+#endif
+#if MYNEWT_VAL(UART_4)
 static void uart_irq4(void);
+#endif
+
 static void (*s_uartirqs[])(void) = {
```
kasjer, line 153: While it's not part of a change s_uartirqs could be const

hw/mcu/nxp/kinetis/src/hal_lpuart.c
```diff
@@ -70,25 +70,112 @@ struct hal_uart {
...
+#   error "UART_2 is not supported on this MCU"
+#   endif
+    if (port == 2) {
+        return &uarts[index];
+    }
+    index++;
+#endif
+#if MYNEWT_VAL(UART_3)
+#   if FSL_FEATURE_SOC_LPUART_COUNT < 4
+#   error "UART_3 is not supported on this MCU"
```
kasjer, line 112: Slight inconsistency here, if `#if` is indented why not indent `#error` as well

kasjer (approved):
LGTM

hw/mcu/nxp/kinetis/src/hal_lpuart.c
utzig: Fixed.

hw/mcu/nxp/kinetis/src/hal_lpuart.c
utzig: It will not, I removed the check.

hw/mcu/nxp/kinetis/src/hal_lpuart.c
utzig: Done.

hw/mcu/nxp/kinetis/src/hal_lpuart.c
utzig: I am not sure I get what you mean exactly, but I added an extra indent to the `error ...` line, although it looks weird to me.

PR #2611, opened 2021-06-03
hw/drivers/spiram: Fix 4 bytes address handling

kasjer:
Even though there is no 4 bytes addressing RAM out there,
because syscfg allowed such value to be set update code
to handle this case correctly to be future proof.

Also restriction was added for dummy bytes.

### DIFF
--- a/hw/drivers/ram/spiram/src/spiram.c
+++ b/hw/drivers/ram/spiram/src/spiram.c
@@ -102,7 +102,7 @@ spiram_write_enable(struct spiram_dev *dev)
 int
 spiram_read(struct spiram_dev *dev, uint32_t addr, void *buf, uint32_t size)
 {
-    uint8_t cmd[5] = { SPIRAM_READ };
+    uint8_t cmd[6] = { SPIRAM_READ };
     int i;
     int rc;
     int cmd_size = 1 + dev->characteristics->address_bytes + dev->characteristics->dummy_bytes;
@@ -147,7 +147,7 @@ spiram_read(struct spiram_dev *dev, uint32_t addr, void *buf, uint32_t size)
 int
 spiram_write(struct spiram_dev *dev, uint32_t addr, void *buf, uint32_t size)
 {
-    uint8_t cmd[4] = { SPIRAM_WRITE };
+    uint8_t cmd[5] = { SPIRAM_WRITE };
     int rc;
     int i;
     bool locked;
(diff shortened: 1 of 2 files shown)


hw/drivers/ram/spiram/src/spiram.c
```diff
@@ -102,7 +102,7 @@ spiram_write_enable(struct spiram_dev *dev)
 int
 spiram_read(struct spiram_dev *dev, uint32_t addr, void *buf, uint32_t size)
 {
-    uint8_t cmd[5] = { SPIRAM_READ };
+    uint8_t cmd[6] = { SPIRAM_READ };
```
vrahane, line 105: Does it make sense to break the limit down into 3/4 byte addressing + max command length or something ?

vikrant-proxy: Given these parameters don't change on the fly, splitting these into 
```
(1 + MYNEWT_VAL(SPIRAM_0_ADDRESS_BYTES) + MYNEWT_VAL(SPIRAM_0_DUMMY_BYTES)) 
```
might make this more readable.
@kasjer can we merge this soon

vrahane: Agreed, that does seem like a better approach.

kasjer: we can't, SPIRAM_0_ADDRESS_BYTES is per instance setting not per driver

vikrant-proxy: then maybe we use `cmd_size` that uses per instance setting?

kasjer: there is no point doing this unless you aim is to have more flash used by the code, slower execution time and more stack used.

vrahane:
```
/* The user is required to specify both based on their config and usage. */

(1 + MYNEWT_VAL(SPIRAM_MAX_ADDRESS_BYTES) + MYNEWT_VAL(SPIRAM_MAX_DUMMY_BYTES))

```

kasjer: I'll take it as a joke.

vrahane: I know its not super intuitive but it is better than 5-6 hardcoded value which seems we have no clue about where it came from.

vrahane: @kasjer  Can you please address this ? Let's get it merged.

kasjer:
There are no 4 bytes address SPI RAM out there restriction can be change to allow only 2,3 bytes instead of 2,3,4

PR #2613, opened 2021-06-05
mbedtls xc32 build fixes

kasjer:
This changes one instance of `#include ""` to `#include <>` to allow
build with xc32-gcc compiler. This is really a issue in the xc32 compiler not
**mbedtls** but it can be fixed by changing `""`  to `<>`.

Other change is to remove include `"os/mynewt.h"` and use `<syscfg/syscfg.h>`
this time is due to symbol found in PIC32MZ... headers that conflicts with some
local symbol used by **mbedtls**.
Including big part of **mynewt** core header files is not really needed to build **mbedtls**
when just having **MYNEWT_VAL** macro expansion is enough.

### DIFF
--- a/crypto/mbedtls/include/mbedtls/config_mynewt.h
+++ b/crypto/mbedtls/include/mbedtls/config_mynewt.h
@@ -32,7 +32,7 @@
 extern "C" {
 #endif
 
-#include "os/mynewt.h"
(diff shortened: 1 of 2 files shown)


vrahane (approved):
LGTM

PR #2614, opened 2021-06-09
nrf5340: Conditional Rev A errata 26 fix

vrahane:
- As per Errata 26, CTRL-AP: APPROTECT.DISABLE and
  SECUREAPPROTECT.DISABLE registers are not functional
- So we need to conditionally disable its usage
- Adding a syscfg so we can rely on UICR config instead of CTRLAP
-  Remove MCU_NET_CORE restriction from APP CORE, does not matter anymore since MCUs are different

Note: Somehow this fix is not required on the netcore side. Hence, only making this fix on the app core side.

** BEFORE **
A `HardFault` would happen at `57             NRF_CTRLAP_S->APPROTECT.DISABLE = NRF_UICR_S->APPROTECT;` in `repos/apache-mynewt-core/hw/mcu/nordic/src/ext/nrfx/mdk/system_nrf5340_application.c` which would be seen in the bootloader.

** AFTER **
No `HardFault`

### DIFF
--- a/hw/mcu/nordic/nrf5340/pkg.yml
+++ b/hw/mcu/nordic/nrf5340/pkg.yml
@@ -104,3 +104,6 @@ pkg.deps.I2C_3':
 
 pkg.deps.NRF5340_EMBED_NET_CORE:
     - "@apache-mynewt-core/hw/mcu/nordic/nrf5340/net_core_image"
+
+pkg.cflags.MCU_NRF5340_EN_APPROTECT_USERHANDLING:
+    - "-DENABLE_APPROTECT_USER_HANDLING"
--- a/hw/mcu/nordic/nrf5340/syscfg.yml
+++ b/hw/mcu/nordic/nrf5340/syscfg.yml
@@ -25,13 +25,14 @@ syscfg.defs:
             - $notnull
         choices:
             - nRF5340_APP
+
     MCU_APP_CORE:
         description: >
             Constant value always set to 1.  It allows to have common
             packages for network and application core that do have
             some differences depending on which core they are build for.
         value: 1
-        restriction: MCU_NET_CORE==0
+
     MCU_FLASH_MIN_WRITE_SIZE:
         description: >
             Specifies the required alignment for internal flash writes.
@@ -87,6 +88,12 @@ syscfg.defs:
         description: >
             Enable instruction and data cache
             Default value is 0, so disabled.
+        value: 0
+
+    MCU_NRF5340_EN_APPROTECT_USERHANDLING:
+        description: >
+            This needs to be set by the BSP/App/Target if the
+            MCU/BSP wants to handle APPROTECT on its own.
         value: 0
 
 # MCU peripherals definitions


wes3 (approved):
LGTM but I have no board to test this on.

hw/mcu/nordic/nrf5340/syscfg.yml
```diff
@@ -25,13 +25,15 @@ syscfg.defs:
...
         choices:
             - nRF5340_APP
+
     MCU_APP_CORE:
         description: >
             Constant value always set to 1.  It allows to have common
             packages for network and application core that do have
             some differences depending on which core they are build for.
         value: 1
         restriction: MCU_NET_CORE==0
```
sjanc, line 35: isn't the real problem a typo above?  restriction_S_   ?

hw/mcu/nordic/nrf5340/pkg.yml
```diff
@@ -104,3 +104,10 @@ pkg.deps.I2C_3':
 
 pkg.deps.NRF5340_EMBED_NET_CORE:
     - "@apache-mynewt-core/hw/mcu/nordic/nrf5340/net_core_image"
+
+# As per Errata 26 for Engineering RevA silicon we should not
+# depend on CTRL-AP: APPROTECT.DISABLE and SECUREAPPROTECT.DISABLE
+# registers since they are not functional. It causes quite a lot of
+# crashes and hardfaults
+pkg.cflags.MCU_NRF5340_REVA_SILICON:
```
sjanc, line 112: I'd not make this as revA workaround but rather a general config as provided by nrfx:  enabled, user controlled and UICR controlled.

sjanc (changes requested):

hw/mcu/nordic/nrf5340/syscfg.yml
vrahane: I guess, I just wanted the build to restart so, I found something that needed fixing :-)

hw/mcu/nordic/nrf5340/pkg.yml
vrahane: done.

hw/mcu/nordic/nrf5340/syscfg.yml
vrahane: done.

vrahane: That restriction does not hold truth anymore since MCUs are different for both cores. I am removing it and updated description above.

vrahane:
@sjanc Can you please review ?

sjanc (commented):
I'd not mention that this is for RevA only as it is perfectly valid for Rev1 if user wants to handle this in code.

Other than comments it looks good to me though

vrahane:
@sjanc done, I have mentioned Errata 26 in the detailed commit message because it is needed for that. Kept the commit message generic enough.

sjanc (approved):

PR #2615, opened 2021-06-11
max3107 shell

kasjer:
This adds few commands for shell to manipulate registers of MAX3107.

Second commit moves MAX3107_0 instance settings to separate package
that will make it more clear what is instance and what is driver setting.


hw/drivers/uart/max3107/max3107_0/pkg.yml
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: hw/drivers/uart/max3107/max3107_0
```
vrahane, line 20: I don't think this should be an instance. It should just be `hw/drivers/uart/max3107`

hw/drivers/uart/max3107/max3107_0/pkg.yml
```diff
@@ -0,0 +1,27 @@
...
+pkg.description: Package adds single instance of MAX3107 device.
+pkg.author: "Apache Mynewt"
+pkg.keywords:
+    - uart
+
+pkg.init:
+    max3107_0_init: 'MYNEWT_VAL(MAX3107_0_SYSINIT_STAGE)'
```
vrahane, line 27: this function is not going to be any different for different instances. So, I would just name it `max3107_init` and also the syscfg should be named `MAX3107_SYSINIT_STAGE`

hw/drivers/uart/max3107/max3107_0/syscfg.yml
```diff
@@ -0,0 +1,125 @@
...
+    MAX3107_0_IRQ_PIN:
+        description: >
+            MAX3107 interrupt pin.
+        value: -1
+
+    MAX3107_0_LDOEN_PIN:
+        description: >
+            MAX3107 LDEON pin.
+        value: -1
+
```
vrahane, line 121: All above instances should not be _0_ I think, it's best to keep only one instance.

hw/drivers/uart/max3107/max3107_0/pkg.yml
kasjer: I disagree

hw/drivers/uart/max3107/max3107_0/pkg.yml
kasjer: I disagree

hw/drivers/uart/max3107/max3107_0/syscfg.yml
kasjer: I disagree

vrahane (approved):
LGTM. Did not realize it was mostly re-organization.

PR #2616, opened 2021-06-14
i2s fixes

kasjer:
Two fixes related to I2S.

- NRF5x versions would assert when i2s_open and i2s_close were issued without transferring any data
- i2s_write did not fill correctly sample count field in i2s_buffer


vrahane (approved):
LGTM

PR #2617, opened 2021-06-14
pic32: Fix tick timer handling

kasjer:
Tick timer interrupt handler assumed it was execute before
next tick interrupt should be fired.
In case when interrupt was handled late (due to other interrupts
or just blocked interrupt for some time) next computed tick time
was already in the past and system time stopped for a very long time.

Now COMPARE register when set is compared to current COUNT register
and if it is in the past next tick time is computed until interrupt
is schedule to be ahead of COUNT register.

Also os_time_advance() takes into account what is the difference of
COUNT and COMPARE registers and in case of late interrupt, g_os_time
can advance more then one tick at a time.


vrahane (approved):
LGTM

PR #2618, opened 2021-06-16
sbrk unification

kasjer:
A lot of copies of **sbrk()** function was kept around.
They differ slightly but mostly it was copy/paste work.

This PR tries to make sbrk same for all platforms.

- most bsp were stripped of own copy of _sbrk.c_ file
- **sbrk()** was placed in _hal/src/sbrk.c_
- STM32 based BSP need additional call to **_sbrkInit()** in startup files

This change is initiated by attempt to change how crash dumps are
stored in flash. It will lead to dump only used memory (RAM that was not
assigned to HEAP does not need to be saved).

### TODO:
ARC and Microchip were not touched yet.
There are some BSP that had yet another way to define sbrk and
they will be dealt later if this PR is agreed upon.


hw/hal/syscfg.yml
```diff
@@ -42,7 +42,11 @@ syscfg.defs:
             If set to 0 software breakpoints placed with HAL_DEBUG_BREAK macro will not
             be executed.
         value: 1
-
+    HAL_SBRK:
+        description: >
+            If set HAL provides standard implementation of _sbrk function.
+            It also provides _sbrkInit function that sets up heap space for malloc.
+        value: 0
```
utzig, line 49: Wouldn't it be better to make this default to `1` and leave the few HALs that must provide their own implementation reset it?

kasjer: done

andrzej-kaczmarek (approved):
lgtm

utzig (approved):
LGTM

PR #2620, opened 2021-06-17
i2s: Assure open fails in case of driver error

kasjer:
When device is not configured correctly i2s_start for
I2S output device may fail.
i2s_start() returning I2S_ERR_NO_BUFFER is rather expected
while other errors may indicate some serious configuration
problem.

With this change failing i2s_start will stop user code form
openning device since it will probably not work anyway.


vrahane (approved):
LGTM

PR #2622, opened 2021-06-21
hw/mcu/nordic: use correct struct type for spi2 and spi3 bus in nrf52 periph

vikrant-proxy:
`bus_spi_hal_dev_create` needs struct `bus_spi_hal_dev` when using bus driver.


vrahane (approved):
LGTM, wonder why these were different.

PR #2624, opened 2021-06-23
hw/mcu/nrf5340_net: Remove enable/disable from hal_spi_init

ncasaril:
This applies when using bus driver and SPIM0 for the networking core, and SPIM0 is initialised from hal_spi_init_hw. 
During ```hal_spi_init``` it called ```hal_spi_enable``` which returs an error because the txrx_cb isn't set.

This change makes hal_spi_init similar on networking core to application core where no extra disabling or enabling occurs. 

Without this change the bus device isn't properly initialised and causes a segfault as soon as you try to use the spi.


vrahane (approved):
The change does make sense. LGTM.

sjanc (approved):

PR #2627, opened 2021-06-23
ipc nrf5340 flash

kasjer:
This adds driver to read and write flash across core domains.
APP core can for example read net core crash dump flash area.

To use this setup bsp.yml file have to have flash area from flash device 2.

This changes priority of syscfg init for flash_map to be executed after IPC.
IPC flash driver uses 2 IPC channels so **IPC_NRF5340_CHANNELS** must
specify value high enough to accommodate those two channels.
By default channels 2 and 3 are set but they can be change in syscfg.
Core that wants to expose its flash to other core does it by adding
```
IPC_NRF5340_FLASH_SERVER: 1
```
to its syscfg values, client core specifies:
```
IPC_NRF5340_FLASH_CLIENT: 1
```

Read and write sized are currently limited to **IPC_NRF5340_BUF_SZ - 16**


vrahane:
@kasjer Build is failing, can you please check what's up ?

vrahane:
@kasjer It's failing because of the following:
```
+newt target create -s test-blinky-nordic_pca10095
+newt target set -s test-blinky-nordic_pca10095 bsp=@apache-mynewt-core/hw/bsp/nordic_pca10095
+newt target set -s test-blinky-nordic_pca10095 app=apps/blinky
+newt build -q test-blinky-nordic_pca10095
Error: hw/drivers/flash/ipc_nrf5340_flash/src/ipc_nrf5340_flash.c:26:10: fatal error: ipc_nrf5340/ipc_nrf5340.h: No such file or directory
 #include <ipc_nrf5340/ipc_nrf5340.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
+rc=1
+[[ 1 -ne 0 ]]
+EXIT_CODE=1
+newt target delete -s -f test-blinky-nordic_pca10095
+for bsp in ${TARGETS}
+[[  ci40 embarc_emsk hifive1 native-armv7 native-mips pic32mx470_6lp_clicker pic32mz2048_wi-fire sensorhub native dialog_cmac  =~ [[:blank:]]nordic_pca10095_net[[:blank:]] ]]
+echo 'Testing bsp=nordic_pca10095_net'
Testing bsp=nordic_pca10095_net
+target=test-blinky-nordic_pca10095_net
+newt target delete -s -f test-blinky-nordic_pca10095_net
+newt target create -s test-blinky-nordic_pca10095_net
+newt target set -s test-blinky-nordic_pca10095_net bsp=@apache-mynewt-core/hw/bsp/nordic_pca10095_net
+newt target set -s test-blinky-nordic_pca10095_net app=apps/blinky
+newt build -q test-blinky-nordic_pca10095_net
Error: hw/drivers/flash/ipc_nrf5340_flash/src/ipc_nrf5340_flash.c:26:10: fatal error: ipc_nrf5340/ipc_nrf5340.h: No such file or directory
 #include <ipc_nrf5340/ipc_nrf5340.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

hw/drivers/flash/ipc_nrf5340_flash/src/ipc_nrf5340_flash.c
```diff
@@ -0,0 +1,580 @@
...
+static struct os_event erase_event = {
+    .ev_cb = flash_erase_cb,
+};
+
+static void
+nrf5340_ipc_flash_server_erase(struct ipc_msg *req)
+{
+    nrf5340_ipc_flash_server_read(&req->param, sizeof(uint32_t));
+    erase_event.ev_arg = req;
+    os_eventq_put(os_eventq_dflt_get(), &erase_event);
```
vrahane, line 521: So, I think it would be better to make this eventq configurable. We do not want this driver to work on the main eventq always.

vrahane:
So, I think this functionality is needed for sure but instead of making it blocking based on a semaphore and mutex, it needs to be made non-blocking. Neither the netcore, nor the appcore should wait for each others response. Instead they should continue with other stuff and maybe call an optional callback when the operation is done. Since we are making this driver generic enough, it is important that we make it superfluous.

If this driver was only specific to coredumps it would make sense to make it blocking.

kasjer:
@vrahane This driver adds way to read/write/erase flash in same way as any other flash provided by any MCU or SPIFLASH driver.
All hal flash interface functions are blocking in mynewt so far.
You may try to introduce asynchronous version for such access, I'm looking forward to review such work since we really need this non-blocking variant.
I see that current flash access mynewt has, can only be used for crash dumps, and with non-blocking functions maybe we can use flash even for FCB or NFFS or firmware upgrade.

sjanc (approved):

hw/bsp/nordic_pca10095_net/src/hal_bsp.c
```diff
@@ -48,13 +48,18 @@ hal_bsp_flash_dev(uint8_t id)
...
     if (id == 0) {
-        return &nrf5340_net_flash_dev;
+        return &nrf5340_flash_dev;
     }
 #if MCUBOOT_MYNEWT
     if (id == 1) {
         return &nrf5340_net_vflash_dev.nv_flash;
     }
 #endif
+#if MYNEWT_VAL(IPC_NRF5340_FLASH_CLIENT)
```
vikrant-proxy, line 58: should this have been `IPC_NRF5340_FLASH_SERVER` given appcore is the client?

PR #2628, opened 2021-06-23
hw/ipc_nrf5340: Fix enabling GPIO passing to network core

sjanc:
Due to way MYNEWT_VAL is defined (always within ()) some preprocesor
magic is required when passing comma separated values.


ncasaril (commented):
Works as intended but perhaps #2629 is a less complicated version of the same?

PR #2629, opened 2021-06-24
hw/ipc_nrf5340: Fix enabling multiple GPIOs passed to network core 

ncasaril:
Due to the way MYNEWT_VAL is defined (always within ()) some preprocesor
magic is required when passing comma separated values.

Heavily Inspired by #2628 but using preprocessor magic to initialise the original vector instead.
Tested with 0, 1, and multiple pins redirected to networking core.

### DIFF
--- a/hw/drivers/ipc_nrf5340/src/ipc_nrf5340.c
+++ b/hw/drivers/ipc_nrf5340/src/ipc_nrf5340.c
@@ -162,7 +162,7 @@ ipc_nrf5340_init(void)
 
 #if MYNEWT_VAL(MCU_APP_CORE)
 #if MYNEWT_VAL(IPC_NRF5340_NET_GPIO)
-    unsigned int gpios[] = { MYNEWT_VAL(IPC_NRF5340_NET_GPIO) };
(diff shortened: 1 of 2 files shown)


sjanc (approved):

vrahane (approved):

vrahane:
LGTM.

andrzej-kaczmarek:
do we really want to relay on code injection and macro magic to handle this? why not just define bitmask for pins?

sjanc:
yeah, bitmap would be cleaner solution

ncasaril:
Now using a bitmask instead of the preprocessor magic. Also updated syscfg comment to match.

vrahane (approved):

sjanc:
although this is already merged, after looking at how user needs to define it now... I think the previous macro-magic was nicer solution...

ncasaril:
@sjanc, I don't think there are that many people using this yet so we could still go back to the magic if everyone agrees?

sjanc:
@ncasaril yeap, I'd go with your original solution.

In long term we can think about extending newt tool to have support for 'set' or 'list' type of sysvals...

ncasaril:
New PR created where the original, non bitmask, solution is used: #2637

PR #2630, opened 2021-06-24
hw/ipc_nrf5340: Assert on ipc_nrf5340_recv if channel is already in use

sjanc:
Calling ipc_nrf5340_recv on channel which is already enabled is always
an application error which leads to confusing behaviour (or crashes).

If application want to change receive handler for specified channel it
shall first unregister old one by passing NULL pointer.


vrahane (approved):
LGTM

PR #2631, opened 2021-06-24
hw/bsp/nordic_pca10095: Add reduced size crash dumps

kasjer:
So far crash dumps had whole RAM image stored in flash.
If flash area assigned to coredumps were to small, end
of RAM was not dumped (including interrupt stack).

This adds possibility to skip dumping unused heap space.
This can reduce size of core dumps.


vrahane (approved):
This looks fine to me. Definitely will help a lot.

PR #2632, opened 2021-06-28
nrf5340/i2c: Fix STOP event handling

kasjer:
After read or write that ended with error (address nack)
interrupts are blocked and stop condition is generated
that will rise EVENTS_STOP later on.

Read or write finishes with this event active.

On next write or read interrupt were enabled before EVENTS_STOP
was cleared resulting in interrupt handling that would signal
semaphore.
In some cases that may lead to data corruption when code reuses buffer
while I2C transaction is ongoing.

This changes clears old events before enabling interrupts.


sjanc (approved):

vrahane (approved):
LGTM

vrahane:
@vikrant-proxy Tired it out. looks good.

PR #2637, opened 2021-06-29
hw/ipc_nrf5340: (Re-)Fix enabling GPIO passing to network core

ncasaril:
The outcome of #2629 used a bitmask to set which GPIOs the application core handed over to the networking core. 
This made for neat code but complicated the syscfg handling. 

This PR undoes #2629, the bitmask change, and uses preprocessor magic to accomplish the same with a simple comma separated syscfg value for specifying pins.  

Feedback welcome again, and this time I'll hang out for consensus before merging.


vrahane:
I am fine with both approaches, the macro approach preferred. Thank you for pushing this again.

vrahane (approved):

sjanc (approved):

PR #2639, opened 2021-07-06
hw/bsp/dwm1001-dev Add LIS2DH12 accelerometer support

seijikun:
Hey guys,

thanks for this awesome piece of software!
This PR is an attempt at adding support for the LIS2DH12 accelerometer sensor, which is integrated with [DWM1001-DEV](https://www.decawave.com/dwm1001dev/datasheet/) boards.
I'm fairly new to mynewt, so this PR is more like a shot in the dark, and mostly copied from other BSPs with the same accelerometer.
The `sensor_itf.si_ints` values are chosen, such that it does not crash - which does not mean that it's working ;). Would be good if somebody with more hardware knowledge could take a look at them (the `.active = true` certainly isn't true, this should be something like `HAL_GPIO_TRIG_RISING` or `HAL_GPIO_TRIG_FALLING`).

Here is an example application:
```c
#include <assert.h>
#include <string.h>
#include <stdio.h>
#include <math.h>
#include "sysinit/sysinit.h"
#include "os/os.h"
#include "bsp/bsp.h"
#include "hal/hal_gpio.h"
#include "hal/hal_bsp.h"
#include "hal/hal_i2c.h"
#include "mcu/nrf52_hal.h"
#ifdef ARCH_sim
#include "mcu/mcu_sim.h"
#endif

#include <config/config.h>

#include <sensor/sensor.h>
#include "sensor/accel.h"
#include <lis2dh12/lis2dh12.h>

#if MYNEWT_VAL(DW1000_DEVICE_0)
#include <dw1000/dw1000_hal.h>
#endif

int accel_data_cb(struct sensor* sensor, void* a, void* databuf, sensor_type_t sensorType) {
    if(!databuf) { return SYS_EINVAL; }
    struct sensor_accel_data* sad = (struct sensor_accel_data*) databuf;
    if (!sad->sad_x_is_valid || !sad->sad_y_is_valid || !sad->sad_z_is_valid) { return SYS_EINVAL; }
    printf("%s: [ secs: %ld usecs: %d cputime: %u ]\n",
                   "LISTENER_CB",
                   (long int)sensor->s_sts.st_ostv.tv_sec,
                   (int)sensor->s_sts.st_ostv.tv_usec,
                   (unsigned int)sensor->s_sts.st_cputime);

    printf("x = %f, y = %f, z = %f\n", sad->sad_x, sad->sad_y, sad->sad_z);
    return 0;
}

static struct sensor_listener sensor_listener = {
    .sl_sensor_type = SENSOR_TYPE_ACCELEROMETER,
    .sl_func = accel_data_cb,  
    .sl_arg = NULL
};

 

int main(int argc, char **argv){
    sysinit();

    struct sensor* accel = sensor_mgr_find_next_bydevname("lis2dh12_0", NULL);
    assert(accel != NULL);

    assert(0 == lis2dh12_set_full_scale(&accel->s_itf, LIS2DH12_FS_2G));
    assert(0 == lis2dh12_set_op_mode(&accel->s_itf, LIS2DH12_OM_HIGH_RESOLUTION));

    sensor_set_poll_rate_ms(accel->s_dev->od_name, 1000 / 50);
    sensor_register_listener(accel, &sensor_listener);

    while (1) {
        os_eventq_run(os_eventq_dflt_get());
    }
    assert(0);
    return 0;
}
```

### DIFF
--- a/hw/bsp/dwm1001-dev/pkg.yml
+++ b/hw/bsp/dwm1001-dev/pkg.yml
@@ -41,3 +41,9 @@ pkg.deps.SOFT_PWM:
 
 pkg.deps.UARTBB_0:
     - "@apache-mynewt-core/hw/drivers/uart/uart_bitbang"
+
+pkg.deps.LIS2DH12_ONB:
+    - "@apache-mynewt-core/hw/drivers/sensors/lis2dh12"
+    
+pkg.init:
+    config_lis2dh12_sensor: 'MYNEWT_VAL(DWM1001_DEV_SYSINIT_STAGE_LIS2DH12)'
--- a/hw/bsp/dwm1001-dev/src/hal_bsp.c
+++ b/hw/bsp/dwm1001-dev/src/hal_bsp.c
@@ -33,6 +33,31 @@
 #include "mcu/nrf52_periph.h"
 #include "bsp/bsp.h"
 
+#if MYNEWT_VAL(LIS2DH12_ONB)
+#include "lis2dh12/lis2dh12.h"
+static struct lis2dh12 lis2dh12;
+#endif
+
+#if MYNEWT_VAL(LIS2DH12_ONB)
+static struct sensor_itf i2c_0_itf_lis = {
+    .si_type = SENSOR_ITF_I2C,
+    .si_num  = 0,
+    .si_addr = 0x19,
+    .si_ints = {
+        { /* TODO: determine correct values */
+            .host_pin = 25,
+            .device_pin = 0,
+            .active = true
+        },
+        { /* TODO: determine correct values */
+            .host_pin = 25,
+            .device_pin = 0,
+            .active = true
+        }
+    }
+};
+#endif
+
 /*
  * What memory to include in coredump.
  */
@@ -93,6 +118,47 @@ hal_bsp_get_nvic_priority(int irq_num, uint32_t pri)
     return cfg_pri;
 }
 
+int
+config_lis2dh12_sensor(void)
+{
+#if MYNEWT_VAL(LIS2DH12_ONB)
+    int rc;
+    struct os_dev *dev;
+    struct lis2dh12_cfg cfg;
+
+    dev = (struct os_dev *) os_dev_open("lis2dh12_0", OS_TIMEOUT_NEVER, NULL);
+    assert(dev != NULL);
+
+
+    memset(&cfg, 0, sizeof(cfg));
+
+    cfg.lc_s_mask = SENSOR_TYPE_ACCELEROMETER;
+    cfg.lc_rate = LIS2DH12_DATA_RATE_HN_1344HZ_L_5376HZ;
+    cfg.lc_fs = LIS2DH12_FS_2G;
+    cfg.lc_pull_up_disc = 1;
+
+    rc = lis2dh12_config((struct lis2dh12 *)dev, &cfg);
+    SYSINIT_PANIC_ASSERT(rc == 0);
+
+    os_dev_close(dev);
+#endif
+    return 0;
+}
+
+static void
+sensor_dev_create(void)
+{
+    int rc;
+    (void)rc;
+
+#if MYNEWT_VAL(LIS2DH12_ONB)
+    rc = os_dev_create((struct os_dev *) &lis2dh12, "lis2dh12_0",
+                       OS_DEV_INIT_PRIMARY, 0, lis2dh12_init,
+                       (void *)&i2c_0_itf_lis);
+    assert(rc == 0);
+#endif
+}
+
 void
 hal_bsp_init(void)
 {
@@ -101,6 +167,8 @@ hal_bsp_init(void)
 
     /* Create all available nRF52840 peripherals */
     nrf52_periph_create();
+
+    sensor_dev_create();
 }
 
 void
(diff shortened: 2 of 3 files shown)


mlaz:
Sounds like you need to look at the module's (DWM1001, not DWM1001-dev) [schematic](https://www.qorvo.com/products/d/da007971) there you should find all the pinout info.

hw/bsp/dwm1001-dev/src/hal_bsp.c
```diff
@@ -93,6 +118,47 @@ hal_bsp_get_nvic_priority(int irq_num, uint32_t pri)
...
+config_lis2dh12_sensor(void)
+{
+#if MYNEWT_VAL(LIS2DH12_ONB)
+    int rc;
+    struct os_dev *dev;
+    struct lis2dh12_cfg cfg;
+
+    dev = (struct os_dev *) os_dev_open("lis2dh12_0", OS_TIMEOUT_NEVER, NULL);
+    assert(dev != NULL);
+
```
kasjer, line 131: extra empty line

hw/bsp/dwm1001-dev/src/hal_bsp.c
```diff
@@ -93,6 +118,47 @@ hal_bsp_get_nvic_priority(int irq_num, uint32_t pri)
...
 
+int
+config_lis2dh12_sensor(void)
+{
+#if MYNEWT_VAL(LIS2DH12_ONB)
+    int rc;
+    struct os_dev *dev;
+    struct lis2dh12_cfg cfg;
+
+    dev = (struct os_dev *) os_dev_open("lis2dh12_0", OS_TIMEOUT_NEVER, NULL);
```
kasjer, line 129: in mynewt code it is custom not to put space after finishing ) in casts like this
```c
dev = (struct os_dev *)os_dev_open(....
```

hw/bsp/dwm1001-dev/src/hal_bsp.c
```diff
@@ -93,6 +118,47 @@ hal_bsp_get_nvic_priority(int irq_num, uint32_t pri)
...
+}
+
+static void
+sensor_dev_create(void)
+{
+    int rc;
+    (void)rc;
+
+#if MYNEWT_VAL(LIS2DH12_ONB)
+    rc = os_dev_create((struct os_dev *) &lis2dh12, "lis2dh12_0",
```
kasjer, line 155: extra space after cast expression

hw/bsp/dwm1001-dev/syscfg.yml
```diff
@@ -62,3 +71,6 @@ syscfg.vals.BLE_CONTROLLER:
     OS_CPUTIME_FREQ: 32768
     OS_CPUTIME_TIMER_NUM: 5
     BLE_LL_RFMGMT_ENABLE_TIME: 1500
+
+syscfg.restrictions:
+    - "LIS2DH12_ONB && I2C_0_PIN_SCL == 28 && I2C_0_PIN_SDA == 29"
```
kasjer, line 76: if it was:
```yml
syscfg.restrictions:
    - "!LIS2DH12_ONB || (I2C_0 == 1 && I2C_0_PIN_SCL == 28 && I2C_0_PIN_SDA == 29)"
```
sensor could be turned on and off as syscfg value implies it could.
notice extra `I2C_0 == 1` that is also needed I guess

hw/bsp/dwm1001-dev/src/hal_bsp.c
```diff
@@ -33,6 +33,31 @@
...
+        { /* TODO: determine correct values */
+            .host_pin = 25,
+            .device_pin = 0,
+            .active = true
+        },
+        { /* TODO: determine correct values */
+            .host_pin = 25,
+            .device_pin = 0,
+            .active = true
+        }
```
kasjer, line 56: I would remove this second table entry, first one seems to be correct

hw/bsp/dwm1001-dev/pkg.yml
```diff
@@ -41,3 +41,9 @@ pkg.deps.SOFT_PWM:
 
 pkg.deps.UARTBB_0:
     - "@apache-mynewt-core/hw/drivers/uart/uart_bitbang"
+
+pkg.deps.LIS2DH12_ONB:
+    - "@apache-mynewt-core/hw/drivers/sensors/lis2dh12"
+    
+pkg.init:
```
kasjer, line 48: if it was changed to:
```yml
pkg.init.LIS2DH12_ONB:
```
whole config_lis2dh12_sensor()  function could be ifdef'ed-out not just its body

kasjer (changes requested):
Minor style (extra spaces and line) changes would be nice.
restriction should be changed to allow turning sensor on and off.

hw/bsp/dwm1001-dev/pkg.yml
kasjer: Fist line of commit message would also be more consistent with what is used in mynewt if it looked like text that was chosen for PR title:
```
hw/bsp/dwm1001-dev: Add LIS2DH12 accelerometer support
```
instead of "Add LIS2DH12 accelerometer support for DWM1001-dev boards"

seijikun:
Sorry, I don't have access to the hardware anymore, so I can't continue on this with testing.
I tried having only one entry in the interrupts array, that crashed on initialization.
I'm going to close this. Sorry again!

PR #2641, opened 2021-07-13
spi_hal: Pull CS line back up if write / read_write operations fail

vikrant-proxy:


vrahane (approved):
This does make sense, without it the CS line will never get de-asserted if an error occurs while writing or reading.

PR #2643, opened 2021-07-26
hw/mcu/nrf52: Accept NULL txbuf in hal_spi_txrx by transmitting 0xFF

ncasaril:
The new bus_spi_write_read function in the bus driver doesn't handle write_reads with a combined length of more than 16 without passing NULL as the txbuf parameter:

https://github.com/apache/mynewt-core/blob/595a074aec88de2dc50688990debedb3c228e89c/hw/bus/drivers/spi_hal/src/spi_hal.c#L242

This change allows nrf52:hal_spi_txrx to handle a NULL txbuf.

### DIFF
--- a/hw/mcu/nordic/nrf52xxx/src/hal_spi.c
+++ b/hw/mcu/nordic/nrf52xxx/src/hal_spi.c
@@ -996,7 +996,7 @@ err:
  *
  *     MASTER: master sends all the values in the buffer and stores the
  *             stores the values in the receive buffer if rxbuf is not NULL.
- *             The txbuf parameter cannot be NULL.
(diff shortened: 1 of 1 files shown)


kasjer:
@ncasaril I'm not opposed to your change, but accepting it would make one MCU implementation ready to work with bus_node_write_read_transact while others would still fail.
We could change them all of course but maybe this solution #2644 would work  for you instead.
Could you please check it it solves the issue (I will verify that it works if you don't have time but it will take some time).
Thank you for your involvement.

ncasaril:
Closing as #2644 solves this problem for all mcus whilst my fix only addresses nrf52.

PR #2644, opened 2021-07-26
hw/bus/spi_hal: Fix bus_spi_write_read

kasjer:
Function bus_spi_write_read() was calling hal_spi_txrx()
with NULL for txbuf for transactions involving more then 16 bytes.
NULL pointer is not valid for this function.
When code was build with SPI_HAL_USE_NOBLOCK enabled, zeroed receive buffer
was already used as argument for transmit buffer.
Same approach is now used when SPI_HAL_USE_NOBLOCK is not set.

### DIFF
--- a/hw/bus/drivers/spi_hal/src/spi_hal.c
+++ b/hw/bus/drivers/spi_hal/src/spi_hal.c
@@ -212,7 +212,7 @@ bus_spi_write_read(struct bus_dev *bdev, struct bus_node *bnode,
 #if MYNEWT_VAL(SPI_HAL_USE_NOBLOCK)
     if (wlength + rlength <= sizeof(buf)) {
         memcpy(buf, wbuf, wlength);
-        memset(buf + wlength, 0, rlength);
+        memset(buf + wlength, 0xFF, rlength);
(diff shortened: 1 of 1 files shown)


ncasaril (commented):
To be consistent with:

https://github.com/apache/mynewt-core/blob/ca3280bcd39b1dc44df307db27f243c9a3b39ab7/hw/bus/drivers/spi_hal/src/spi_hal.c#L145

the memset should probably be setting it to 0xFF instead of 0?

kasjer:
> To be consistent with:
> 
> https://github.com/apache/mynewt-core/blob/ca3280bcd39b1dc44df307db27f243c9a3b39ab7/hw/bus/drivers/spi_hal/src/spi_hal.c#L145
> 
> the memset should probably be setting it to 0xFF instead of 0?

@ncasaril thanks for pointing out inconsistency

ncasaril:
This will work for me and as you pointed out it has fewer repercussions than having to update the hal_spi_txrx for all mcus. I'll withdraw PR #2643 in favour of this one. Thanks @kasjer!

ncasaril:
I've now tested this change on nrf52 working with large reads / writes to a spi device (dw1000).

vrahane (approved):
LGTM

PR #2646, opened 2021-07-26
sys/console: Stop history auto search when echo is off

kasjer:
When echo is off history auto search should not output anything.
This allows to have **newtmgr** upload working while code
is build with history auto search support.


vrahane (approved):
LGTM

PR #2647, opened 2021-07-26
hw/mcu/nordic/nrf5340: Update netcore build script syscfg names

vikrant-proxy:
Update build script syscfg names to reflect nRF5340 netcore version and signing key


vrahane (approved):
LGTM

PR #2648, opened 2021-07-29
sys/console/rtt_console: Fix CONSOLE_INPUT guards in RTT console

vikrant-proxy:
Move console_rx_restart within CONSOLE_INPUT guards to compile correctly with CONSOLE_INPUT turned off


vrahane (approved):
LGTM

PR #2649, opened 2021-07-29
sys/console/minimal: Fix RTT console compiler error for minimal console

vikrant-proxy:
Compile `console_out` for UART console only. RTT console has it's own implementation of `console_out` which conflicts if RTT_CONSOLE is used with console/minimal

### DIFF
--- a/sys/console/minimal/src/console.c
+++ b/sys/console/minimal/src/console.c
@@ -134,6 +134,7 @@ end:
     return rc;
 }
 
+#if MYNEWT_VAL(CONSOLE_UART)
 int
 console_out(int c)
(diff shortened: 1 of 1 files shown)


vrahane (approved):
LGTM

kasjer:
This is not correct way to do this.
rtt console should have function name change from **console_out** to **console_out_nolock**.
This was done for all other cases 2 years back and it looks like rtt minimal console was left out.

vikrant-proxy:
Thanks @kasjer ! Can you take a look at https://github.com/apache/mynewt-core/pull/2656

PR #2650, opened 2021-07-30
os/cortex_m: Fix semaphore release hazard

kasjer:
Context switch is performed in low priority interrupt.
If higher priority interrupt code releases semaphore
BEFORE g_current_task is updated in context switch,
os_sem_release() function can see that task
that entered os_sem_pend() and initiated context switch
is still running and there is no need to switch tasks.
As a result task is marked as ready.
Then high priority interrupt ends and context switch is
finished leaving lower priority task running.
When this happens and current task is idle it may take
a while till reschedule is done again leaving task that
should be running in ready state.

To prevent this interrupts are now masked while context switch
code check current task against ready list.
Interrupts are enabled again once g_current_task is updated, rest of
context switch may be performed (as it was) with interrupts enabled.

### DIFF
--- a/kernel/os/src/arch/cortex_m0/m0/HAL_CM0.s
+++ b/kernel/os/src/arch/cortex_m0/m0/HAL_CM0.s
@@ -155,14 +155,19 @@ PendSV_Handler:
         .fnstart
         .cantunwind
 
-        LDR     R3,=g_os_run_list   /* Get highest priority task ready to run */
-        LDR     R2,[R3]             /* Store in R2 */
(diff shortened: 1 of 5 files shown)


andrzej-kaczmarek (approved):

kasjer:
alternative solution #2651

mkiiskila:
LGTM

mkiiskila (approved):

vrahane (approved):
Tried this out as well, works fine so far, going with majority.

kasjer:
alternative solution was accepted

PR #2651, opened 2021-07-30
kernel/os: Fix semaphore release hazard

kasjer:
Context switch is performed in low priority interrupt.
If higher priority interrupt code releases semaphore
BEFORE g_current_task is updated in context switch,
os_sem_release() function can see that task
that entered os_sem_pend() and initiated context switch
is still running and there is no need to switch tasks.
As a result task is marked as ready.
Then high priority interrupt ends and context switch is
finished leaving lower priority task running.
When this happens and current task is idle it may take
a while till reschedule is done again leaving task that
should be running in ready state.

To prevent this, condition to reschedule is now changed
so that if current task is same as task that should be woken
context switch is also initiated just in case switching
already started and g_current_task was not updated.

### DIFF
--- a/kernel/os/src/os_sem.c
+++ b/kernel/os/src/os_sem.c
@@ -88,8 +88,16 @@ os_sem_release(struct os_sem *sem)
         rdy->t_flags &= ~OS_TASK_FLAG_SEM_WAIT;
         os_sched_wakeup(rdy);
 
-        /* Schedule if waiting task higher priority */
-        if (current->t_prio > rdy->t_prio) {
+        /*
+         * Schedule if waiting task higher priority.
+         * current->t_prio == rdy->t_prio means that current task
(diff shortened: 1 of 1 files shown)


kasjer:
this is alternative solution to #2650

vrahane:
I did try out this one and seems to be working fine.

andrzej-kaczmarek:
I prefer this one since it's simpler, fixes problem for all archs and does not have overhead on context switch.

andrzej-kaczmarek (approved):

vrahane (approved):

vrahane:
While both solutions work fine. I prefer this one.

mlaz (approved):
LGTM

wes3 (commented):
LGTM

PR #2653, opened 2021-08-03
nrf5340: Set default HCI transport and flow control

sjanc:
Using nrf5340 IPC HCI transport requires enabled controller to host
flow control for proper operations. Enabling it by default make it
easier for users.


vrahane (approved):
LGTM

PR #2659, opened 2021-08-11
Update da1469x vbus

kasjer:
VBUS changes were handled locally by charger code.
USB code also needs to know about such events.

Now common VBUS code handles interrupt and inform
charger and/or USB module about such events

This requires: https://github.com/hathach/tinyusb/pull/1094 to be merged first for **auto** and **ignore** modes (selected by **DA1469X_USB_VBUS_HANDLING** syscfg).
Older TinyUSB can be used only with **custom** mode as there may be missing function in TinyUSB driver.


michal-narajowski:
@kasjer Please rebase this branch on the latest master and it should fix the build issue on osx.

andrzej-kaczmarek (approved):

PR #2661, opened 2021-08-14
stm32f4 stm32f7 vbus detection fixes

kasjer:
- STM32F7xx MCUs differs in some USB registers from F4.
 VBUS detection was not working for this reason on F7
- USB ID pin is not configurable
- STM32F4 VBUS detection code incorrectly configured PA9 and set GCCFG
  register values when VBUS detection was enabled


utzig (approved):
LGTM

PR #2664, opened 2021-08-18
Updates to the kinetis TRNG driver

utzig:
* trng: kinetis: Use statically allocated stack

Change TRNG poller task to use a statically allocated stack; allocating this way uses the same mechanism as other Mynewt standard tasks, and avoids the requirement of a heap.

* trng: kinetis: update TRNG to be interrupt driven

When TRNG HW (not RNGA) is present, it can be used in an interrupt driven manner, where interrupts are generated for data available or on errors. With this commit the poller task is updated so it only runs when woken by an interrupt instead of polling constantly to fill in the cache. This fixes issues with task starvation, since the poller task is very high priority and the TRNG is very slow in the standard configuration.

A new syscfg was added to allow faster generation of entropy by changing the value of `KINETIS_TRNG_ENTROPY_DELAY`.

This commit also refactors the TRNG queue to be simpler and better abstracted.

PS: the commit does not touch the functionality of RNGA based HW (eg K64F) and was tested on both FRDM-K64F and FRDM-K82F.

### DIFF
--- a/hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
+++ b/hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
@@ -35,6 +35,22 @@
     do {                                                              \
         (base)->MCTL |= (TRNG_MCTL_PRGM_MASK | TRNG_MCTL_ERR_MASK);   \
     } while (0)
+#define TRNG_CLEAR_AND_ENABLE_INTS(base)                              \
+    do {                                                              \
+        (base)->INT_CTRL &= ~(TRNG_INT_CTRL_HW_ERR_MASK |             \
+                              TRNG_INT_CTRL_ENT_VAL_MASK);            \
+        (base)->INT_MASK |= TRNG_INT_MASK_HW_ERR_MASK |               \
+                            TRNG_INT_MASK_ENT_VAL_MASK;               \
+    } while (0)
+#define TRNG_CLEAR_AND_ENABLE_ENTROPY_INT(base)                       \
+    do {                                                              \
+        (base)->INT_CTRL &= ~TRNG_INT_CTRL_ENT_VAL_MASK;              \
+        (base)->INT_MASK |= TRNG_INT_MASK_ENT_VAL_MASK;               \
+    } while (0)
+#define TRNG_DISABLE_ENTROPY_INT(base)                                \
+    do {                                                              \
+        (base)->INT_MASK &= ~TRNG_INT_MASK_ENT_VAL_MASK;              \
+    } while (0)
 #else
 #error "Unsupported TRNG interface"
 #endif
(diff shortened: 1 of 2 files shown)


hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
```diff
@@ -35,88 +35,138 @@
...
-static uint16_t rng_cache_in;
-static struct os_mutex rng_cache_mu;
-static os_stack_t *pstack;
+#define TRNG_CACHE_LEN MYNEWT_VAL(KINETIS_TRNG_CACHE_LEN)
+static struct {
+    uint8_t cache[TRNG_CACHE_LEN];
+    uint16_t out;
+    uint16_t in;
+    uint16_t used;
+    struct os_mutex mu;
```
kasjer, line 67: This order of fields with cache being first is prone to leave some padding before **out** and before **mu**

hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
```diff
@@ -35,88 +35,138 @@
...
-   running = false;
 }
 
 static size_t
 kinetis_trng_read(struct trng_dev *trng, void *ptr, size_t size)
 {
     size_t num_read;
+    uint8_t *u8p;
+
+    CACHE_LOCK();
```
kasjer, line 154: CACHE_LOCK could be after next two initialization lines

hw/drivers/trng/trng_kinetis/syscfg.yml
```diff
@@ -30,3 +30,10 @@ syscfg.defs:
         value: 0
         restrictions:
             - '!KINETIS_TRNG_USE_RNGA'
+    KINETIS_TRNG_ENTROPY_DELAY:
+        description: >
+            The entropy delay can be used to improve the speed of number
+            generation with the compromise of worse entropy. 3200 is the
+            default value, but it can be reduced for faster value
+            generation.
```
kasjer, line 38: I have not read documentation of TRNG block. Maybe description could specify what the units are.

kasjer (approved):
LGTM,

Strange that we missed local variable in kinetis_trng_start() way back.

hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
utzig: I moved it to the end.

hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
utzig: Fixed.

hw/drivers/trng/trng_kinetis/syscfg.yml
utzig: I rephrased it including the "units".

hw/drivers/trng/trng_kinetis/src/trng_kinetis.c
kasjer: You could check if now you don't have padding before mu.

utzig:
> Strange that we missed local variable in kinetis_trng_start() way back.

We did not, the diff just looks weird, but the variable is in the function above that one.

kasjer:
> > Strange that we missed local variable in kinetis_trng_start() way back.
> 
> We did not, the diff just looks weird, but the variable is in the function above that one.

I meant that code:
```c
static inline void
kinetis_trng_start(void)
{
    struct os_event evt;
...
    evt.ev_queued = 0;
    evt.ev_arg = NULL;
    (void)os_eventq_put(&rng_evtq, &evt);
}
```
...put **evt** from stack into the queue **rng_evtq** and get away with it

PR #2665, opened 2021-08-18
Update MCUboot repo and branch

utzig:
MCUboot was moved to mcu-tools org, and now uses "main" instead of "master" branch.


vrahane (approved):
LGTM

PR #2667, opened 2021-08-20
[DNM] hw: mcu: kinetis: Add initial hal_timer support

utzig:
Implement initial hal_timer interface using the LPTMR modules in the Kinetis MCU line.

WIP: this was tested on the FRDM-K82F, still needs testing on the FRDM-K64F, and needs more patches to be send in separate PRs.


hw/mcu/nxp/kinetis/src/hal_timer.c
```diff
@@ -0,0 +1,428 @@
...
+kinetis_tmr_read(struct kinetis_hal_tmr *tmr)
+{
+    return ((uint32_t)(tmr->overflow << 16) + LPTMR_GetCurrentTimerCount(tmr->base));
+}
+
+static void
+kinetis_tmr_set_period(struct kinetis_hal_tmr *tmr, uint32_t tick)
+{
+    /* XXX only set the lower 16-bits */
+    LPTMR_SetTimerPeriod(tmr->base, tick);
```
kasjer, line 87: Function LPTMR_SetTimerPeriod() should not be called when tick is 0 due to assert inside.
If I'm understanding it correctly with 32768 clock it can happen every 31 hours but I may be wrong.

hw/mcu/nxp/kinetis/src/hal_timer.c
```diff
@@ -0,0 +1,428 @@
...
+    .base = LPTMR0,
+};
+#endif
+#if MYNEWT_VAL(TIMER_1)
+struct kinetis_hal_tmr kinetis_tmr1 = {
+    .base = LPTMR1,
+};
+#endif
+
+static IRQn_Type kinetis_timer_irqs[] = LPTMR_IRQS;
```
kasjer, line 62: **kinetis_timer_irqs** could be **const**, maybe **kinetis_timers** also

hw/mcu/nxp/kinetis/src/hal_timer.c
```diff
@@ -0,0 +1,428 @@
...
+        } else {
+            break;
+        }
+    }
+
+    timer = TAILQ_FIRST(&tmr->hal_timer_q);
+    if (removed && timer) {
+        kinetis_tmr_set_period(tmr, timer->expiry);
+    } else {
+        tmr->overflow++;
```
kasjer, line 163: I'm not sure if overflow should not be handled differently.
If timer reset itself to run again in after LPTIMER overflowed, when first timer was checked again it would had value without overflow incremented so it would increment it without calling callback.
And in the next cycle timer would fire but it could be delayed while 16 bit period.

hw/mcu/nxp/kinetis/src/hal_timer.c
```diff
@@ -0,0 +1,428 @@
...
+    uint32_t psr;
+    uint32_t freq;
+
+    psr = tmr->base->PSR;
+    switch (psr & LPTMR_PSR_PCS_MASK) {
+    case 0:
+        freq = MCGIRCLK_HZ;
+        /* if the clock divisor is not bypassed, apply the divider */
+        if (!(psr & LPTMR_PSR_PBYP_MASK)) {
+            freq >>= ((psr >> LPTMR_PSR_PRESCALE_SHIFT) & 0xffff) + 1;
```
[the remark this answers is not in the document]
utzig: Done.

hw/mcu/nxp/kinetis/src/hal_timer.c
utzig: Done.

hw/mcu/nxp/kinetis/src/hal_timer.c
utzig: Yeah, good catch, it's interesting the CMR (compare) register cannot be zero, and also there is no control on the value o CNR (the counter), so I just checked the lower 16 bits for zero and incremented. Not sure how you came up with the 31 hours though, to it would seem more like 32768/3600 = 9.1 hours.

hw/mcu/nxp/kinetis/src/hal_timer.c
utzig: I didn't exactly understand your use case, but I can see incrementing overflow could be skipped under some circumstances. I think I put it in an else due to some unrelated bug while debugging, and now I remove that else and moved the overflow incrementing to just before clearing the interrupt; re-tested everything (high and low speed) and it all seems to work correctly.

hw/mcu/nxp/kinetis/src/hal_timer.c
kasjer: 31 hours was a mistake it's more like 36,4 hours (expiry is 32bits long) overflows at 4,294,967,296 ticks assuming 32768 ticks per second it makes 4,294,967,296 / 32768 (ticks in second) / 3600(seconds in hour) = 36,4 hours.

Anyway knowing how LPTMR_SetTimerPeriod() asserts when tick == 0 it could be that writing 0 to CMR register will not work (is not clear to me that CMR can't have 0 value) if that's the case I accept your solution that sometimes tick will be delayed.
But maybe writing 0 is OK and interrupt will happen and assert is for sanity only.
In that case knowing how it works and that 16 bits are used inside the function, something like this could be considered.
```c
LPTMR_SetTimerPeriod(tmr->base, (tick & 0xffff) | 0x10000);
```

hw/mcu/nxp/kinetis/src/hal_timer.c
```diff
@@ -0,0 +1,431 @@
...
+            break;
+        }
+    }
+
+    timer = TAILQ_FIRST(&tmr->hal_timer_q);
+    if (removed && timer) {
+        kinetis_tmr_set_period(tmr, timer->expiry);
+    }
+
+    tmr->overflow++;
```
kasjer, line 168: I don't think it will work as expected.
- If this was Free-Running mode (which I guess it is not) then overflow variable in the current form could hold upper 16 bits of 32 bits timer, but then it would need to be incremented not on every interrupt but only if overflow really occurred.
- It this is not Free-Running mode (I assume it is base on **LPTMR_GetDefaultConfig()** call), then overflow variable would not work as intended since counter overflows on different values (not every 0xFFFF but whenever CMR==CNR) which can be rather small. In this case reading timer will return some strange values.

hw/mcu/nxp/kinetis/src/hal_timer.c
utzig: The datasheet says writing zero does not enable to timer and keeps CTF (the interrupt flag) on, so I think the way they thought this out is that it should always start counting from zero up to CMR.

PR #2669, opened 2021-08-20
hw: bsp: frdm-k82f: update LED definitions

utzig:
Use standard LED_n nomenclature from color name based.

### DIFF
--- a/hw/bsp/frdm-k82f/include/bsp/bsp.h
+++ b/hw/bsp/frdm-k82f/include/bsp/bsp.h
@@ -30,10 +30,11 @@ extern uint8_t __DATA_ROM;
 #define RAM_SIZE            0x40000
 
 /* RBG LED pins */
-#define LED_RED_PIN         MCU_GPIO_PORTC(8)
-#define LED_GREEN_PIN       MCU_GPIO_PORTC(9)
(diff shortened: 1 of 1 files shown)


hw/bsp/frdm-k82f/include/bsp/bsp.h
```diff
@@ -33,7 +33,9 @@ extern uint8_t __DATA_ROM;
 #define LED_RED_PIN         MCU_GPIO_PORTC(8)
 #define LED_GREEN_PIN       MCU_GPIO_PORTC(9)
 #define LED_BLUE_PIN        MCU_GPIO_PORTC(10)
+
 #define LED_BLINK_PIN       LED_RED_PIN
+#define LED_2               LED_GREEN_PIN
```
kasjer, line 38: What happened to LED_1? :frowning:

utzig: That is not used anywhere afaik, I am not sure why BSPs export it but I'll do it here too.

utzig: I switched to using the standard LED_n nomenclature, thanks for the suggestion.

kasjer (approved):
I was just wandering why start with 2 and skip 1 and 3.
I did not mind color coded LEDs.

PR #2671, opened 2021-08-21
apps/sensors_test: Remove blinking tasks

kasjer:
Application defined 2 tasks to blink 1 led.
It also provided default pin (15) for led if
BSP did not provide LED_BLINK_PIN definition.

Amount of RAM used for tasks and stacks prevents
this application to run on MCUs with 20KB of RAM.

This:
- replaces tasks with callout reducing RAM used.
- removes local hard coded definition of LED,
  so pin 15 may be used for something else.

### DIFF
--- a/apps/sensors_test/src/main.c
+++ b/apps/sensors_test/src/main.c
@@ -32,10 +32,6 @@
 #include <reboot/log_reboot.h>
 #include <id/id.h>
 
-#ifndef LED_BLINK_PIN
-#define LED_BLINK_PIN 15
-#endif
-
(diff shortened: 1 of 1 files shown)


utzig:
Looks like it copied the blinking code from slinky.

utzig (approved):

vrahane:
Yes, that is correct. It was meant to be same as slinky so that other devs could pick it up quite easily. I do agree, reducing the memory usage does make sense.

vrahane (approved):

PR #2674, opened 2021-09-01
sys/id: Add override facility for functions and storage

agross-korg:
This patch adds an override facility to the sys/id package.  This allows someone
to override the functions and storage location to implement their own id 
config area.

### DIFF
--- a/sys/id/include/id/id.h
+++ b/sys/id/include/id/id.h
@@ -24,6 +24,8 @@
 extern "C" {
 #endif
 
+extern struct conf_handler id_conf;
+
 #if MYNEWT_VAL(ID_SERIAL_PRESENT)
 /*
  * Maximum expected serial number string length.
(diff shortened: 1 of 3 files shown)


wes3 (approved):
I am fine with these changes although I might mention somewhere that if you override what is there you must provide your own id_conf handler since conf_register(&id_conf) is still called (which I presume is intentional).

agross-korg:
> I am fine with these changes although I might mention somewhere that if you override what is there you must provide your own id_conf handler since conf_register(&id_conf) is still called (which I presume is intentional).

Good point.  I'll add that to the commit message

kasjer:
This is interesting, almost all functionality of package is removed, maybe some other package should be used instead. Not that I mind, not my area.

agross-korg:
> This is interesting, almost all functionality of package is removed, maybe some other package should be used instead. Not that I mind, not my area.

Yes, this was to get around some packages using the id_serial storage directly.  Perhaps there is a better way, but if there is I didn't find it.

agross-korg:
Going to abandon these changes.  The better solution was fixing all of the packages using the id/XXXXX entries to use the generic config APIs instead of directly accessing the id/XXXX storage via global variables.

PR #2675, opened 2021-09-01
stm32 PWM fixes

kasjer:
This updates STM32 PWM usability:
- fixes return value from` stm32_pwm_set_frequency()` (**pwm_test** application assert failure)
- fixes build for **STM32F0** with PWM driver
- adds **PWM** peripherals configuration to **nucleo-l073rz** board


utzig (approved):
LGTM

PR #2676, opened 2021-09-02
pwm/stm32_pwm: Fix PWM using advanced timers

JustineKH:
Advanced timer PWMs require the MOE bit in the TIMX_BDTR register to be set to enable output.
Enabling the compare preload function (CCxP in the TIMx_CCER register) is not supported in code for non-advanced timers. It should be explicitly disabled for advanced timers.


kasjer (approved):
Verified to work with several MCUs

PR #2677, opened 2021-09-02
stm32l0: Use common pin macros

kasjer:
STM32L0 mcu.h had its own version of MCU_GPIO_PORTx macros.
All others ST MCUs use common definition from stm32_common/mcu.h

This common definition file als provides more macros that are
required for alternate pin function handling.

This change removes STM32L0 copy of pin macros and includes common
definitions that all others MCUs use.


utzig (approved):
LGTM

PR #2678, opened 2021-09-02
apps/pwm_test: Fix pin handling for STM32

kasjer:
STM32 PWM driver encodes alternate pin function
at higher bits of pin definition.

Definition of pins with alternate function > 0 are 16 bits long,
where bits 12:15 specify function.

pwm_shell assumed that 8 bits is enough for pin number, but it will
not work for most ST cases.

This changes local variable type to accommodate ST encoding of pins
with alternate function.


utzig (approved):
LGTM

PR #2682, opened 2021-09-16
da1469x: Increase size of loadable binaries in development mode

agross-korg:
This patch increases the maximum supported size of ram loaded binaries using the
serial loading protocol.

Images < 64k in size use the original load method.  Any binary exceeding 64k but less than 128k needs to use the extended load sequence.


kasjer (approved):
LGTM

PR #2683, opened 2021-09-21
Memfault-SDK integration with MyNewt

t3zeng:
This set of changes allows mynewt users to pull in the memfault sdk using the `MEMFAULT_ENABLE` syscfg which gives access to the memfault sdk mynewt port.

Memfault coredumps have been implemented for arm cortex m4 and m33 for now. More can be added as they get implemented and tested in the future.

This PR requires the mynewt port changeset to get merged to the memfault-sdk:
https://github.com/memfault/memfault-firmware-sdk/pull/19


vrahane:
Hi all,

I have already reviewed the original changes and suggested this approach since it seemed the right way to do things rather than having mynewt maintain a version of the memfault sdk. Let us know your thoughts on this approach.

repository.yml
```diff
@@ -127,3 +127,11 @@ repo.deps:
             master: 0-dev
             mynewt_1_8_0_tag: 0.1.0
             mynewt_1_9_0_tag: 0.2.0
+
+    mynewt-memfault-sdk:
+        type: github
```
vrahane, line 132: Fix indentation here to be consistent.

t3zeng: I'm unable to see the difference in indent here.

vrahane:
One other way we thought of getting this functionality was doing a conversion from native coredump format to memfault coredump format but that seemed too wieldy.

repository.yml
vrahane: Actually, you are right. It is fine. All good.

wes3 (approved):
I have no issues with this approach

andrzej-kaczmarek:
I see few problems with this PR:
- There are commits which introduce changes only to be removed in subsequent one, this should be cleaned up first
- Do I understand correctly that change in `repository.yml` will make newt to download this 3rd party repo for each new project? If so, I do not really think this is a good idea.
- syscfg name used in kernel/os that introduces dependency to MemFault package is defined in that package so it does not make any sense

utzig:
> * Do I understand correctly that change in `repository.yml` will make newt to download this 3rd party repo for each new project? If so, I do not really think this is a good idea.

Would make more sense to integrate it like was done with tinyusb: https://github.com/apache/mynewt-core/tree/master/hw/usb/tinyusb

t3zeng:
> > * Do I understand correctly that change in `repository.yml` will make newt to download this 3rd party repo for each new project? If so, I do not really think this is a good idea.
> 
> Would make more sense to integrate it like was done with tinyusb: https://github.com/apache/mynewt-core/tree/master/hw/usb/tinyusb

If I am understanding this correctly, you are suggesting some documentation indicating how to add the repo via project.yml and remove it from repository.yml correct?

wes3:
@andrzej-kaczmarek and @utzig I missed the repository.yml changes so I am with you there.

utzig:
> If I am understanding this correctly, you are suggesting some documentation indicating how to add the repo via project.yml and remove it from repository.yml correct?

Exactly.

vrahane:
I think the tinyusb strategy is good, alternatively we could have a callback for the trap handler and register it via a package for memfault. What do you guys think of that ?

t3zeng:
> * syscfg name used in kernel/os that introduces dependency to MemFault package is defined in that package so it does not make any sense

My goal is to call the memfault fault handling api in the mynewt coredump area if `MEMFAULT_ENABLE` is set in the project. The idea is that it is a no-op if the user did not set this syscfg and if they did then that pathway to the memfault sdk opens up for use. Is there precedent on accomplishing something like this?

utzig:
> My goal is to call the memfault fault handling api in the mynewt coredump area if `MEMFAULT_ENABLE` is set in the project. The idea is that it is a no-op if the user did not set this syscfg and if they did then that pathway to the memfault sdk opens up for use. Is there precedent on accomplishing something like this?

I believe that although this might look a bit weird it should work; if the package is not included by the user the syscfg will be unset, if it is included it will be either set or unset according to what was configured, so it's not a problem per se, apart from the fact that we have a new syscfg that for someone who doesn't know what it means will be at a loss.

vrahane:
Alternate approach here: https://github.com/apache/mynewt-core/pull/2686

PR #2684, opened 2021-09-22
mcu: kinetis: fix QSPI HW reset race

utzig:
Lock the OS when calling `QSPI_SoftwareReset`; this API sets bits that result in a QSPI HW reset, loops for a while, and manually resets the same bits. If a context switch happens while those bits are set, and the next task reads the QSPI flash through the AHB mapping, this read results in zeroes being returned instead of the expected data.

Also removed a non-required busy wait after the HW reset.


kasjer (approved):
LGTM

PR #2690, opened 2021-10-04
hw/mcu/cmac: Fix hal_timer_read

andrzej-kaczmarek:
This is regression after c9f72793d712365e6bdc0b5a3386f6e3caebee11.

LLT value stored for given HAL value needs to correspond precisely to
HAL value converted to LLT, otherwise conversion from HAL to LLL will
not be accurate and calculated times e.g. for scheduling may be off by
up to 1 tick.


wes3 (approved):
LGTM

PR #2695, opened 2021-10-13
Fix nrf51 clock init

danielkucera:


kasjer:
I think code change would be good but...
For now maybe it fixes something (not clear what) and it breaks build for other nrf51 based BSPs (nordic_pca10028, nordic_pca10028-16k bbc_microbit, bmd200, calliope_mini, nrf51-blenano, nrf51-arduino_101, vbluno51).

To make it look good all affected BSP should be updated.
Commit message is just too enigmatic.
It states that if fixes something but it is not clear what was broken.
Changes are similar to what was applied to NRF52 but at least two things were introduce:
- clock selection with choice in yml
- and fix for something in clock setting, it would be good if commit message explained that.

kasjer:
Closing on author suggestion.

PR #2706, opened 2021-11-01
mcu/nrf5340_net: Add ipc_nrf5340 dependency for nrf5340_net_vflash_init

t3zeng:
`nrf5340_net_vflash_init` is currently using a struct (`ipc_shared`) defined in `ipc_nrf5340/ipc_nrf5340_priv.h`.

When building, there will be an error due to the fact that the `ipc_nrf5340` package is not a dependency so the header cannot be found.


vrahane (approved):
LGTM

PR #2711, opened 2021-11-05
mcu/nrf5340: Add RAM location check to calls using easyDMA

vikrant-proxy:
Add checks to confirm that the buffer points to a location in RAM given that easyDMA can only access RAM locations

### DIFF
--- a/hw/mcu/nordic/nrf5340_net/src/hal_spi.c
+++ b/hw/mcu/nordic/nrf5340_net/src/hal_spi.c
@@ -691,7 +691,7 @@ hal_spi_txrx_noblock(int spi_num, void *txbuf, void *rxbuf, int len)
     }
 
     if (rxbuf != NULL && !nrfx_is_in_ram(rxbuf)) {
-        return SYS_EINVAL;
+        return EINVAL;
     }
 
 #if MYNEWT_VAL(SPI_0_MASTER)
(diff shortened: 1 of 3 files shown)


vrahane (approved):
I am fine with the change, quite useful actually as we have seen. @kasjer The spiflash DFU issue was actually because of this. Let’s wait for others approval.

hw/mcu/nordic/nrf5340_net/src/hal_spi.c
```diff
@@ -685,10 +686,14 @@ hal_spi_txrx_noblock(int spi_num, void *txbuf, void *rxbuf, int len)
 {
     struct nrf5340_net_hal_spi *spi = &nrf5340_net_hal_spi0;
 
-    if (spi_num != 0 || (spi->txrx_cb_func == NULL) || (len == 0)) {
+    if (spi_num != 0 || (spi->txrx_cb_func == NULL) || (len == 0) || !nrfx_is_in_ram(txbuf)) {
         return EINVAL;
     }
 
+    if (rxbuf != NULL && !nrfx_is_in_ram(rxbuf)) {
+        return SYS_EINVAL;
```
kasjer, line 694: This looks slightly strange to return SYS_EINVAL and EINVAL so close together for similar reason.
Maybe some unification could be applied, I'm not sure which way though.

kasjer (approved):
Even with those return error discrepancies it should be useful, I guess this kind of error happens only in development and is not really runtime error that will popup later.

hw/mcu/nordic/nrf5340_net/src/hal_spi.c
vikrant-proxy: ah yes, let me be consistent with the existing convention.

vikrant-proxy: addressed

PR #2716, opened 2021-11-08
util: Add sensitive data mynewt util to hide data in coredumps

t3zeng:


kernel/os/src/arch/cortex_m33/os_fault.c
```diff
@@ -213,6 +217,10 @@ os_default_irq(struct trap_frame *tf)
     uint32_t orig_sp;
 #endif
 
+#if MYNEWT_VAL(MEMPROTECT_CLEAR_DATA)
```
kasjer, line 220: if MTB buffer is used, having this call before mtb_stop() is called can clobbers most of MTB content with loop data that is not relate to crash.

util/memprotect/src/memprotect.c
```diff
@@ -0,0 +1,35 @@
...
+
+extern int __mynewt_protected_start, __mynewt_protected_end;
+
+/**
+ * Clear out sensitive memory area
+ */
+void
+memprotect_clear_data(void)
+{
+    uint32_t * ptr = (uint32_t *) &__mynewt_protected_start, idx;
```
kasjer, line 30: It's not common practice in mynewt to put two variables in one line especially when once is pointer and second not.

util/memprotect/src/memprotect.c
```diff
@@ -0,0 +1,35 @@
...
+
+    for (idx = 0; i < (uint32_t *) &__mynewt_protected_end; idx+= sizeof(* ptr)) {
+        ptr[idx] = MYNEWT_VAL(MEMPROTECT_PATTERN);
```
kasjer, line 33: does it even compile?
i does not seem to be declared around

kasjer (commented):
Usual way for not including some part of RAM in core dump is to modify hal_bsp.c dump_cfg[] array with regions that should be dumped. In that case array will have more then one region and sensitive data can be put by linker script in some specific area that is not mentioned in dump_cfg.

util/memprotect/src/memprotect.c
t3zeng: `MEMPROTECT_PATTERN` is in `util/memprotect/syscfg.yml`

kernel/os/src/arch/cortex_m33/os_fault.c
t3zeng: I see, I will move it to after mtb_stop

t3zeng:
> Usual way for not including some part of RAM in core dump is to modify hal_bsp.c dump_cfg[] array with regions that should be dumped. In that case array will have more then one region and sensitive data can be put by linker script in some specific area that is not mentioned in dump_cfg.

If I understand this correctly, given that `dump_cfg` is a whitelist rather than a blacklist, I would need to still put the sensitive data in one area via linker script and attribute, and then include all areas except the sensitive part of RAM in every hal_bsp.c file. 

My concern on this is that there is a very large number of hal_bsp.c and more to come in the future as more hardware is supported and custom coredump callbacks used with `OS_COREDUMP_CB` may not necessarily be able to use this feature out of the box

util/memprotect/src/memprotect.c
t3zeng: ok I can split these out

util/memprotect/src/memprotect.c
vrahane: `i` is not defined anywhere.

t3zeng: Oh, I see. I swapped out i for idx prior to putting the code up and missed it. Fixed it

kernel/os/pkg.yml
```diff
@@ -45,7 +46,7 @@ pkg.deps.OS_SYSVIEW:
     - "@apache-mynewt-core/sys/sysview"
 
 pkg.deps.OS_CRASH_LOG:
-    - "@apache-mynewt-core/sys/reboot"
+    - "@apache-mynewt-core/sys/reboot"
```
vrahane, line 49: Remove the extra space added there.

kasjer:
> @kasjer We have been speaking about this and it was my suggestion to look at defining sections instead of adding more code for protected data so that we could generically define data structures and variables in a specified section. It seems to be the best way to make it user friendly. `dump_cfg` is also good. Though it is little weird to define variables in a region that is not present in `dump_cfg`. Do you know of an easy to way to do that ?

Considering that you already wanted to add one special section for sensitive data that should not go to core dump
having such function in bsp does what you want with same changes to linker scripts.
```c
const struct hal_bsp_mem_dump *
hal_bsp_core_dump(int *area_cnt)
{
    dump_cfg[0].hbmd_start = &_ram_start;
    dump_cfg[0].hbmd_size = (int)&__mynewt_protected_start - (int)&_ram_start;
    dump_cfg[1].hbmd_start = &__mynewt_protected_end;
    dump_cfg[1].hbmd_size = (int)&_ram_start + RAM_SIZE - (int)&__mynewt_protected_end;

    *area_cnt = 2;
    return dump_cfg;
}
```
I think some of the reviewers already use such approach and they may have some more insight.

t3zeng:
> > @kasjer We have been speaking about this and it was my suggestion to look at defining sections instead of adding more code for protected data so that we could generically define data structures and variables in a specified section. It seems to be the best way to make it user friendly. `dump_cfg` is also good. Though it is little weird to define variables in a region that is not present in `dump_cfg`. Do you know of an easy to way to do that ?
> 
> Considering that you already wanted to add one special section for sensitive data that should not go to core dump having such function in bsp does what you want with same changes to linker scripts.
> 
> ```c
> const struct hal_bsp_mem_dump *
> hal_bsp_core_dump(int *area_cnt)
> {
>     dump_cfg[0].hbmd_start = &_ram_start;
>     dump_cfg[0].hbmd_size = (int)&__mynewt_protected_start - (int)&_ram_start;
>     dump_cfg[1].hbmd_start = &__mynewt_protected_end;
>     dump_cfg[1].hbmd_size = (int)&_ram_start + RAM_SIZE - (int)&__mynewt_protected_end;
> 
>     *area_cnt = 2;
>     return dump_cfg;
> }
> ```
> 
> I think some of the reviewers already use such approach and they may have some more insight.

@kasjer, would this solution cause issues with `os_addr_is_ram` returning 0 on protected addresses? Also, if variables on the protected ram area are on the stack, wouldn't you still be able to see them on the backtrace

wes3:
@t3zeng 
1) os_addr_is_ram returning 0, if I am understanding what you are asking , does not seem to be a problem (although I could be wrong). Seems like os_stacktrace is just trying to display text addresses from ram areas that will be in the coredump. The protected data is not in .text so it would not get displayed and thus is not a problem.

2) Again, assuming I am understanding your question, yes, I guess it is possible that if the protected variables are in registers on the stack they will get displayed in os_fault.c. However, I think a bigger problem is this: if those variables are accessed and they are placed on a stack (a tasks stack, for example) the will get placed in the coredump since the stacks are in ram and will be part of the coredump.

I have gone through some of the code and I think I prefer the approach where the dump_cfg is modified to exclude the protected area but I am not sure I have considered everything yet.

vrahane:
So, I think for protected variables on the stack, there are couple of options, others can comment more on based on their experiences but I can think of the following:

1. Best bet would be to define them in the `mynewt_protected` section
2. Have a protected section per stack somehow, this would need to be done from the linker script with defining dummy sections per stack and updating `StackTop`, `StackLimit`, `ProtectedStackStart1` and `ProtectedStackEnd1` for example and when a trap comes in, clear the individual protected stack sections. Again, this would mean a lot of changes to the existing stuff, so not sure if that is the route we want to take here.
3. A combined protected section for all stacks and clearing that section (Somewhat same as 1, but easily identifiable with a dedicated section)

mkiiskila:
There are already sections defined for regions of RAM which users can use to keep things out of coredump, aka storage of secret stuff. These are 'sec_data_core', 'sec_bss_core' and 'sec_bss_nz_core'. Not all users care about this, so it was left for the BSP to define whether these get included in the core or not (dump_cfg, as others have pointed out).

At the moment I'd vote no for adding another section, given that these already exist.

Regarding stack... yes. That can be a problem. General practice is to clobber the local variable which held the secret after use. Which the crypto libraries we include do. One low cost option would be to have a per-task flag which could be set just before secrets are about to be used in stack, and cleared after secret use is done. Then you could decide to either exclude that stack, or clobber (easier) before including it in the core/stacktrace. 

With the current implementation we have, if I was really worried about secrets in my use-case, and I was using mynewt, I'd move all the security computations to be done by a single task, and place it's stack in 'sec_bss_core' section. Define dump_cfg to exclude 'sec_XXX'.  And not have OS_CRASH_STACKTRACE set.

t3zeng:
> There are already sections defined for regions of RAM which users can use to keep things out of coredump, aka storage of secret stuff. These are 'sec_data_core', 'sec_bss_core' and 'sec_bss_nz_core'. Not all users care about this, so it was left for the BSP to define whether these get included in the core or not (dump_cfg, as others have pointed out).
> 
> At the moment I'd vote no for adding another section, given that these already exist.
> 
> Regarding stack... yes. That can be a problem. General practice is to clobber the local variable which held the secret after use. Which the crypto libraries we include do. One low cost option would be to have a per-task flag which could be set just before secrets are about to be used in stack, and cleared after secret use is done. Then you could decide to either exclude that stack, or clobber (easier) before including it in the core/stacktrace.
> 
> With the current implementation we have, if I was really worried about secrets in my use-case, and I was using mynewt, I'd move all the security computations to be done by a single task, and place it's stack in 'sec_bss_core' section. Define dump_cfg to exclude 'sec_XXX'. And not have OS_CRASH_STACKTRACE set.

Thanks for the insight @mkiiskila 

While I was looking into `sec_data_core ` and `sec_bss_core ` I came across these secret sections as well:`sec_data_secret` and `sec_bss_secret`

What is the difference in behavior between these areas?

mkiiskila:
Ah, yes. I misspoke, those 2 sections were the ones I meant! Thanks @t3zeng for pointing it out. Those 2 sections are meant for storing secret stuff.

vrahane:
So, I guess we do not need this PR then since it can already be handled in the BSP, so no new section is needed.

t3zeng:
I fiddled around with `sec_data_secret` and looking at the memory map, it doesn't appear to move variables anywhere. Looking at the code, this seems to make sense as I don't see the macro push data to any sections nor does there seem to be areas in the linker script (eg. apache-mynewt-core/hw/mcu/nordic/nrf5340/nrf5340.ld) for secret variables to pass to `dump_cfg`. 

Based on what I'm seeing, it looks like the intended use case is to:
1. add a custom area to the linker script myself
2. define `sec_data_secret` as that area in bsp.h
3. Update `dump_cfg` so that coredumps capture around the protected memory area
4. Add an API call to `os_default_irq` to clobber variables in sensitive area so that the stack doesn't show the values

Would that be correct?

mkiiskila:
Thanks for looking over this. Yes for 1-3, no for 4. The linker script indeed is where you'd insert symbols around secret data and secret bss areas, and then take those into account within dump_cfg.
Both of those items are under BSP control; we have not had an example of this being done within the BSPs in the trees, and (as you've discovered) is not specifically called out in the documentation.

Which leaves us with item #4. I have not done anything for this as I considered the risk to be small'ish. My reasoning was that a) probability of crashing during computation when secret was being used is low and b) raw secret would be held in the areas not included in coredump, and what is in the stack is only small parts of it (4 bytes in reg) and data which has been obfuscated by going (partly) through security algo.

So there is a bit of  possibility peephole. One could argue that if you can induce crashes at specific times AND get your hands on corefiles OR stacktrace appearing on the console, then you could re-construct the secret from pieces. This might be unacceptable for some use cases.

If you (or someone else?) is interested in plugging this hole, I suspect that would help some users!

mkiiskila:
Also, adding documention about the current way of doing things would be liked by many. Even modifying a BSP to demonstrate this use would be nice.

kasjer:
One of the points of having coredump is to see stack for tasks. If stack data is a security concern, stack itself can be put in crash-excluded area instead of destroying it, so I also don't see benefit of 4.

t3zeng:
The stack is still useful to have but depending on where you crash it's possible to put protected data onto the stack. In cases like these you don't want to exclude the entire stack, it would be better to see it as normal for debugging purposes but with sensitive data like keys clobbered. Hiding the entire stack in a crash excluded area would mitigate a lot of the benefits of coredumps

PR #2717, opened 2021-11-12
hw/drivers/adc_nrf52: Prevent NULL pointer access

vikrant-proxy:
Assign the device pointer before un-configuring ADC instance. This prevents `global_adc_dev` from being NULL inside `channel_unconf()` via `init_instance_unconf()`


vrahane (approved):
LGTM

PR #2718, opened 2021-11-15
Fix nrf534 ipc with old boot

kasjer:
IPC and virtual flash configuration was changed to be TrustZone non-secure friendly.
Those modification broke compatibility with existing net core bootloader.
This restores compatibility with old bootloader.

To enable old bootloader compatibility app core application have to be compiled with
`IPC_NRF5340_PRE_TRUSTZONE_NETCORE_BOOT: 1`

### DIFF
--- a/hw/drivers/ipc_nrf5340/syscfg.yml
+++ b/hw/drivers/ipc_nrf5340/syscfg.yml
@@ -48,6 +48,16 @@ syscfg.defs:
             eg "LED_1, LED_2" or "1, 2". Further GPIO configuration should be
             done by Network Core.
         value: ""
+    IPC_NRF5340_PRE_TRUSTZONE_NETCORE_BOOT:
+        description: >
+            Set this to one only in case of the bootloader predates TrustZone
+            changes. With ARM TrusZone support added NRF_IPC_NS block is used
+            for both secure and non-secure application. Pre-TrustZone code
+            was always running in secure mode and network core code always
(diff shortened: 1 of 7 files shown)


hw/drivers/ipc_nrf5340/syscfg.yml
```diff
@@ -48,6 +48,17 @@ syscfg.defs:
...
+            for both secure and non-secure application. Pre-TrustZone code
+            was always running in secure mode and network core code always
+            accessed NRF_IP_S (including code that is running in bootloader).
+            NRF_IP_S->GPMEM[] was used to pass address of net core application
+            image to net bootloader.
+        value: 0
 
 syscfg.restrictions:
     - "!BSP_NRF5340 || BSP_NRF5340_NET_ENABLE"
+    - "!IPC_NRF5340_PRE_TRUSTZONE_NETCORE_BOOT || MCU_APP_SECURE"
```
t3zeng, line 64: Given that `MCU_APP_SECURE` is undefined on the netcore side, you would hit this restriction if you set `IPC_NRF5340_PRE_TRUSTZONE_NETCORE_BOOT` to 1

t3zeng:
Tried this branch on nrf5340 hw and the `IPC_NRF5340_PRE_TRUSTZONE_NETCORE_BOOT` works as expected

kasjer:
Just one commit after rebase

sjanc (approved):

PR #2723, opened 2021-11-19
mgmt: Fix package dependency

kasjer:
There were two sections pkg.deps.CRASH_TEST_MGMT:
in such case only second section is used by newt tool.

cborattr/cborattr.h was included but cborattr package was
not specified.


vrahane (approved):
Makes sense, although I have used it recently. The dependency probably was getting included somewhere else. LGTM.

PR #2724, opened 2021-11-24
hw/bsp: Add BSP for nRF52840 Dongle pca10059

kasjer:
BSP is mostly copy of pca10056 with buttons and LEDs
definitions.
UART is turned off by default


utzig (approved):
LGTM

PR #2726, opened 2021-11-26
util: Add BUSY_LOOP_HANDLER macro

mkasenberg:
Needed by babblesim to move time forward in busy loops.


mkasenberg:
@andrzej-kaczmarek

PR #2728, opened 2021-12-01
tinyusb/dfu: Add boot_preboot for DFU in bootloader

kasjer:
When mcuboot have BOOT_PREBOOT set it executes boot_preboot() function.
dfu package provides this function boot_preboot() that checks selected GPIO
to see if DFU mode should be activated.


utzig (approved):
LGTM

PR #2729, opened 2021-12-01
baselibc: Call mynewt_main() instead of main()

andrzej-kaczmarek:


apps/bleprph_oic/src/main.c
```diff
@@ -362,7 +362,7 @@ static const oc_handler_t omgr_oc_handler = {
  * @return int NOTE: this function should never return!
  */
 int
-main(void)
+mynewt_main(int argv, char **argv)
```
kasjer, line 365: more canonical way would be `(int argc, char *argv[])`

kernel/os/src/os.c
```diff
@@ -82,6 +82,14 @@ int g_os_started;
 #define MIN_IDLE_TICKS  (MYNEWT_VAL(OS_IDLE_TICKLESS_MS_MIN) * OS_TICKS_PER_SEC / 1000)
 #define MAX_IDLE_TICKS  (MYNEWT_VAL(OS_IDLE_TICKLESS_MS_MAX) * OS_TICKS_PER_SEC / 1000)
 
+extern int main(int, char **);
+
+__attribute__((weak)) int
+mynewt_main(int argc, char **argv)
```
kasjer, line 88: With this **weak** `mynewt_main()` version even though application has own version of mynwet_main(), linker for arm-nono-gcc and xc32-gcc often (not always) chooses this weak version and then complains that `main` is missing.
```
repos/apache-mynewt-core/kernel/os/src/os.c:90: undefined reference to `main'
```
It seems to work better with applications from apache-mynewt-core but when iptest was build for black_vet6 with new mynewt_main() in main.c present weak version was used.
For local applications like blinky or others it newer worked for me.

So I think that if we are going to go this way it would be better to drop this weak version of find some way that is more stable.

kasjer: Solution could be separate package with just one file with weak implementation.

sjanc:
fixed upstream

PR #2733, opened 2021-12-02
hw/mcu: Fix cmac build on Windows

kasjer:
Having include ../../../../../../apache-mynewt-core... in
os_arch_arm.c breaks build on Windows.

Adding include path to package seems to fix the problem.

```sh
Executing scripts/create_cmac_bin.sh
Target successfully built: targets/dialog_cmac
Executing scripts/build_libcmac.sh
Building target @apache-mynewt-nimble/targets/dialog_cmac
Error: CreateFile h\:\work\mynewt\repos\apache-mynewt-core\kernel\os\src\os_priv.h: The filename, directory name, or volume label syntax is incorrect.
```

### DIFF
--- a/hw/mcu/dialog/cmac/pkg.yml
+++ b/hw/mcu/dialog/cmac/pkg.yml
@@ -31,3 +31,5 @@ pkg.deps:
     - "@apache-mynewt-core/hw/hal"
     - "@apache-mynewt-core/hw/mcu/dialog"
     - "@apache-mynewt-nimble/nimble/drivers/dialog_cmac"
+
+pkg.cflags: -I@apache-mynewt-core/kernel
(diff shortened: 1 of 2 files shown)


hw/mcu/dialog/cmac/pkg.yml
```diff
@@ -31,3 +31,5 @@ pkg.deps:
     - "@apache-mynewt-core/hw/hal"
     - "@apache-mynewt-core/hw/mcu/dialog"
     - "@apache-mynewt-nimble/nimble/drivers/dialog_cmac"
+
+pkg.cflags: -Irepos/apache-mynewt-core/kernel
```
andrzej-kaczmarek, line 35: iirc `-I@apache-mynewt-core/kernel` should also work

kasjer: done

andrzej-kaczmarek (approved):

PR #2741, opened 2021-12-16
stlink: Allow STM32_Programmer_CLI for flashing

kasjer:
stlink: Allow STM32_Programmer_CLI for flashing

ST provides command line tool for flashing devices as a part of STM32CubeIDE.

In some cases this tool works better than currently used st-flash tool.
st-flash sometimes has problem detecting flash memory:

```
Error: Downloading bin/targets/nucleo-f411re-blinky/app/apps/blinky/blinky.img to 0x8020000
st-flash 1.7.0-11-ge662da0
2021-12-16T14:25:02 INFO common.c: stm32f411re: 128 KiB SRAM, 0 KiB flash in at least 16 KiB pages.
Unknown memory region
```

For same hardware setup ST provided tool has no problem (openocd also works fine but requires script selection).

When user sets environment variable `STLINK_USE_STM32_PROGRAMMER_CLI=1`,
**STM32_Programmer_CLI** will be used instead of st-flash tool.

### DIFF
--- a/hw/scripts/stlink.sh
+++ b/hw/scripts/stlink.sh
@@ -41,7 +41,11 @@ stlink_load () {
 
     echo "Downloading" $FILE_NAME "to" $FLASH_OFFSET
 
-    st-flash --reset write $FILE_NAME $FLASH_OFFSET
+    if [ "$STLINK_USE_STM32_PROGRAMMER_CLI" == 1 ] ; then
+      STM32_Programmer_CLI -c port=SWD -d $FILE_NAME $FLASH_OFFSET -rst
+    else
+      st-flash --reset write $FILE_NAME $FLASH_OFFSET
+    fi
     if [ $? -ne 0 ]; then
         exit 1
     fi


hw/scripts/stlink.sh
```diff
@@ -41,7 +41,11 @@ stlink_load () {
 
     echo "Downloading" $FILE_NAME "to" $FLASH_OFFSET
 
-    st-flash --reset write $FILE_NAME $FLASH_OFFSET
+    if which STM32_Programmer_CLI >/dev/null ; then
+      STM32_Programmer_CLI -c port=SWD -d $FILE_NAME $FLASH_OFFSET -rst
```
utzig, line 45: It's a bit weird to force the use of the ST tool instead of using the open-source one provided by stlink (in a script called stlink.sh!). Have you considered using an environment variable, eg `STLINK_USE_STM32_PROGRAMMER_CLI=1` or something that people could enable the optional use?

kasjer: I though stlink is more like debugger name in that context, that's why I did not create separate script for this.
Even when ST environment is installed STM32_Programmer_CLI is buried inside some cryptic plugin folder:
/c/st/STM32CubeIDE_1.7.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.win32_2.0.0.202105311346/tools/bin/

It will not be used by accident, user has to add this to PATH variable and when it's done user intention is obvious with no additional settings.

utzig: > It will not be used by accident, user has to add this to PATH variable and when it's done user intention is obvious with no additional settings.

```
$ which STM32_Programmer_CLI
/home/utzig/STMicroelectronics/STM32Cube/STM32CubeProgrammer/bin/STM32_Programmer_CLI
```

Yes it will.

utzig (approved):
LGTM

PR #2744, opened 2022-01-03
stm32: Add missing package when RTC is used for systick

kasjer:
When OS_TICKS_USE_RTC is 1 hal_os_tick.c includes datetime/datetime.h
but appropriate package was not added now it is added when required.


utzig (approved):
LGTM

PR #2746, opened 2022-01-05
hw/ipc_nrf5340: Add support for reseting Network Core

sjanc:
This allows application core to reset Network core and re-sync IPC
state.


kasjer (approved):
LGTM

PR #2747, opened 2022-01-10
mcu/nrf52: Add reset pin configuration

kasjer:
NRFX provides setting to make reset pin work.
Without this setting appropriate GPIO is just GPIO.
When setting is present, pin acts as hardware reset.

This adds syscfg value that can enable reset pin.


kasjer:
I was blind

PR #2749, opened 2022-01-12
hw/ipc_nrf5340: Adjust default number of enabled channels

sjanc:
Increase default on Application core to 5 to cover "default" BLE
usecase: 2x for HCI, 2x for Virtual Flash and 1x for Keep Alive.

Increase default on Network core to 16 (max) since on that core only
"struct ipc_channel" is kept in RAM (rest is on Application core) and
true number of channels is read from AppCore IPC config in runtime.


kasjer (approved):
LGTM

PR #2755, opened 2022-01-26
hal: Modify hal_system_restart

wes3:
Two changes to the code: clear ICPR register (to clear any pending interrupts) and also leave global interrupts enabled when calling hal_system_start. The idea behind this change is that the cortex-M has global interrupts enabled on reset and that this should be the state when the system is restarted.

For now, I only modified the code in the hw/mcu/dialog directory to have a discussion regarding this change in case folks object. I am not sure why there is a both a hal_system_start and hal_system_restart function but given there were two I decided to change restart only. It appears that mcuboot for mynewt calls hal_system_start and it should probably call hal_system_restart or we should abandon hal_system_restart and have hal_system_start clear ICER and ICPR and leave primask with interrupts enabled.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_system_start.c
+++ b/hw/mcu/dialog/da1469x/src/hal_system_start.c
@@ -62,18 +62,23 @@ hal_system_start(void *img_start)
 void
 hal_system_restart(void *img_start)
 {
-    uint32_t primask __attribute__((unused));
     int i;
 
     /*
-     * Disable interrupts, and leave them disabled.
-     * They get re-enabled when system starts coming back again.
+     * NOTE: on reset, PRIMASK should have global interrupts enabled so
+     * the code disables interrupts, clears the interrupt enable bits,
+     * clears any pending interrupts, then enables global interrupts
+     * so processor looks like state it would be in if it reset.
      */
-    __HAL_DISABLE_INTERRUPTS(primask);
-
+    __disable_irq();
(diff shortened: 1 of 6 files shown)


vrahane:
I sort of agree, since we generally use hal_system_start() so, I think we should probably clear ICPR and ICER there and use primask.

Calling hal_system_restart() from hal_system_start() is a little bit weird in my opinion but either ways I think this change is useful.

wes3:
@vrahane Not sure you were saying this, but I was not suggesting calling hal_system_restart from hal_system_start. As of now, hal_system_restart calls hal_system_start which is fine assuming we want to keep hal_system_restart. Thinking more about this, I think it fine to have both functions as there may be a case where you dont want to clear all interrupts. I just think there are some places where hal_system_restart should be called as opposed to hal_system_start.

kasjer (commented):
Looks good,

wes3:
@kasjer yes once folks agreed with the changes these will get ported to different MCUs

andrzej-kaczmarek:
that makes sense, so will just wait for updated PR with changes to other CM MCUs

wes3:
I modified some of the other MCUs. I did not do anything to code that did not have a hal_system_restart() function or somewhat similar code so in effect this only ended up modifying the dialog and nrf mcu implementations.

kasjer (approved):
LGTM

vrahane (approved):

andrzej-kaczmarek (approved):

PR #2760, opened 2022-01-30
kernel/os: Add support for 31250 Hz for os_cputime and os_tick

andrzej-kaczmarek:
We'll need 31250 Hz/tps for upcoming changes to Dialog CMAC code.


kasjer (approved):
LGTM, however as it is special case for 1MHz (which is also divisible by power of 2) maybe it could benefit of inline as is the case for 1MHz.

andrzej-kaczmarek:
@kasjer good idea, will change

PR #2761, opened 2022-01-30
hw/*/cmac: Use 31250 clock instead of 32768

andrzej-kaczmarek:
Since 31250, unlike 32768, is a factor of 1000000 we can use much simpler conversions between LL timer and hal_timer/os_cputime/os_tick values than we currently have. In fact, all we have to is to shift by 5 bits and thus we can remove all those nasty workaround that were added to speed up current conversions that suffered from lack of integral division support on CM0+.

Also, since LL timer is 37 bits and we shift by 5 bits, it means we have full 32-bit value of os_cputime readily available just as if we have a native 32-bit timer running at 31250Hz.


andrzej-kaczmarek:
see https://github.com/apache/mynewt-nimble/pull/1155

kasjer (approved):

kasjer:
Changes make throughput tests work without crashing CMAC

PR #2762, opened 2022-01-30
OSX support for detect_programmer

cqexbesd:
Not sure what this gets used for but I had already started to patch it before I realised it wasn't my problem and perhaps the code is useful for others.

### DIFF
--- a/hw/scripts/common.sh
+++ b/hw/scripts/common.sh
@@ -88,6 +88,27 @@ detect_programmer() {
         # extract the VID:PID list for connected USB devices
         USB_DEV=$(lsusb | cut -f6 -d' ')
 
+    elif [ $(which system_profiler) ] ; then
+
+        # extract the VID:PID list for connected USB devices on OSX
+        # and get them in the same format as lsusb from Linux
+        USB_DEV=$(system_profiler SPUSBDataType 2>/dev/null | awk '
+            /^[ \t]+Product ID:/ {
+                # remove leading 0x from hex number
+                sub(/^0x/, "", $3);
+                PID=$3;
+            }
+            /^[ \t]+Vendor ID:/ {
+                # remove leading 0x from hex number
+                sub(/^0x/, "", $3);
+                printf("%s:%s\n", PID, $3);
+            }
+        ')
+
+    fi
+
+    if [ -n "$USB_DEV" ]; then
+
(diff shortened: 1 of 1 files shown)


hw/scripts/common.sh
```diff
@@ -88,6 +88,14 @@ detect_programmer() {
         # extract the VID:PID list for connected USB devices
         USB_DEV=$(lsusb | cut -f6 -d' ')
 
+    elif [ $(which system_profiler) ] ; then
+
+        # extract the VID:PID list for connected USB devices on OSX
+        USB_DEV=$(system_profiler SPUSBDataType 2>/dev/null | awk '/^[ \t]+Product ID:/ { PID=$3 } /^[ \t]+Vendor ID:/ { printf("%s:%s\n", PID, $3) }')
```
vrahane, line 94: Thanks for the contribution, however, I do think this statement should be reduced into smaller consecutive statements to make it consistent with rest of the script.

kasjer: For me this line looks perfectly OK.
I don't see anything that could be split.
Same approach is use few line above where `lsusb` is piped to `cut` and here `system_profiler` is piped to `awk` in very nice way including error redirection to **/dev/null**

I don't have OSX to check for system_profiler output but awk part is very clear and there is no single character I would written better.
@vrahane please reconsider

vrahane: @kasjer  Functionality wise, might be fine but just from a style perspective it is a very long line. In the entire script people have taken efforts to reduce the length of the statements. So, I would expect the same here. For example:
https://linuxhint.com/split-long-bash-command/

kasjer: @vrahane if your comment was about reformatting I would not object (@cqexbesd surely knows how to use `\`), but **smaller consecutive statements** did not apply here in my opinion.

vrahane: Either reformatting or using smaller consecutive statements, both are fine as long as it’s smaller. 151 characters in one line is not really a good practice.

cqexbesd: Reformatted, and a bug fix added.

vrahane (approved):
Works fine. It does output the list of VID and PIDs in USB_DEV correctly.

PR #2767, opened 2022-02-08
hw/nrf53: Add option to use internal load caps for lfxo

andrzej-kaczmarek:


hw/mcu/nordic/nrf5340/src/hal_system.c
```diff
@@ -86,6 +86,17 @@ hal_system_clock_start(void)
     regval = CLOCK_LFCLKSTAT_STATE_Running << CLOCK_LFCLKSTAT_STATE_Pos;
 
 #if MYNEWT_VAL_CHOICE(MCU_LFCLK_SOURCE, LFXO)
+
+#if MYNEWT_VAL_CHOICE(MCU_LFCLK_XO_INTCAP, external)
+    NRF_OSCILLATORS->XOSC32KI.INTCAP = OSCILLATORS_XOSC32KI_INTCAP_INTCAP_External;
```
kasjer, line 91: Maybe **nrfx** `nrf_oscillators_lfxo_cap_set()` could be used here?

kasjer (approved):

hw/mcu/nordic/nrf5340/src/hal_system.c
andrzej-kaczmarek: I'll leave as-is since we use registers everywhere, but probably rewriting some of the code to nrfx would be a good idea anyway...

PR #2773, opened 2022-02-16
hw/mcu/dialog: Reset UART before reconfiguration

kasjer:
If TX FIFO is not empty (and maybe other conditions) during programming
new baudare UART can endup in BUSY state that will never end.

This adds reset of UART prior to configuring it in similar way as is
already done in uart_da1469x.c

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_uart.c
+++ b/hw/mcu/dialog/da1469x/src/hal_uart.c
@@ -572,6 +572,10 @@ hal_uart_config(int port, int32_t baudrate, uint8_t databits, uint8_t stopbits,
          */
         da1469x_uart_set_rx_pin_as_gpio(uart);
     }
+    /* Reset UART before setting up new baudrate */
+    regs->UART2_SRR_REG = UART2_UART2_SRR_REG_UART_UR_Msk |
+                          UART2_UART2_SRR_REG_UART_RFR_Msk |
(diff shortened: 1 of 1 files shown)


wes3 (approved):
Changes look good. Question: were these changes made in case hal_uart_config was called when the fifos had data in them or something like that (to prevent garbage being transmitted or received, for example)? If that is the case I am wondering about the ordering of some things in this function. For example, maybe NVIC interrupts should be disabled and cleared before setting the clock registers. If not, all good.

kasjer:
> Changes look good. Question: were these changes made in case hal_uart_config was called when the fifos had data in them or something like that (to prevent garbage being transmitted or received, for example)? If that is the case I am wondering about the ordering of some things in this function. For example, maybe NVIC interrupts should be disabled and cleared before setting the clock registers. If not, all good.

It come out when I prepared changes for 6MHz UART; Main purpose was to have high speed UART debug traces that could be used for analyzing BLE related-timing (or at least I thing it was the purpose).
I did it to trace stuff going on USB and timing of data was supposed to be as fast as possible so basically debugging code did not use hal UART code at all just wrote single bytes to UART data register.
It turned out that USB (when cable was connected) started to generate events before main function had time to open UART, then when main started configuration, UART already had something in the TX FIFO and during configuration BUSY interrupt fired.
So in real usage it will not be a problem I guess.

PR #2777, opened 2022-02-18
hw/mcu/da1469x: Add support for both edges GPIO interrupt trigger

kasjer:
While hardware can be setup to detect only rising or falling
edge it is possible to add HAL_GPIO_TRIG_BOTH support
handled by software.

This re-arms edge of interrupt in GPIO ISR to detect next change.

### DIFF
--- a/hw/mcu/dialog/da1469x/src/hal_gpio.c
+++ b/hw/mcu/dialog/da1469x/src/hal_gpio.c
@@ -83,6 +83,7 @@ struct hal_gpio_irq {
     int pin;
     hal_gpio_irq_handler_t func;
     void *arg;
+    hal_gpio_irq_trig_t trigger;
 };
 
 static struct hal_gpio_irq hal_gpio_irqs[HAL_GPIO_MAX_IRQ];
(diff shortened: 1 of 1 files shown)


wes3 (commented):
These comments may not be completely thought out or I could be missing something here. It seems possible to miss an edge here as you clear and then reset the trigger (between calling WKUP_CLEAR_PX and hal_gpio_trigger_irq_on_next_edge). Is this possible?

Well, I will go ahead and approve. I assume you do not care about this case; the interrupt occurred and the driver will get notified that something changed and you do not care that the edge changed during the interrupt. Looks like it was approved. :-)

agross-korg (approved):
So in general, this is good enough to set the next edge that will trigger the interrupt.  However, it won't tell you how many edges occurred between the previous irq and this one.  Due to latency and priority, the edge could have transitioned again.  This code handles a transition while setting the edge, which is good.  So overall, this is ok.

kasjer:
@wes3 @andrzej-kaczmarek Thanks for the review.
I was just trying to put it in words what Andy wrote so nicely.
Event with hardware both edge detection it's not possible to detect all edges to to processing time limit.
Real concern is not to leave detector in a state that slow change will be missed (that could happen if rising edge was expected but hi state was already present and transition to lo state could be missed).
Right now I thing we should be ready for this. If rapid changes happen during the process of decision it may happen that edge will be detected but that should not spoil anything.

PR #2778, opened 2022-02-25
hw/mcu/nrf: Add common startup code

andrzej-kaczmarek:
This adds common startup code to nRF MCU package so it can be reused by
BSP. This avoids multiple duplicates of the same file in all BSPs.

Also proper configuration is added so standard and split startup code
is not included in build at the same time. This fixes issue when link
option -whole-archive is used.


kasjer (approved):
Not tested, LGTM

andrzej-kaczmarek:
fixed build issue

PR #2779, opened 2022-02-25
hw/drivers/rtt: Update to rev 24346

andrzej-kaczmarek:


kasjer (approved):
LGTM provided that this is pure code upgrade provided by SEGGER

andrzej-kaczmarek:
only paths in #includes were adjusted, otherwise it's an exact copy

PR #2782, opened 2022-02-28
usb msc fat view

kasjer:
Code provides implementation of USB MSC protocol.

When msc_fat_view package is included in provides USB descriptor with MSC function.
Code simulates FAT disk with custom files that can provide information about MCU.
For now readme.txt with some basic data.
Two files with view of SLOT0 are also provided.

It is base for firmware updated over USB when drag/drop could be used.


kasjer:
Another version will come soon

PR #2786, opened 2022-03-05
TinyUSB and external pull up on D+

kasjer:
For black_vet6 USB_DP_HAS_EXTERNAL_PULL_UP is set to 1 to reflect hardware design.
Without this USB host may not noticed that device rebooted and re-enumeration is required

For ST based hardware (both synopsys or stm32_fsdev) external pullup handling was working in
mynewt OS but when USB was built into bootloader GPIO handling of D+ was to fast because
os_time_delay() did not work.
For bootloader case os_cpu_time_delay is used.


utzig (approved):
LGTM

PR #2788, opened 2022-03-05
tinyusb: Add hardware initialization for stm32wb55

kasjer:
TinyUSB uses same driver for STM32WB55 as for STM32F1 but there
are differences in interrupts and power settings that call
for separate package.


utzig (approved):
LGTM

PR #2795, opened 2022-03-16
Fix flash hole in Adafruit Feather nrf52 BSP

wtfuzz:
The bootloader linker script currently has a flash length of `0x4000`, while the image linker has an origin of `0x8000`.

This closes the hole and doubles the usable bootloader space. This allows `mcuboot` with serial uploads enabled to fit.


sjanc (changes requested):
this area is used for reboot log,  check bsp.yml

if you need more space for bootloader you'd have to shrink something else

wtfuzz:
Thanks for pointing that out. I've updated the linker script and bsp.yml to use a 24kB bootloader region, and shrunk the reboot log region to 8kB similar to the pinetime layout.

sjanc (approved):

PR #2798, opened 2022-03-16
hw/drivers/flash/spiflash: add P25Q32H chip

StarGate01:
Similar to #2582 , this adds support for the P25Q32H SPI NOR flash chip found in some P8(b) smartwatch variants. 

Datasheet: https://www.puyasemi.com/uploadfiles/2018/08/201808071524332433.pdf , see p. 70 for IDs.


kasjer (approved):
Looks, good, thanks for doc link and contribution

PR #2799, opened 2022-03-16
sensors/lis2dh12: Minor change to support lis2dk10

benmccrea:
Change to use read-modify-write when setting bit 7 of CTRL_REG0. This allows the driver to be used with lis2dk10, since reserved bit values in this register location on lis2dk10 are different and must not be changed. 

The lis2dh12 and lis2dk10 are almost identical and main difference is that lis2dh12 supports both SPI and I2C, whereas lis2dk10 supports only I2C. There is just 1 register map difference; lis2dh12 has CTRL_REG0 documented in the datasheet and the same register location in lis2dk10 is documented as reserved.


hw/drivers/sensors/lis2dh12/src/lis2dh12.c
```diff
@@ -640,13 +640,20 @@ int
 lis2dh12_pull_up_disc(struct sensor_itf *itf, uint8_t disconnect)
 {
     uint8_t reg;
+    int rc;
 
-    reg = 0;
+    rc = lis2dh12_readlen(itf, LIS2DH12_REG_CTRL_REG0, &reg, 1);
+    if (rc) {
```
vrahane, line 646: @benmccrea I think it would be better to make the read of the bit conditional given it is not applicable for lis2dh12.

benmccrea: @vrahane Thanks, I did it this way to make it work for both lis2dk10 and lis2dh12. If you want, I can add a syscfg and use the old behavior unless the syscfg is set.

vrahane: Yes, I think that would be better given the register is reserved for the lis2dk10

benmccrea: @vrahane I added LIS2DH12_LIS2DK10_COMPAT. Let me know if you have any other concerns or can approve.

vrahane (approved):
LGTM

PR #2813, opened 2022-04-19
MbedTLS: Suppress gcc 10 warning

kasjer:
Debug build compiled with gcc 10.x (arm and riscv) produces
warning:
repos/apache-mynewt-core/crypto/mbedtls/src/hmac_drbg.c:134:11: error: 'ret' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  134 |     return( ret );

This warning is incorrect and now is suppressed.

### DIFF
--- a/crypto/mbedtls/pkg.yml
+++ b/crypto/mbedtls/pkg.yml
@@ -25,5 +25,7 @@ pkg.keywords:
     - ssl
     - tls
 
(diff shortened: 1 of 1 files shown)


utzig:
## Style check summary

sjanc (approved):

PR #2814, opened 2022-04-19
bsp/olimex-pic32-hmz144: Add missing oscillator frequency

kasjer:
SYSTEM_CLOCK_OSC_FREQ was missing from bsp definition
resulting in build error.


utzig:
## Style check summary

PR #2826, opened 2022-04-30
stm32 wb hal upgrade

kasjer:
STM hal files were quite old (1.2)
This upgrade is needed due to bug in hardware where
FLASH_FLAG_OPTVERR is always on after boot and was not really cleared,
resulting in flash access failures.
It would be possible to clear this bit in mynewt but since
ST already fixed this in their repo there is no need for local fix.


utzig:
## Style check summary

sjanc (approved):

PR #2831, opened 2022-05-09
bsp/p-nucleo-wb55-dongle: Update BUTTON_1 in bootloader

kasjer:
When BUTTON_1_AS_STM32_DFU is set (that should be used only in bootloaer)
hal_bsp redirect execution to STM32 DFU bootloader to allow writing
bootloader or application.
Function boot_preboot was defined always even if functionality was not enabled.
This prevented alternative boot_preboot() functionality that can be provided
by other package.
Now hal_bsp.c only provides boot_preboot() when BUTTON_1_AS_STM32_DFU is set.


utzig:
## Style check summary

sjanc (approved):

PR #2833, opened 2022-05-09
tinyusb/cdc: Add fifo size configuration option

kasjer:
So far CDC RX/TX fifo size was matching endpoint size.
Now it is possible to specify those in syscfg.yml.
This can improve console logging.
Default FIFO sizes remain unchanged.


utzig:
## Style check summary

sjanc (approved):

PR #2834, opened 2022-05-10
tinyusb: USB function enable rework

kasjer:
USB function in TinyUSB stack is enabled by one of the following
definitions: CFG_TUD_CDC, CFG_TUD_HID, CFG_TUD_MSC, CFG_TUD_MIDI,
CFG_TUD_MSC, CFG_TUD_VENDOR, CFG_TUD_USBTMC, CFG_TUD_DFU_RT, CFG_TUD_DFU,
USBD_BTH need to be set to non-zero value.

So far some of those could be set in syscfg (definition were provided
in std_descriptors package). Following syscfg values were there:
USBD_HID, USBD_CDC, USBD_MSC, USBD_DFU, UDBD_BTH.

In order to enable standard functionality two things had to be set.
For example to have USB console one had to include hw/usb/tinyusb/cdc_console
(this could be done by settings CONSOLE_USB: 1) and separate settings in
target or application was needed USBD_CDC: 1.
For BT one had to set BLE_TRANSPORT_HS: usb and USBD_BTH: 1.

Now that any package can overwrite setting that is defined but no set
it is possible to for package that implements some USB functionality
to enabled it in TinyUSB stack.

This change leaves all previously defined USBD_xxx: values to not defined
and packages can just enabled it when they are included.

There is no need to have USBD_CDC: 1 when CONSOLE_USB: 1 is present.
For BTH separate commit will go to nimble repository.

### DIFF
--- a/hw/usb/tinyusb/cdc_console/syscfg.yml
+++ b/hw/usb/tinyusb/cdc_console/syscfg.yml
@@ -23,6 +23,7 @@ syscfg.defs:
         value: '"Mynewt console"'
 
 syscfg.vals:
+    USBD_CDC: 1
 
 syscfg.restrictions:
     - "USBD_CDC"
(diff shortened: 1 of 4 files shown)


andrzej-kaczmarek (approved):

sjanc (approved):

utzig:
## Style check summary

PR #2837, opened 2022-05-13
mempool: Fix check for duplicate free

kasjer:
When OS_MEMPOOL_CHECK was set to 1, code tried to
verify that block is not already in free list.
If during iteration over free list, memory block
was taken from the pool (from interrupt or higher
priority task) and already filled with data
mb_next could point to anything resulting in
MPU or hard fault exception.

Now critical section is added to prevent list
modification during check.


utzig:
## Style check summary

sjanc (approved):

PR #2839, opened 2022-05-17
nrf5340: avoid have 2 interrupts for each IPC message

bvdberg:
The Nimble IPC messages are written in 2 parts: the cmd byte and the rest.
It is unneccesary to already interrupt the other core for the first byte,
since it will be interrupted by the 2nd write already.
(a 2nd patch to Nimble will use this new function)

### DIFF
--- a/hw/drivers/ipc_nrf5340/include/ipc_nrf5340/ipc_nrf5340.h
+++ b/hw/drivers/ipc_nrf5340/include/ipc_nrf5340/ipc_nrf5340.h
@@ -21,6 +21,7 @@
 #define _HW_DRIVERS_IPC_NRF5340_H
 
 #include <stdint.h>
+#include <stdbool.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -69,6 +70,21 @@ void ipc_nrf5340_recv(int channel, ipc_nrf5340_recv_cb cb, void *user_data);
  * @return            0 on success and negative error on failure
  */
 int ipc_nrf5340_send(int channel, const void *data, uint16_t len);
+
+/**
+ * Sends data over specified IPC channel. IPC uses ring buffer for data passing.
+ * If IPC_NRF5340_BLOCKING_WRITE is not enable and there is not enough space to
+ * store data SYS_ENOMEM error is returned.
+ *
+ * @param channel     IPC channel number to send on
+ * @param data        Data to be sent over IPC. If this is NULL only signal is
+ *                    sent
+ * @param len         Data length
+ * @param last        If this is the last send, so other side can be notified
+ *
+ * @return            0 on success and negative error on failure
+ */
+int ipc_nrf5340_write(int channel, const void *data, uint16_t len, bool last);
 
 /**
  * Reads data from IPC ring buffer to specified flat buffer. Should be used only
--- a/hw/drivers/ipc_nrf5340/src/ipc_nrf5340.c
+++ b/hw/drivers/ipc_nrf5340/src/ipc_nrf5340.c
@@ -392,6 +392,12 @@ ipc_nrf5340_recv(int channel, ipc_nrf5340_recv_cb cb, void *user_data)
 int
 ipc_nrf5340_send(int channel, const void *data, uint16_t len)
 {
+    return ipc_nrf5340_write(channel, data, len, true);
+}
+
+int
+ipc_nrf5340_write(int channel, const void *data, uint16_t len, bool last)
+{
     struct ipc_shm *shm;
     uint16_t frag_len;
     uint16_t space;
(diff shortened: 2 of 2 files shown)


utzig:
## Style check summary

### Our coding style is [here!](https://github.com/apache/mynewt-core/blob/master/CODING_STANDARDS.md)


#### hw/drivers/ipc_nrf5340/src/ipc_nrf5340.c
<details>

```diff
@@ -405,12 +405,14 @@
         /* assert since that will always fail for non-blocking write
          * indicating use error
          */
-        if (space == 0) return SYS_ENOMEM;
+        if (space == 0) {
+            return SYS_ENOMEM;
+        }
 #endif
     } while (space == 0);
 
     ipc_nrf5340_shm_write(shm, &data, 1);
-    // dont notify other side! (will be done by later calls to ipc_nrf5340_send)
+    /* dont notify other side! (will be done by later calls to ipc_nrf5340_send) */
 
     return 0;
 }
```

</details>

hw/drivers/ipc_nrf5340/include/ipc_nrf5340/ipc_nrf5340.h
```diff
@@ -56,6 +56,18 @@ void ipc_nrf5340_reset(void);
...
+ * Sends single byte data  over specified IPC channel. IPC uses ring buffer for data passing.
+ * If IPC_NRF5340_BLOCKING_WRITE is not enable and there is not enough space to
+ * store data SYS_ENOMEM error is returned. This function does not notify the other side
+ *
+ * @param channel     IPC channel number to send on
+ * @param data        Data to be sent over IPC.
+ *
+ * @return            0 on success and negative error on failure
+ */
+int ipc_nrf5340_send_no_notify(int channel, uint8_t data);
```
sjanc, line 69: this looks like very specific case (1 byte of data),  maybe add more generic  "int ipc_nrf5340_write(int channel, const void *data, uint16_t len)" instead?     and then for completeness also  "void ipc_nrf5340_notify(int channel)" that would only notify with no data?

sjanc (changes requested):
I like the idea of reducing number of IPC interrupts,  some comments on API though

hw/drivers/ipc_nrf5340/include/ipc_nrf5340/ipc_nrf5340.h
bvdberg: Yes that would also be fine, I'll change to that

kasjer: Maybe `int ipc_nrf5340_write(int channel, const void *data, uint16_t len, bool last)`
then current `ipc_nrf5340_send()` could just call it with **true**;
and new function could grab current code of ipc_nrf5340_send() and have something like that
```c
if (last) {
    NRF_IPC->TASKS_SEND[channel] = 1; 
}
```
just an idea to thing about

bvdberg: I think this is indeed better, less functions on the API

bvdberg: I pushed the new version

kasjer:
@bvdberg I think that ipc_nrf5340_send() is already used by some users (not only by nimble and what is in mynewt core) that's why I propose to have another function and send ipc_nrf5340_send() would just used it.
Changing prototype is not a way to go I think.
I'm sorry if it was not clear in my previous comment.

bvdberg:
sounds fair, i have added ipc_nrf5340_write() with the bool last argument. The ipc_nrf5340_send() now uses write(.., true)

hw/drivers/ipc_nrf5340/include/ipc_nrf5340/ipc_nrf5340.h
```diff
@@ -68,7 +69,22 @@ void ipc_nrf5340_recv(int channel, ipc_nrf5340_recv_cb cb, void *user_data);
  *
  * @return            0 on success and negative error on failure
  */
-int ipc_nrf5340_send(int channel, const void *data, uint16_t len);
+int ipc_nrf5340_send(int channel, const void *data, uint16_t len, bool last);
```
kasjer, line 72: oops, still modified :)

bvdberg:
doh! I did build though. Ok fixed :)

kasjer (commented):
LGTM, let's wait for @sjanc

utzig:
## Style check summary

sjanc (changes requested):
This will break blocking send if data is bigger than internal shm size... I think notification should stil be done from withing loop (note that for non-blocking config this loop iterates only once anyway)

kasjer:
> This will break blocking send if data is bigger than internal shm size... I think notification should stil be done from withing loop (note that for non-blocking config this loop iterates only once anyway)

Maybe changing from:
```c
    NRF_IPC->TASKS_SEND[channel] = 1;
```
to:
```c
    if (last || len > space) {
        NRF_IPC->TASKS_SEND[channel] = 1;
    }
```

utzig:
## Style check summary

bvdberg:
Well spotted. Modified as @kasjer proposed.

utzig:
## Style check summary

bvdberg:
I made all the changes, new it should be perfect ;)

sjanc (approved):

PR #2841, opened 2022-05-19
tinyusb: include bsp.h in FAT View

kasjer:
With this inclusion it is possible to use pin
defined in bsp.h (i.e. BUTTON_1) as pin for enabling
MSC FAT View in bootloader.


utzig:
## Style check summary

sjanc (approved):

PR #2842, opened 2022-05-25
da1469x: Specify custom location of da1469x_flash_loader

jrotkiewicz:
User may now use the mynewt value FLASH_LOADER_TARGET in bootloader target
to specify the da1469x_flash_loader target location.


jrotkiewicz:
@kasjer

PR #2846, opened 2022-06-03
sensors/bmp388: Add compatibility for bmp390

benmccrea:
Allow sensors/bmp388 driver to accept the bmp390 chip ID. 

The Bosch BMP390 an updated pressure sensor which is compatible with the older BMP388. The only difference from a driver perspective is the chip ID, where the BMP390 chip ID is 0x60 and BMP388 is 0x50.

### DIFF
--- a/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
+++ b/hw/drivers/sensors/bmp388/include/bmp388/bmp388.h
@@ -65,6 +65,8 @@ extern "C" {
 
 /**\name BMP3 chip identifier */
 #define BMP3_CHIP_ID      0x50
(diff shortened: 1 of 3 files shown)


vrahane (commented):
The code looks fine but I think it’s better if you also specify it in the pkg.yml file.
https://github.com/apache/mynewt-core/blob/838519bffb5e575c6e46a5c2811af645ab601bfa/hw/drivers/sensors/bmp388/pkg.yml#L26

benmccrea:
@vrahane Thanks. I added bmp390 to the pkg.yml file. I also fixed a bug (chip ID test is done in 2 other places).

PR #2850, opened 2022-06-14
mempool: Fix OS_MEMPOOL_SIZE when OS_MEMPOOL_GUARD is on

kasjer:
OS_MEMPOOL_SIZE returned incorrect memory size for mpool buffer when
OS_MEMPOOL_GUARD was enabled.
OS_MEMPOOL_GUARD requires extra space between memory blocks and
OS_MEMPOOL_SIZE (and OS_MEMPOOL_BYTES) returned values that did not
include this extra space.
During os_mempool_init_internal() memory outside expected buffer
was clobbered.


rymanluk (approved):
LGTM :)

mkiiskila (approved):
Thanks for fixing this!

PR #2858, opened 2022-07-04
hw/drivers/flash/spiflash: Add option to ignore JEDEC ID

StarGate01:
This PR introduces three new **optional** configuration parameters: `SPIFLASH_IGNORE_MANUFACTURER`, `SPIFLASH_IGNORE_MEMORY_TYPE`, and `SPIFLASH_IGNORE_MEMORY_CAPACITY`. (Default: `0`).

These parameters selectively disable the whitelist checks of the SPI Flash JEDEC IDs.

### Rationale

I am currently [porting the InfiniTime OS](https://github.com/InfiniTimeOrg/InfiniTime/pull/1050) (written for the [PineTime smartwatch](https://www.pine64.org/pinetime/)) to a popular Chinese smartwatch series called "P8". `mynewt-core` is used for the bootloader.

These P8 watches come in a whole zoo of variants, each being compatible with the firmware but containing sometimes a bit different chips due to the silicon shortage and general chip availability on the global market. This is usually no problem to handle.

However, the SPI driver of `mynewt-core` enforces a strict whitelist on the reported SPI Flash JEDEC ID. I see how this is useful in normal applications where the hardware configuration is known in detail and compatibility has to be ensured, however this leads to my bootloader refusing to work on prior unknown / new smartwatch variants, bricking them.

In the past, others and I have added definitions for SPI IDs (e.g. #2798 , #2582 ), but I can't keep up with all the models and I don't want to brick any more watches due to a purely whitelisting (not technical) issue. 

I am open to suggestions, I think these optional configuration parameters would be quite useful to have.

### DIFF
--- a/hw/drivers/flash/spiflash/src/spiflash.c
+++ b/hw/drivers/flash/spiflash/src/spiflash.c
@@ -1374,12 +1374,24 @@ spiflash_identify(struct spiflash_dev *dev)
          * different pins, or of different type.
          * It is unlikely that flash depended packaged will work correctly.
          */
-        assert(manufacturer == supported_chips[0].fc_jedec_id.ji_manufacturer &&
-               memory_type == supported_chips[0].fc_jedec_id.ji_type &&
-               capacity == supported_chips[0].fc_jedec_id.ji_capacity);
-        if (manufacturer != supported_chips[0].fc_jedec_id.ji_manufacturer ||
-            memory_type != supported_chips[0].fc_jedec_id.ji_type ||
-            capacity != supported_chips[0].fc_jedec_id.ji_capacity) {
+        assert(MYNEWT_VAL(SPIFLASH_IGNORE_MANUFACTURER) ||
+            manufacturer == supported_chips[0].fc_jedec_id.ji_manufacturer);
+        if (!(MYNEWT_VAL(SPIFLASH_IGNORE_MANUFACTURER) ||
+              manufacturer == supported_chips[0].fc_jedec_id.ji_manufacturer)) {
+            rc = -1;
+            goto err;
+        }
+        assert(MYNEWT_VAL(SPIFLASH_IGNORE_MEMORY_TYPE) ||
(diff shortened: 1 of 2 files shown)


StarGate01:
@kasjer Is this something you would consider merging? The feature would be opt-in.

kasjer:
Hi, I am out of office, idea is good so definitely I will support it, I
have not seen the code, I will take a look next week.

Thanks for your submittion.
Jerzy

czw., 21 lip 2022, 10:17 użytkownik Christoph Honal <
***@***.***> napisał:

kasjer:
@StarGate01 maybe you could change your commit a little bit to be more consistent with mynewt style that requires opening brace `{` to be on the same line as `if`.
Since new syscfg values have default value 0 it can be safely used in plain `if` (not `#if`)
Instead of having:
```c
#if !MYNEWT_VAL(SPIFLASH_IGNORE_MANUFACTURER)
        assert(manufacturer == supported_chips[0].fc_jedec_id.ji_manufacturer);
        if (manufacturer != supported_chips[0].fc_jedec_id.ji_manufacturer) {
            rc = -1;
            goto err;
        }
#endif
```
we would have:
```c
        assert(!MYNEWT_VAL(SPIFLASH_IGNORE_MANUFACTURER) &&
               manufacturer == supported_chips[0].fc_jedec_id.ji_manufacturer);
        if (!MYNEWT_VAL(SPIFLASH_IGNORE_MANUFACTURER) &&
            manufacturer != supported_chips[0].fc_jedec_id.ji_manufacturer) {
            rc = -1;
            goto err;
        }
```
generate code would be the same and compiler would check syntax that otherwise would not be checked in some build depending on ignore flags and mynewt style would be preserved without additional pre-processor conditions added for style sake only.

You can amend your commit and force push.

StarGate01:
@kasjer Thank you for the style comments; I have applied your suggestions.

kasjer:
@StarGate01 thanks for contributions

PR #2863, opened 2022-08-18
tinyusb: Allow for several CDC interfaces

kasjer:
This adds way to have several CDC interfaces (COM ports or /dev/ttyACMx devices) on one USB devices.
This will allow to have USB console with other serial based interface like HCI over serial connection.

cdc package is added that routes TinyUSB CDC devices callbacks to multiply interface implementations.

Some USB string descriptor handling is also updated so string indices for interfaces and DFU slot are no longer hardcode.

This also adds standard descriptor for CDC HCI that can be used by future nimble transport.

### DIFF
--- a/hw/usb/tinyusb/std_descriptors/include/tusb_config.h
+++ b/hw/usb/tinyusb/std_descriptors/include/tusb_config.h
@@ -72,11 +72,54 @@ extern "C" {
 #define CFG_TUD_ENDPOINT0_SIZE   MYNEWT_VAL(USBD_EP0_SIZE)
 
 /* ------------- CLASS ------------- */
-#if MYNEWT_VAL(USBD_CDC)
-#define CFG_TUD_CDC              MYNEWT_VAL(USBD_CDC)
+/*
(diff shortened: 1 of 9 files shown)


hw/usb/tinyusb/std_descriptors/include/tusb_config.h
```diff
@@ -118,7 +153,13 @@ extern "C" {
 #define CFG_CDC_CONSOLE          0
 #endif
 
-#define CFG_TUD_CDC              ((CFG_CDC) + (CFG_CDC_CONSOLE))
+#if MYNEWT_VAL(USBD_CDC_HCI)
```
wes3, line 156: Is USBD_CDC_I defined in any syscfg.yml file? I could not find it but maybe it is in a different PR or commit or I missed it?

wes3 (commented):
Built and tested with CDC console and it works on my platform. Had one comment.

wes3 (approved):
LGTM. I did not test HCI over usb but I did test CDC console and it works.

sjanc (approved):
tested HCI over USB with blehci on linux, works great!

PR #2866, opened 2022-08-25
riscv: Fix initialized data

kasjer:
Startup script has only one loop that copies initialized data and
code that should execute from RAM.
It implies that variables in .data .ram_text .srodata .sdata and such
has same layout in flash and in RAM.
Those data are packed by default in flash so output sections
(.data .srodata .sdata) if they start with something that requires
alignment change may end up with wrong data being copied to
statically initialized data.
One way to prevent such case is to copy each output section in separate
loop. But it is also possibly to add output section attribute ALIGN_WITH_INPUT
and then linker will fill necessary space in flash image.

### DIFF
--- a/hw/bsp/hifive1/hifive1.ld
+++ b/hw/bsp/hifive1/hifive1.ld
@@ -128,14 +128,14 @@ SECTIONS
     PROVIDE( _data = . );
   } >ram AT>flash
 
-  .data          :
+  .data          :  ALIGN_WITH_INPUT
   {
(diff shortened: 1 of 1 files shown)


mkiiskila (approved):
Changes lgtm.

I would have also considered removing .lalign section, as I can't see a reason for it's existence. .dalign section also seems unneeded, we could move the _data symbol creation inside .data output section. 

However, I'm ok with changes as they are.

PR #2869, opened 2022-09-06
hw/mcu: Add family and series defs to syscfg

andrzej-kaczmarek:
Those allow for conditional configurations to be implemented.
It may be possible to retrieve the same information via vendor provided
defs in header files, but with syscfg defs we can also use this in pkg
or syscfg deps.


andrzej-kaczmarek:
creating as a draft for now to think about it more :)

sjanc (approved):

PR #2870, opened 2022-09-06
coremark: Compute number of iteration automatically by default

kasjer:
Default value 200 for COREMARK_ITERATIONS is not good enough for most BSP/MCUs.
Making default value 0 makes coremark pre-compute number of iterations needed for 10-20s of benchmark time.

this change will make most (maybe all) BSP work out of the box.


utzig (approved):
LGTM

sjanc (approved):

PR #2871, opened 2022-09-12
hw/bsp: Make all scripts executable

sjanc:
Non-executable scripts prevents newt commands to work correctly on linux.


kasjer (approved):
LGTM

PR #2873, opened 2022-09-12
pic32 flash access fixes

kasjer:
This fixes:
- unaligned address writes that are not needed (FCB uses unaligned writes).
- writes that are not multiply of 4 are now allowed
- flash read uses uncached segment to avoid wrong data read after writes (CRC failure in FCB usage)

### DIFF
--- a/hw/mcu/microchip/pic32mz/src/hal_flash.c
+++ b/hw/mcu/microchip/pic32mz/src/hal_flash.c
@@ -104,23 +104,47 @@ pic32mz_flash_read(const struct hal_flash *dev, uint32_t address,
     return 0;
 }
 
+/**
+ * Function returns address to 4 byte aligned data
+ * @param src      pointer to source data
+ * @param aligned_buffer address of aligned buffer to use in case src in unaligned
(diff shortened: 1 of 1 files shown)


hw/mcu/microchip/pic32mz/src/hal_flash.c
```diff
@@ -104,23 +104,45 @@ pic32mz_flash_read(const struct hal_flash *dev, uint32_t address,
...
+ * @param num_bytes number of bytes to provide.
+ * @return src if it is 4 byte aligned otherwise aligned_buffer
+ */
+const uint32_t *
+aligned_ptr(const void *src, uint32_t aligned_buffer[4], uint32_t num_bytes)
+{
+    if (((uint32_t)src & 3) == 0) {
+        return (const uint32_t *)src;
+    }
+    if (num_bytes > sizeof(aligned_buffer)) {
```
sjanc, line 121: hmm this sizeof will return 4  (as this is array function parameter)

kasjer: indeed, fixed now

sjanc (approved):

PR #2876, opened 2022-09-16
nrfx_glue: Support atomic and static assert

kasjer:
Some nrfx functionality uses NRFX_STATIC_ASSERT and NRFX_ATOMIC_xxxx functions.

Those were not provided in mynewt even though they are present in Nordic nrfx_glue.h template.

Now mote nrfx can be use if needed.


mlaz (approved):
LGTM

PR #2885, opened 2022-09-26
da1469x: Add option to not disable watchdog on init

agross-korg:
This patch adds a facility to control whether or not the Dialog watchdog is disabled when the hal_watchdog_init function is being called.  There are some use cases where the watchdog needs to remain running from a previous application or bootloader.

### DIFF
--- a/hw/mcu/dialog/da1469x/syscfg.yml
+++ b/hw/mcu/dialog/da1469x/syscfg.yml
@@ -454,6 +454,11 @@ syscfg.defs:
         restrictions:
             - '!GPADC_BATTERY'
 
+    MCU_WATCHDOG_DISABLE_ON_INIT:
+        description: >
+          'Disable watchdog on init'
+        value: 1
+
 syscfg.vals:
     OS_TICKS_PER_SEC: 128
 
--- a/hw/mcu/dialog/da1469x/src/hal_watchdog.c
+++ b/hw/mcu/dialog/da1469x/src/hal_watchdog.c
@@ -27,7 +27,10 @@ int
 hal_watchdog_init(uint32_t expire_msecs)
 {
     SYS_WDOG->WATCHDOG_CTRL_REG = SYS_WDOG_WATCHDOG_CTRL_REG_WDOG_FREEZE_EN_Msk;
+
+#if MYNEWT_VAL(MCU_WATCHDOG_DISABLE_ON_INIT)
     GPREG->SET_FREEZE_REG |= GPREG_SET_FREEZE_REG_FRZ_SYS_WDOG_Msk;
+#endif
 
 #if MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, RCX)
     g_hal_watchdog_reload_val = expire_msecs / 21;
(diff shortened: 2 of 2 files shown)


hw/mcu/dialog/da1469x/syscfg.yml
```diff
@@ -454,6 +454,11 @@ syscfg.defs:
         restrictions:
             - '!GPADC_BATTERY'
 
+    WATCHDOG_DISABLE_ON_INIT:
+        description: >
+          'Enable watchdog on init'
+        value: 1
```
kasjer, line 460: Name **WATCHDOG_DISABLE_ON_INIT** is not really consistent with description. One can wander which one is true.
Dialog freezing watchdog is quite unique future (usually when watchdog is enabled it's impossible to stop it till reset).
I would name it rather more explicit like WATCHDOG_FREEZE_ON_INIT but that's just me (and I will not insist).
But changing description to something like 'Freez watchdog on init, watchdog will be unfrozen by hal_watchdog_enable()' could be more descriptive.

andrzej-kaczmarek: I'd also prefix this option with `MCU_` as it will be more consistent with other syscfg options that are specific to particular MCU rather than a generic ones

agross-korg: Yeah I kind of went back and forth with the name as well.  So some watchdogs are auto-started on reset.  Some are not.  Some are like doomsday devices and you can't freeze them.  This Dialog one is kind of an odd bird.  The implementation we currently have freezes the watchdog during init.  So to keep the default behavior of the driver, I added the DISABLE_ON_INIT and made it the default (=1). 

It didn't matter to me whether we were freezing it or not on init, what mattered to me was that this was the behavior that was coded into the init function and that people are currently used to.  I am quite happy to just remove the freeze altogether.

agross-korg: The background reason of why I am doing any of this.... So the watchdog is enabled coming out of the ROM bootloader (either by reset or otherwise).  The mcuboot bootloader was freezing it and leaving us in a state where the watchdog was not running for the entiretly of the bootloader.  During production, we are soldering batteries on to the PCBs and the intermittent voltage connection during soldering was allowing chips to boot and then they get hung with a power fluctuation.  The only way to fix these was to somehow manage to attach via JTAG or disconnect the battery or let the battery run out.

Leaving the watchdog running, it could recover itself.

agross-korg: Even if we re-enabled the watchdog, there is a window where a power fluctuation could hang us inbetween the _init and the _enable of the watchdog.

agross-korg: Updated both and force pushed change.

kasjer (approved):
This is interesting it looks like Dialog watchdog works according to API description. I guess for other MCUs watchdog is started when `hal_watchdog_init()` is called even though in the API description of `hal_watchdog_init()` it clearly states that `hal_watchdog_enable()` should be called. So to me it looks like mcuboot code is missing call to this function. Function `hal_watchdog_enable()` is called in **os.c** and **fl.c** but not in mcuboot and probably it works because most init functions start watchdog without need to call enable explicitly.
Point about small window of time when problem can show up is still valid so for me this looks good.

PR #2890, opened 2022-09-29
Add ETH support to PIC32 boards

kasjer:
- pic32mz_periph_create() now creates eth device if ETH_0 is specified
- olimex-pic32-hmz144 now includes config settings for RMII/MII
- olimex-pic32-emz64 includes RMII settings and also gets configuration for on-bard LAN8710 (interrupt pin, and reset pin)


hw/mcu/microchip/pic32mz/src/pic32mz_periph.c
```diff
@@ -372,11 +384,23 @@ pic32mz_periph_i2c_devs(void)
     }
 }
 
+static void
+pic32mz_periph_create_eth(void)
+{
+#if MYNEWT_VAL(ETH_0)
+    int rc;
+    rc = pic32_eth_init(&eth0_cfg);
+    assert(0);
```
sjanc, line 393: this looks odd,   assert(rc == 0) maybe?

sjanc (changes requested):

hw/mcu/microchip/pic32mz/src/pic32mz_periph.c
kasjer: Polishing before commit not always works as intended.
Fixed.

sjanc (approved):

PR #2894, opened 2022-10-04
mcu/nrf5340: Enable GPIOTE, TIMERx for nrfx

kasjer:
This just enables some peripherals so nrfx functions will be available.


kasjer:
I change my mind about this. All can be accomplished using pkg.cflags at application level

PR #2896, opened 2022-10-05
ublox-bmd-345: Update syscfg to not use defunct configs

sjanc:
Some configuratio were renamed from _LL_ to _FEM_.


kasjer (approved):
LGTM, tested to work

PR #2899, opened 2022-10-11
mcu/nrf5340: Add secure functions package

kasjer:
New package will contain secure functions.
Those functions will allow access to secure only peripherals.

To start MCU assignment to GPIO function tfm_gpio_pin_mcu_select
allows to redirect PIN to net core.

tfm_uicr_otp_read() allows to read UICR->OTP from non-secure code.

When tfm package build with bootloader or secure application
functions are treated as every other function.

When tfm is build with non-secure application function bodies
are not present in the build and are expected to be linked with
library that provides secure gateways.

tfm can be also build into bootloader in this case it's the
bootloader that provides secure functions.

ELF that provides implementation of secure functions must be
build with TFM_EXPORT_NSC: 1 syscfg value.

### DIFF
--- a/hw/mcu/nordic/nrf5340/pkg.yml
+++ b/hw/mcu/nordic/nrf5340/pkg.yml
@@ -29,6 +29,7 @@ pkg.deps:
     - "@apache-mynewt-core/hw/mcu/nordic"
     - "@apache-mynewt-core/hw/cmsis-core"
     - "@apache-mynewt-core/hw/hal"
+    - "@apache-mynewt-core/hw/mcu/nordic/nrf5340/tfm"
 
 pkg.deps.BUS_DRIVER_PRESENT:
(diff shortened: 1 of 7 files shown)


hw/mcu/nordic/nrf5340/syscfg.yml
```diff
@@ -126,6 +126,13 @@ syscfg.defs:
             Application code to runs in secure mode.
             This settings must have same value for bootloader and application build.
         value: 1
+    MCU_NET_GPIO:
```
sjanc, line 129: hmm there is already MCU_GPIO_NET

kasjer: Thanks, old MCU_GPIO_NET reused. Old code from startup that used it is now removed.

sjanc (approved):

PR #2902, opened 2022-10-12
hw/ipc_nrf5340: Some updates

andrzej-kaczmarek:


hw/drivers/ipc_nrf5340/src/ipc_nrf5340.c
```diff
@@ -470,6 +470,25 @@ ipc_nrf5340_available_buf(int channel, void **dptr)
     return 0;
 }
 
+uint16_t
+ipc_nrf5340_data_available_get(int channel)
```
sjanc, line 471: there is already ipc_nrf5340_available() for this

hw/drivers/ipc_nrf5340/include/ipc_nrf5340/ipc_nrf5340.h
```diff
@@ -131,6 +131,32 @@ uint16_t ipc_nrf5340_available(int channel);
  */
 uint16_t ipc_nrf5340_available_buf(int channel, void **dptr);
 
+/**
+ * Returns number of free space available in IPC ring buffer.
```
sjanc, line 151: copy-paste?  same as ipc_nrf5340_data_available_get() and doesn't match implementation

hw/drivers/ipc_nrf5340/src/ipc_nrf5340.c
andrzej-kaczmarek: see the last commit in this PR

sjanc: hmm fair enough,  but it could use ipc_nrf5340_shm_get_data_length() internally anyway ;)

andrzej-kaczmarek: I'll think about it ;)

sjanc (approved):

PR #2903, opened 2022-10-14
sys/console: Reset text attributes on init

kasjer:
When console has colors (for now only for software cursor) color or other text attributes can be changed from the default when device reboots.
Now escape sequnce to restore default text attributes is added to first time initialization of console.


andrzej-kaczmarek (approved):
works as advertised

PR #2908, opened 2022-10-22
tinyusb: Fix long file name handling

kasjer:
New directory entries with long file name were not handled correctly and could lead to stack corruption.
Now file names up to 78 characters are handled.
And longer file names should not result in device
misbehavior.


sjanc (approved):

andrzej-kaczmarek:
tested, works fine

PR #2913, opened 2022-12-03
loramac: fix broken rx1 interval

danielkucera:
`lora_mac_process_radio_rx_timeout` is called from `lora_mac_process_rx_win2_timeout` which is triggered when RX2/join2 window starts. The problem here is that there still can be unfinished rx1 window downlink transmission (delay between rx1 and rx2 window start is fixed to 1sec and the message air time can be longer).
In that case, we don't want to put radio to sleep - that would cancel the receive and the message would be dropped


kasjer (approved):
LGTM, thanks

PR #2915, opened 2022-12-22
apps: Allow stub/full implementation of packages in syscfg

m-gorecki:


apps/bleprph_oic/pkg.yml
```diff
@@ -30,10 +30,10 @@ pkg.deps:
     - "@apache-mynewt-nimble/nimble/host/services/gatt"
     - "@apache-mynewt-nimble/nimble/host/store/config"
     - "@apache-mynewt-nimble/nimble/transport"
-    - "@apache-mynewt-core/sys/console/full"
-    - "@apache-mynewt-core/sys/log/full"
+    - "@apache-mynewt-core/sys/console/"
```
andrzej-kaczmarek, line 33: remove trailing `/`

apps/bleprph_oic/syscfg.yml
```diff
@@ -25,6 +25,10 @@ syscfg.defs:
         value: 0
 
 syscfg.vals:
+    CONSOLE_MODE: full
```
andrzej-kaczmarek, line 28: I do not recall if we have `CONSOLE_MODE` instead of `CONSOLE_IMPLEMENTATION` for a reason, but if it's just for 'historical reasons' we should add `CONSOLE_IMPLEMENTATION` and deprecate `CONSOLE_MODE`.

@kasjer ^^ ?

sys/stats/pkg.yml
```diff
@@ -0,0 +1,31 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: sys/stats
+pkg.description: API common to both the full and stub log implementations.
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
```
andrzej-kaczmarek, line 23: https

andrzej-kaczmarek (changes requested):

apps/bleprph_oic/syscfg.yml
sjanc: I'm fine with renaming this to CONSOLE_IMPLEMENTATION to keep things consistent

kasjer: I would deprecate _MODE it was first attempt. Later on for logs _IMPLEMENTATION was chosen as it more clear what it does.

m-gorecki:
This PR is dependent on https://github.com/apache/mynewt-nimble/pull/1447

Target building tests are failing because blehci app uses defunct setting, which is changed in the PR linked above.

sys/console/syscfg.yml
```diff
@@ -17,10 +17,17 @@
...
         choices:
             - full
             - minimal
             - stub
+
+    # Defunct settings
+
+    CONSOLE_MODE:
+        description: Use 'CONSOLE_IMPLEMENTATION' instead
+        defunct: 1
```
sjanc, line 33: I think this should initially be "deprecated" instead of defunc

andrzej-kaczmarek (approved):

sjanc (approved):

PR #2955, opened 2023-03-07
sys/console: Add default CONSOLE_IMPLEMENTATION value

m-gorecki:
Until now CONSOLE_IMPLEMENTATION had no default value, which was causing build errors when implementation package wasn't specified.

### DIFF
--- a/sys/console/syscfg.yml
+++ b/sys/console/syscfg.yml
@@ -20,7 +20,7 @@ syscfg.defs:
     CONSOLE_IMPLEMENTATION:
         description: >
             Selects console implementation package full, minimal or stub.
-        value:
+        value: full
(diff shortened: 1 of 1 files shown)


sys/console/syscfg.yml
```diff
@@ -20,7 +20,7 @@ syscfg.defs:
     CONSOLE_IMPLEMENTATION:
         description: >
             Selects console implementation package full, minimal or stub.
-        value:
+        value: stub
```
andrzej-kaczmarek, line 23: I'd go for 'full' as it may be a bit counter-intuitive to have non-functional console by default

kasjer (approved):
LGTM

PR #2963, opened 2023-03-20
pic32 use spi_hal bus driver for SPI

kasjer:
This is simple refactor of hal_spi so it can be used with spi_hal bus driver.

When bus driver bus/drivers/spi_hal is added to the target pic32mz_periph_create() function creates proper devices
so applications using bus driver interface can be built.

### DIFF
--- a/hw/mcu/microchip/pic32mz/src/hal_spi.c
+++ b/hw/mcu/microchip/pic32mz/src/hal_spi.c
@@ -59,7 +59,7 @@ struct hal_spi {
     int                       rxcnt;
     hal_spi_txrx_cb           callback;
     void                      *arg;
-    const struct mips_spi_cfg *pins;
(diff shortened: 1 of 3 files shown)


hw/mcu/microchip/pic32mz/src/hal_spi.c
```diff
@@ -705,3 +705,10 @@ hal_spi_abort(int spi_num)
 
     return 0;
 }
+
+int
+hal_spi_init_hw(uint8_t spi_num, uint8_t spi_type,
+                const struct hal_spi_hw_settings *cfg)
+{
+    return hal_spi_init(spi_num, (void *)cfg, spi_type);
```
sjanc, line 713: nitpick: maybe cast to const void * ?

sjanc (approved):

PR #2972, opened 2023-04-04
Use syscfg setting for sysinit stage numbers

supervillain101:
This allows the target to rearrange package initialization order via syscfg override

### DIFF
--- a/hw/bus/pkg.yml
+++ b/hw/bus/pkg.yml
@@ -25,4 +25,4 @@ pkg.keywords:
 pkg.deps:
 
 pkg.init:
-    bus_pkg_init: 100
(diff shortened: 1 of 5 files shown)


andrzej-kaczmarek:
I'm not a big fan of allowing user to configure init priority for core packages. What is the reason you need to change it? If priorities are not correct because of some dependency it should be fixed instead. If you need to have some other init done before/after other init then we do support this via `$after` and `$before` syntax (https://github.com/apache/mynewt-newt/pull/457).

wes3 (approved):
LGTM. Is there any reason why we should not merge this one?

andrzej-kaczmarek:
as I said, I'm not a big fan of such changes but I won't block them from being merged - feel free to merge at your discretion

PR #2974, opened 2023-04-05
stm32l0: switch to GitHub sdk

kasjer:
mcu/stm32l0: Switch SDK to use external github repository

Two bsp had to have more ST HAL modules enabled to build


sjanc (commented):
this should also cleanup .rat-excludes and LICENSE of files that are being removed

kasjer:
> this should also cleanup .rat-excludes and LICENSE of files that are being removed

one global exclude for folder **stm32l0xx** changed to two files that still are based on ST BSD license

sjanc (approved):

PR #2983, opened 2023-04-17
hw/drivers: Add ipc_cmac driver

andrzej-kaczmarek:
This adds ipc_cmac driver which is basically cmac_driver moved from nimble repo and with some extra changes to make configuration easier.

The structure of image header for CMAC image is modified to allow host detect configuration of CMAC image. This means there's no need to sync configuration on M33 and CMAC targets manually, it's enough to configure CMAC image and M33 will detect configuration in runtime.

### DIFF
--- a/hw/drivers/ipc_cmac/diag/pkg.yml
+++ b/hw/drivers/ipc_cmac/diag/pkg.yml
@@ -17,20 +17,12 @@
 # under the License.
 #
 
-pkg.name: hw/mcu/dialog/cmac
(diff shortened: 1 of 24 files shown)


andrzej-kaczmarek:
requires https://github.com/apache/mynewt-nimble/pull/1486

sjanc:
looks ok but remove this tmp commit :-)

sjanc (approved):

PR #2990, opened 2023-05-04
lps33thw: add syscfg to make statistics optional

benmccrea:
Adds a syscfg, `LPS33THW_STATS`, to make use of statistics in the driver optional. The syscfg defaults to 1 (no modification to existing behavior). Some people will not care to use statistics for memory reasons.

I also want to note, the way statistics are currently implemented in the driver prevents it from being used with more than 1 sensor at a time, because `lps33thw_init` will attempt to register the statistics again the second time it is called. This will result in a sysinit panic:
```
    /* Initialise the stats entry */
    rc = stats_init(STATS_HDR(g_lps33thwstats),
        STATS_SIZE_INIT_PARMS(g_lps33thwstats, STATS_SIZE_32),
        STATS_NAME_INIT_PARMS(lps33thw_stat_section));
    SYSINIT_PANIC_ASSERT(rc == 0);

    /* Register the entry with the stats registry */
    rc = stats_register(dev->od_name, STATS_HDR(g_lps33thwstats));
    SYSINIT_PANIC_ASSERT(rc == 0);
 ```
It would be preferable to change this code so that each sensor has its own statistics, or if all sensors are to share the same statistics, then only call `stats_init` and `stats_register` for the first sensor.

### DIFF
--- a/hw/drivers/sensors/lps33thw/pkg.yml
+++ b/hw/drivers/sensors/lps33thw/pkg.yml
@@ -41,7 +41,7 @@ pkg.deps.BUS_DRIVER_PRESENT:
     - "@apache-mynewt-core/hw/bus/drivers/i2c_common"
     - "@apache-mynewt-core/hw/bus/drivers/spi_common"
 
-pkg.req_apis:
+pkg.req_apis.LPS33HW_STATS:
     - stats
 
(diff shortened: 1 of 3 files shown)


wes3 (approved):
LGTM

PR #2997, opened 2023-05-19
Add downloader selection

kasjer:
So far (almost) every BSP had its own version of download script.
Changing download software specially for ST devices was cumbersome.
Lately it was possible to override download script in target but
it still is more complicated then just changing syscfg value.

This adds common hw/scripts/download.sh file that can use all available
options to download image.

sh/scripts is converted to package to allow syscfg definitions that can
be overriden in BSPs or targets.

jlink, openocd, stutil, stm32_programmer_cli and pyocd can be selected.

Two example BSP were converted to common download script to see if it will catch up.


rymanluk:
@kasjer could you update all other bsp so it use new way?

hw/bsp/nordic_pca10095_net/bsp.yml
```diff
@@ -29,7 +29,7 @@ bsp.linkerscript.BOOT_LOADER.OVERWRITE:
     - "hw/bsp/nordic_pca10095_net/boot-nrf5340_net.ld"
     - "@apache-mynewt-core/hw/mcu/nordic/nrf5340_net/nrf5340_net.ld"
 bsp.part2linkerscript: "hw/bsp/nordic_pca10095_net/split-nordic_pca10095_net.ld"
-bsp.downloadscript: "hw/bsp/nordic_pca10095_net/nordic_pca10095_net_download.sh"
+bsp.downloadscript: "@apache-mynewt-core/hw/scripts/download.sh"
```
mkiiskila, line 32: Is the repo name needed in the location for this BSP for some reason?

hw/scripts/download.sh
```diff
@@ -0,0 +1,88 @@
...
+
+case "${MYNEWT_VAL_MYNEWT_DOWNLOADER}" in
+  "ezflashcli")
+    . $CORE_PATH/hw/scripts/ezflashcli.sh
+    common_file_to_load
+    ezflashcli_load
+    ;;
+  "jlink")
+    . $CORE_PATH/hw/scripts/jlink.sh
+    JLINK_DEV=${MYNEWT_VAL_JLINK_TARGET}
```
mkiiskila, line 48: Should we complain if the required variable is not set? The same applies to openocd case.

This might save users some debug time...

mkiiskila (approved):
I like the idea. It might also be helpful for 'download.cmd' to exist, for devs using windows?

I just had minor comments, nothing stopping this from going ahead.

hw/bsp/nordic_pca10095_net/bsp.yml
kasjer: no, I will remove it for consistency

hw/scripts/download.sh
kasjer: good idea, fix on the way

kasjer:
> I like the idea. It might also be helpful for 'download.cmd' to exist, for devs using windows?

Originally .cmd files were needed on Windows because `newt` tool executed scripts just like other commands.
All the .cmd files were exactly the same executing `bash` with arguments.
For a log time now newt on Windows when _some_script.sh_ file is to be executed executes `bash some_script.sh` instead.
The only reason to provide .cmd files again would be allow flashing devices from naked cmd.exe, but to make it work user would have to specify a lot of environment variables (that newt tool currently sets up) and still have bash.exe available in PATH variable directories.
I don't expect anyone is hardcore enough to do this this way.
All development for those scripts was done on Windows (at this point as a remainder I will make sure that those scripts actually have +x attribute set which they probably don't have since they were developed on NTFS).

In short as Windows user I don't see any reason to add .cmd files unless I'm missing something.

hw/scripts/download.sh
kasjer: done

hw/bsp/nordic_pca10095_net/bsp.yml
kasjer: done

PR #2999, opened 2023-05-23
hw/ipc_nrf5340: Support for flexible memory sharing

kasjer:
This adds easy way to add application specific shared memory regions that will be defined on application core and can be easily found on network core.
This is done so linker scripts for network and application core does not have to be carefully synchronized.


rymanluk:
@andrzej-kaczmarek do we need it ?:)

PR #3001, opened 2023-05-25
Be consistent in console write order for '\n' and '\r'

k-bharadwaj:
In the current implementation of the UART and CDC consoles, there can be a discrepency in the order of the Linefeed and Carriage Return print order. For consistency, we change this to always be "\r\n" for both variants of the implementation. This is useful in automated echo tests for verifying payloads.


wes3 (approved):
LGTM

PR #3005, opened 2023-05-31
Add support for alternative libc

kasjer:
Every compiler.yml had ld.flags with -specs-nosys.specs which is fine
for baselibc.
Changing to newlib-nano provided by compiler could lead to conflict.
Now -specs=nosys.specs is moved to baselibc settings and can
be different if baselibc is not used.

Package for newlib-nano is now added it:
- defines: -speac=nano.spec
- provides required glue functions for system calls

Most BSP for arm switch to be dependent to libc instead of libc/baselibc


andrzej-kaczmarek (approved):
lgtm

PR #3008, opened 2023-06-06
lvgl/trackball: Add cursor image

kasjer:
As temporary solution character from font was used as cursor
image.
Now there is normal triangle shaped cursor.
Additionally cursor will change to drag cursor when button
(track ball) is held for specific number of ms.

Trackball drag time reduced to 500ms


rymanluk (approved):
LGTM

PR #3013, opened 2023-06-12
lvgl: Add syscfgs for configuring fonts

m-gorecki:
This is useful if default font used by the lvgl based application has to be changed.


kasjer (approved):
LGTM

PR #3015, opened 2023-06-12
ci: Add workflow for testing various GCC versions

sjanc:
This will help make sure we build correctly with more GCC version.


kasjer (approved):
LGTM

PR #3016, opened 2023-06-13
hw/bus/drivers/spi_da1469x: Allow for 16-bit spi dma transfers

wes3:
The previous da1469x spi driver only allowed for 8-bit DMA transfers as the current spi config code does not support 16-bit transfers. However, there may be a need for this so added code to check to see if the word size is 16-bits and, if so, do 16-bit transfers. Note that no other configurations are expected currently and it is assumed if the spi word size is not 8-bit it is 16-bit.

Note that I think the hal spi and the da1469x code should be updated with the ability to do 16-bit spi but that change is a bit more intensive so just wanted to do this for now.


wes3:
something is odd here. Closing PR.

PR #3017, opened 2023-06-14
hw/bus/drivers/spi_da1469x: Allow for 16-bit SPI transfers

wes3:
Changes allow for 16-bit DMA transfers for spi. Note that the current hal and other configuration code do not allow the spi to be configured for 16 bits. However, if this is configured outside the provided API this code will allow for the proper size transfer.

I do think that we should add 16-bit spi transfers to the HAL (for all hal spi) but this is a larger undertaking and I think this change would be independent (and the same) without doing that which is why I am putting in this PR without adding 16-bit spi for HAL

### DIFF
--- a/hw/bus/drivers/spi_da1469x/src/spi_da1469x.c
+++ b/hw/bus/drivers/spi_da1469x/src/spi_da1469x.c
@@ -53,7 +53,7 @@ STATS_NAME_END(spi_da1469x_stats_section)
 #define MIN_DMA_SIZE 8
 
 /* A value of 1 for the word size means 16-bit words */
-#define SPI_CTRL_REG_16BIT_WORD (1 << SPI_SPI_CTRL_REG_SPI_WORD_Pos) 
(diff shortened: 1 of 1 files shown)


kasjer:
While code looks OK when only PR title is considered what will happen if transfer length is less then `MIN_DMA_SIZE`. Then even though SPI is configured for 16 bits, interrupt handler will operate on bytes transmitting extra bytes (I suppose) and dropping every other received byte.

wes3:
@kasjer That is a good point; I did not consider what would happen in the case DMA is not used. Let me review the code.

kasjer (approved):
Not tested but LGTM

wes3:
I did some testing; here are some stats. I do a very large DMA transfer (programming an fpga and it succeeds) and then some other short ones. The "read" command consists of two bytes written and three read.

/* Send a read command */
013588 read_bytes: 3
013588 written_bytes: 104179
013589 dma_transferred_bytes: 104169

/* Send a read command */

014982 Read data: 00:00:03
compat> stat spi_da1469x_0

015598 read_bytes: 6
015598 written_bytes: 104181
015599 dma_transferred_bytes: 104169

PR #3024, opened 2023-06-18
baselibc: use mynewt_main instead of main

kasjer:
This is extension of #2729 with difference being lack of weak function that seems not work every time.
In incorporates original commit by @andrzej-kaczmarek.

This change was motivated by awkward way native mynewt build was working.
Some implementations of main function looked like this
```c
int
main(int argc, char **argv)
{
#ifdef ARCH_sim
    mcu_sim_parse_args(argc, argv);
#endif
   ...
```
where `mcu_sim_parse_args` called main function recursively after parsing arguments and preparing native environment.

Now mynewt main function is called `mynewt_main` and is called from startup code (start.c) while native bsp has `main` function that parses command line arguments and then calls mynewt application `mynewt_main`


sjanc:
related blinky fix: https://github.com/apache/mynewt-blinky/pull/19

sjanc (approved):

sjanc (approved):

PR #3025, opened 2023-06-19
hw/mcu/cmac: Fix os_tick handling

andrzej-kaczmarek:
The os_tick is handled by 36_10_EQ_Y and 9_0_EQ_Y comparators which
requires extra handling to match full value on 2 separate comparators.
This could also trigger an assert on wakeup in cmac_timer_slp_disable
which expects that no interrupt is fired during that process. This is to
guarantee that we complete wake up before timer expires. However, since
we initially match on 36_10_EQ_Y only which has resolution of 1024us
it's possible that it will trigger an interrupt even though we woke up
before complete EQ_Y value matches.

To fix this and simplify the process we can use single EQ_Y comparator
to match on complete value, just as we do for EQ_X.


wes3 (approved):
LGTM but I have not tested it. We will test it and see if we see more asserts with the Y compare

PR #3029, opened 2023-06-27
hw/scripts: Add fallback to gdb-multiarch

kasjer:
This adds gdb detection so if gdb-multiarch is installed and arm-none-eabi or some of the riscv versions are not available gdb-multiarch will be used.


sjanc (approved):
tested, works good :)

PR #3033, opened 2023-07-07
Increase size of some `struct fcb` members

k-bharadwaj:
- Moving the sector_cnt and scratch_cnt members to a uint16_t allows for a counts > 255.
- Moving the version to a uint16_t allows for a more flexible scheme of versioning.

Component: Flash Circular Buffer


fs/fcb/include/fcb/fcb.h
```diff
@@ -36,33 +36,33 @@ extern "C" {
 #include <os/os_mutex.h>
 #include <flash_map/flash_map.h>
 
-#define FCB_MAX_LEN	(CHAR_MAX | CHAR_MAX << 7) /* Max length of element */
+#define FCB_MAX_LEN (CHAR_MAX | CHAR_MAX << 7) /* Max length of element */
```
kasjer, line 39: No need to covert tabs to spaces here if you change does not affect this line

fs/fcb/include/fcb/fcb.h
```diff
@@ -36,33 +36,33 @@ extern "C" {
...
  */
 struct fcb_entry {
-    struct flash_area *fe_area;	/* ptr to area within fcb->f_sectors */
-    uint32_t fe_elem_off;	/* start of entry */
-    uint32_t fe_data_off;	/* start of data */
-    uint16_t fe_data_len;	/* size of data area */
+    struct flash_area *fe_area; /* ptr to area within fcb->f_sectors */
+    uint32_t fe_elem_off;   /* start of entry */
+    uint32_t fe_data_off;   /* start of data */
+    uint16_t fe_data_len;   /* size of data area */
```
kasjer, line 49: Same here, no need to change tabs

fs/fcb/include/fcb/fcb.h
```diff
@@ -36,33 +36,33 @@ extern "C" {
...
     struct flash_area *f_sectors; /* Array of sectors, must be contiguous */
 
     /* Flash circular buffer internal state */
-    struct os_mutex f_mtx;	/* Locking for accessing the FCB data */
+    struct os_mutex f_mtx;  /* Locking for accessing the FCB data */
     struct flash_area *f_oldest;
     struct fcb_entry f_active;
     uint16_t f_active_id;
-    uint8_t f_align;		/* writes to flash have to aligned to this */
+    uint8_t f_align;        /* writes to flash have to aligned to this */
```
kasjer, line 65: For this structure it's fine something changed so tabs conversion is ok here

fs/fcb/include/fcb/fcb.h
```diff
@@ -127,7 +127,7 @@ int fcb_is_empty(struct fcb *fcb);
  */
 int
 fcb_offset_last_n(struct fcb *fcb, uint8_t entries,
-        struct fcb_entry *last_n_entry);
+                  struct fcb_entry *last_n_entry);
```
kasjer, line 130: no need to just reformat

kasjer (changes requested):
Extending sector count to uint16_t seems reasonable.
However this PR should have just one commit with only affected files changes.
Please rebase you work instead of merge then force push you branch.
If you need assistance with rebasing let me know.

mkiiskila:
Thanks for the merge request, however...

I agree with kasjer. This merge request should have 2 line which are getting changed; the sizes of sector counts
from uint8_t -> uint16_t. Please close this merge request, and open another one with that change only.
That kind of pull request will get approved.

MPU6050 changes must be removed from this merge request, that is totally unrelated to this change.
Also, the reason for those changes is not included in the request description. You can open another one for those.

 Whitespace changes within FCB are unnecessary, it makes the request noisier, and therefore more tedious to review.
I would refrain from doing formatting changes like that as part of functional change.

k-bharadwaj:
Looks like uncrustify messed things up - closing this to put up a cleaner patch. Thanks for reviewing!

PR #3035, opened 2023-07-12
mcu/stm32h7: Fix flash write/erase

kasjer:
Flash min write size was incorrectly set to 1 while for most H7 MCUs it has to be 32 bytes.

HAL_FLASH_Program uses different arguments then the reset of MCU with non-linear flash sectors.
Write now respects minimum write size.

STM32H7 hal requires eraseInit.Banks field to be set.


wes3 (approved):
LGTM but I did not test this on a board with an H7xx chip

PR #3036, opened 2023-07-12
hw/drivers/mmc: Fix Multiple Block Read CMD18

kasjer:
This PR:
- fixes CMD18 multiple block read where read control token and CRC were handled correctly only if single block was read
- adds code that switches to higher speed after initialization (was marked as TODO:)

### DIFF
--- a/hw/drivers/mmc/src/mmc.c
+++ b/hw/drivers/mmc/src/mmc.c
@@ -65,22 +65,11 @@
 
 static uint8_t g_block_buf[BLOCK_LEN];
 
-static struct hal_spi_settings mmc_settings = {
-    .data_order = HAL_SPI_MSB_FIRST,
-    .data_mode  = HAL_SPI_MODE0,
-    /* XXX: MMC initialization accepts clocks in the range 100-400KHz */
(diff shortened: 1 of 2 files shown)


hw/drivers/mmc/src/mmc.c
```diff
@@ -173,36 +162,34 @@ send_mmc_cmd(struct mmc_cfg *mmc, uint8_t cmd, uint32_t payload)
...
-    struct mmc_cfg *mmc;
-
     /* TODO: create new struct for every new spi mmc, add to SLIST */
-    mmc = &g_mmc_cfg;
-    mmc->spi_num = spi_num;
+    struct mmc_cfg *mmc = &g_mmc_cfg;
     mmc->ss_pin = ss_pin;
-    mmc->spi_cfg = spi_cfg;
-    mmc->settings = &mmc_settings;
+    mmc->mmc_spi_cfg = *(struct mmc_spi_cfg *)spi_cfg;
```
sjanc, line 180: nitpick: this cast is no longer needed (as spi_cfg is not void * anymore)

sjanc (approved):

PR #3040, opened 2023-07-14
hw/drivers/flash/spiflash: Make SPI mode configurable

kasjer:
SPI mode for hardcoded to HAL_SPI_MODE3, most SPI flashes also support HAL_SPI_MODE0.
Now user can specify which mode to use.

### DIFF
--- a/hw/bsp/black_vet6/src/hal_bsp.c
+++ b/hw/bsp/black_vet6/src/hal_bsp.c
@@ -290,7 +290,7 @@ extern const struct hal_flash stm32_flash_dev;
 struct bus_spi_node_cfg flash_spi_cfg = {
     .node_cfg.bus_name = MYNEWT_VAL(BSP_FLASH_SPI_BUS),
     .pin_cs = MYNEWT_VAL(SPIFLASH_SPI_CS_PIN),
-    .mode = BUS_SPI_MODE_3,
+    .mode = MYNEWT_VAL(SPIFLASH_SPI_MODE),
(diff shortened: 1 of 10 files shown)


wes3 (commented):
I have one comment: there are some bsps where the default value is mode 0. It appears that the new mode for all will be 3 since the new default is 3. If I am reading this right and it is true, should we override the default mode in those bsps where the default mode was mode 0?

kasjer:
@wes3 good catch, you are technically correct. Two BSP in question are nucleo-h723 and nucleo-u575 those are recent addition to mynewt and we added support for spi flash just to make example how it works.
None of the Nucleo boards I know of (those two included) has any SPI Flash on board so it does not spoil anything I think and now mode is configurable.
I made option just to make it easier on Logic analyzer when lot of other device uses MODE0.
Clock idle state 0 or 1 could be related to power consumption so user can choose which one to use while for SPI Flash it does not matter much.

wes3:
@kasjer Are you waiting for approval on this? I think a coding style check failed.

kasjer:
> @kasjer Are you waiting for approval on this? I think a coding style check failed.

I could reformat section that style checker complains about but this is style was approved way back.
Whole file could be re-formatted but we usually don't do this just to make style checker happy.

So offending code changes value keeping format style consistent with the rest of surrounding lines.
So basically what I'm trying to say that this change does not introduce new style error just keeps existing spaces.

wes3 (approved):
LGTM

PR #3041, opened 2023-07-14
Add parse_arg module

sjanc:
This module is based on current implementation from btshell application. With some API cleanups and unifications.


kasjer (approved):
LGTM

PR #3046, opened 2023-07-30
hw/mcu/dialog: Fix DC/DC for CMAC

andrzej-kaczmarek:
The code to enable DC/DC on CMAC was not updated to new transport and IPC driver.


wes3 (approved):
LGTM

PR #3050, opened 2023-08-02
MbedTLS update to 2.28.3 and conversion to external module

sjanc:


sjanc:
@mkiiskila @agross-korg there are some Mynewt specific mbedtls APIs added by you, I tried to keep then intact but since there are no users of those in core I'm not sure if those work OK.

Any chance you could add at least basic tests for those 2 functions?

andrzej-kaczmarek (approved):

PR #3056, opened 2023-08-16
repo: Add dependency to TinyUSB

kasjer:
User had to add dependecy to TinyUSB in project.yml. Now repository dependency is added.


sjanc (approved):

andrzej-kaczmarek:
why not add this to tinyusb package instead?

sjanc:
yeah, we already discussed this offline that it should be in package :)

sjanc:
#3069 was merged instead

PR #3059, opened 2023-08-17
mcu/nordic: Require LF clock source to be specified

sjanc:
This is typically specified by BSP and omitting to do so is an error.

### DIFF
--- a/hw/mcu/nordic/nrf51xxx/syscfg.yml
+++ b/hw/mcu/nordic/nrf51xxx/syscfg.yml
@@ -101,6 +101,8 @@ syscfg.defs:
         description: >
             Selected source for low frequency clock (LFCLK).
         value:
+        restrictions:
+            - $notnull
         choices:
             - LFRC      # 32.768 kHz RC oscillator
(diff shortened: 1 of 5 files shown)


andrzej-kaczmarek (approved):

PromyLOPh:
(Continued from Slack) My mistake was to explicitly use LFRC though and `$notnull` would not catch that. I was wondering whether a restriction like `MCU_LFCLK_SOURCE == 'LFRC' && !BLE_CONTROLLER` would work, since BLE clearly does not work with the low-accuracy LFRC.

sjanc:
> (Continued from Slack) My mistake was to explicitly use LFRC though and `$notnull` would not catch that. I was wondering whether a restriction like `MCU_LFCLK_SOURCE == 'LFRC' && !BLE_CONTROLLER` would work, since BLE clearly does not work with the low-accuracy LFRC. (edit) It should be noted that the LFRC can be calibrated, which makes BLE reliably work for me.

yeap, I updated PR with this restriction

PR #3060, opened 2023-08-18
Fix build error with arm gcc 12.3.1

jrotkiewicz:
While trying to build with arm-none-eabi-gcc version 12.3.1 and setting LIBC: nano, the compiler did not recognise static_assert.

### DIFF
--- a/hw/mcu/nordic/include/nrfx_glue.h
+++ b/hw/mcu/nordic/include/nrfx_glue.h
@@ -67,7 +67,7 @@ extern "C" {
  *
  * @param expression  Expression to evaluate.
  */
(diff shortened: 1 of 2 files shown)


kasjer:
Little context here.

static_assert is C++ keyword
_Static_assert is C keyword

C standard specified that static_assert macro should expand to _Static_assert. That was the case in arm-none-eabi-gcc assert.h and same strategy was implemented in baselib.c

Additionally NRFX_STATIC_ASSERT, when it was implemented in mynewt, chose to map to static_assert.
Now it seems that C standard drops this static_assert definition. So we just change NRF static assert to C standard _Static_assert.

For C++ we just specify _Static_assert if it's not already defined

sjanc (approved):

PR #3061, opened 2023-08-24
stm32f1/spi: Add definition missing from HAL

kasjer:
It seems that HAL_SPI_ERROR_FRE was defined is some previous versions of HAL for STMF1.
Now it's gone but register bit is still there and there is not Errata stating that bit should not be touched.

SPI block is shared among all STM products and all of them have this UDR bit that is mapped to HAL_SPI_ERROR_FRE in ST HAL.

Without this change HAL SPI will not build for STM32F1 devices.


wes3 (approved):
LGTM

PR #3065, opened 2023-09-13
sys/config: fix CONF_FLOAT

wes3:
The check in conf_value_from_str for CONF_FLOAT is incorrect. FLT_MIN is not the smallest negative float; it is the smallest positive float. Removing the check for min/max value as it is not needed in this case.


vrahane (approved):
LGTM

PR #3068, opened 2023-09-14
sys/log: Add option to make the global log index increment sequentially

benmccrea:
Adds a syscfg `LOG_SEQUENTIAL_IDX`. When enabled the global log index won't be incremented when adding logs of `LOG_TYPE_STREAM` which are not persisted (e.g. console logs). 

The global index is only useful for persisted logs, and if it is incremented for non-persisted logs there will be jumps in the log indexes of the persisted logs. This makes detection of missing logs a challenging task. By incrementing the global index only for persisted logs, it becomes sequentially increasing, which makes detection of missing/dropped logs trivial for upstream consumers of the logs.

This PR also adds a syscfg `LOG_SHELL_SHOW_INDEX` to enable displaying the log index in the 'log' shell command output.

### DIFF
--- a/sys/log/full/src/log.c
+++ b/sys/log/full/src/log.c
@@ -540,7 +540,18 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
 
     OS_ENTER_CRITICAL(sr);
 #if MYNEWT_VAL(LOG_GLOBAL_IDX)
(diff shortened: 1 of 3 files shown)


vrahane (approved):
This is optional and seems to be fine. We will let others comment, if no concerns I will merge it in.

PR #3084, opened 2023-09-28
hw/mcu/dialog: added critical section to dma acquire peripheral

dustinfranco:


wes3 (changes requested):
Looks good but this needs a change to the commit message. The commit message should have as the title: hw/mcu/dialog: <title> and not FWW-xxx in it.

PR #3090, opened 2023-10-05
hw/usb/tinyusb: usb cdc console and tinyusb library sysinit before console uses it

vrahane:
hw/usb/tinyusb: Fix sysinit levels so that tinyusb gets initialized
before console uses it

- Changing cdc_console sysinit to 502 adding a new syscfg
  `CONSOLE_USB_CDC_SYSINIT_STAGE` for CDC USB console
  and use that instead
- Also add restrictions so that usbd is never inited after
  cdc_console

### DIFF
--- a/hw/usb/tinyusb/cdc_console/syscfg.yml
+++ b/hw/usb/tinyusb/cdc_console/syscfg.yml
@@ -25,5 +25,12 @@ syscfg.defs:
 syscfg.vals:
     USBD_CDC_CONSOLE: 1
 
+syscfg.defs:
+    CONSOLE_USB_CDC_SYSINIT_STAGE:
+        description: >
+          Initialize USB CDC Console at the specified sysinit level
+        value: 502
+
 syscfg.restrictions:
     - "USBD_CDC_CONSOLE"
+    - CONSOLE_USB_CDC_SYSINIT_STAGE > USBD_SYSINIT_STAGE
--- a/hw/usb/tinyusb/cdc_console/pkg.yml
+++ b/hw/usb/tinyusb/cdc_console/pkg.yml
@@ -32,4 +32,4 @@ pkg.deps:
     - "@apache-mynewt-core/hw/usb/tinyusb/cdc"
 
 pkg.init:
-    usb_cdc_console_pkg_init: 'MYNEWT_VAL(CONSOLE_SYSINIT_STAGE)'
+    usb_cdc_console_pkg_init: 'MYNEWT_VAL(CONSOLE_USB_CDC_SYSINIT_STAGE)'


hw/usb/tinyusb/syscfg.yml
```diff
@@ -33,7 +33,7 @@ syscfg.defs:
     USBD_SYSINIT_STAGE:
         description: >
             Sysinit stage for USB device functionality.
-        value: 500
+        value: 21
```
andrzej-kaczmarek, line 36: add syscfg for this sysinit and change priority in your local target if needed

hw/usb/tinyusb/cdc_console/syscfg.yml
```diff
@@ -25,5 +25,12 @@ syscfg.defs:
 syscfg.vals:
     USBD_CDC_CONSOLE: 1
 
+syscfg.defs:
+    CONSOLE_USB_CDC_SYSINIT_STAGE:
+        description: >
+          Tiny USB needs to be initialized before Console USB CDC is \
+          initialized
```
andrzej-kaczmarek, line 32: this lacks description of what it configures

andrzej-kaczmarek (changes requested):

hw/usb/tinyusb/syscfg.yml
vrahane: CONSOLE_SYSINIT_STAGE Is set to 20 by default and that same level(syscfg) is used for CDC console. Tinyusb needs to be initialized before the CDC console if we want the CDC console to work fine by default, else it is going to have race conditions which is what we are seeing.

andrzej-kaczmarek: I understand that but are you sure that changing default value for tinyusb from 500 to 21 will not introduce any side effects in other configurations (<- @kasjer)? You can add conditional restriction instead so newt will print error if CDC console is enabled and USBD_SYSINIT_STAGE >= CONSOLE_USB_CDC_SYSINIT_STAGE and then change USBD_SYSINIT_STAGE in your local config accordingly.

kasjer: I'm not sure what will happen if we move USB initialization to 21. So maybe it is safer to keep default priority as @andrzej-kaczmarek suggested.

vrahane: @andrzej-kaczmarek I am fine closing the PR if you think this is not needed. As you said I can very well fix it in my local target. I just wanted to fix it in master/upstream as well.

vrahane: Well, I have tested it and seems to be working fine. @benmccrea can second me on that.

andrzej-kaczmarek: I think adding CONSOLE_USB_CDC_SYSINIT_STAGE still makes sense as otherwise you won't be able to set proper priority order in scenario you mentioned (correct me if I'm wrong). I'm just reluctant to change default configuration for all other cases as well if we're not sure what could be the side-effects.

vrahane: Though, if you guys feel nervous about   initializing it that early, cdc_console will have to be initialized later which will be in the 500s which will be too late. I am fine either ways.

vrahane: One more thing to consider is all hardware stuff gets initialized quite early and only tinyusb which in turn initializes the USB controller for da1469x in our specific case gets initialized in the 500s.

vrahane: Having said that I understand the concerns, I will just add CONSOLE_USB_CDC_SYSINIT_STAGE  and set it to 500+ sysinit level. I will change our targets locally to initialize both early on. We (mynewt) can probably make a well tested change later to do it early on.

vrahane: Having said that I understand the concerns, I will just add CONSOLE_USB_CDC_SYSINIT_STAGE  and set it to 500+ sysinit level. I will change our targets locally to initialize both early on. We (mynewt) can probably make a well tested change later to do it early on.

hw/usb/tinyusb/cdc_console/syscfg.yml
vrahane: updated

hw/usb/tinyusb/syscfg.yml
vrahane: done

andrzej-kaczmarek:
lgtm, but please fix the commit message. also the same restriction added in two different places is redundant.

vrahane:
@andrzej-kaczmarek  Done

andrzej-kaczmarek (approved):

PR #3094, opened 2023-10-13
kernel/os: Fix issue that causes WDOG reset reason to be lost

benmccrea:
In the default irq handler, hal_watchdog_tickle() was called before doing a core dump but on Dialog da1469x this causes the reset reason to be lost, and reboot log was recording the reset reason as HAL_RESET_SOFT instead of HAL_RESET_WDOG.

The issue is fixed by calling hal_watchdog_disable() instead of hal_watchdog_tickle() before doing a core dump.

This PR also adds hal_watchdog_disable() to the HAL API for all MCUs (for now it is implemented only for Dialog da1469x and CMAC).

### DIFF
--- a/hw/hal/include/hal/hal_watchdog.h
+++ b/hw/hal/include/hal/hal_watchdog.h
@@ -52,6 +52,11 @@ int hal_watchdog_init(uint32_t expire_msecs);
 void hal_watchdog_enable(void);
 
 /**
+ * Stops the watchdog.
+ */
+void hal_watchdog_disable(void);
+
+/**
  * Tickles the watchdog.   This needs to be done periodically, before
  * the value configured in :c:func:`hal_watchdog_init()` expires.
  */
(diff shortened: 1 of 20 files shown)


vrahane (approved):
I tried it out. It generates a coredump fine on a watchdog with the correct reboot reason. I will let others comment as well. If there are no issues, will merge it in. FYI, this is what caused the issue: https://github.com/apache/mynewt-core/pull/2440

andrzej-kaczmarek:
If you replace `tickle` with `disable` and leave `disable` not implemented for all platforms but dialog, it will possibly break those platforms as in mentioned issue. Also watchdog cannot be disabled on some platforms, in particular nrf52, so this won't work.

So perhaps what can be done instead is to make call to `tickle` optional via syscfg and then use coredump callback to disable watchdog via registers to avoid having an API that is not implemented on any platform.

benmccrea:
@andrzej-kaczmarek Thanks, I agree with you. We'll find a different way. We can use os_coredump_cb as you suggest, or we might be able to modify how watchdog reboot reason is detected for da1469x so that it isn't lost by the tickle. 
Closing this unmerged.

PR #3095, opened 2023-10-16
kernel/os: Fix issue that causes WDOG reset reason to be lost

benmccrea:
In the default irq handler, hal_watchdog_tickle() was called before doing a core dump but on Dialog da1469x this causes the reset reason   to be lost, and reboot log was recording the reset reason as HAL_RESET_SOFT instead of HAL_RESET_WDOG.

 The issue is fixed by calling hal_watchdog_disable() instead of hal_watchdog_tickle() before doing a core dump.

 This PR also adds hal_watchdog_disable() to the HAL API for all MCUs (for now it is implemented only for Dialog da1469x and CMAC).


benmccrea:
Closing (accidentally opened)

PR #3100, opened 2023-11-03
kernel/os: Add default irq callback

wes3:
Added a MYNEWT_VAL named OS_DEFAULT_IRQ_CB which, when enabled, will call a user-defined callback as the first function called in os_default_irq(). This will allow users to define custom functionality when a default irq occurs.

### DIFF
--- a/kernel/os/include/os/os_fault.h
+++ b/kernel/os/include/os/os_fault.h
@@ -36,6 +36,10 @@ void __assert_func(const char *file, int line, const char *func, const char *e)
 typedef void (*coredump_cb_t)(void *tf);
 #endif
 
+#if MYNEWT_VAL(OS_DEFAULT_IRQ_CB)
+void os_default_irq_cb(void);
+#endif
+
 #if MYNEWT_VAL(OS_CRASH_FILE_LINE)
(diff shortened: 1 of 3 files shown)


wes3:
@benmccrea @andrzej-kaczmarek BTW: I should mention that this PR came about due to the hal_watchdog_tickle() issue for dialog mcu's wiping out the reset reason. I did not want to mess with the os_coredump_cb as was mentioned in that other PR which was closed. I wanted this to be called as the first function in os_default_irq() and be independent of the other MYNEWT_VALs. I only added this to os_fault.c in cm33 as I wanted to see what others thought about this and if desired we can add it to the other os_fault.c modules.

vrahane (approved):
LGTM

benmccrea (approved):

andrzej-kaczmarek:
I'm not a big fan of this, but I won't oppose it since I don't have any better idea ;)

wes3:
@andrzej-kaczmarek Honestly, I am not the biggest fan of this myself but I was having a hard time thinking of a better way to do this. The only other thought that came to mind was to replace the NMI irq for dialog mcu's. Currently, os_arch_arm.c installs the default irq by doing a NVIC_SetVector. I would need to look at the order at which things are done but maybe this could be done after os_arch_os_init() is called in some dialog specific code. That is my only other thought.

PR #3114, opened 2023-11-30
ci: Build application samples

sjanc:
Make sure samples are building with degfault BSP
(currently nordic_pca10056).


kasjer (approved):
LGTM

PR #3125, opened 2023-12-21
util/stream: Add streaming API

kasjer:
This API provides interface to streams it should replace util/streamer.


rymanluk (approved):
LGTM

PR #3128, opened 2024-01-15
sys/log/modlog: Add modlog hexdump

m-gorecki:
This adds MODLOG_HEXDUMP macros which allow to write specified number of bytes' values in a string format into the log modules.


kasjer (approved):
LGTM

PR #3133, opened 2024-01-30
mcu/nordic: Set flags before running SPI callback

Uks2:
When a non-blocking SPI transfer has finished, our `spi->spi_xfr_flag` and the hardware interrupt flag are now reset before the `spi->txrx_cb_func` is run.  This allows the callback to start a new SPI transfer.

I've tested the nRF52 changes and they're fine.  I'm making the bold assumption that the other nRFs work the same.


kasjer (approved):
Seems reasonable to me, not tested though.

PR #3148, opened 2024-02-21
tinyusb/msc_fat_view: Fix custom entries support

kasjer:
For now msc_fat_view provided several FAT root entries that could be enabled in syscfg.

There is function to add new root dir entries msc_fat_view_add_dir_entry(), however even if user added custom entry it would not show up since all entries were refreshed when during test_unit_ready by call to init_disk_data that silently dropped non standard entries and created root from scratch.

Now root population is divided in to two functions. init_disk_data() that provides standard enties is called when packaged is initiated and update_disk_data() is called when actual sectors for host are generated and this function includes user provided entries.

With this change core dump entries have to be initiated after msc_fat_view_pkg_init and any time before tinyusb_start


rymanluk (approved):
LGTM, also tested and works fine

PR #3149, opened 2024-02-22
tinyusb: Change revision to include NRF5x ISO fixes

kasjer:
Two important ISO fixes for NRF5x were added to TinyUSB repository:
- 68bb85840 nrf5x: Handle ISOOUT CRC errors
- f0ddf8d10 dcd_nrf5x: ISO OUT handling

There is not release with those changes yet so commit hash is used for now.

### DIFF
--- a/hw/usb/tinyusb/tinyusb_sdk/pkg.yml
+++ b/hw/usb/tinyusb/tinyusb_sdk/pkg.yml
@@ -41,6 +41,6 @@ pkg.source_dirs:
 repository.tinyusb:
     type: github
     branch: master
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek (approved):

sjanc (approved):

kasjer:
Due to repository.yml in TinyUSB this change while reasonable does not work and project.yml has to specify version and we have to wait till TinyUSB tag shows up and repository.yml is updated. It would be good for other external repository that don't have repository.yml file.

PR #3153, opened 2024-03-06
Depend on latest MCUboot stable release on master branch

sjanc:
There is no point in using MCUboot from main branch even on core master. Core releases are depending on latest stable at point of release so we move 'backward' anyway. It is better to test on with what will likely be used for release.


kasjer (approved):
LGTM

PR #3156, opened 2024-03-06
hw/bsp/nordic_pca10090: Use selectable downloader

kasjer:
Download was not working due to missing device specifier for jlink. Now it is corrected and nrfjprog can also be used for flashing.


sjanc (approved):
flashing works now

PR #3161, opened 2024-03-08
mcu/nrf51: Fix low frequency clock selection

kasjer:
When XTAL_32768/XTAL_RC/XTAL_32768_SYNTH were deprecated and MCU_LFCLK_SOURCE was chosen to be used instead, code that actually enabled clocks still depended on syscfg values that were never set.

This only changes preprocessor conditions to use MCU_LFCLK_SEOURCE.


sjanc (approved):
tested, works OK!

PR #3170, opened 2024-03-22
tinyusb/msc_fat_view: Add API to eject/insert medium

kasjer:
msc_fat_view adds some predefined entries to root directory. If application decides to add more entries after USB starts and host system already read and cached some data
media eject/insert sequence can inform system that old directory structure and FAT is not valid.

msc_fat_view_add_dir_entry() function will do it automatically but it is also possible to report no medium with
msc_fat_view_media_eject() function
msc_fat_view_media_insert() will inform host system that media is present.

syscfg value MSC_FAT_VIEW_AUTO_INSERT when set to 1 will report media as present during system init.


rymanluk (approved):
LGTM

PR #3171, opened 2024-03-25
hmcu/stm32: Add link_tables to all ld scripts

kasjer:
newt tool can create link_tables.ld.h file from
pkg.link_table section
This modifies all STM32 linker scripts to have this functionality.

G4 and G0 already have common linker script so they are already covered.


kasjer:
superseded by #3171

PR #3174, opened 2024-03-28
repository.yml: Remove 0.0.1 version

m-gorecki:
0.0.1 version was used for historical reasons and is no longer necessary. It was also causing issue with newt info command. When commit with release tag happens to be the HEAD commit of master branch, newt info would print version 0.0.1 instead of release version from tag.


sjanc (approved):
alright, let see if this doesn't break anything :)

PR #3175, opened 2024-03-28
hw/bsp: Enable double precision for STM32F767 and STM32H723

kasjer:
Two boards have MCU with double precision FPU.
This simply enables double precision instruction generation for those two boards.


rymanluk (approved):
LGTM

PR #3177, opened 2024-03-29
link: Add link_tables to all ld scripts

kasjer:
newt tool can create link_tables.ld.h file from
pkg.link_table section
This modifies all linker scripts to have this
functionality.

supersedes #3171


rymanluk (approved):
LGTM

PR #3179, opened 2024-03-29
util: Add coremark repository

kasjer:
This remove coremark code from core repository and adds external githun repository with unmodified code.

apps/coremark is now stripped to minimum

Coremark is not package in util/coremark and can
be included in any application it also provides
shell command 'coremark' that can be execute
when SHELL_TASK is enabled.


rymanluk (approved):
LGTM

PR #3184, opened 2024-04-05
sys/flash_map: Allow build with 1 flash image area

kasjer:
When bootloader alone can write new image to device (no swap needed) there is no need to define fake
flash area anymore.

This allows build when there is no definition for secondary swap image flash area.


rymanluk (approved):
LGTM

PR #3191, opened 2024-04-11
hw/mcu/stm32g4: Remove unused linker script

kasjer:
For stm32g4 autogenerated linker script is used.
File stm32g4xx.ld was not used is now removed.


kasjer:
duplicated

PR #3194, opened 2024-04-11
ci: Add native to ignored BSPs in bootloader build test

m-gorecki:
After the merge of https://github.com/apache/mynewt-newt/commit/a798fcc687dd81857de7bb92e77711c9a74f0692
newt passes to the linker the same .a files as before, but in different order and it causes linker error. It means that previously this test was passing just by accident.

It does not make sense to build bootloader for native target anyway, so this test will simply be skipped.


m-gorecki:
Closing, because it's duplicating PR #3193

kasjer (approved):

PR #3199, opened 2024-04-14
hw/mcu/stm32h7: Use linear sector flash driver

kasjer:
STM32H7 actually has liner flash with sectors of 128kB. There is no need for stm32_flash_sectors table in BSP.

Additionally non-linear version that was used so far would not compile for some H7 chips that have different FLASH CR register and where FLASH_VOLATEG_RANGE_3 is not defined.

Liner sector variant was chosen when FLASH_PAGE_SIZE was defined. While H7 and H5 don't have this definition they do have FLASH_SECTOR_SIZE and when this definition is present FLASH_IS_LINEAR is also set to 1

### DIFF
--- a/hw/mcu/stm/stm32h7xx/src/hal_flash.c
+++ b/hw/mcu/stm/stm32h7xx/src/hal_flash.c
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <syscfg/syscfg.h>
+#include <mcu/stm32_hal.h>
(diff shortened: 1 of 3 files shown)


hw/mcu/stm/stm32h7xx/src/hal_flash.c
```diff
@@ -0,0 +1,59 @@
...
+
+    (void)PageError;
+
+    if (!(sector_address & (FLASH_SECTOR_SIZE - 1))) {
+        eraseinit.TypeErase = FLASH_TYPEERASE_SECTORS;
+#ifdef FLASH_BANK_2
+        if ((sector_address - dev->hf_base_addr) < (STM32_FLASH_SIZE / 2)) {
+            eraseinit.Banks = FLASH_BANK_1;
+        }
+        else {
```
sjanc, line 41: nitpick:  odd formating

hw/mcu/stm/stm32h7xx/src/hal_flash.c
```diff
@@ -0,0 +1,59 @@
...
+#include <syscfg/syscfg.h>
+#include <mcu/stm32_hal.h>
+#include "hal/hal_flash_int.h"
+
+#define STM32_FLASH_SIZE      (MYNEWT_VAL(STM32_FLASH_SIZE_KB) * 1024)
+
+int
+stm32_mcu_flash_erase_sector(const struct hal_flash *dev, uint32_t sector_address)
+{
+    FLASH_EraseInitTypeDef eraseinit;
```
sjanc, line 29: no need to zero this on init? some fields can be left uninitialized

sjanc (approved):

hw/mcu/stm/stm32h7xx/src/hal_flash.c
kasjer: fixed

hw/mcu/stm/stm32h7xx/src/hal_flash.c
kasjer: Only VoltageRange can be uninitialized and it't value is ignored when FLASH_CR_PSIZE is not defined so it't safe

PR #3200, opened 2024-04-15
sys/config: Make floating point support optional

kasjer:
Adding CONF_FLOAT support could increase code size significantly if floating point are not used otherwise.

This makes support optional (enabled by default when HARDFLOAT is present).


wes3 (approved):
LGTM

PR #3222, opened 2024-05-16
hw/mcu/dialog: Add 1MHz high-speed mode to hal i2c

wes3:
Adds 1MHz i2c speed to the i2c hal. The 1MHz speed requires high speed mode on the chip. In order for this to work it appears that the RESTART_EN bit has to be set in the control register or a tx abort condition occurs (bit 8 in tx abort source reg).

I tried to find some decent documentation on enabling high-speed mode for the da1469x. I also looked at some SDK versions that I had and I did not see that this bit was being set. When I tried simply setting the speed to 1MHz I was seeing this abort condition and read the chip doc and it appears that the RESTART_EN bit needs to be set for high-speed mode to work. Tested it with a da1469x equivalent chip and another chip that works with 1MHz i2c and things seem ok.


kasjer (approved):
LGTM

PR #3223, opened 2024-05-18
stm32/hal_watchdog: Fix timeout calculations

kasjer:
All STM32 devices use LSI for watchdog functionality. LSI frequency varies between MCUs and it's approximation is define by LSI_VALUE in stm32xxx_hal_conf.h.
Code was assuming 32768 which is ok for devices that have LSI rate at 30kHz.
Some device has 37kHz others 40kHz it those cases calculated time could be too much off resulting in reset by watchdog.

Now code uses LSI_VALUE adding 10% to avoid cases when LSI runs faster then nominal.


wes3 (approved):
LGTM

PR #3225, opened 2024-05-18
Fixes for ina219/ina226

kasjer:
For both drivers shell code has warning in new compilers

For INA219 voltage calculation was not correct for 32V or 16V range depending on
random bit.


kasjer:
Stats indent complain is not worth fixing. Style checker should be updated if anyone can do this.

sjanc (approved):

PR #3229, opened 2024-05-21
pwm/pwm_da1469x: Preserve timer state on frequency change.

wes3:
Read the current timer state (timer control reg) and preserve its running state when setting frequency. The only bits that should be set in the control register are the bits to enable the timer in the proper mode and enable the clock.

The rational behind this change is that some users may want to be able to change the frequency "on the fly" without having to diable pwm, change frequency and then re-enable pwm. I do realize that with this chip if you change frequency the duty cycle changes so changing frequency on the fly also changes the duty cycle but this was always the case.

Note that this should not change the previous functionality (other than adding some extra code to read the control register and mask out the SYS_CLK_EN bit. This change could have been done using a syscfg value to preserve the old code and make this new code present only present if the syscfg was set but it seems unnecessary. However, if folks think there should be a syscfg to make this behavior optional that would be fine.


vrahane:
We did speak about this and seems fine to me. Haven't tested it.

vrahane (approved):

andrzej-kaczmarek (approved):

PR #3230, opened 2024-05-21
sys/config: Add watchdog tickle before each conf_fcb_var_read

JustineKH:
If there are a lot of config variables, a watchdog can be triggered by the multiple `conf_fcb_var_read`s.

### DIFF
--- a/sys/config/src/config_fcb.c
+++ b/sys/config/src/config_fcb.c
@@ -29,6 +29,7 @@
 #include "config/config_fcb.h"
 #include "config/config_generic_kv.h"
 #include "config_priv.h"
+#include "hal/hal_watchdog.h"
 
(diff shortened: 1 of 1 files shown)


vrahane (approved):
LGTM

wes3 (approved):
I am fine with the changes; just wondering if we need a tickle in the outer loop but I suspect not.

vrahane:
@wes3  That's a good point. If the flash is filled with all 0s, we might keep looping for a while without tickling the watchdog. Maybe we should tickle the watchdog at the beginning of the outer loop as well as inner loop.

wes3 (approved):
LGTM

vrahane (approved):

PR #3231, opened 2024-05-23
stylecheck: Add uncrastify exception for STATS

kasjer:
Uncrustify removes leading spaces from STATS.
This simple awk scrip is executed after uncrustify inserts spaces when expected


sjanc (approved):
okay, I guess we will have to leave this hack for time being :)

PR #3233, opened 2024-05-24
util/stream: Add pumping functions

kasjer:
This extends in_stream and out_stream to provide pumping
functions

Additionally stream_pump() function is added to allow easy
pumping data between streams.

### DIFF
--- a/hw/drivers/i2s/src/i2s_stream.c
+++ b/hw/drivers/i2s/src/i2s_stream.c
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


hw/drivers/i2s/src/i2s_stream.c
```diff
@@ -0,0 +1,95 @@
...
+    int written = 0;
+    uint32_t available = istream_available(istream);
+    uint32_t sample_count;
+
+    if (available < count) {
+        count = available;
+    }
+    sample_count = count / i2s->sample_size_in_bytes;
+
+    while (count) {
```
sjanc, line 44: this looks bit odd since count is never modified inside loop

util/stream/src/stream.c
```diff
@@ -107,3 +108,30 @@ istream_read(struct in_stream *istream, uint8_t *buf, uint32_t count)
...
+    if (ostream->vft->pump_from) {
+        pumped = ostream->vft->pump_from(ostream, istream, count);
+    } else if (istream->vft->pump_to) {
+        /* When input stream has pump_to used it */
+        pumped = istream->vft->pump_to(istream, ostream, count);
+    } else {
+        /* Both stream don't provide pump functions, use local buffer for pumping data */
+        uint8_t buf[16];
+        pumped = 0;
+        while (count) {
```
sjanc, line 126: same here

kasjer: fixed

hw/drivers/i2s/src/i2s_stream.c
kasjer: loop was not needed and now is removed it could be replaced with if

sjanc (approved):

PR #3253, opened 2024-06-17
hw/mcu/nordic: Add common GPIO HAL for nRF MCUs

m-gorecki:
Common HAL package will be extended in the future
to provide common HAL API for other peripherals.
This also makes GPIO HAL use the API provided by nrfx, instead of using MCUs registers directly.


sjanc (approved):

kasjer (approved):
LGTM, not tested though

PR #3289, opened 2024-07-19
sys/config: Add littlefs configuration

m-gorecki:
Until now CONFIG_LITTLEFS was only defined,
but wasn't doing anything. Now littlefs
is configured and used as backend storage.

Tested with slinky app config.


kasjer (approved):
LGTM. tested OK

PR #3295, opened 2024-08-07
log_fcb walk back

kasjer:
This adds functionality to walk back in fcb and in log_fcb
Additional options to `log` cmd
`-s <n>` skips n logs before prints are started
`-r` start printing from the latest logs (work only for FCB based logs)

```
log -n 10 -r reboot_log
```
will print 10 latest logs from reboot log

```
log -s 1000 reboot_log
```
will print logs skipping first 1000 from reboot_log

### DIFF
--- a/fs/fcb/src/fcb_getnext.c
+++ b/fs/fcb/src/fcb_getnext.c
@@ -20,6 +20,47 @@
 
 #include "fcb/fcb.h"
 #include "fcb_priv.h"
+#include "defs/error.h"
+
+#if MYNEWT_VAL_FCB_BIDIRECTIONAL
+#define FCB_STEP_BACK(loc) ((loc)->fe_step_back)
+#else
+#define FCB_STEP_BACK(loc) (false)
+#endif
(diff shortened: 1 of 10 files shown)


fs/fcb/src/fcb_getnext.c
```diff
@@ -51,63 +90,377 @@ fcb_getnext_area(struct fcb *fcb, struct flash_area *fap)
...
+        if (cache->cache_data == NULL) {
+            cache->sector_ix = 0xFFFF;
+            cache->cache_size = 0;
+            cache->cache_count = 0;
+        } else {
+            cache->cache_size = new_size;
         }
+    }
+
+    if (cache->cache_data) {
```
sjanc, line 310: this could use some comment:
elem_size is converted to host, but other params (ix, offset) are used as is
also elem_size is (32bit size?) int but it is stored into cache_data which is uint16

If cache is kept in host order, maybe this function should get all params in host order?

kasjer: inconsistent endianness removed.
function argument change to uint16_t to match calling argument

sjanc (approved):

PR #3302, opened 2024-09-02
hw/bsp: pca10059 add readme.md

kasjer:
This adds readme.md files that explains how to use board without hardware debugger when factory programmed Open DFU Bootloader is present.


sjanc (approved):

jfclere:
Many thanks

PR #3307, opened 2024-09-09
hw/drivers/bme280: Fix writes to I2C

kasjer:
I2C writes when driver was enabled clear MSB of register address.

This prevented any write to control registers.

Now registers MSB is cleared on write only for SPI.


jfclere:
That fixes my problem with bme280 and i2c.

sjanc (approved):

PR #3312, opened 2024-09-30
[wip] hw/drivers: Add ipc_icbmsg

mkasenberg:


hw/mcu/nordic/nrf5340/syscfg.yml
```diff
@@ -579,6 +579,17 @@ syscfg.defs:
             Adds default MPU configuration package
         value: 0
 
+    IPC_NRF5340_PRE_TRUSTZONE_NETCORE_BOOT:
```
MariuszSkamra, line 582: Looks like `IPC_NRF5340_PRE_TRUSTZONE_NETCORE_BOOT` is already defined in `drivers/ipc_nrf5340/syscfg.yml`. Shouldn't this be removed from there? Otherwise we'll have two duplicated definition.

PR #3314, opened 2024-10-02
mcu/stm32: Fix clock configuration for non STM32F4xx

kasjer:
Only STM32F411 devices have PLLI2S PLLM parameter.

Also TIPPresSelection is not always present but since it was initialized with 0 initializing line is omitted.


sjanc (approved):
I guess setting RTCClockSelection=0 could also be omitted

PR #3318, opened 2024-10-04
Add NVIC_DisableAll and Cortex_DisableAll functions

kasjer:
Function `NVIC_DisableAll()` disable all interrupts on NVIC controller.
`Cortex_DisableAll()` disables all NVIC interrupts and stops SytTick that also can generate interrupt.

Call to `Cortex_DisableAll()` is added to all hal_bsp_deinit() functions that are executed before control is passed from the bootloader ot the application.

In some cases hal_bsp_deinit was added.

`hal_system_init()` was added to platforms that did not have this function.
Now `hal_system_init()` calls `NVIC_Relocate()` (previously NVIC_Relocate was often called from `SystemInit()` that is usually provided by MCU manufacturer. So mynewt had copies of those files with local modifications that made it harder to maintain in case of upgrade.

Same prototypes
```c
void NVIC_Relocate(void);
void NVIC_SetVector(IRQn_Type IRQn, uint32_t vector);
uint32_t NVIC_GetVector(IRQn_Type IRQn);
```
were placed in many headers and now are put in just one.

NXP MK8xF now uses common startup code with auto-generated linker script like ST and Nordic chips.


sjanc (approved):
but please rebase before merge to trigger CI

PR #3321, opened 2024-10-17
DA1469x: Clock time shifts on reboots

apc067:
If the LP clock's frequency is not an integer multiple of the specified OS tick frequency, the period of OS ticks will be incorrect due to integer division truncation errors (to the tune of almost 1% for RCX), resulting in OS Time
gradually creeping away from the actual time. This PR fixes the issue, by maintaining the fractional OS tick value and advance OS Time accordingly.

This PR also improves on the implementation of the DA1469x clock driver and updates the RTC divider registers after the selected LP clock is calibrated.

### DIFF
--- a/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
+++ b/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
@@ -62,6 +62,21 @@ void da1469x_clock_sys_xtal32m_switch_safe(void);
 void da1469x_clock_sys_rc32m_disable(void);
 
 /**
+ * Enable RC32M
+ */
+void da1469x_clock_sys_rc32m_enable(void);
+
+/**
+ * Switch sys_clk to RC32M
+ */
+void da1469x_clock_sys_rc32m_switch(void);
+
+/**
+ * Disable XTAL32K
+ */
+void da1469x_clock_lp_xtal32k_disable(void);
+
+/**
  * Enable XTAL32K
  */
 void da1469x_clock_lp_xtal32k_enable(void);
@@ -72,6 +87,21 @@ void da1469x_clock_lp_xtal32k_enable(void);
  * Caller shall ensure XTAL32K is already settled.
  */
 void da1469x_clock_lp_xtal32k_switch(void);
+
+/**
+ * Disable RC32K
+ */
+void da1469x_clock_lp_rc32k_disable(void);
+
+/**
+ * Enable RC32K
+ */
+void da1469x_clock_lp_rc32k_enable(void);
+
+/**
+ * Switch lp_clk to RC32K
+ */
+void da1469x_clock_lp_rc32k_switch(void);
 
 /**
  * Enable RCX
@@ -96,6 +126,11 @@ void da1469x_clock_lp_rcx_calibrate(void);
 void da1469x_clock_lp_rc32k_calibrate(void);
 
 /**
+ * Calibrate XTAL32K
+ */
+void da1469x_clock_lp_xtal32k_calibrate(void);
+
+/**
  * Calibrate RC32M
  */
 void da1469x_clock_lp_rc32m_calibrate(void);
@@ -109,6 +144,11 @@ uint32_t da1469x_clock_lp_rcx_freq_get(void);
  * Get calibrated (measured) RC32K frequency
  */
 uint32_t da1469x_clock_lp_rc32k_freq_get(void);
+
+/**
+ * Get calibrated XTAL32K frequency
+ */
+uint32_t da1469x_clock_lp_xtal32k_freq_get(void);
 
 /**
  * Get calibrated (measured) RC32M frequency
--- a/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
+++ b/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
@@ -131,6 +131,11 @@ void da1469x_clock_lp_rc32k_calibrate(void);
 void da1469x_clock_lp_xtal32k_calibrate(void);
 
 /**
+ * Calibrate selected LP clock
+ */
+void da1469x_clock_lp_calibrate(void);
+
+/**
  * Calibrate RC32M
  */
 void da1469x_clock_sys_rc32m_calibrate(void);
@@ -149,6 +154,11 @@ uint32_t da1469x_clock_lp_rc32k_freq_get(void);
  * Get calibrated XTAL32K frequency
  */
 uint32_t da1469x_clock_lp_xtal32k_freq_get(void);
+
+/**
+ * Get seleted LP clock's frequency
+ */
+uint32_t da1469x_clock_lp_freq_get(void);
 
 /**
  * Get calibrated (measured) RC32M frequency
--- a/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
+++ b/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
@@ -133,7 +133,7 @@ void da1469x_clock_lp_xtal32k_calibrate(void);
 /**
  * Calibrate RC32M
  */
-void da1469x_clock_lp_rc32m_calibrate(void);
+void da1469x_clock_sys_rc32m_calibrate(void);
 
 /**
  * Get calibrated (measured) RCX frequency
@@ -153,7 +153,7 @@ uint32_t da1469x_clock_lp_xtal32k_freq_get(void);
 /**
  * Get calibrated (measured) RC32M frequency
  */
-uint32_t da1469x_clock_lp_rc32m_freq_get(void);
+uint32_t da1469x_clock_sys_rc32m_freq_get(void);
 
 /**
  * Disable RCX
--- a/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
+++ b/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
@@ -171,6 +171,11 @@ uint32_t da1469x_clock_sys_rc32m_freq_get(void);
 void da1469x_clock_lp_rcx_disable(void);
 
 /**
+ * Set the RTC dividers
+ */
+void da1469x_clock_lp_set_rtc_divs(void);
+
+/**
  * Enable AMBA clock(s)
  *
  * @param mask
--- a/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
+++ b/hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
@@ -173,7 +173,7 @@ void da1469x_clock_lp_rcx_disable(void);
 /**
  * Set the RTC dividers
  */
-void da1469x_clock_lp_set_rtc_divs(void);
+void da1469x_clock_lp_set_rtc_divs(uint32_t rtc_clock_freq);
 
 /**
  * Enable AMBA clock(s)
--- a/hw/mcu/dialog/da1469x/src/da1469x_clock.c
+++ b/hw/mcu/dialog/da1469x/src/da1469x_clock.c
@@ -111,7 +111,8 @@ da1469x_clock_sys_xtal32m_init(void)
      */
     xtalrdy_cnt = MYNEWT_VAL(MCU_CLOCK_XTAL32M_SETTLE_TIME_US) * XTALRDY_IRQ_FREQ_MAX / 1000000;
 
-    CRG_XTAL->XTALRDY_CTRL_REG = (xtalrdy_cnt << CRG_XTAL_XTALRDY_CTRL_REG_XTALRDY_CNT_Pos) |
+    CRG_XTAL->XTALRDY_CTRL_REG = ((xtalrdy_cnt << CRG_XTAL_XTALRDY_CTRL_REG_XTALRDY_CNT_Pos) &
+                                  CRG_XTAL_XTALRDY_CTRL_REG_XTALRDY_CNT_Msk) |
                                  (DA1469X_XTALRDY_CLK_SEL << CRG_XTAL_XTALRDY_CTRL_REG_XTALRDY_CLK_SEL_Pos);
 }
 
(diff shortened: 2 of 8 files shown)


kasjer (commented):
Number of functions are moved around that makes it harder to see what is new and what was really modified/added.

hw/mcu/dialog/da1469x/include/mcu/da1469x_clock.h
```diff
@@ -73,52 +123,91 @@ void da1469x_clock_lp_xtal32k_enable(void);
...
-uint32_t da1469x_clock_lp_rc32m_freq_get(void);
+void da1469x_clock_lp_calibrate(void);
 
 /**
- * Disable RCX
+ * Initialize selected LP RC clock
  */
-void da1469x_clock_lp_rcx_disable(void);
+void
+da1469x_lpclk_rc_init();
```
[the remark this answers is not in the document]
apc067: Fixed.

hw/hal/include/hal/hal_system.h
```diff
@@ -95,6 +95,16 @@ const char *hal_reset_cause_str(void);
...
+/**
+ * Calibrate selected LP clock
+ */
+void hal_system_lpclk_calibrate(void);
+
+/**
+ * Notify about LP clock update
+ */
+void hal_system_lpclk_updated(void);
+
```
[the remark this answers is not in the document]
apc067: After a discussion with my team I realized that I had the wrong understanding about what should be added to this HAL API - now fixed.

hw/hal/include/hal/hal_os_tick.h
```diff
@@ -49,6 +49,12 @@ void os_tick_init(uint32_t os_ticks_per_sec, int prio);
  */
 void os_tick_idle(os_time_t n);
 
+/**
+ * Calculate the OS tick parameters.
+ *
+ * @param os_ticks_per_sec The number of OS ticks per second
+ */
+void hal_os_tick_calc_params(uint32_t os_ticks_per_sec);
```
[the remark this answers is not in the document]
apc067: Looking at the other services in hal_os_tick.c, I still believe this function belongs here. This is a new addition to the API, for the OS tick dividers to be re-tuned (case in point, after an oscillator re-calibration). It can be implemented for the other MCU's as well, on a need-to-do basis.

apc067:
> Number of functions are moved around that makes it harder to see what is new and what was really modified/added.

I see your concern; I separated the original commit into two, one only rearranges the functions with no changes whatsoever, and the second one adds the missing services to the API.

hw/hal/include/hal/hal_os_tick.h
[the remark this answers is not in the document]
kasjer: I'm not convinced.
There is no code (in public repository) that requires this prototype. Even file in this commit does not need this.
If I was to implement this function for other MCU I would not know what is the purpose, why OS needs it and when it should be called.
To make your case it would be nice to have other code that utilizes this functionality. For now this function could be static in your hal_os_tick.c without any prototype.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
```diff
@@ -547,6 +556,21 @@ da1469x_clock_lp_freq_get(void)
 #endif
 }
 
+static void
+da1469x_clock_lp_set_rtc_divs(void)
+{
+    /* Please see section 2.20.4 for details */
```
kasjer, line 562: Could you be more specific what section 2.20.4 is?

kasjer:
If your first commit does not change functionality is it really needed?

hw/mcu/dialog/da1469x/src/da1469x_clock.c
```diff
@@ -305,11 +355,18 @@ da1469x_clock_calibrate(uint8_t clock_sel, uint16_t ref_cnt)
...
+da1469x_clock_sys_rc32m_calibrate(void)
 {
     g_mcu_clock_rc32m_freq = da1469x_clock_calibrate(1, 100);
 }
 
+void
+da1469x_clock_lp_xtal32k_calibrate(void)
+{
+    g_mcu_clock_xtal32k_freq =
+        da1469x_clock_calibrate(2, MYNEWT_VAL(MCU_CLOCK_RCX_CAL_REF_CNT));
```
andrzej-kaczmarek: this should have separate syscfg

hw/mcu/dialog/da1469x/src/da1469x_lpclk.c
```diff
@@ -38,6 +38,7 @@ static void
 da1469x_lpclk_settle_tmr_cb(void *arg)
 {
     da1469x_clock_lp_xtal32k_switch();
+    da1469x_clock_lp_xtal32k_calibrate();
```
andrzej-kaczmarek, line 41: Calibration of XTAL32K should be optional and disabled by default - we should assume after settling it's accurate and has its native frequency.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
```diff
@@ -342,29 +385,36 @@ da1469x_clock_calibrate(uint8_t clock_sel, uint16_t ref_cnt)
...
 
-    return 32000000 * ref_cnt / ref_val;
+    return CAL_REF_FREQ * ref_cnt / ref_val;
 }
 
 void
 da1469x_clock_sys_rc32m_calibrate(void)
 {
-    g_mcu_clock_rc32m_freq = da1469x_clock_calibrate(1, 100);
+    g_mcu_clock_rc32m_freq = da1469x_clock_calibrate(DA1469X_CALIB_RC32M, RC32M_CAL_REF_CNT);
```
andrzej-kaczmarek, line 419: ref cnt should be configurable via syscfg for consistency

hw/mcu/dialog/da1469x/src/da1469x_clock.c
```diff
@@ -534,3 +534,27 @@ da1469x_clock_lp_rcx_freq_get(void)
...
+
+void
+da1469x_clock_lp_calibrate(void)
+{
+#if MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, RCX)
+    da1469x_clock_lp_rcx_calibrate();
+#elif MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, RC32K)
+    da1469x_clock_lp_rc32k_calibrate();
+#elif MYNEWT_VAL_CHOICE(MCU_LPCLK_SOURCE, XTAL32K)
+    da1469x_clock_lp_xtal32k_calibrate();
```
andrzej-kaczmarek, line 582: as in other commit, this should be optional

hw/hal/include/hal/hal_os_tick.h
andrzej-kaczmarek, line 1: This code should be optional as it only applies to RCX. For devices which use XTAL32K and e.g. 128 OS ticks per seconds there's really no need to have extra overhead for ticks calculation.

andrzej-kaczmarek (changes requested):
Please don't move code around in `.c` files - this breaks git annotate

hw/hal/include/hal/hal_os_tick.h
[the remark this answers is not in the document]
apc067: As a matter of fact, the reason I went thru the trouble of carving this out into a function & adding it to the API is that I wish to use it from our code.

hw/mcu/dialog/da1469x/src/da1469x_lpclk.c
apc067: See below...

hw/hal/include/hal/hal_os_tick.h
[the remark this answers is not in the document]
andrzej-kaczmarek: I don't really see why external code would need to call this. This calculates parameters based on current lpclk frequency so it just has to be done after lpclk frequency has changed e.g. after calibration so it can be called from that routine directly.

hw/hal/include/hal/hal_os_tick.h
apc067: I think a good way to handle this would be for the caller to either call hal_os_tick_calc_params() or not. (In other words, this API itself could be agnostic about which oscillator is being used - it just cares about the frequency.)

apc067:
> Please don't move code around in `.c` files - this breaks git annotate

Yeah, keeping the git annotate may indeed outweigh the benefits of having the functions in a neat order. To that extent, should I keep the "da1469x/clock: Eliminate magic numbers" commit, or should I leave those lines unchanged as well?

hw/mcu/dialog/da1469x/src/da1469x_clock.c
apc067: Sure.

hw/hal/include/hal/hal_os_tick.h
andrzej-kaczmarek: I added this comment to `.h` by mistake - it should apply to `.c` file (i.e. ticks calculation). Sorry for confusion.

andrzej-kaczmarek:
> > Please don't move code around in `.c` files - this breaks git annotate
> 
> Yeah, keeping the git annotate may indeed outweigh the benefits of having the functions in a neat order. To that extent, should I keep the "da1469x/clock: Eliminate magic numbers" commit, or should I leave those lines unchanged as well?

Keep it, removing magic numbers is fine.

hw/hal/include/hal/hal_os_tick.h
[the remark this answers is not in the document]
andrzej-kaczmarek: Actually calling this from  `da1469x_lpclk_updated` should be enough.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
```diff
@@ -29,39 +29,92 @@
...
      */
-    xtalrdy_cnt = MYNEWT_VAL(MCU_CLOCK_XTAL32M_SETTLE_TIME_US) * 1000 / 4085;
+    xtalrdy_cnt = MYNEWT_VAL(MCU_CLOCK_XTAL32M_SETTLE_TIME_US) * XTALRDY_IRQ_FREQ_MAX / 1000000;
 
-    reg = CRG_XTAL->XTALRDY_CTRL_REG;
-    reg &= ~(CRG_XTAL_XTALRDY_CTRL_REG_XTALRDY_CNT_Msk);
-    reg |= CRG_XTAL_XTALRDY_CTRL_REG_XTALRDY_CLK_SEL_Msk;
-    reg |= xtalrdy_cnt;
-    CRG_XTAL->XTALRDY_CTRL_REG = reg;
+    CRG_XTAL->XTALRDY_CTRL_REG = (xtalrdy_cnt << CRG_XTAL_XTALRDY_CTRL_REG_XTALRDY_CNT_Pos) |
```
vrahane, line 116: nit-pick: nothing major, I think we should use the masks here specially because values are coming from a syscfg.

apc067: Good point; thanks!

apc067:
Addressed all current reviewer comments.

hw/hal/include/hal/hal_os_tick.h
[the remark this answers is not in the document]
apc067: Agreed & done.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
apc067: Done.

hw/mcu/dialog/da1469x/src/da1469x_lpclk.c
apc067: Done.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
apc067: Done.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
apc067: Done - the conditional is applied within `da1469x_clock_lp_xtal32k_calibrate()`.

vrahane (approved):
I tested it. Seems to be working great. Once all the comments are addressed, we can merge it in.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
```diff
@@ -286,6 +336,17 @@ da1469x_clock_lp_rc32k_calibrate(void)
...
 }
 
+void
+da1469x_clock_lp_xtal32k_calibrate(void)
+{
+#if MYNEWT_VAL(MCU_CLOCK_XTAL32K_ALLOW_CALIB)
+    g_mcu_clock_xtal32k_freq =
+        da1469x_clock_calibrate(2, MYNEWT_VAL(MCU_CLOCK_RCX_CAL_REF_CNT));
+#else
+    g_mcu_clock_xtal32k_freq = XTAL32M_FREQ;
```
andrzej-kaczmarek, line 407: XTAL32K_FREQ

andrzej-kaczmarek (changes requested):
one issue left, otherwise lgtm.

please also fix coding style to pass CI

sjanc:
yeap, looks ok after this typo and minor coding style is fixed:)
Thanks!

hw/mcu/dialog/da1469x/src/da1469x_clock.c
apc067: One of these typos still managed to creep through - good find!

apc067: Fixed by amending the original commit.

hw/mcu/dialog/da1469x/src/da1469x_clock.c
apc067: Done.

kasjer: I could not find "DA1469x design doc" anywhere are you sure it is available?

apc067: The actual file name is "Lithium_Dialog_Design_Documentation_RevE.pdf". I don't remember where I got it from; I believe it may have been DM'd to me.

andrzej-kaczmarek: this is described in section 34.3 of DA1469x datasheet so please update comment accordingly

apc067: OK, I see... looks like the design doc was fashioned into the actual datasheet. Comment updated.

apc067: Actually, the Lithium Design Doc is specific to our version of the MCU - however, the features mynewt-core covers apply to DA1469x, so referencing that datasheet is indeed more appropriate.

PR #3324, opened 2024-10-23
sys/config: Add option to skip dup check when saving values

andrzej-kaczmarek:
By default value is written to config storage only if it changed since last stored value. This however requires that all stored values are read to determine last value and in case of conf_save this is done for each saved value which can take a lot of time.

This adds option to skip that dup check and simply stores new value. This results in much faster writes at the expense of load times as there are more copies of the same value stored. Also conf_save will always write all exported values so this increase number of writes to flash. However, e.g. in case of large FCB used as config storage dup check on write can take much longer than simple read or store without dup checks.

### DIFF
--- a/sys/config/src/config_store.c
+++ b/sys/config/src/config_store.c
@@ -186,6 +186,7 @@ conf_get_stored_value(char *name, char *buf, int buf_len)
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


kasjer (approved):

vrahane (approved):
LGTM

PR #3330, opened 2024-10-29
sys/shell: Fix the Tasks table header row

apc067:
This commit removes the "flg" label from the Tasks table header, after the column itself was removed in commit 6c918ac ("Remove not used oti_flags from struct os_task_info").


kasjer:
> This commit removes the "flg" label from the Tasks table header, after the column itself was removed in commit [64e8e16](https://github.com/apache/mynewt-core/commit/64e8e167b2c82bf50b505122fc8fbce8c5355677) ("Remove not used oti_flags from struct os_task_info").

ID is incorrect in commit (short version) and in PR description

apc067:
Aw, the classic case of a text copy not going through - fixed in both places.

kasjer (approved):

PR #3338, opened 2024-11-06
kernel/os: Increase mutex test timeout

m-gorecki:
On slower VMs (for example GHA's mac-os) tests were failing, as previous timeout was not enough.


kasjer (approved):
LGTM

sjanc (approved):

PR #3349, opened 2024-11-26
kernel/os: Update os_assert_cb to match assert_func

andrzej-kaczmarek:
This updates os_assert_cb prototype to match assert_func. This allows to get proper assert information in callback e.g. for logging.

Note that this breaks API, but since updates in other projects are trivial I'd like to get this merged. A good example why this is useful is a blehci app which can use os_assert_cb to log asserts system-wide via HCI vs event: https://github.com/apache/mynewt-nimble/pull/1930


kasjer (dismissed):

kasjer:
`os_assert_cb()` is called in many more **os_fault.c** files

kasjer (approved):

PR #3358, opened 2024-12-18
lvgl: Add support for FSMC as 16 bit interface

kasjer:
STM32 FSMC peripheral (for accessing external memories) is used to write to LCD with 8080 II 16 bit interface. It allows for maximum performance when LCD is connected to dedicated pins.


rymanluk (approved):
LGTM

PR #3359, opened 2024-12-20
hw/mcu/stm32f4: Add shell command to show clocks

kasjer:
This adds package that can display all enabled clocks. Similar to what was already done for stm32f1


rymanluk (approved):
LGTM

PR #3362, opened 2025-01-15
STM32F7 I2S Support

kasjer:
This adds support for I2S for STM32F7 devices

Change adds i2s driver based on STM32F4

Two nucleo STM32F7 BSP boards configuration for I2S (PLL and ST HAL)


kasjer:
Style check failed due to complaint from uncrustify that can't be unreasonable configured for certain macros

kasjer:
> this looks very similar to f4, wondering if that could be some common package in future?

It's a possibility, once common part is clearly found.
DMA and clocks are different peripheral itself (based on SPI) is mostly the same across most (all?) ST devices.

PR #3364, opened 2025-01-24
bsp/stm32f74x: Fix RAM and DTCM description

kasjer:
Start or RAM and size of DTCM were taken from STM32F76x series.

DTCM should be 64KB
and RAM1 should start at 0x20010000


sjanc (approved):

harteware07:
These changes fixed my problems that the generated binary crashed immediately. Thank you!

PR #3370, opened 2025-02-19
tinyusb/msc_fat_view: 

apc067:
Change order of package inits, so that the coredump file(s) are added to MSC-FAT-View successfully.

### DIFF
--- a/hw/usb/tinyusb/msc_fat_view/pkg.yml
+++ b/hw/usb/tinyusb/msc_fat_view/pkg.yml
@@ -76,7 +76,9 @@ pkg.link_tables:
 pkg.init.'(!BOOT_LOADER && TINYUSB_AUTO_START!=0)':
     msc_fat_view_pkg_init: $before:tinyusb_start
 
+pkg.init.'(!BOOT_LOADER && TINYUSB_AUTO_START==0)':
+    msc_fat_view_pkg_init: MYNEWT_VAL(MSC_FAT_VIEW_SYSINIT_STAGE)
+
 pkg.init.'!BOOT_LOADER && MSC_FAT_VIEW_COREDUMP_FILES':
     msc_fat_view_coredump_pkg_init:
         - $before:msc_fat_view_pkg_init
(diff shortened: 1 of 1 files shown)


hw/usb/tinyusb/msc_fat_view/pkg.yml
```diff
@@ -78,5 +78,5 @@ pkg.init.'(!BOOT_LOADER && TINYUSB_AUTO_START!=0)':
 
 pkg.init.'!BOOT_LOADER && MSC_FAT_VIEW_COREDUMP_FILES':
     msc_fat_view_coredump_pkg_init:
-        - $after:msc_fat_view_pkg_init
```
kasjer, line 81: @apc067 you could also remove next line so there is only one **$before**

You could also add following lines if you want:
```yml
pkg.init.'(!BOOT_LOADER && TINYUSB_AUTO_START==0)':
    msc_fat_view_pkg_init: MYNEWT_VAL(MSC_FAT_VIEW_SYSINIT_STAGE)
```
those would remove build error if TINYUSB_AUTO_START is 0.
This setting is intended to delay start of USB so application can do some more before USB is detected, application can call `tinyusb_start()` manually if for example BLE stack should be activate first.

If you are not confident with my suggestion just leave single **$before** and I will add those additional lines later.

apc067: OK, I see... being before `msc_fat_view_pkg_init` already implies being before `tinyusb_start` as well.

As for the `pkg.init.'(!BOOT_LOADER && TINYUSB_AUTO_START==0)'` clause, the idea is that `MSC_FAT_VIEW_SYSINIT_STAGE` is set to a very conservative (high) value, but if there is TINYUSB_AUTO_START, the `USBD_SYSINIT_STAGE` is serving as an "anchor" for instead, Is this correct?

I implemented your recommended changes.

kasjer (approved):
LGTM, tested OK

PR #3381, opened 2025-03-03
log_fcb: Init all usage of fcb_entry on stack

kasjer:
In several cases struct fcb_entry was not initialized before usage.
In some cases it could lead to unpredictable behaviour.


vrahane (approved):
LGTM, it would be good to understand what bugs you observed or how the tests were failing ?

sjanc (approved):

PR #3382, opened 2025-03-06
fs/fcb: FCB has some async issues, fixing some obvious ones

vrahane:
We are seeing some logging issues related to rotates, on analyzing the FCB code, it seems not protecting FCB access using the mutex might be one of the reasons and causes.

### DIFF
--- a/fs/fcb/src/fcb.c
+++ b/fs/fcb/src/fcb.c
@@ -131,6 +131,17 @@ fcb_free_sector_cnt(struct fcb *fcb)
 }
 
 int
+fcb_is_empty_nolock(struct fcb *fcb)
+{
+    bool ret = false;
+
+    ret = (fcb->f_active.fe_area == fcb->f_oldest &&
+           fcb->f_active.fe_elem_off == sizeof(struct fcb_disk_area));
+
+    return ret;
+}
+
+int
 fcb_is_empty(struct fcb *fcb)
 {
     int rc = 0;
@@ -262,13 +273,13 @@ fcb_offset_last_n(struct fcb *fcb, uint8_t entries,
 
     i = 0;
     memset(&loc, 0, sizeof(loc));
-    while (!fcb_getnext(fcb, &loc)) {
+    while (!fcb_getnext_nolock(fcb, &loc)) {
         if (i == 0) {
             /* Start from the beginning of fcb entries */
             *last_n_entry = loc;
         } else if (i > (entries - 1)) {
             /* Update last_n_entry after n entries and keep updating */
-            fcb_getnext(fcb, last_n_entry);
+            fcb_getnext_nolock(fcb, last_n_entry);
         }
         i++;
     }
@@ -293,7 +304,7 @@ fcb_clear(struct fcb *fcb)
         return FCB_ERR_ARGS;
     }
 
-    while (!fcb_is_empty(fcb)) {
+    while (!fcb_is_empty_nolock(fcb)) {
         rc = fcb_rotate(fcb);
         if (rc) {
             break;
--- a/fs/fcb/src/fcb.c
+++ b/fs/fcb/src/fcb.c
@@ -110,6 +110,12 @@ fcb_free_sector_cnt(struct fcb *fcb)
 {
     int i;
     struct flash_area *fa;
+    int rc = 0;
+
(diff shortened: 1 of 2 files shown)


fs/fcb/src/fcb.c
```diff
@@ -118,14 +124,29 @@ fcb_free_sector_cnt(struct fcb *fcb)
...
 int
 fcb_is_empty(struct fcb *fcb)
 {
-    return (fcb->f_active.fe_area == fcb->f_oldest &&
+    int rc = 0;
+    bool ret = false;
+
+    rc = os_mutex_pend(&fcb->f_mtx, OS_WAIT_FOREVER);
+    if (rc && rc != OS_NOT_STARTED) {
+        return FCB_ERR_ARGS;
```
sjanc, line 141: Would be good to note in header file that negative value means error and not that it is not empty (non-zero value)

(alternative would be to make this   int fcb_is_empty(struct fcb *fcb, bool *empty) for more explicit error vs return value

fs/fcb/src/fcb.c
```diff
@@ -226,8 +247,14 @@ fcb_offset_last_n(struct fcb *fcb, uint8_t entries,
...
 {
     struct fcb_entry loc;
+    int rc = 0;
     int i;
 
+    rc = os_mutex_pend(&fcb->f_mtx, OS_WAIT_FOREVER);
+    if (rc && rc != OS_NOT_STARTED) {
+        return FCB_ERR_ARGS;
+    }
+
```
sjanc, line 257: there is fcb_getnext_nolock() which could be used in loop below to avoid nested locking

fs/fcb/src/fcb.c
```diff
@@ -110,6 +110,12 @@ fcb_free_sector_cnt(struct fcb *fcb)
 {
     int i;
     struct flash_area *fa;
+    int rc = 0;
+
+    rc = os_mutex_pend(&fcb->f_mtx, OS_WAIT_FOREVER);
+    if (rc && rc != OS_NOT_STARTED) {
+        return FCB_ERR_ARGS;
```
sjanc, line 117: Would be good to note in header file that negative value means error (or make cnt output parameter similar to suggestion in fcb_is_empty())   (either is fine for me)

fs/fcb/src/fcb.c
```diff
@@ -257,14 +286,21 @@ fcb_offset_last_n(struct fcb *fcb, uint8_t entries,
 int
 fcb_clear(struct fcb *fcb)
 {
-    int rc;
+    int rc = 0;
+
+    rc = os_mutex_pend(&fcb->f_mtx, OS_WAIT_FOREVER);
+    if (rc && rc != OS_NOT_STARTED) {
+        return FCB_ERR_ARGS;
+    }
```
sjanc, line 294: loop below could use _nolock() variants to avoid nested locking

sjanc (commented):
Hi,

To avoid nested locking I'd suggest to introduce _nolock private variants of functions that are used under mutex similar to how fcb_getnext_nolock() and fcb_getnext() are handled.

and commit message could be a bit more descriptive, eg

```
fs/fcb: Fix FCB  async issues

FCB  API can be called from multiple threads. Add missing mutex locking.
```

fs/fcb/src/fcb.c
vrahane: I think using the return code is good enough, will just update the header to include the updated return code.

fs/fcb/src/fcb.c
vrahane: Will update the header file with the description.

vrahane:
@sjanc  I have updated the PR addressing some review comments. Please take a look.

fs/fcb/src/fcb.c
```diff
@@ -257,14 +286,21 @@ fcb_offset_last_n(struct fcb *fcb, uint8_t entries,
...
 
-    rc = 0;
     while (!fcb_is_empty(fcb)) {
```
sjanc, line 296: so if there is no fsb_is_empty_nolock() this should handle error checking from  fsb_is_empty()

same for fcb_free_sector_cnt() and other that had mutex added if those are used internally

vrahane: Added `fcb_is_empty_nolock()`

PR #3394, opened 2025-03-21
crypto/mbedtls: Use config defining flag globally

m-gorecki:
Now -DMBEDTLS_USER_CONFIG_FILE=<mbedtls/config_mynewt.h> flag is used globally for each build that depends on mbedtls package. This way we won't have to add this flag in each package that uses mbedtls.


kasjer (approved):
LGTM

PR #3398, opened 2025-03-26
boot/startup: Add optional stack fill in startup

kasjer:
Code was already present in default startup code for NRF5340. Now is available for all devices with common startup.

It allows to verify real main stack requirements (interrupt stack) and stack used in bootloader.

### DIFF
--- a/boot/startup/syscfg.yml
+++ b/boot/startup/syscfg.yml
@@ -34,6 +34,11 @@ syscfg.defs:
             For bootloader it's main stack, for application this stack is used by interrupts
             and exceptions.
         value: 768
+    MAIN_STACK_FILL:
+        description: >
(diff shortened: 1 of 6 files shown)


boot/startup/syscfg.yml
```diff
@@ -34,6 +34,10 @@ syscfg.defs:
             For bootloader it's main stack, for application this stack is used by interrupts
             and exceptions.
         value: 768
+    MAIN_STACK_FILL:
+        description: >
+            Fill main stack (interrupt stack) with pattern on startup.
+        value: 0
```
andrzej-kaczmarek, line 40: non-zero value could be used as a pattern

kasjer: Update as requested

andrzej-kaczmarek (approved):

PR #3399, opened 2025-03-26
Enable common startup code and autogenerated linker script in NRF5340

kasjer:
This adds the first step to utilize common startup code and autogenerated linker scripts for NRF53 devices.

NRF5340 already had option to use nonstandard startup code.
With this change it is possible to use this method however it will be remove in the future.
User will still have option to disable common startup code and linker script.

Additionally embedded net core image uses link_table feature from newt tool.

This also adds two boards with NRF5340 device from Ublox and Fanstel.


sjanc (approved):
CI error should be fixed by https://github.com/apache/mynewt-core/pull/3409

andrzej-kaczmarek (approved):

PR #3400, opened 2025-03-27
sys/log, fs/fcb: Add support for trailers per entry in logs for fcb

vrahane:
- Trailer callbacks can be enabled using
  LOG_FLAGS_TRAILER_SUPPORT.
- This is needed since we cannot update the log header
  without breaking backwards compatibility
- Adding a trailer keeps the log header in place and the entry
  intact. On an upgrade or downgrade, log entries are still
  readable.
- Trailer callbacks can be registered using log_trailer_cbs_register().
- Trailer data is managed by the callbacks
- Add trailer buffer append/free callback
- Add reset_data callback for trailer data
- Allocate trailer buffer in log_append_prepare() and
  free the trailer buffer with log_trailer_free() in
  log.c. Both of them being callbacks to be implemented
  by the user.
- Add 2 byte trailer length field after trailer at
  the end of the log entry.
Note: This is a replacement for the number of entries PR, https://github.com/apache/mynewt-core/pull/3168 and most of the comments should be pre-handled in this code. the trailer can be implemented by the user of this feature as they want privately only keeping the trailer handling code upstream.

### DIFF
--- a/sys/log/full/include/log/log.h
+++ b/sys/log/full/include/log/log.h
@@ -77,6 +77,46 @@ struct log_storage_info {
 };
 #endif
 
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER)
+
+#define LOG_TRAILER_LEN_SIZE 2
+
+/** @typedef log_trailer_free_func_t
+ * @brief Callback that frees the trailer buffer, to be called by user of the
+ * library once append returns, either successfully or unsuccessfully
+ *
+ * @param log                 The log the trailer is being appended to
+ * @param om                  mbuf pointer to be freed
+ * @param arg                 Argument to be passed to the trailer handler
+ *
+ * @return                    0 on success, non-zero on failure
+ */
+typedef int log_trailer_free_func_t(struct log *log, struct os_mbuf *om, void *arg);
+
+/** @typedef log_trailer_get_data_func_t
+ * @brief Callback that is executed each time the corresponding log entry is
+ * appended to
+ *
+ * @param log                 The log the trailer needs to be appended to
+ * @param om                  mbuf pointer to be set to mbuf which contains
+ *                            trailer data
+ * @param arg                 Argument to be passed to the trailer handler
+ *
+ * @return                    0 on success, non-zero on failure
+ */
+typedef int log_trailer_get_data_func_t(struct log *log, struct os_mbuf **om, void *arg);
+
+/** @typedef log_trailer_reset_data_func_t
+ * @brief Callback used to reset trailer data per log
+ *
+ * @param log                 The log the trailer is to be read from
+ * @param arg                 Argument to be passed to the trailer handler
+ *
+ * @return                    0 on success, non-zero on failure.
+ */
+typedef int log_trailer_reset_data_func_t(struct log *log, void *arg);
+#endif
+
 typedef int (*log_walk_func_t)(struct log *, struct log_offset *log_offset,
         const void *dptr, uint16_t len);
 
@@ -99,6 +139,7 @@ typedef int (*lh_append_mbuf_body_func_t)(struct log *log,
 typedef int (*lh_walk_func_t)(struct log *,
         log_walk_func_t walk_func, struct log_offset *log_offset);
 typedef int (*lh_flush_func_t)(struct log *);
+typedef uint16_t (*lh_read_entry_len_func_t)(struct log *log, const void *dptr);
 #if MYNEWT_VAL(LOG_STORAGE_INFO)
 typedef int (*lh_storage_info_func_t)(struct log *, struct log_storage_info *);
 #endif
@@ -118,6 +159,7 @@ struct log_handler {
     lh_walk_func_t log_walk;
     lh_walk_func_t log_walk_sector;
     lh_flush_func_t log_flush;
+    lh_read_entry_len_func_t log_read_entry_len;
 #if MYNEWT_VAL(LOG_STORAGE_INFO)
     lh_storage_info_func_t log_storage_info;
 #endif
@@ -131,8 +173,9 @@ struct log_handler {
 /* Image hash length to be looged */
 #define LOG_IMG_HASHLEN 4
 
-/* Flags used to indicate type of data in reserved payload*/
+/* Flags used to indicate type of data in reserved payload */
 #define LOG_FLAGS_IMG_HASH (1 << 0)
+#define LOG_FLAGS_TRAILER  (1 << 1)
 
 #if MYNEWT_VAL(LOG_VERSION) == 3
 struct log_entry_hdr {
@@ -204,6 +247,20 @@ STATS_SECT_END
 #define LOG_STATS_INCN(log, name, cnt)
 #endif
 
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER)
+/* Trailer support callbacks */
+struct log_trailer_handler {
+    /* Frees the trailer buffer */
+    log_trailer_free_func_t *lth_trailer_free;
+    /* Trailer append callback used to append trailer to the log entry */
+    log_trailer_get_data_func_t *lth_trailer_get_data;
+    /* Trailer reset data callback used to reset the trailer data
+     * per log which is defined by the application registering these callbacks
+     */
+    log_trailer_reset_data_func_t *lth_trailer_reset_data;
+};
+#endif
+
 struct log {
     const char *l_name;
     const struct log_handler *l_log;
@@ -225,6 +282,16 @@ struct log {
      * at init
      */
     log_walk_func_t l_init_cb;
+#endif
+
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER)
+    /* Log trailer support */
+    /* Void argument for callbacks registeration */
+    void *l_tr_arg;
+    /* Trailer mbuf */
+    struct os_mbuf *l_tr_om;
+    /* Trailer handler with callbacks */
+    struct log_trailer_handler *l_th;
 #endif
 };
 
@@ -523,6 +590,105 @@ int log_read(struct log *log, const void *dptr, void *buf, uint16_t off,
 
 #if MYNEWT_VAL(LOG_INIT_CB)
 /**
+ * Reads entry length from the specified log.
+ *
+ * @return                      The number of bytes of entry length; 0 on failure.
+ */
+uint16_t
+log_read_entry_len(struct log *log, const void *dptr);
+
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER)
+
+/**
+ * @brief                Read log trailer of a specific entry
+ *
+ * @param log            The log to read trailer data from
+ * @param dptr           Data pointer to the log entry
+ * @param buf            Pointer to the buffer to read the trailer into
+ *
+ * @return 0 on success; nonzero on failure.
+ */
+uint16_t
+log_read_trailer(struct log *log, const void *dptr, uint8_t *buf);
+
+/**Add commentMore actions
+ * @brief Returns trailer data in an mbuf specified by the caller
+ *
+ * @param log                   The log to append to
+ * @param om                    mbuf pointer to be set to buffer which contains
+ *                              trailer data
+ * @param arg                   Argument to be passed to the trailer handler
+ *
+ * @return                      0 on success; nonzero on failure.
+ */
+static inline int
+log_trailer_get_data(struct log *log, struct os_mbuf **om, void *arg)
+{
+    if (log->l_th && log->l_th->lth_trailer_get_data) {
+        return log->l_th->lth_trailer_get_data(log, om, arg);
+    }
+    return SYS_ENOTSUP;
+}
+
+/**
+ * @brief Callback that frees the trailer buffer, to be called by user of the library
+ * once append returns, either successfully or unsuccessfully
+ *
+ * @param log                   The log the trailer is being appended to
+ * @param om                    mbuf pointer to be freed
+ * @param arg                   Argument to be passed to the trailer handler
+ *
+ * @return                      0 on success; nonzero on failure.
+ */
+static inline int
+log_trailer_free(struct log *log, struct os_mbuf *om, void *arg)
+{
+    if (log->l_th && log->l_th->lth_trailer_free) {
+        return log->l_th->lth_trailer_free(log, om, arg);
+    }
+    return SYS_ENOTSUP;
+}
+
+/**
+ * @brief Resets the trailer data in the log, calling user callback
+ *
+ * @param log                   The log to reset.
+ * @param arg                   Argument to be passed to the trailer handler
+ *
+ * @return                      0 on success; nonzero on failure.
+ */
+static inline int
+log_trailer_reset_data(struct log *log, void *arg)
+{
+    if (log->l_th && log->l_th->lth_trailer_reset_data) {
+        return log->l_th->lth_trailer_reset_data(log, arg);
+    }
+    return SYS_ENOTSUP;
+}
+
+/**
+ * @brief Register trailer callbacks
+ *
+ * @param log            The log to read from.
+ * @param lth            The trailer handler to register.
+ * @param arg            Argument to be passed to the trailer handler
+ *
+ * @return 0 on success; nonzero on failure.
+ */
+static inline void
+log_trailer_cbs_register(struct log *log, struct log_trailer_handler *lth,
+                         void *arg)
+{
+    if (!lth || !log) {
+        return;
+    }
+
+    log->l_th = lth;
+    log->l_tr_arg = arg;
+}
+#endif
+
+/**
  * @brief Register Log init callback to be called by log_read_last_hdr
  * while reading the last log entry at init
  *
@@ -554,7 +720,7 @@ int log_read_hdr(struct log *log, const void *dptr, struct log_entry_hdr *hdr);
  * @brief Reads the header length
  *
  * @param hdr Ptr to the header
- * 
+ *
  * @return Length of the header
  */
 uint16_t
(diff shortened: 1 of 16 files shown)


vrahane:
Note: I fixed the coding styling issues but the compliance check is showing stale information.

andrzej-kaczmarek:
Please cleanup this PR so there are no commits that fix code added in other commits - it makes it more difficult to review.

Please also add some more verbose commit message that explains this feature since it's not a trivial fix.

vrahane:
@andrzej-kaczmarek  I have updated the PR. Please take a look.

sys/log/full/src/log.c
```diff
@@ -571,6 +691,10 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
     }
 #endif
 
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER_SUPPORT)
+    ue->ue_flags |= LOG_FLAGS_TRAILER_SUPPORT;
```
andrzej-kaczmarek, line 695: Not sure what is the purpose of this flag. I assume this should be an indication whether trailer is appended to log entry or not, but it's set for each entry even in case application decides not to append a trailer. How can application figure out if an entry has a trailer when log is read?

sys/log/full/src/log_cbmem.c
```diff
@@ -40,21 +51,61 @@ log_cbmem_append_body(struct log *log, const struct log_entry_hdr *hdr,
...
+#endif
     };
 
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER_SUPPORT)
+    if (hdr->ue_flags & LOG_FLAGS_IMG_HASH) {
+        sg->entries[1].flat_len = LOG_IMG_HASHLEN;
+    }
+
+    if (hdr->ue_flags & LOG_FLAGS_TRAILER_SUPPORT) {
+        rc = log_trailer_append(log, trailer, &trailer_len, NULL, NULL);
```
andrzej-kaczmarek, line 81: Usage of `log_trailer_append` is inconsistent in logs implementation. Here it seems that the underlying callback should write to `trailer` buffer since it's later written via `sg` struct. In FCB implementation the same parameter is not used so I assume callback should write data directly to FCB.

sys/log/full/syscfg.yml
```diff
@@ -196,6 +202,10 @@ syscfg.defs:
         restrictions:
             - LOG_FCB_BOOKMARKS
             - LOG_FCB
+    LOG_MAX_TRAILER_LEN:
+        description: >
+            Maximum length of trailer that can be appended to a log entry
+        value: "LF_MAX_ALIGN * 3"
```
andrzej-kaczmarek, line 208: I'd prefer that this is just a number, not something that is actually an injected code.

sys/log/full/syscfg.yml
```diff
@@ -63,6 +63,12 @@ syscfg.defs:
             1 - enable.
         value: 0
 
+    LOG_FLAGS_TRAILER_SUPPORT:
+        description: >
+            Enable logging TLV with custom data types in every log entry
```
andrzej-kaczmarek, line 68: Description is not correct.

sys/log/full/include/log/log.h
```diff
@@ -531,11 +660,10 @@ int log_read_hdr(struct log *log, const void *dptr, struct log_entry_hdr *hdr);
  * @brief Reads the header length
  *
  * @param hdr Ptr to the header
- * 
+ *
  * @return Length of the header
  */
-uint16_t
-log_hdr_len(const struct log_entry_hdr *hdr);
+uint16_t log_hdr_len(const struct log_entry_hdr *hdr);
```
andrzej-kaczmarek, line 666: This PR has a lot of unrelated code changes (e.g. code style changes) which should be removed. You can create separate PR for those changes.

sys/log/full/include/log/log.h
```diff
@@ -131,25 +213,48 @@ struct log_handler {
...
+#endif
+
+#define LOG_FCB_EXT_HDR_SIZE LOG_BASE_ENTRY_HDR_SIZE + LOG_IMG_HASHLEN + \
+    LF_MAX_ALIGN
+
+#ifndef LOG_FCB_FLAT_BUF_SIZE
+/* Assuming the trailer fits in this, an arbitrary value */
+#define LOG_FCB_FLAT_BUF_SIZE (LOG_FCB_EXT_HDR_SIZE > LF_MAX_ALIGN * 3) ? \
+    LOG_FCB_EXT_HDR_SIZE : LF_MAX_ALIGN * 3
+#endif
```
andrzej-kaczmarek, line 254: I don't really understand the logic behind above calculations and magic values.

sys/log/full/src/log.c
```diff
@@ -320,21 +324,41 @@ log_read_hdr_walk(struct log *log, struct log_offset *log_offset, const void *dp
         }
     }
 
+    if (arg->hdr->ue_flags & LOG_FLAGS_TRAILER_SUPPORT && log->l_th) {
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER_SUPPORT)
+        if (log->l_th->log_process_trailer) {
```
andrzej-kaczmarek, line 329: This is the only place where `process_trailer` callback is called. How can application actually access the trailer data when reading the log in a generic way?

andrzej-kaczmarek (changes requested):
It's quite hard to understand how this code should work and how it should be used since there's not much description for the new APIs and the usage of various callbacks is inconsistent. Also there are a lot of unrelated changes which makes it even harder to figure out which parts are relevant.

My general understanding is that it was supposed to work smth like:
For each new entry a callback is called so application can append an arbitrary data to a log entry. This data is effectively appended as a part of log body so there's a flag indicating presence of a trailer so the data can be later restored.

But it doesn't seem to do that, e.g. the flag is set for each entry regardless of the presence of the trailer, there's no way to read the trailer data from the log except for one single case handled in the code and so on. See inline comments for more details.

sys/log/full/include/log/log.h
```diff
@@ -204,6 +309,16 @@ STATS_SECT_END
...
 
+/* Trailer support callbacks */
+struct log_trailer_handler {
+    log_trailer_len_func_t *log_trailer_len;
+    log_trailer_data_len_func_t *log_trailer_data_len;
+    log_trailer_append_func_t *log_trailer_append;
+    log_process_trailer_func_t *log_process_trailer;
+    log_trailer_mbuf_append_func_t *log_trailer_mbuf_append;
+    log_trailer_reset_data_func_t *log_trailer_reset_data;
+};
```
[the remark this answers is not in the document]
vrahane: - So, earlier I had `l_trailer_arg` which you wanted me to make private so, I moved that out. Now, there needs to be a way to reset that trailer data which is what `trailer_reset_data` allows. If you have any other way to perform a reset, please let me know. This is called from `log_flush`.
- `trailer_len` helps user of the library to get the length of the trailer as the name suggest.
- `trailer_data_len` helps user of the library get the length of the trailer data. This was used earlier before I got rid of `l_trailer_arg` but now since the callbacks take care of the trailer data. I can remove it.

sys/log/full/src/log_cbmem.c
```diff
@@ -88,9 +140,21 @@ log_cbmem_append_mbuf_body(struct log *log, const struct log_entry_hdr *hdr,
     if (hdr->ue_flags & LOG_FLAGS_IMG_HASH) {
         sg.entries[1].flat_len = LOG_IMG_HASHLEN;
     }
+
+    if (hdr->ue_flags & LOG_FLAGS_TRAILER_SUPPORT) {
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER_SUPPORT)
```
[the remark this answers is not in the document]
vrahane: The reason is I don't want to impose the addition of extra memory thats needed for the implementation of the function to someone who is not using the trailer support.

sys/log/full/src/log_cbmem.c
vrahane: That's a good catch. Thank you. I will check the existence of `log->l_th` to set the flag.

sys/log/full/src/log.c
vrahane: That is exactly why `l_trailer_arg` was there. If an application needs to process the data, they will have to implement the process callback in which the data will get read during the correct function calls.

sys/log/full/syscfg.yml
vrahane: If the max alignment changes, the max trailer length is going to get affected because alignment plays a big role in fitting the trailer, the implementation of a trailer can be a TLV, a reverse TLV or static data but in order for it to be read successfully from the end, alignment matters. I would prefer to keep it this way if you don't feel strongly.

sys/log/full/src/log.c
vrahane: That's a good catch. Thank you. I will check the existence of `log->l_th` to set the flag.

sys/log/full/src/log_cbmem.c
vrahane: I see what you mean but the way cbmem is used we actually use the scatter gather implementation to add optional data. For example: the image hash is also added as one entry to the scatter gather array. So, that's what I did for the trailer as well. There is no flat buffer implementation for logging to cbmem, so, we do not really need to implement it that way. Unless you feel otherwise, I would like to keep it this way for the sake of simplicity and consistency.

vrahane:
@andrzej-kaczmarek I understand the concerns, while I can address some of them, I have added additional comments to make the `log_trailer_handler{}` little more easier to understand. I have also address some of the comments you mentioned. Please take a look. I have explained my reasoning for others.

sys/log/full/src/log.c
vrahane: Fixed it.

sys/log/full/syscfg.yml
vrahane: Fixed it.

sys/log/full/src/log_cbmem.c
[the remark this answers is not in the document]
andrzej-kaczmarek: My point is, the `LOG_FLAGS_TRAILER_SUPPORT` only makes sens if trailer support is enabled so checking it outside `#ifdef` doesn't make any sense.

sys/log/full/src/log_cbmem.c
andrzej-kaczmarek: The same append callback expects the application to either write the data to the log or return it in provided buffer depending on the log type - how is that "for the sake of simplicity and consistency"?

sys/log/full/include/log/log.h
andrzej-kaczmarek: I guess `LF_MAX_ALIGN * 3` is the same as *default* value for syscfg so if that one is changes, the whole calculation breaks.

sys/log/full/syscfg.yml
andrzej-kaczmarek: The alignment should be handle by the log framework and I should not really need to care about alignment when setting this number, especially that different log storage can have different alignment requirements.

sys/log/full/src/log.c
andrzej-kaczmarek: I don't really understand what you are trying to do here.

If you call `log_walk` to read the log there should be a way for the application to get the trailer data for each entry. This can be either by a callback which `log_walk` should call for each entry or (I think preferably) there should be an API that application can use to read that trailer data.

sys/log/full/src/log.c
```diff
@@ -692,7 +679,9 @@ log_append_prepare(struct log *log, uint8_t module, uint8_t level,
 #endif
 
 #if MYNEWT_VAL(LOG_FLAGS_TRAILER_SUPPORT)
-    ue->ue_flags |= LOG_FLAGS_TRAILER_SUPPORT;
+    if (log->l_th) {
+        ue->ue_flags |= LOG_FLAGS_TRAILER_SUPPORT;
```
andrzej-kaczmarek, line 683: This has the same issue as previous code. If application registers callbacks, the flag will be set for each new log entry, even if the application doesn't append any trailer data via the callback.

sys/log/full/include/log/log.h
```diff
@@ -311,11 +301,23 @@ STATS_SECT_END
...
+     * which gets called at various places where the log entry gets read
+     */
     log_process_trailer_func_t *log_process_trailer;
+    /* Trailer mbuf append callback used to append trailer to the log entry
+     * using the mbufs
+     */
     log_trailer_mbuf_append_func_t *log_trailer_mbuf_append;
+    /* Trailer reset data callback used to reset the trailer data
+     * per log which is defined by the application registering these callbacks
+     */
```
andrzej-kaczmarek, line 320: As I mentioned in some other comment, there's no "user data" in the callback registration API.

andrzej-kaczmarek (changes requested):
Please remove unrelated code from this PR to make it easier to review.
The fix commits should be squashed with their corresponding commits before pushing new version of PR - it's possible to diff PR revisions using GitHub easily if needed.

I think the major issue with this change is the convoluted flow for the callbacks. It seems like the application has to take care of lots of details it should not really care about, e.g. underlying log storage used (https://github.com/apache/mynewt-core/pull/3400#discussion_r2079569218), log storage alignment (https://github.com/apache/mynewt-core/pull/3400#discussion_r2079585469 https://github.com/apache/mynewt-core/pull/3400#discussion_r2079582407), restoring trailer data (https://github.com/apache/mynewt-core/pull/3400#discussion_r2079591435), but actually there's no way for anyone to figure this out without going deep dive into the code.

The idea seems quite simple though, smth like:
- the `append` callback (or equivalent) is called for each log entry so application can provide the trailer data; the log framework takes care of writing this to actual log storage
- some callback or API is used during `log_walk` to provide trailer data to the application; the log framework takes care of reading the trailer data from the log entry if available
That should be really sufficient to do the same in a much easier way, unless there's something I'm missing here.

sys/log/full/src/log.c
vrahane: If the application has registered trailer callbacks for that particular log, that means the application knows how to handle the trailer data. I really don't get your point. Please provide an alternate approach.

sys/log/full/include/log/log.h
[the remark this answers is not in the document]
vrahane: Well, we do need to flush the data. So, do you have an alternate approach in mind ?

sys/log/full/src/log_cbmem.c
vrahane: The `log_cbmem` implementation is what I am talking about. I tried to keep it consistent with that. That's how its implemented.

sys/log/full/src/log_cbmem.c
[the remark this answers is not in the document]
vrahane: done

vrahane:
I addressed review comment based on:

"Please remove unrelated code from this PR to make it easier to review.
The fix commits should be squashed with their corresponding commits before pushing new version of PR - it's possible to diff PR revisions using GitHub easily if needed."

vrahane:
I cannot tell why the tests are failing on CI, they are succeeding locally:
```
Passed tests: [sys/log/full/selftest/align4 sys/log/full/selftest/fcb2_bookmarks sys/log/full/selftest/perlogidx sys/log/full/selftest/align2 sys/log/full/selftest/align1_imghash sys/log/full/selftest/fcb2_align1 sys/log/full/selftest/align8 sys/log/full/selftest/fcb_bookmarks_mixed sys/log/full/selftest/fcb2_watermarks sys/log/full/selftest/fcb_bookmarks sys/log/modlog/selftest sys/log/full/selftest/align1]
All tests passed
```

vrahane:
@sjanc  Can you please take a look at the unit tests and style check, no sure whats going on there. Thank you.

vrahane:
Ok, I think that fixed the tests. As for the updated style check, I think that needs to account for 79 characters in the header. It wants me to wrap s line which is not necessary.

sys/log/full/include/log/log.h
```diff
@@ -218,6 +272,21 @@ struct log {
 #endif
 #if MYNEWT_VAL(LOG_STATS)
     STATS_SECT_DECL(logs) l_stats;
+#endif
+    /* Custom log init callback to be called by the last hdr
+     * read function to read custom data from log entries
+     * at init
+     */
+    log_walk_func_t l_init_cb;
```
andrzej-kaczmarek, line 280: make this also a syscfg and disabled by default

sys/log/full/include/log/log.h
```diff
@@ -218,6 +272,21 @@ struct log {
...
+
+#if MYNEWT_VAL(LOG_FLAGS_TRAILER)
+    /* Log trailer support */
+    /* Void argument for callbacks registeration */
+    void *l_tr_arg;
```
andrzej-kaczmarek, line 285: this argument shall be passed to all callbacks, otherwise it doesn't make any sense

sys/log/full/src/log_fcb.c
```diff
@@ -324,20 +331,15 @@ log_fcb_start_append(struct log *log, int len, struct fcb_entry *loc)
...
 }
 
-/**
- * Calculates the number of message body bytes that should be included after
- * the entry header in the first write.  Inclusion of body bytes is necessary
- * to satisfy the flash hardware's write alignment restrictions.
- */
 static int
-log_fcb_hdr_body_bytes(uint8_t align, uint8_t hdr_len)
+log_fcb_hdr_trailer_bytes(uint16_t align, uint16_t len)
```
andrzej-kaczmarek, line 335: should not be renamed since it apparently is still used when trailer support is disabled

andrzej-kaczmarek (changes requested):

sys/log/full/include/log/log.h
vrahane: done

vrahane:
@andrzej-kaczmarek  I have made the changes. Please take a look.

PR #3408, opened 2025-04-08
log_console: Add option to print time stamp as clock time

kasjer:
When LOG_CONSOLE_PRETTY_WITH_TIMESTAMP is set to 1 timestamp of the logs is print in human readable format instead of really big numbers.

### DIFF
--- a/sys/log/full/syscfg.yml
+++ b/sys/log/full/syscfg.yml
@@ -52,6 +52,11 @@ syscfg.defs:
             Turned off by default when CONSOLE_TICKS is on.
         value: 1
 
+    LOG_CONSOLE_PRETTY_TIMESTAMP_CLOCK:
+        description: >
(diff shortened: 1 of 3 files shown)


sys/log/full/syscfg.yml
```diff
@@ -52,6 +52,11 @@ syscfg.defs:
             Turned off by default when CONSOLE_TICKS is on.
         value: 1
 
+    LOG_CONSOLE_PRETTY_TIMESTAMP_CLOCK:
+        description: >
+            Print timestamp as time yyyy-mm-ddThh:ss:mm.tttttttt
```
andrzej-kaczmarek, line 57: `tttttt`

kasjer: changed from 8 to 6 characters, also `t` to `u` (as for micro seconds)

andrzej-kaczmarek (approved):

PR #3414, opened 2025-04-25
log_fcb: optimize log_fcb_copy

kasjer:
Local function log_fcb_copy() did not use information about start location that was already available.
Function log_fcb_rtr_erase() found first element to copy and then used only part of what wa returned.
In turn log_fcb_copy() walk log again skipping entries.

It would not work correctly if more then on sector was used for source fcb and requested number of
preserved entries span more then one sector.

Code is shorter, faster and uses less stack


vrahane (approved):
LGTM, pretty straight-forward.

PR #3417, opened 2025-04-30
sys/log: Simplify module coloring scheme

apc067:
This simplified module coloring scheme ensures that all modules receive a color, and that the colors remain the same after reboots. It also differentiates between system- vs. per-user modules.

Notes:

- I chose color inversion (ANSI `7;`) instead of explicit background color specification so that it can be agnostic to the typical white-on-black vs. black-on-white terminal text preference (granted that in each case, the normal colors have been configured in the terminal app to be well visible).
- From the module colors, I excluded black & white for the above reason, and red as the color used for serious modlog levels.

### DIFF
--- a/sys/log/full/src/log_console.c
+++ b/sys/log/full/src/log_console.c
@@ -52,63 +52,49 @@ log_console_get(void)
 }
 
 #if MYNEWT_VAL(LOG_CONSOLE_PRETTY)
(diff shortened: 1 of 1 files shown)


kasjer (approved):
LGTM

andrzej-kaczmarek (approved):

PR #3422, opened 2025-05-19
LVGL driver for ST7796S display

kasjer:
Initial support for ST7798S displays with 320x480 resolution.

This also fixes rotation for other drivers


kasjer:
Style check failed due to unreasonable limitation of uncristify

PR #3424, opened 2025-06-01
Shell commands and modules added during link time

kasjer:
### Shell commands registration
So far shell commands could be added with `shell_cmd_register()` and modules (set of commands) could be added with `shell_register()`

This approach requires function call usually added in package initialization function.
Registering commands requires some RAM two tables:
```c
static struct shell_module shell_modules[MYNEWT_VAL(SHELL_MAX_MODULES)];
static struct shell_cmd compat_commands[MYNEWT_VAL(SHELL_MAX_COMPAT_COMMANDS) + 1];
```
Number of modules that can be registered is defined by two mynewt values `SHELL_MAX_MODULES` and `SHELL_MAX_COMPAT_COMMANDS`.
When code tries to add more commands or modules system will crash with assert(0).

### Link time table for commands registration

Now when link time table can be used shell commands and modules can be added to be available in shell.
Commands can be added with macro `MAKE_SHELL_CMD()` and modules can be added using `SHELL_MODULE_WITH_TABLE()`.
Both register commands and modules does not use RAM and do not need function calls.

For backward compatibility `shell_cmd_register()` and `shell_register()` can still be used.
To minimize memory usage syscfg values `SHELL_MAX_MODULES` and `SHELL_MAX_COMPAT_COMMANDS` can be set to 0.

### Benefits of new shell commands registration
- Memory usage is reduced both RAM and flash
- commands are in alphabetical order (they were displayed in order they were registered)
- Targets don't need to specify maximum number of command and modules so adding some package will not result in unexpected assert(0) that instruct you to increase limits.
### Scope of this PR
- Macros for module and commands creation
- Rework of shell command handling to use link time tables
- Update of many packages (not all yet) to use new way of making shell commands
### Comparison of size for target that with many commands
Code without this changes
```
objsize
   text	   data	    bss	    dec	    hex	filename
 121368	   1744	  30572	 153684	  25854	bin/targets/blackpill411ce-slinky/app/@apache-mynewt-core/apps/slinky/slinky.elf
```
Same application after change are applied:
```
objsize
   text	   data	    bss	    dec	    hex	filename
 120844	   1496	  29988	 152328	  25308	bin/targets/blackpill411ce-slinky/app/@apache-mynewt-core/apps/slinky/slinky.elf
```

### DIFF
--- a/kernel/os/include/os/link_tables.h
+++ b/kernel/os/include/os/link_tables.h
@@ -95,7 +95,7 @@
     __attribute__((section("." #table_name), used))
 /* Macro to create attribute for table with name that may affect order of elements */
 #define LINK_TABLE_ELEMENT_SECTION(table_name, elem)                          \
-    __attribute__((section("." #table_name "." #elem)))
+    __attribute__((section("." #table_name "." #elem), used))
 /* Macro to create element that goes to link table.
  * It could be used like this:
  * LINK_TABLE_ELEMENT(example1_table, foo5) = {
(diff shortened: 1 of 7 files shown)


kasjer:
@rymanluk, PR is stripped down now to shell changes only and two examples that shows how os commands **tasks, mpool, reset...** are converted and module **prompt**.
Changes to many other commands that are in mynewt-core will follow once basics are accepted.

PR #3448, opened 2025-07-02
bus: Node config change fix attempt

apc067:


apc067:
Not the good approach. :-)

PR #3456, opened 2025-07-08
sys/log: Trailer buffer pointer, data and callbacks should not be zeroed out on registration

vrahane:
The issue here is that when the most recent log entry gets read, it needs to read the trailer data as well. Zeroing out the trailer info breaks it since the trailer data for last entry won't get read.

This probably got added as part of the review comments which broke the initialization.


sjanc:
Hi,

I have trouble understanding this change.  How this is used exactly? `log_trailer_cbs_register()` (l_tr_om would be left uninitialized anyway) is to be called before `log_register()` ?  (which would be a bit untypical design choice, since  `log_register()` is used to initialize `struct log` in first place)

Unfortunately there is no code that would use this (either unit test or maybe a dedicated simple sample) in upstream tree..
Could you show code snipped on how to use this?

Also please fix commit message (subject line length and missing commit body)

vrahane:
@sjanc  I have added more info in the commit message. Trailer buffer pointer, data and callbacks should not be zeroed out
  on registration. This is because `log_register()` does not just initialize the fields, it also reads the most recent log entry to
  update the index and also call a custom init callback which could be rading the trailer.
- The `trailer_alignment` of 0 was making the offset for the `trailer_len` to be wrong. With this change we do not try to copy padding for the trailer_alignment if it is 0. This change along with the removal of zeroing out the trailer info in the log structure was causing the tests to fail since the offset was wrong. Correcting the `trailer_len` offset fixes everything.
We do have internal tests for our custom trailers which I have been utilizing to test this. The initialization is done the following way:

```
logs_trailer_register_data(&g_mylog, &g_mylog_trailer_data).   <---- This is our custom trailer data registration function (our own API)
                             
log_trailer_cbs_register(&g_mylog, &g_l_th, NULL);
log_register_init_cb(&g_mylog_sum, logs_read_last_hdr_trailer_init_cb);                                                        
log_register("mylog", &g_mylog, &log_fcb_handler, &mylog_fcb, LOG_SYSLEVEL);
logs_ne_init(&g_mylog);    <----- Initializes trailer data variable locally if trailer is not present (our own API)
```
`logs_read_last_hdr_trailer_init_cb()` does the custom trailer data init on `log_register()` after reading the most recent log entry and trailer data. Reading the trailer data needs to be handled here to help any logs that get logged before application libraries can actually initialize the local trailer data variables.

sjanc (approved):

PR #3457, opened 2025-07-09
bsp/apollo3_evb: Fix blinky build

Slawek-skrawek:
LED_BLINK_PIN was defined as MYNEWT_VAL for this board but now syscfg val LED_BLINK_PIN has default value LED_BLINK_PIN This causes build error:
Error: In file included from apps/blinky/src/main.c:25: apps/blinky/src/main.c: In function 'mynewt_main': repos/apache-mynewt-core/hw/bsp/apollo3_evb/include/bsp/bsp.h:51:25: error: implicit declaration of function 'MYNEWT_VAL' [-Wimplicit-function-declaration]
   51 | #define LED_BLINK_PIN   MYNEWT_VAL(LED_0_PIN)
      |                         ^~~~~~~~~~

To fix this expression MYNEWT_VAL(LED_0_PIN)
is changed to MYNEWT_VAL_LED_0_PIN


m-gorecki (commented):
Would be nice to have all pins definitions changed like this. Then it will be possible to select other LEDs in blinky app syscfg. For example:
`  LED_BLINK_PIN: LED_3
`

PR #3459, opened 2025-07-10
fs/fs: Fix dependecy to mcu manager

kasjer:
There was incorrect dependency to mcumgr file system management commands

It was
@apache-mynewt-mcumgr/cmd/fs/port/mynewt
instead of
@apache-mynewt-mcumgr/cmd/fs_mgmt/port/mynewt


vrahane (approved):
LGTM

PR #3462, opened 2025-07-12
bsp/native: Fix linker script for native build

kasjer:
Previously linker script was generated by gcc on Ubuntu. It turned out that it will not work on Fedora builds.

Now link time tables are ADDED to default linker script and default linker script is not replaced by arbitrary version that works only with some toolchains.

Key factor is that .rodata section is an anchor point to put link time tables entries.


sjanc (approved):
tested on F42

PR #3465, opened 2025-07-14
hw/scripts: Add shortcut for accessing STLink via USB S/N

Slawek-skrawek:
`STLINK_SN=<serial>` env allows to specify STLink USB S/N. This is convenient shortcut for `'--extrajtagcmd "--serial <serial>"'` for st-flash or `'--extrajtagcmd "sn=<serial>"'` for STM32_Programmer_CLI.


kasjer (approved):
LGTM

PR #3468, opened 2025-07-17
hw/scripts: Add adapter selection for opencod

wiktorkwiatkowski:
Added support for selecting a specific debug probe using USB serial number via environment variables.
Script creates new .cfg file where serial number is stored, and appends it to the CFG variable that include all .cfg files. Once done the file is deleted.

In case an error appears the .cfg file with S/N is not deleted and user can check whether wrong number didnt cause the error.

This is useful when multiple boards are connected and openocd cannot choose which one is target.

Supported variables:
- **STLINK_SN** for ST-Link
- **JLINK_SN** for J-Link

Example:
`STLINK_SN=0666FF495157808667171335 newt load nucleo-f767zi-blinky`

### DIFF
--- a/hw/scripts/openocd.sh
+++ b/hw/scripts/openocd.sh
@@ -23,6 +23,20 @@ CFG_RESET="reset halt"
 # FILE_NAME must contain the name of the file to load
 # FLASH_OFFSET must contain the offset in flash where to place it
 #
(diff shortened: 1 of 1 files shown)


kasjer (approved):
Verified to work also on Windows with openocd from pacman repository

PR #3470, opened 2025-07-17
sys/log: Improve the console log level indication

apc067:
Added the option to configure the string & color of MODLOG levels in the console printout from sysconfig; also made text highlighting attribute (for system log modules & highlighted log levels) configurable.

Since the maximum MODLOG level is often being used to ensure unconditional logging, it too received its own configuration support.

### DIFF
--- a/sys/log/full/src/log_console.c
+++ b/sys/log/full/src/log_console.c
@@ -55,10 +55,10 @@ log_console_get(void)
 
 #if MYNEWT_VAL(LOG_CONSOLE_PRETTY_WITH_COLORS)
 
-#define CSI                     "\x1b["
-#define CSE                     "m"
-
-#define INV_COLOR               "7;"
+#define ANSI_CSI               "\x1b["
+#define ANSI_SEP               ";"
+#define ANSI_CSE               "m"
+#define ANSI_COLOR_RESET       ANSI_CSI "0" ANSI_CSE
 
 #define COLOR_BLACK             30
 #define COLOR_RED               31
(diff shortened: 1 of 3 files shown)


kasjer (commented):
`datarec` is not a concept known to mynewt-core

apc067:
> `datarec` is not a concept known to mynewt-core

Yes, good point - commit message corrected.

kasjer (approved):
If you have time, you could consider adding colors to syscfg for people with different test.
Name 'MAX' could also be configurable as I assume that in your case it mean something specific and it just happen that LOG_LEVEL_MAX was chosen.

sys/log/full/src/log_console.c
```diff
@@ -122,8 +126,12 @@ static const char * const log_level_str[] = {
     "WRN",
     "ERR",
     "CRI",
+/* Add new custom log levels here */
+    "MAX",
 };
 
+static int real_log_levels = ARRAY_SIZE(log_level_str) - 1;
```
sjanc, line 133: lets make this const (it is compile-time const anyway)

apc067:
> If you have time, you could consider adding colors to syscfg for people with different test. Name 'MAX' could also be configurable as I assume that in your case it mean something specific and it just happen that LOG_LEVEL_MAX was chosen.

That's a good idea; will implement it under this PR

apc067:
> While clang format is not perfect and gives sometime strange output, overall it should be taken into account for obvious fixes
> 
> here eg. LOG_LEVEL_STR() abomination can be ignored but others (eg comments indent) should be taken into account.
> 
> We need to move slowly towards coherent style enforced by clangd format..

Will fix the comment indent. As for the LOG_LEVEL_STR() butchery, it looks visually amazing - could be sold as a post-modernist poem. :-)

sjanc (approved):

PR #3480, opened 2025-07-24
usb/tinyusb: Add msc_flash package to expose flash over USB

kasjer:
Package adds functionality that can present USB disk to the HOST system read and writes target flash or flash area.

This can be used to directly write or read a lot of data without msc_fat_view (that is visible in host system as FAT disk).


vrahane (approved):
Tried it out, LGTM.

PR #3482, opened 2025-07-24
spiflash: Add dependency to stats when needed

kasjer:
When SPIFLASH_STAT was set to 1 code tries to
include stats/stats.h but direct dependency was
not there.
It worked if other packages included stats.

Now code compiles when SPIFLASH_STAT is 1 and
nothing else wanted stats.


vrahane (approved):
LGTM

PR #3496, opened 2025-09-02
tinyusb/da146xx/usb_monitor: Add usb monitor for da146xx

vrahane:
- Add USB monitor to check if USB is connected to DA146xx
- A callback can be registered to detect USB connection state
- tud_connected() is based on the USB SETUP packet which monitors the connectivity with the host instead of just monitoring VBUS.

### DIFF
--- a/hw/usb/tinyusb/da146xx/include/tusb_hw.h
+++ b/hw/usb/tinyusb/da146xx/include/tusb_hw.h
@@ -22,8 +22,27 @@
 
 #define CFG_TUSB_MCU OPT_MCU_DA1469X
 
+#include <os/mynewt.h>
 #include <syscfg/syscfg.h>
 #include <stdbool.h>
+
+#if MYNEWT_VAL(DA146XX_USB_MONITOR)
+/**
+ * Set event queue for USB monitoring
+ */
+void tinyusb_da146xx_usb_monitor_evq_set(struct os_eventq *evq);
+
+/**
+ *
+ * Register callback for USB state monitor changes
+ */
+void tinyusb_da146xx_usb_monitor_register_cb(void (*cb)(bool connected));
+#endif
+
+/**
+ * Check if USB is connected (based on keep-alive and VBUS if used)
+ */
+bool tinyusb_da146xx_is_connected(void);
 
 /* Function should be exported from TinyUSB */
 void tusb_vbus_changed(bool present);
--- a/hw/usb/tinyusb/da146xx/src/da1469x.c
+++ b/hw/usb/tinyusb/da146xx/src/da1469x.c
@@ -24,21 +24,24 @@
 
 #include <device/usbd.h>
 
-static bool vbus_present;
+#if MYNEWT_VAL(DA146XX_USB_MONITOR)
 
-static void
-USBD_IRQHandler(void)
-{
-    tud_int_handler(0);
-}
+#ifdef MYNEWT_VAL_DA146XX_USB_MONITOR_EVQ
+struct os_eventq MYNEWT_VAL(DA146XX_USB_MONITOR_EVQ);
+#endif
+
+#include <da146xx_usb_monitor/da146xx_usb_monitor.h>
+#endif
+
+static bool g_vbus_present;
 
 static void
 tinyusb_vbus_changed(bool present)
 {
-    if (vbus_present == present) {
+    if (g_vbus_present == present) {
         return;
     }
-    vbus_present = present;
+    g_vbus_present = present;
 
     if (present) {
         da1469x_pd_acquire(MCU_PD_DOMAIN_SYS);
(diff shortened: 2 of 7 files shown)


hw/usb/tinyusb/da146xx/include/tusb_hw.h
```diff
@@ -22,9 +22,28 @@
...
 
+#include <os/mynewt.h>
 #include <syscfg/syscfg.h>
 #include <stdbool.h>
 
+#if MYNEWT_VAL(USB_MONITOR)
+/**
+ * Set event queue for USB monitoring
+ */
+void tinyusb_usb_monitor_evq_set(struct os_eventq *evq);
```
sjanc, line 33: this is not used anywhere

hw/usb/tinyusb/da146xx/include/tusb_hw.h
```diff
@@ -22,9 +22,28 @@
...
+ *
+ * Register callback for USB state monitor changes
+ */
+void tinyusb_usb_monitor_register_cb(void (*cb)(bool connected));
+#endif
+
+/**
+ * Check if USB is connected (based on keep-alive and VBUS if used)
+ */
+bool tinyusb_is_connected(void);
```
sjanc, line 45: does this need to be public (and why in da146xx header?) ?  it seems to be used only internally by da1469x.c

hw/usb/tinyusb/da146xx/include/tusb_hw.h
```diff
@@ -22,9 +22,28 @@
...
+/**
+ * Set event queue for USB monitoring
+ */
+void tinyusb_usb_monitor_evq_set(struct os_eventq *evq);
+
+/**
+ *
+ * Register callback for USB state monitor changes
+ */
+void tinyusb_usb_monitor_register_cb(void (*cb)(bool connected));
```
sjanc, line 39: why this is in da146xx specific code?

sjanc:
maybe you could describe what usecase you are trying to achieve with this?

hw/usb/tinyusb/usb_monitor/syscfg.yml
```diff
@@ -0,0 +1,31 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+    USB_MONITOR_RATE_MS:
+        description: 'USB monitor check rate in milliseconds'
+        value: 500
+
+    USB_MONITOR_TIMEOUT_MS:
```
sjanc, line 25: this is also not used

vrahane: Removed it.

hw/usb/tinyusb/da146xx/include/tusb_hw.h
vrahane: Made it da146xx specific code.

hw/usb/tinyusb/da146xx/include/tusb_hw.h
vrahane: It is only needed if the eventq for the callout should not be the default eventq.

vrahane:
@sjanc  so, this helps in the case where VBUS handling for da146xx is custom, VBUS handling is asynchronous and so if one wants to monitor USB state, it has to be in synchronous fashion. What this means is based on configuration for VBUS handling we monitor USB state along with VBUS or we just monitor the VBUS if USB_MONITOR is not enabled. Also, made this code DA146xx specific, so there is no confusion about the behavior for other MCUs.

hw/usb/tinyusb/da146xx/src/da1469x.c
```diff
@@ -66,9 +90,59 @@ tinyusb_hardware_init(void)
...
-    }
-    else if (MYNEWT_VAL_CHOICE(DA1469X_USB_VBUS_HANDLING, ignore)) {
+    } else if (MYNEWT_VAL_CHOICE(DA1469X_USB_VBUS_HANDLING, ignore) ||
+               MYNEWT_VAL_CHOICE(DA1469X_USB_VBUS_HANDLING, custom)) {
         /* If VBUS handling is ignored notify TinyUSB that VBUS is on */
-        tinyusb_vbus_changed(true);
+#if MYNEWT_VAL(DA146XX_USB_MONITOR)
+        /* With USB Monitor enabled, don't immediately notify VBUS on
+         * Wait for actual USB activity (Setup packet) to be detected
+         * by USB Monitor before notifying TinyUSB of VBUS presence.
```
kasjer, line 99: This comment looks strange.

VBUS notification is normally used to connect USB device to  host by connection pull up resistor.
If application for some reason ignores VBUS (present of not) and makes call to tusb_connect(0) to start activity on USB, I don't see any reason to later notify USB about VBUS that will call tusb_connect again.
First call will start USB activity regardless of VBUS.
VBUS is not used for anything else apart from initiating connection and disconnecting when it's gone.

vrahane:
alright, we are moving this internally.

PR #3501, opened 2025-09-24
hw/mcu/nordic: Bump nrfx to v3.14.0

andrzej-kaczmarek:


kasjer:
Changing `nrf_saadc_burst_set` to `nrf_saadc_channel_burst_set` will probably fix build problem

andrzej-kaczmarek:
added skip style check since nrfx_glue is copied from nrfx

PR #3502, opened 2025-09-26
DA1469x: Fix crash when LPCLK is set to XTAL32K

apc067:
Fix the root-cause of `g_mcu_clock_xtal32k_freq` not getting updated with a proper frequency value.


apc067:
> looks good, could you just fix code style bot complains?
> 
> I'd suggest to remove 'de-facto' and make comment only 1 space indent, maybe it will fit line limit than

Done.

PR #3504, opened 2025-09-30
File system rework

kasjer:
This is start of a rework of file system related code.

There was a few drawbacks in current implementation of file system especially disk related (this does not touch FCB).
- disk interface read and write used 32 bit byte addresses limiting usage to SD cards up to 4GB which is not much in current era. Now disk interface will use block address (instead of bytes) so it is more consistent with how disks are accessed (SD cards)
- filesystem ops is now split to filesystem and filesystem ops (only functions), so same file system ops can be used by several file systems (few partitions or two SD cards)
- code adds mount points so file systems can be mounted and unmouted
- disk interface that is currently used for FAT only will be used also to expose SD cards to USB
- Code adds cli command for mount points
- MMC driver updated to work with new disk API
- MMC now uses modlog

### DIFF
--- a/fs/disk/include/disk/disk.h
+++ b/fs/disk/include/disk/disk.h
@@ -22,7 +22,8 @@
 
 #include <stddef.h>
 #include <inttypes.h>
-#include "os/mynewt.h"
+#include <os/mynewt.h>
+#include <os/link_tables.h>
 
 #ifdef __cplusplus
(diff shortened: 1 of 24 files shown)


sjanc:
looks ok to me, lets rebase and get this in

sjanc (approved):

PR #3534, opened 2025-11-03
    hw/scripts/nrfutil: make traits optional and improve device selection

wiktorkwiatkowski:
Traits are no longer by default. They are now optional and only used when the user explicitly wants to flash multiple boards at once. DFU mode is controlled through NRFUTIL_DFU_MODE, and NRFUTIL_DFU_SN can be used to select a specific DFU device when several are connected.

### DIFF
--- a/hw/scripts/nrfutil.sh
+++ b/hw/scripts/nrfutil.sh
@@ -38,8 +38,14 @@ nrfutil_load () {
         echo "Missing filename"
         exit 1
     fi
+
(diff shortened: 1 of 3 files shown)


hw/scripts/nrfutil.sh
sjanc: we should also double check if traits are used correctly

from what I understand traits are used only for device matching, not for selecting type of update to be performed

PR #3535, opened 2025-11-04
hw/bsp/dialog_da1469x-dk-pro: select linker script by MCU_TARGET

wiktorkwiatkowski:
Pick the linker script dynamically based on the syscfg MCU_TARGET value. This makes flashing DA14691 easier since it has a different RAM size than other DA1469x chips.


hw/bsp/dialog_da1469x-dk-pro/include/bsp/bsp.h
```diff
@@ -25,6 +25,9 @@
 
 #ifdef __cplusplus
 extern "C" {
```
sjanc: unrelated  change

sjanc (changes requested):
other than those spurious empty lines looks OK to me

kasjer (approved):

PR #3538, opened 2025-11-05
mcu/stm32: Fix hal_nvreg for STM32F1

kasjer:
This fixes 3 issues for hal_nvreg for STM32F1 devices.
- STM32F1 registers are 16 bits not 32 like in other series.
- Backup registers for ST HAL for STM32F1 start from 1 not 0.
- PWR clock needs to be turned on if it was not before

### DIFF
--- a/hw/mcu/stm/stm32_common/src/hal_nvreg.c
+++ b/hw/mcu/stm/stm32_common/src/hal_nvreg.c
@@ -17,6 +17,8 @@
  * under the License.
  */
 
+#include <stdbool.h>
+#include <os/mynewt.h>
 #include <hal/hal_nvreg.h>
 #include <mcu/stm32_hal.h>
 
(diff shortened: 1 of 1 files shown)


hw/mcu/stm/stm32_common/src/hal_nvreg.c
```diff
@@ -30,23 +32,38 @@
...
+#endif
 
 void
 hal_nvreg_write(unsigned int reg, uint32_t val)
 {
 #if PWR_ENABLED
     RTC_HandleTypeDef hrtc = { .Instance = RTC };
     if (reg < HAL_NVREG_MAX) {
+#if defined(__HAL_RCC_PWR_IS_CLK_ENABLED)
+        bool disable_pwr = __HAL_RCC_PWR_IS_CLK_ENABLED();
```
sjanc: this looks like there is `!` missing in condition,  ie to disable only if it wasn't already enabled

also, for parity I'd put both enable() and disable() calls under same condition

PR #3551, opened 2025-12-01
hw/bsp: Enable nRF5340 network core when HCI transport is selected

sjanc:
There is no point in not enabling networking core when HCI transport is selected. This makes it a bit more user friendly since one doesn't have to do this in target.


sjanc:
#AutoPTS run mynewt nrf53 SM/PER/SCJW/BV-02-C

PR #3602, opened 2026-04-07
Add support for STM32H5 devices with nucleo-h563zi board

kasjer:
This adds initial support for STM32H563 device.

Example BSP is nucleo-h563zi board


sjanc:
looks ok, but please update rat and LICENSE

sjanc (approved):

PR #3625, opened 2026-05-04
ci: automate HW CI status updates

wiktorkwiatkowski:
This adds script and workflow to keep hardware test statuses up to date. It relies on HTML comment tags within the README to identify and replace the badge section during each run


.github/scripts/update_badges.py
```diff
@@ -0,0 +1,84 @@
+import urllib.request
```
sjanc, line 20: please add proper licence header

.github/workflows/update_hw_ci_badges.yml
```diff
@@ -0,0 +1,35 @@
+name: Update HW CI Badges
```
sjanc, line 20: licence

sjanc (changes requested):

sjanc (approved):

PR #3628, opened 2026-05-06
sys/config: Add static conf handlers support

kasjer:
So far conf_handlers could be registered at run time.
This approach delays loading of configuration till
application start processing events.

This limits what can be configured in config.

Now system supports static configuration,
conf_handler can be put in link time table
so conf_handler is ready from the very start
and configuration can be loaded earlier.

CONFIG_AUTO_LOAD syscfg value allows to load
configuration during initialization of config
subsystem as static handlers are already in place.

Dynamic handlers are still supported while discuraged.

sys/id nad sys/fault packages are converted to use static registration

stm32_eth now support configuration via sys/config as an example
how to use it

### DIFF
--- a/hw/drivers/lwip/stm32_eth/syscfg.yml
+++ b/hw/drivers/lwip/stm32_eth/syscfg.yml
@@ -24,9 +24,30 @@ syscfg.defs:
         value: 240
 
     STM32_MAC_ADDR:
+        description: "Use STM32_ETH_MAC_ADDR instead"
+        defunct: 1
+
+    STM32_ETH_MAC_ADDR:
         description: "Use given array of 6 bytes for MAC address."
-        value: ((uint8_t[6]){0x00, 0x01, 0x01, 0x02, 0x02, 0x03})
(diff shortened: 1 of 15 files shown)


hw/drivers/lwip/stm32_eth/syscfg.yml
```diff
@@ -24,9 +24,30 @@ syscfg.defs:
         value: 240
 
     STM32_MAC_ADDR:
+        description: "Use STM32_ETH_MAC_ADDR instead"
+        defunct: 1
+
+    STM32_ETH_MAC_ADDR:
         description: "Use given array of 6 bytes for MAC address."
-        value: ((uint8_t[6]){0x00, 0x01, 0x01, 0x02, 0x02, 0x03})
+        value: 0x00,0x01,0x01,0x02,0x02,0x03
```
andrzej-kaczmarek, line 32: what about using simple uint64 format, like we have in nimble?

see BLE_LL_PUBLIC_DEV_ADDR

sjanc (approved):

PR #3639, opened 2026-05-15
hw/mcu/nrf54l: Add GRTC as hal timer TIMER_5

mkasenberg:
Support for RTC10 has been removed since nrfx 3.10.0


hw/mcu/nordic/nrf54lxx/src/hal_timer.c
```diff
@@ -126,6 +136,95 @@ nrf_read_timer_cntr(NRF_TIMER_Type *hwtimer)
...
+    if (delta_ticks <= 0) {
+        /* Force interrupt to occur as we may have missed it */
+        NVIC_SetPendingIRQ(bsptimer->tmr_irq_num);
+        return;
+    }
+
+    /*
+     * Schedule compare relative to current absolute GRTC time.
+     */
+    expiry64 = cntr + hal_rtc_ticks_to_grtc((uint32_t)delta_ticks);
```
andrzej-kaczmarek: this is probably incorrect. timer value should be always aligned with "logical" rtc tick, otherwise the tick is set to be anywhere 0..+31us vs. actual tick time

PR #3640, opened 2026-05-15
tinyusb: Fix power requirement description

kasjer:
**USBD_CONFIGURATION_MAX_POWER** description stated that it used 2 ma units just as is stored in device
descriptor. But it really is in 1 mA and is converted in code before it land up in descriptor.

This fixes description so it's not misleading.


sjanc (approved):
Could be extended with note that it is stil in 2mA steps

PR #3642, opened 2026-05-19
mgmt/imgmgr: Add option to enable frugal list in mcumgr

andrzej-kaczmarek:
This option is then used by mcumgr to set default value for its syscfg, as for other options.


andrzej-kaczmarek:
related to https://github.com/apache/mynewt-mcumgr/pull/193

sjanc (approved):

PR #3673, opened 2026-06-11
util/ring_buffer: Extend API for easy usage

kasjer:
- This adds few functions that make working with ring_buffer easier.

- uart_console changed to use initialization function instead of direct struct filed access


sjanc (approved):
would be good to get some unittests for this mdoule too

PR #3674, opened 2026-06-12
Tinycrypt removal

sjanc:


hw/drivers/flash/enc_flash/ef_mbedtls/include/ef_mbedtls/ef_mbedtls.h
```diff
@@ -30,9 +30,9 @@ extern "C" {
 #endif
 
 /*
- * Tinycrypt specific version of the flash device.
+ * MbedTLS specific version of the flash device.
  */
-struct eflash_tinycrypt_dev {
+struct eflash_mbedtls_dev {
```
kasjer, line 35: `etd_dev`, `etd_key` took names from **e**flash_**t**inycrypt_**d**ev maybe etd_ could be dropped as it does seems to be helpful, see `log_entry_hdr` that was left with some strange prefix before.
Same probably happened to `eflash_da1469x_dev` that is filled with strange prefixes.

kasjer (changes requested):
Other then filed name prefix looks good. Tested OK.

kasjer (approved):

PR #3685, opened 2026-06-25
Rename .bssnz to .noinit to fix linker wildcard conflicts

wiktorkwiatkowski:
Previously, the .bss.core.nz section was placed before .bss to avoid the *(.bss*) linker wildcard. This placement assigned it a low RAM address, causing the bootloader to accidentally overwrite these variables during startup. Moving the section after .bss was not an option because the wildcard would still catch it. Renaming the section to .noinit solves this problem. It allows the section to be safely placed after .bss at a higher memory address, ensuring the variables survive system resets.


sjanc (approved):

kasjer (commented):
LGTM, I will check some boards to see if it builds.

Some alignment is strange

PR #3711, opened 2026-07-20
shell: Fix LINK_TABLE start pointer alignment in shell module tables

stenslae:
In Mynewt, linker tables are used to group structs together. However, if the linker doesn't perfectly align the start of the section to the struct's size boundary, C pointer arithmetic miscalculates the array bounds. This manually aligns the pointer.


kasjer (approved):
@stenslae Thank you for your fix.

PR #3719, opened 2026-07-30
mtest: Add watchdog test

wiktorkwiatkowski:


test/mtest/mtest_watchdog/src/main.c
```diff
@@ -0,0 +1,62 @@
...
+        (void)boot_status;
+    }
+}
+
+MTEST_SUITE(watchdog_test)
+{
+    MTEST_RUN_INIT(watchdog_test);
+    switch (reset_count) {
+    case 0:
+        watchdog_test_case_1();
```
sjanc, line 43: while those functions should never return I'd still add break for sake of coding style

sjanc (commented):
just minor comment, otherwise looks ok

test/mtest/mtest_watchdog/src/main.c
wiktorkwiatkowski: https://github.com/apache/mynewt-core/blob/2705534e37259f1684d6fc360adfc7b0bcd08da4/test/mtest/mtest_watchdog/src/watchdog_test_cases.c#L46-L48

`watchdog_test_case_1();` actually does return when the time counter is at least 2x `WATCHDOG_INTERVAL`. 
Adding a `break` statement would break the test flow, and `watchdog_test_case_2();` would never execute.

wiktorkwiatkowski: I can change the switch statement to something like this:
```
if (reset_count == 0) { 
        watchdog_test_case_1();
        watchdog_test_case_2();
    } else if (reset_count == 1) {
        watchdog_test_case_2();
    } else {
        boot_status = 0;
        reset_count = 0;
    } 
```
It will work the same but will get rid of the fallthrough in the switch statement.

sjanc: switch is fine,  but comment if  fall-through is on purpose

sjanc (approved):

PR #3723, opened 2026-08-10
Revert "shell: Fix LINK_TABLE start pointer alignment in shell module…

kasjer:
#3711 turned out to be not correct solution.

While it works for native Linux 64-bit build, is breaks normal target builds.

It's true that affected table element is 32 bytes long. But size of this structure is not the reason why elements are aligned to 32 bytes so manual alignment to structure size is unfortunately wrong solution.
For 32 bit ARM builds this alignment does not happen.

Working on other solution for this...

This reverts commit 1431593f7a2c46c71edd9d3b4e4d11db7ee037bb.

### DIFF
--- a/sys/shell/src/shell.c
+++ b/sys/shell/src/shell.c
@@ -31,30 +31,19 @@
 
 #define SHELL_PROMPT "shell"
 
-static const struct shell_cmd *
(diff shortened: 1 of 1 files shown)


sjanc (approved):

sjanc:
@stenslae FYI

PR #3738, opened 2026-09-07
Bump NRFX to 4.5

kasjer:
Changes to build mynewt code with nrfx 4.5

NRFX does not have now files that remove all content
depending on NRFX_XXXX_ENABLED flags.

Mynewt SDK package for NRFX selects now only relevant
files during build instead of building everything.

Some minor changes to adjust name changes in structures.

### DIFF
--- a/hw/mcu/nordic/syscfg.yml
+++ b/hw/mcu/nordic/syscfg.yml
@@ -1,0 +1,199 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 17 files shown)


hw/mcu/nordic/syscfg.yml
```diff
@@ -0,0 +1,199 @@
...
+        description: Enable NRFX PWM driver
+        value: 0
+
+    NRFX_QDEC_ENABLED:
+        description: Enable NRFX QDEC driver
+        value: 0
+
+    NRFX_QSPI_ENABLED:
+        description: Enable NRFX QSPI driver
+        value:
```
sjanc: is this left without default value on purpose?

kasjer: It turns out that for SDK package value can be overridden not only in bsp, app or target. MCU for NRF5340 turns this on.
I will set default value to 0.

